PR #5154, opened 2015-04-14
Define constants for maximum username/email length

MpDzik:


sebastienros:
I don't think it makes sense to take this one on 1.9-int. It's safe but at the same time not critical for shipping. Would you mind create the same PR targeting 1.x ?

MpDzik:
I created a new PR #5156. Closing this one.

Issue #4905, opened 2014-11-17
Taxonomy Term count doesn't decrease


orchardbot:
@rodpl created:
https://orchard.codeplex.com/workitem/21076

Orchard 1.8.x

Taxonomy Term count doesn't reduce its quantity when term is selected and then deselected for content item.

PR #5155, opened 2015-04-14
#4905: Fixed the Taxonomy term counter does not decrease.

benschi11:
I think the problem is that the TermsPartHandler triggers the TermCountProcessor as a Task async. So the ContentItem is unpublished. We need to retrieve the TermPart with the VersionOption "VersionOptions.Latest" instead of the default "VersionOptions.Published".

Link to codeplex: https://orchard.codeplex.com/SourceControl/network/forks/benschi11/orchardtaxonomycount/contribution/8129

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs
@@ -91,7 +91,8 @@ namespace Orchard.Taxonomies.Handlers {
         
         // Fires off a processing engine task to run the count processing after the request so it's non-blocking.
         private void RecalculateCount(IProcessingEngine processingEngine, ShellSettings shellSettings, IShellDescriptorManager shellDescriptorManager, TermsPart part) {
-            processingEngine.AddTask(shellSettings, shellDescriptorManager.GetShellDescriptor(), "ITermCountProcessor.Process", new Dictionary<string, object> { { "termsPartId", part.ContentItem.Id } });
+            var termPartRecordIds = part.Terms.Select(t => t.TermRecord.Id).ToArray();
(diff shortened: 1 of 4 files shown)


sebastienros:
Don't you think this change would create the opposite issue, which is that if you create a Draft version it would increment the counter ?

benschi11:
As far as I can see the ITermCountProcessor gets only called if a TermPart is published, unpublished or deleted. So that should be a problem or am I wrong?

```C#
OnPublished<TermsPart>((context, part) => RecalculateCount(processingEngine, shellSettings, shellDescriptorManager, part));
OnUnpublished<TermsPart>((context, part) => RecalculateCount(processingEngine, shellSettings, shellDescriptorManager, part));
OnRemoved<TermsPart>((context, part) => RecalculateCount(processingEngine, shellSettings, shellDescriptorManager, part));
```
This events call this method:

```C#
// Fires off a processing engine task to run the count processing after the request so it's non-blocking.
private void RecalculateCount(IProcessingEngine processingEngine, ShellSettings shellSettings, IShellDescriptorManager shellDescriptorManager, TermsPart part) {
	processingEngine.AddTask(shellSettings, shellDescriptorManager.GetShellDescriptor(), "ITermCountProcessor.Process", new Dictionary<string, object> { { "termsPartId", part.ContentItem.Id } });
}
```

benschi11:
I think there are more changes to make. I found another bug:
1) Create a new content item with a TaxonomyField and "Allow new terms to be created" is true
2) Publish the content item with the the terms "test1" and "test2"
3) the count of "test1" and the count of "test2" is 1 (it is correct)
4) Edit the content item and remove the tag "test2"
5) The count of "test1" is 1 and the count of "test2" is also 1 (incorrect)

I thought about fixing the bug with working with the VersionNumbers.

sebastienros:
Also, please target 1.8.x. The integration branch is supposed to be frozen, unless a regression is found, or we like you so much that we'll take it.

benschi11:
I think I need to close this one and create a new one? Or is there a other way?

sebastienros:
I have no idea. Feel free to close it, it might be easier.

benschi11:
Will create a new PR for the 1.8.x branch.

PR #5156, opened 2015-04-15
Define constants for maximum username/email length

MpDzik:


Jetski5822:
Are we sure the Email Length should be validated interms of a total length? should we not validate the domain length and then the total length?

Jetski5822:
http://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address

MpDzik:
I'm not sure. I did this refactoring because I am importing users into Orchard from external data sources and I need to validate the input data. I want to duplicate Orchard's validation logic, but without hardcoding magic numbers. That's why I needed to introduce these constants.

pszmyd:
I guess we already have a 255-char limit already because of the default nvarchar length used.

Issue #5148, opened 2015-04-10
Default recipe error logged on setup (simple fix included)


orchardbot:
@jtkech created:
https://orchard.codeplex.com/workitem/21330

This happens when trying to delete the recipe directory that isn't empty yet

A simple fix, in RecipeStepQueue.cs - Dequeue(), is to replace
if (stepIndex < 1) {

With
if (stepIndex < 0) {

Best

sebastienros:
Approved. Can you please do the change? You can even do it directly from github now.

jtkech:
ok

jtkech:
Done
https://github.com/OrchardCMS/Orchard/pull/5171

jtkech:
Done in the right Milestone
https://github.com/OrchardCMS/Orchard/pull/5176

PR #5171, opened 2015-04-16
#5148: Default recipe error logged on setup

jtkech:
This happens when trying to delete the recipe directory that isn't empty yet


sebastienros:
Please follow the milestone the bug was assigned to, 1.8.x. Can you target 1.8.x with this PR instead of 1.9-int?

jtkech:
Ok

PR #5176, opened 2015-04-17
#5148: Default recipe error on setup (Milestone: 1.8.x)

jtkech:
This happens when trying to delete the recipe directory that isn't empty yet


src/Orchard.Web/Modules/Orchard.Recipes/Services/RecipeStepQueue.cs
```diff
@@ -57,7 +57,7 @@ public class RecipeStepQueue : IRecipeStepQueue {
                 _appDataFolder.DeleteFile(stepPath);
             }
 
-            if (stepIndex < 1) {
+            if (stepIndex < 0) {
```
sebastienros: It's actually an else from the previous block. But that's fine.

jtkech:
Oh yes, it's because I wanted to replace only one character ;)

PR #5177, opened 2015-04-18
Search Widget Enhancements

sfmskywalker:
This PR refactors the Search Settings as well as Search Widget to enable users to choose which index to search.

The Search Settings has been modified such that the user can configure the searchable fields for all available indexes.

The Admin Content Search settings UI has been greatly simplified by taking advantage of above changes. All that this setting needs now is which index to search for the admin search - the fields are configured are stored as part of the SearchSettingsPart for the desired index.


src/Orchard.Web/Modules/Orchard.Search/Controllers/SearchController.cs
```diff
@@ -45,23 +46,24 @@ public class SearchController : Controller {
         public ILogger Logger { get; set; }
         dynamic Shape { get; set; }
 
-        public ActionResult Index(PagerParameters pagerParameters, string q = "") {
+        public ActionResult Index(PagerParameters pagerParameters, string searchIndex = null, string q = "") {
```
Jetski5822: Breaking Change. I think if we are going to do this, we might want to think about creating a SearchParameters object?

sfmskywalker: How is this a breaking change? The _searchIndex_ argument is optional.

Jetski5822: Ohh its on a Controller, my bad. Not a breaking change :)

sfmskywalker: No worries. :) Thanks for checking!

PR #5178, opened 2015-04-19
Remove Duplicate code that could be cleanedup for better maintenance.

vairam-svs:
Remove Duplicate code that could be cleanedup for better maintenance.

### DIFF
--- a/src/Orchard/UI/Resources/MetaEntry.cs
+++ b/src/Orchard/UI/Resources/MetaEntry.cs
@@ -13,23 +13,19 @@ namespace Orchard.UI.Resources {
 
         public MetaEntry(string name = null, string content = null, string httpEquiv = null, string charset = null)
         {
-            if (!String.IsNullOrEmpty(name))
-            {
+            if (!String.IsNullOrEmpty(name)) {
                 Name = name;
             }
 
-            if (!String.IsNullOrEmpty(content))
-            {
+            if (!String.IsNullOrEmpty(content)) {
                 Content = content;
             }
 
-            if (!String.IsNullOrEmpty(httpEquiv))
-            {
+            if (!String.IsNullOrEmpty(httpEquiv)) {
                 HttpEquiv = httpEquiv;
             }
 
-            if (!String.IsNullOrEmpty(charset))
-            {
+            if (!String.IsNullOrEmpty(charset)) {
                 Charset = charset;
             }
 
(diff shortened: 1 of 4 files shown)


src/Orchard/UI/Resources/MetaEntry.cs
```diff
@@ -5,6 +5,35 @@
 namespace Orchard.UI.Resources {
     public class MetaEntry {
         private readonly TagBuilder _builder = new TagBuilder("meta");
+        
+        public MetaEntry()
+        {
+
+        }
```
Jetski5822: do we need a default ctor?

src/Orchard/UI/Resources/MetaEntry.cs
```diff
@@ -5,6 +5,35 @@
...
+            {
+                HttpEquiv = httpEquiv;
+            }
+
+            if (!String.IsNullOrEmpty(charset))
+            {
+                Charset = charset;
+            }
+
+        }
```
Jetski5822: Can you please fix brackets so they are on the first line :)

src/Orchard/UI/Resources/MetaEntry.cs
```diff
@@ -5,6 +5,35 @@
...
     public class MetaEntry {
         private readonly TagBuilder _builder = new TagBuilder("meta");
+        
+        public MetaEntry()
+        {
+
+        }
+
+        public MetaEntry(string name = null, string content = null, string httpEquiv = null, string charset = null)
+        {
```
Jetski5822: do they have to be null? surely they can be left with out the default null?

src/Orchard/UI/Resources/MetaEntry.cs
vairam-svs: It is also being used in 5 references with no argument with optional,
except for the following, everything else can be changed to with parameters since they are using name, content only and they are in order and don't have to pass null for rest of them.
public static MetaEntry Combine(MetaEntry meta1, MetaEntry meta2, string contentSeparator) {
            var newMeta = new MetaEntry();
            Merge(newMeta._builder.Attributes, meta1._builder.Attributes, meta2._builder.Attributes);
            if (!String.IsNullOrEmpty(meta1.Content) && !String.IsNullOrEmpty(meta2.Content)) {
                newMeta.Content = meta1.Content + contentSeparator + meta2.Content;
            }
            return newMeta;
        }

src/Orchard/UI/Resources/MetaEntry.cs
vairam-svs: ha, that code was a copy from its original location. will make that change

src/Orchard/UI/Resources/MetaEntry.cs
vairam-svs: that was to meet the requirement of the calling methods and the code segment to keep arguments optional.

sebastienros: Could you also target 1.x instead of master ? Thanks. Everything else looks good to me.

PR #5179, opened 2015-04-19
Moving default SiteSettings values to a handler.

Ermesx:
Added missing timezone data and fixed conflicts.


Jetski5822:
GTG

sebastienros:
The only concern I would have with this change could be that there might be some code paths where a new Site is created and it's settings shouldn't be initialized. This might explain why the code was done this way in the first place. However I don't think this is the case, so I m merging, and it's on 1.x so it gives us some time to analyze the impact.

Issue #2536, opened 2012-05-26
UseDefaultCredentials implemented incorrectly in EmailMessagingChannel


orchardbot:
**Znowman** created:
https://orchard.codeplex.com/workitem/18709

We had problems sending mail through the SMTP server we have been given by our customer. That's because of the following line  of code in the latest orchard 1.4.2 in the file EmailMessagingChannel:

smtpClient.UseDefaultCredentials = !smtpSettings.RequireCredentials;
if (!smtpClient.UseDefaultCredentials && !String.IsNullOrWhiteSpace(smtpSettings.UserName)) {
                    smtpClient.Credentials = new NetworkCredential(smtpSettings.UserName, smtpSettings.Password);
                }

We have to pass the credentials but still have to set UseDefaultCredentials to false to enable mailing through the given smtp server.
Seems strange but see the documentation for details: 
http://msdn.microsoft.com/en-us/library/system.net.mail.smtpclient.usedefaultcredentials.aspx

I suggest to create a checkbox in the backend of the email settings so the admin can manage the UseDefaultCredentials  property himself. Many thanks!

Issue #5024, opened 2015-02-11
SMTP credentials have to be set to 'null'


orchardbot:
**mwentz** created:
https://orchard.codeplex.com/workitem/21195

I am using the Workflow and Email Messaging modules to send an email to myself when a user fills out a custom form. In the logs I get a "5.7.1 Client does not have permissions to send as this sender" error when I don't set to require credentials, but it works just fine when I set it to require credentials and I use 'null' for the username and password. It also works to require credentials and put in valid credentials, but unfortunately the address that we want to use for the "from" address was set up for Exchange in a way that it does not have credentials associated with it.

I am migrating the site from a Sitefinity site on the same server with the same SMTP settings and it works fine without requiring credentials in Sitefinity.

PR #5200, opened 2015-04-24
Fix for bugs #2536 and #5024

keesd:
Fixed the Orchard behavior of passing windows credentials to e-mail server when no credentials where entered.

 Added an option to the smptsettings configuration to explicitly configure windows authentication.
This fixes bugs #2536 and #5024.


src/Orchard.Web/Modules/Orchard.Email/Models/SmtpSettingsPart.cs
```diff
@@ -41,6 +41,12 @@ public class SmtpSettingsPart : ContentPart {
             set { this.Store(x => x.RequireCredentials, value); }
         }
 
+        public bool UseDefaultCredentials
+        {
```
sfmskywalker: Please use K&R style formatting (opening braces on the same line).

sfmskywalker:
The fix looks great and I like the added feature to properly use the ```UseDefaultCredentials``` (Use Windows authentication). However, I'd like to see the two radio buttons right underneath each other, and toggling the credentials input fields when the user selects the "Specify username/password" option. This can be done without writing javascript when using the ```data-controllerid``` attribute on a container element surrounding the username and password inputs (the value of that attribute would be the ID of the linked radio button). The script enabling this data- attribute comes from the "ShapesBase" script resource.

keesd:
I know, but the radiobuttons are rendered with a name-attribute that contains a ‘ .’ (dot). This causes problems in the .js code that does the showing/hiding based on the data-controllerids.

This would require a fix in base.js, line 178
$("[name=" + controller.attr("name") + "]").click(function () { $("[name=" + $(this).attr("name") + "]")

the $[name] query would require quotes around the controller.attr("name") . 

Can I include this fix and reissue the pull request?

sebastienros:
Please target 1.x, or dev if 1.x is no more.

sebastienros:
Ping ?

keesd:
resubmitted in pr #5257

keesd:
Resubmitted in dev branch

Issue #5113, opened 2015-03-21
SecureSocketsLayerService: Incorrect Hostname when port number is included (fix included)


orchardbot:
@jtkech created:
https://orchard.codeplex.com/workitem/21284

Related to this issue #20935 that has been resolved but introduces another issue. As @mbroskamp said, with a hostname = "localhost:44300", the UriBuilder.Host becomes "44300"

A quick fix was to replace in MakeInsecure()
var builder = new UriBuilder(insecureHostName.Trim('/') + path) {

With
var builder = new UriBuilder(insecureHostName.Split(':').First().Trim('/') + path) {

And doing the same in MakeSecure()

Note: The port number is correctly set afterwards

Best

PR #5202, opened 2015-04-25
#5113: SecureSocketsLayer, incorrect Hostname when port number is included

jtkech:
For example, with a hostname = "localhost:44300", the UriBuilder.Host becomes "44300"


src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Services/SecureSocketsLayerService.cs
```diff
@@ -214,7 +214,7 @@ public class SecureSocketsLayerService : ISecureSocketsLayerService {
             var settings = GetSettings();
             if (settings == null) return path;
             var insecureHostName = settings.InsecureHostName;
-            var builder = new UriBuilder(insecureHostName.Trim('/') + path) {
+            var builder = new UriBuilder(insecureHostName.Split(':').First().Trim('/') + path) {
```
MpDzik: Shouldn't we have some unit tests for this?

PR #5206, opened 2015-04-26
Make Orchard.Core.Common.Utilities.LazyField<T> [Obsolete]

### LINKED ISSUE #765: There are two LazyField<T>
@sebastienros created:
https://orchard.codeplex.com/workitem/16933

One is in Common, the other in Framework, and there is a difference between the two. Should't we have only one in Framework ? Or is the difference meaningful ?

### LINKED ISSUE #3746: LazyField<T> is duplicated in codebase
__hkui__ created:
https://orchard.codeplex.com/workitem/19918

Orchard.Core.Common.Utilities.LazyField<T>
Orchard.ContentManagement.Utilities.LazyField<T>

### LINKED ISSUE #5191: Deprecate, then remove Orchard.Core.Common.Utilities.LazyField
...since there is another LazyField, in Orchard.ContentManagement.Utilities. The latter seem to be used everywhere any way.
There is a slight difference between the two, the loader delegate in the one in Orchard.ContentManagement.Utilities also gets fed the existing value of the field. This is quite strange however since the LazyField can only already have a value when the loader delegate is executed, when it got its value set previously (and actually it's also necessary that the loader should be set after the value was set).

Ermesx:
It closes #765 and closes #3746 and closes #5191.

I removed _value parameter from loader delegate because it was pointless (_value was always null if noone run setter before loader) and it is used nowhere except storage filters.

Please check if all is well.
W.

### DIFF
--- a/src/Orchard/ContentManagement/Utilities/LazyField.cs
+++ b/src/Orchard/ContentManagement/Utilities/LazyField.cs
@@ -3,7 +3,7 @@
 namespace Orchard.ContentManagement.Utilities {
     public class LazyField<T> {
         private T _value;
-        private Func<T, T> _loader;
+        private Func<T> _loader;
         private Func<T, T> _setter;
 
         public T Value {
@@ -11,7 +11,7 @@ namespace Orchard.ContentManagement.Utilities {
             set { SetValue(value); }
         }
 
-        public void Loader(Func<T, T> loader) {
+        public void Loader(Func<T> loader) {
             _loader = loader;
         }
 
@@ -21,7 +21,7 @@ namespace Orchard.ContentManagement.Utilities {
 
         private T GetValue() {
             if (_loader != null) {
-                _value = _loader(_value);
+                _value = _loader();
                 _loader = null;
             }
             return _value;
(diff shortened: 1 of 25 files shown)


src/Orchard/ContentManagement/Utilities/LazyField.cs
```diff
@@ -11,7 +11,7 @@ public class LazyField<T> {
             set { SetValue(value); }
         }
 
-        public void Loader(Func<T, T> loader) {
+        public void Loader(Func<T> loader) {
```
MpDzik: This is a breaking change.

Ermesx: Hmmm... ok :)
Maybe should I create new LazyfField class and make obsolete both old ?
Agree ?

src/Orchard/ContentManagement/Utilities/LazyField.cs
```diff
@@ -11,17 +11,23 @@ public class LazyField<T> {
...
         }
 
-        public void Loader(Func<T, T> loader) {
+        public void Loader(Func<T> loader) {
             _loader = loader;
         }
 
+        [Obsolete]
+        public void Loader(Func<T, T> loader)
+        {
```
Ermesx: Oh god... when I back home I will fix formatting.
BTW. I think Orchard should use StyleCop it is extremely usefull.

src/Orchard/ContentManagement/Utilities/LazyField.cs
MpDzik: I guess that's a good idea. However, @sebastienros has the final word in this matter.

Issue #5187, opened 2015-04-21
Error using Azure Table Storage in 1.9-int - OData 5.6.2 vs 5.6.3


dcinzona:
We are storing some data in Azure Table Storage and when we try to execute CreateIfNotExists, we get the following error: 

`Could not load file or assembly 'Microsoft.Data.OData, Version=5.6.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies`

We are using the Azure library files in the lib folder ( Azure Storage 4.3.0.0 and Microsoft.Data.OData 5.6.3).  It seems as though there's a hard reference to OData 5.6.2.0 somewhere?  I've looked and can't seem to find it...

Any ideas (or could this be a bug in the Azure SDK)?

dcinzona:
I've also posted an issue on Microsofts Azure SDK repo: https://github.com/Azure/azure-storage-net/issues/125

DanielStolt:
Truly perplexing - looks like your assembly binding redirect just doesn't take!

The only time I've ever seen that happening is when running code in the RoleEntry.OnStart() method in an Azure web role (cloud service) because this code is executed using a different process which has its own App.config. Could this be the case by any chance? But then again, you say it repros locally in IIS Express as well, so that shouldn't be it.

You should enable fusion logging at this point and examine the log. This will tell you exactly how it's determining which version to load, and where it is trying to find that version of the assembly.

dcinzona:
Agreed Daniel, it's weird.  I've tested this on a local IIS deployment (on my dev machine) and on Azure Websites.  The issue persists on both.  When I'm back in the office on Monday, I test again with fusion logging enabled.  It has me baffled.
I'll report back here with my findings. 
Thanks!

dcinzona:
Pulled this from the fuslogvw tool:

### For Microsoft.Data.OData 5.6.2.0:

```
*** Assembly Binder Log Entry  (5/1/2015 @ 3:20:21 PM) ***

The operation failed.
Bind result: hr = 0x80131040. No description available.

Assembly manager loaded from:  C:\Windows\Microsoft.NET\Framework64\v4.0.30319\clr.dll
Running under executable  c:\windows\system32\inetsrv\w3wp.exe
--- A detailed error log follows. 

=== Pre-bind state information ===
LOG: DisplayName = Microsoft.Data.OData, Version=5.6.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35
[...]

PR #5222, opened 2015-05-01
Added assembly bindings for OData 5.6.3 - Fix for #5187

dcinzona:
Microsoft.WindowsAzure.Storage tries to load Microsoft.Data.OData 5.6.2 unless there's a binding in app.config or web.config.  This binding needs to be in the root web.config of the application.  It does not work in the module web.config.


DanielStolt:
I see you initially had this one targeted for 1.9-int, and then retargeted to 1.x. I think this one is blocking and needs to go into 1.9-int, so in the interest of time, I have applied the corrections and pushed to 1.9-int, rather than have you re-retarget this PR. The fix will be merged into 1.x, so I'm closing this PR.

dcinzona:
Sounds good! Thanks!

PR #5231, opened 2015-05-02
Fix for issue #5219

giannik:
will need to bump the version numbers from 1.8.1.0  to 1.9 when released.


sfmskywalker:
I think this should go into the dev branch, since the master branch will not contain the changes made to the project references - that was done in 1.x.

sfmskywalker:
It looks like we can add assembly references without specifying a version number. In my test, this works:

```xml
<add assembly="Orchard.Framework"/>
<add assembly="Orchard.Core"/>
```

giannik:
yes, removing version number  works for me too. 
Shall i make a new pull request in 1.x or will you handle it ?

sfmskywalker:
That'd be great. Thanks!

giannik:
done.
new pull request #5232

Issue #5235, opened 2015-05-02
Awaiting activities not cleared on workflow updating (fix included)


jtkech:
__Repro:__
 - Create a worflow with a starting activity (e.g Content Published)
 - Add a timer activity (e.g 5 minutes)... And save your workflow
 - Publish a content item to start your workflow and wait a little
 - Then, edit, optionally update, and save again your workflow
 - At this point, your workflow has been updated and stopped

__Issue:__
 - 1) The workflow is still listed as running in the "Workflows Definitions" page
 - 2) And in the "Running workflows" page related to the item, you get an exception

__Analysis:__
 - 1) The timer awaiting Activity isn't cleared when the workflow is saved
 - 2) This awaiting activity is linked to an activity that doesn't exist

__Fix:__
In AdminController.cs, EditPost() action, after this line

    workflowDefinitionRecord.ActivityRecords.Clear();

Add this line

    workflowDefinitionRecord.WorkflowRecords.Clear();

This way, the related workflows and their awaiting activities will be cleared

Best

sebastienros:
I can't find where, but I am pretty sure we commented on the same issue explaining the recommended fix. The idea is to display a confirmation message on Save if there are any running instances,. And if accepted then go on with deleting the instances.

jtkech:
Oh yes, but, when editing a worflow definition, there is no info about the associated workflows in the current view model. Anyway, I think we need to see if they are running instances at the time the user clicks on the "Save" button

So, to get this needed sync info on "Save", I've added an ajax get request that targets a new controller action. But not sure I've done it the right Orchard way, with right names... Let me know

jtkech:
Fixed, the related PR has been committed

PR #5251, opened 2015-05-06
#5235: Awaiting activities not cleared on saving

jtkech:
After editing and updating a Workflow, the Workflow is stopped. But the workflow is still listed as running in the "Workflows Definitions" page. And in the "Running workflows" page related to a content item, you get an exception

By clearing WorkflowRecords in the submit.Save controller action, the related workflows and their awaiting activities are cleared


sebastienros:
Thanks, after some discussions about this issue we think that it needs to be improved.

1- Make Workflow Definitions cloneable, so that we can run existing workflow instances while editing the new definition. A user can then wait for all instances to be finished, or force the deletion (see next point).
2- Display a warning/confirmation popup when editing or deleting a workflow definition which has running instances. If the user decides to force the deletion, then your current code is also executed (deleting the running instances of this definition). Please look at how the `delete` actions do that already.

sebastienros:
Also, please target 1.9.x

PR #5256, opened 2015-05-07
#5235: Awaiting activities not cleared on saving

jtkech:
By clearing WorkflowRecords in the submit.Save controller action, the related workflows and their awaiting activities are cleared

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
@@ -265,6 +265,7 @@ namespace Orchard.Workflows.Controllers {
             var activitiesIndex = new Dictionary<string, ActivityRecord>();
 
             workflowDefinitionRecord.ActivityRecords.Clear();
+            workflowDefinitionRecord.WorkflowRecords.Clear();
 
             foreach (var activity in state.Activities) {
                 ActivityRecord activityRecord;
--- a/src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
+using System.Security.Authentication;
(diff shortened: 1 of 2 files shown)


sebastienros:
Please take the comments in the issue into account. Thanks.

jtkech:
Oh yes, but, when editing a worflow definition, there is no info about the associated workflows in the current view model. Anyway, I think we need to see if they are running instances at the time the user clicks on the "Save" button

So, to get this needed sync info on "Save", I've added an ajax get request that targets a new controller action. But not sure I've done it the right Orchard way, with right names... Let me know

src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
```diff
@@ -159,6 +159,12 @@ IFormManager formManager
             return RedirectToAction("Edit", new { workflowDefinitionRecord.Id });
         }
 
+        public JsonResult State(int? id) {
+            var workflowDefinitionRecord = id.HasValue ? _workflowDefinitionRecords.Get(id.Value) : null;
+            var isRunning = workflowDefinitionRecord != null ? workflowDefinitionRecord.WorkflowRecords.Any() : false;
```
sebastienros: Can be:

`var isRunning = workflowDefinitionRecord != null && workflowDefinitionRecord.WorkflowRecords.Any()`

src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
```diff
@@ -159,6 +159,12 @@ IFormManager formManager
             return RedirectToAction("Edit", new { workflowDefinitionRecord.Id });
         }
 
+        public JsonResult State(int? id) {
```
sebastienros: If it's a GET then we need to check for the permissions:
`if (!Services.Authorizer.Authorize(StandardPermissions.SiteOwner, T("Not authorized to edit workflows")))
                 return new HttpUnauthorizedResult();`

jtkech:
Done, but I can't return an HttpUnauthorizedResult(), not compatible to a JsonResult. So I've used an AuthenticationException(). Note: If you want I can use an [HttpPost]

PR #5258, opened 2015-05-07
Various proposals

vesko-k:


vesko-k:
Proprosal 3. Orchard.Widgets - check access permissions when dealing with widgets or layers, will work if the user has appropriate permission set in Contents Feature or Content Permissions part is added to the widget/layer content definitions with right default permissions set.
Unfortunately there is no other way to set widgets/layers content definition access permissions like other creatable types via admin users panel.

PR #5262, opened 2015-05-11
Orchard.ContentPermissions - Allow to apply default permissions if content item access control is not enabled.

vesko-k:


sebastienros:
Two remarks:
1- Can you please revert all the formatting changes you did? You need to follow the formatting conventions
2- Could you explain a little bit more what the goal of the feature is? Screenshots?

Thanks

vesko-k:
In brief, the main purpose of changes is to achieve more flexible security in widgets/layers management.  By now there is only one security permission responsible for that (ManageWidgets). Although there is content definition for widgets, there is not possibility to manage content permissions like other creatable types, which means that there is no way to tune widgets creation and layer management per widget/layer for a role.
From the other hand there is ContentPermissions module who can be used to achieve that with not much modifications. 
For widgets the idea is to add a ContentPermissions part to the widget type and to tune permissions for that type and to check them when user manages widgets/layers. 
Similarly this approach can be applied to the Layer type (to attach ContentPermissions part to the layers definition). Such we can give content permissions for each layer we create and to check permission against the user who is trying to manage it.
Concerning ContentPermissions my idea is to use content definition security permission settings of the module to be used optionally like default permissions set if security override part in the content item is not enabled or still content is not created. Thus when I add content permissions part to the widget/layer type, I can block the other users to have ability to enable it and to user default permission for the whole widget/layer type. 

I apologize if I have not expressed clearly enough, but this is not my native language.

sebastienros:
What is "content definition security permission settings" ?
You can revert you formatting changes by hand, or using git on the specific lines your want to revert.

vesko-k:
Under "content definition security permission setting" I mean content part settings of the ContentPermissions module. Maybe the definition is not accurate, but I call it so because configuration is done in the content type definition (creating or editing a content type).

sebastienros:
Can you give a demo on a Tuesday meeting?

sebastienros:
Please reopen once you are ready to show us how it works.

PR #5267, opened 2015-05-11
Autoroute patterns localization

Skrypt:
Resending for dev branch (1.x renamed)


Skrypt:
Working on it guys don't worry

Skrypt:
This is ready you can test it out first. I added some nice jQuery to show/hide patterns. And also added the option to use the patterns per culture or not.

sebastienros:
- Remove all "en-US" from migrations, leave the Culture empty, and assume it's the SiteCulture if empty.
- Define the default values (without culture) in the Create methods of the migrations, and actually migrate the previous patterns to the new ones, without losing them.

src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
```diff
@@ -37,5 +52,23 @@ public class Migrations : DataMigrationImpl {
...
         }
+
+        public int UpdateFrom3() {
+
+            SchemaBuilder.AlterTable("AutoroutePartRecord", table => table
+                .AddColumn<bool>("UseCulturePattern", c => c.WithDefault(false))
+            );
+
+            // populate the AutorouteSettings.DefaultPatternDefinitions AutoroutePart settings
+            foreach (var autoroutePart in _contentManager.Query<AutoroutePart, AutoroutePartRecord>().List()) {
```
sebastienros: Loop over the Content Types, and for each type, update its Part settings.

src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
```diff
@@ -37,5 +52,23 @@ public class Migrations : DataMigrationImpl {
...
+                var settings = autoroutePart.Settings.GetModel<AutorouteSettings>();
+                string patternIndex = autoroutePart.Settings["AutorouteSettings.DefaultPatternIndex"];
+                settings.DefaultPatterns = new List<DefaultPattern> { new DefaultPattern { PatternIndex = patternIndex, Culture = "" } };
```
sebastienros: Not even sure you need to update the value as you can do it while the object is used the first time.

src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
```diff
@@ -37,5 +52,23 @@ public class Migrations : DataMigrationImpl {
 
             return 3;
         }
+
+        public int UpdateFrom3() {
+
+            SchemaBuilder.AlterTable("AutoroutePartRecord", table => table
```
sebastienros: This should be in `Create` and then return 4.

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettings.cs
```diff
@@ -39,18 +45,58 @@ public class AutorouteSettings {
                 return _patterns;
             }
 
-            set { 
+            set {
                 _patterns = value;
                 PatternDefinitions = new JavaScriptSerializer().Serialize(_patterns.ToArray());
```
sebastienros: We know it's my fault, but please fix it by using the serialization service. Or Json.net directly.

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettings.cs
```diff
@@ -39,18 +45,58 @@ public class AutorouteSettings {
...
 
+        /// <summary>
+        /// A serialized Json array of <see cref="DefaultPattern"/> objects
+        /// </summary>
+        public string DefaultPatternDefinitions { get; set; }
+
+        public List<DefaultPattern> DefaultPatterns {
+            get {
+                if (_defaultPatterns == null) {
+                    _defaultPatterns = new JavaScriptSerializer().Deserialize<DefaultPattern[]>(DefaultPatternDefinitions).ToList();
```
sebastienros: Serialization service.

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettings.cs
```diff
@@ -39,18 +45,58 @@ public class AutorouteSettings {
...
+                //We split the values from the radio button returned values
+                int i = 0;
+                foreach (DefaultPattern defaultPattern in _defaultPatterns) {
+                    if (defaultPattern.Culture.Split('|').Count() > 1) {
+                        _defaultPatterns[i].PatternIndex = defaultPattern.Culture.Split('|').Last();
+                        _defaultPatterns[i].Culture = defaultPattern.Culture.Split('|').First();
+                    }
+                    i++;
+                }
+                DefaultPatternDefinitions = new JavaScriptSerializer().Serialize(_defaultPatterns.ToArray());
```
sebastienros: here too

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettingsEvents.cs
```diff
@@ -25,8 +28,60 @@ public class AutorouteSettingsHooks : ContentDefinitionEditorEventsBase {
...
+                            Culture = settings.DefaultSiteCulture
+                        });
+                    }
+                    current++;
+                }
+
+                //We add a pattern for each culture if there is none
+                if (!settings.Patterns.Where(x => x.Culture == culture).Any()) {
+                    //We add the default pattern from migrations
+                    if (settings.Patterns.Where(x => x.Culture == "").Any()) {
```
sebastienros: Risky, use `String.IsNullOrEmpty` just in case there are nulls.

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettingsEvents.cs
```diff
@@ -25,8 +28,60 @@ public class AutorouteSettingsHooks : ContentDefinitionEditorEventsBase {
...
+                        //we add the RoutePattern and we set the culture since there is none defined
+                        RoutePattern migrationRoutePattern = settings.Patterns.Where(x => x.Culture == "").First();
```
sebastienros: `String.IsNullOrEmpty`, everywhere

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettingsEvents.cs
```diff
@@ -39,31 +94,37 @@ public class AutorouteSettingsHooks : ContentDefinitionEditorEventsBase {
...
-                        Description = "my-title",
-                        Pattern = "{Content.Slug}"
-                    });
-
-                    _notifier.Warning(T("A default pattern has been added to AutoroutePart"));
+                //If there is no default pattern for each culture we set default ones
+                List<RoutePattern> newPatterns = new List<RoutePattern>();
+                int current = 0;
+                foreach (string culture in settings.SiteCultures) {
+                    if (settings.Patterns.Any(x => x.Culture == culture)) {
```
sebastienros: `String.Equals(a,b, StringComparison.OrdinalIgnoreCase)`

src/Orchard.Web/Modules/Orchard.Blogs/Migrations.cs
```diff
@@ -39,8 +39,8 @@ public class Migrations : DataMigrationImpl {
                     .WithPart("AutoroutePart", builder => builder
                         .WithSetting("AutorouteSettings.AllowCustomPattern", "True")
                         .WithSetting("AutorouteSettings.AutomaticAdjustmentOnEdit", "False")
-                        .WithSetting("AutorouteSettings.PatternDefinitions", "[{\"Name\":\"Title\",\"Pattern\":\"{Content.Slug}\",\"Description\":\"my-blog\"}]")
-                        .WithSetting("AutorouteSettings.DefaultPatternIndex", "0"))
+                        .WithSetting("AutorouteSettings.PatternDefinitions", "[{\"Name\":\"Title\",\"Pattern\":\"{Content.Slug}\",\"Description\":\"my-blog\",\"Culture\":\"\"}]")
```
sebastienros: Remove all these default values in the migrations, and handle the case where there is now patterndefinition when you try to read it. At that time only create a default one, which will be saved if the user clicks Save in the UI. Or you actually CLONE the current ones for all the available cultures.

src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
```diff
@@ -105,8 +105,8 @@ public class TaxonomyService : ITaxonomyService {
                     .WithPart("AutoroutePart", builder => builder
                         .WithSetting("AutorouteSettings.AllowCustomPattern", "true")
                         .WithSetting("AutorouteSettings.AutomaticAdjustmentOnEdit", "false")
-                        .WithSetting("AutorouteSettings.PatternDefinitions", "[{Name:'Taxonomy and Title', Pattern: '{Content.Container.Path}/{Content.Slug}', Description: 'my-taxonomy/my-term/sub-term'}]")
```
sebastienros: Undo

src/Orchard.Web/Modules/Orchard.Taxonomies/Migrations.cs
```diff
@@ -36,8 +36,8 @@ public class Migrations : DataMigrationImpl {
                 .WithPart("AutoroutePart", builder => builder
                 .WithSetting("AutorouteSettings.AllowCustomPattern", "True")
                 .WithSetting("AutorouteSettings.AutomaticAdjustmentOnEdit", "False")
-                .WithSetting("AutorouteSettings.PatternDefinitions", "[{\"Name\":\"Title\",\"Pattern\":\"{Content.Slug}\",\"Description\":\"my-taxonomy\"}]")
```
sebastienros: Undo

src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
```diff
@@ -190,8 +190,8 @@ public class Migrations : DataMigrationImpl {
                      .WithPart("AutoroutePart", builder => builder
                         .WithSetting("AutorouteSettings.AllowCustomPattern", "True")
                         .WithSetting("AutorouteSettings.AutomaticAdjustmentOnEdit", "False")
-                        .WithSetting("AutorouteSettings.PatternDefinitions", "[{\"Name\":\"Title\",\"Pattern\":\"{Content.Slug}\",\"Description\":\"my-projections\"}]")
```
sebastienros: Undo

src/Orchard.Web/Modules/Orchard.Pages/Migrations.cs
```diff
@@ -14,8 +14,8 @@ public class Migrations : DataMigrationImpl {
                 .WithPart("AutoroutePart", builder => builder
                     .WithSetting("AutorouteSettings.AllowCustomPattern", "True")
                     .WithSetting("AutorouteSettings.AutomaticAdjustmentOnEdit", "False")
-                    .WithSetting("AutorouteSettings.PatternDefinitions", "[{\"Name\":\"Title\",\"Pattern\":\"{Content.Slug}\",\"Description\":\"my-page\"}]")
```
sebastienros: Undo

src/Orchard.Web/Modules/Orchard.Lists/Migrations.cs
```diff
@@ -11,10 +11,10 @@ public class Migrations : DataMigrationImpl {
                     .WithPart("ContainerPart")
                     .WithPart("MenuPart")
                     .WithPart("AutoroutePart", builder => builder
-                        .WithSetting("AutorouteSettings.AllowCustomPattern", "True")
```
sebastienros: Undo

src/Orchard.Web/Modules/Orchard.DynamicForms/Migrations.cs
```diff
@@ -66,5 +66,15 @@ public class Migrations : DataMigrationImpl {
                 .DisplayedAs("Form Widget"));
             return 1;
         }
+
```
sebastienros: Undo

src/Orchard.Web/Modules/Orchard.CustomForms/Migrations.cs
```diff
@@ -12,8 +12,8 @@ public class Migrations : DataMigrationImpl {
                     .WithPart("AutoroutePart", builder => builder
                         .WithSetting("AutorouteSettings.AllowCustomPattern", "True")
                         .WithSetting("AutorouteSettings.AutomaticAdjustmentOnEdit", "False")
-                        .WithSetting("AutorouteSettings.PatternDefinitions", "[{\"Name\":\"Title\",\"Pattern\":\"{Content.Slug}\",\"Description\":\"my-form\"}]")
```
sebastienros: Undo

src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
Skrypt: You are right. autoroutePart.Settings["AutorouteSettings.DefaultPatternIndex"]; is still in the database and we can still get it. The migration should take care of it for not seeing any reference of the old setting in the code. If we lazy update it and we program an autogenerateAlias function it will also need to batch update them one by one later on. I prefer having it updated from the start instead of needing to add overhead of validations. But that's your call in the end ;)

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettingsEvents.cs
Skrypt: I could add this everywhere I can find Culture == culture in that module too. Just in case someone edits the value by hand in the database and he uses lowercase and that we know that they are always like this : fr-CA <-- the bad french ;)

src/Orchard.Web/Modules/Orchard.Blogs/Migrations.cs
Skrypt: Since I have now the option to enable/disable the culture patterns ; I just added the current patterns to the default culture leaving the other cultures with the default pattern "{Content.Slug}". So, by default the culture patterns will never be enabled. And after the migration it will also be set to false. So, it should not break any actual patterns. Though, if you say that you want me to CLONE those old patterns into every cultures already installed we could do that also in the migration. But there is also the fact that we enable the culture patterns for making a distinction between each culture patterns. So they will need to edit/save them anyhow to do that.

Skrypt:
I fixed everything as asked on today meeting. I added some comments, let me know what you think ; I can do other changes, if needed, pretty quick.

Skrypt:
You are right. autoroutePart.Settings["AutorouteSettings.DefaultPatternIndex"]; is still in the database and we can still get it. The migration should take care of it for not seeing any reference of the old setting in the code. If we lazy update it and we program an autogenerateAlias function it will also need to batch update them one by one later on. My preference is having it updated from the start instead of needing to add overhead of validations. But that's your call in the end ;)

Also, since I have now the option to enable/disable the culture patterns ; I just added the current patterns to the default culture leaving the other cultures with the default pattern "{Content.Slug}". So, by default the culture patterns option will never be enabled. And after the migration it will also be set to false. So, it should not break any actual patterns. Though, if you say that you want me to CLONE those old patterns into every cultures already installed we could do that also in the migration. But there is also the fact that we need to enable the culture patterns for making a distinction between each culture patterns. So they will need to edit/save them anyhow to do that.

@sfmskywalker @sebastienros

Skrypt:
Here it is as requested.

src/Orchard.Web/Modules/Orchard.Autoroute/Settings/AutorouteSettings.cs
```diff
@@ -12,45 +13,95 @@ namespace Orchard.Autoroute.Settings {
...
         /// A serialized Json array of <see cref="RoutePattern"/> objects
         /// </summary>
         public string PatternDefinitions { get; set; }
-        
+
         public List<RoutePattern> Patterns {
             get {
                 if (_patterns == null) {
-                    _patterns = new JavaScriptSerializer().Deserialize<RoutePattern[]>(PatternDefinitions).ToList();
+                    _patterns = new DefaultJsonConverter().Deserialize<RoutePattern[]>(PatternDefinitions).ToList();
```
sebastienros: Look at what other settings do to serialize in JSon

sebastienros:
Last comment is to check how to serialize the settings in JSON.

src/Orchard.Web/Modules/Orchard.Autoroute/Drivers/AutoroutePartDriver.cs
```diff
@@ -20,54 +25,99 @@ public class AutoroutePartDriver : ContentPartDriver<AutoroutePart> {
...
+
+                if (aspect != null) {
+                    itemCulture = aspect.Culture;
+                }
+            }
+
+            if (settings.UseCulturePattern) {
+                //if we are creating from a form post we use the form value for culture
+                HttpContextBase context = _httpContextAccessor.Current();
+                if (!String.IsNullOrEmpty(context.Request.Form["Localization.SelectedCulture"])) {
```
sfmskywalker: This is a major code smell, accessing Request.Form from the AutorouteService. Is there no other way?

src/Orchard.Web/Modules/Orchard.Autoroute/Drivers/AutoroutePartDriver.cs
```diff
@@ -20,54 +25,99 @@ public class AutoroutePartDriver : ContentPartDriver<AutoroutePart> {
...
+            if (!settings.Patterns.Any(x => String.Equals(x.Culture, itemCulture, StringComparison.OrdinalIgnoreCase))) {
+                settings.Patterns = new List<RoutePattern> { new RoutePattern { Name = "Title", Description = "my-title", Pattern = "{Content.Slug}", Culture = itemCulture } };
+            }
 
-                _notifier.Warning(T("No route patterns are currently defined for this Content Type. If you don't set one in the settings, a default one will be used."));
+            // if the content type has no defaultPattern for autoroute, then use a default one
+            if (!settings.DefaultPatterns.Any(x => String.Equals(x.Culture, itemCulture, StringComparison.OrdinalIgnoreCase))) {
+                //if we are in the default culture check the old setting
+                if (String.Equals(itemCulture, _cultureManager.GetSiteCulture(), StringComparison.OrdinalIgnoreCase)) {
+                    if (!String.IsNullOrEmpty(part.TypePartDefinition.Settings["AutorouteSettings.DefaultPatternIndex"])) {
```
sfmskywalker: This causes a `KeyNotFoundException` when setting up a new site.
Is there a reason we're not adding DefaultPatternIndex to the AutoroutePartSettings class so that we can use the `settings`variable instead?

sfmskywalker: I refactored this bit because it was breaking the autoroute part feature for new Orchard installations, so you may want to get latest from dev and update your PR, sorry.

src/Orchard.Web/Modules/Orchard.Autoroute/Drivers/AutoroutePartDriver.cs
Skrypt: Since the LocalizatioPart is not already attached to the content type I found no other way to get this value.

src/Orchard.Web/Modules/Orchard.Autoroute/Drivers/AutoroutePartDriver.cs
Skrypt: I thought I tested this scenario a lot of time but that could have happened with all the changes I made in the end.

Skrypt: Just tested it and you are right it cause a null reference error. I approve the change you made it fix it. Sorry, I forgot to test this scenario after adding the lazy update.

sfmskywalker: No worries, it happens. Thank you for the awesome contribution!

Issue #4963, opened 2015-01-10
French translation and admin navigation


orchardbot:
@Skrypt created:
https://orchard.codeplex.com/workitem/21134

When we add the french translation to Orchard and that we want to create multiple menu in the admin navigation section. The translation provided in the .po files is making the menu switching to fail with a javascript error.

In Orchard.Web/Core/Navigation/Views/Admin/Index.cshtml
at line 138
var leaveConfirmation = '@T("Some items where not saved.")';

should be
var leaveConfirmation = "@T("Some items where not saved.")";

To prevent translations that are using ' char like : Certains éléments n'ont pas été enregistrés.
To fail. The opposite should never happen with " char since in the .po files the string's are encapsulated inside "string".

Skrypt:
The pull request is merged in 1.9.x branch I'm closing this.

PR #5269, opened 2015-05-12
Fixing #4963

Skrypt:
To prevent translations that are using ' char like : Certains éléments n'ont pas été enregistrés.
To fail.

Fixing #4963


sebastienros:
the solution is to use `HttpUtility.JavascriptEncode`

sebastienros:
At the same time please check for other occurrences of `'@` and `"@`.

Skrypt:
Done, in contrary of what I was thinking there was only 2 occurences of this in the code. Plus, I fixed the javascript to display those correctly when using encoded chars. For some reason, when using a message in a confirm you need to put it in a HTML element for him to decode it in HTML else it will display the encoded string.

sebastienros:
Please target 1.9.x

Issue #5224, opened 2015-05-02
[1.9-int] Response header Content-Type missing charset when cached by Orchard.OutputCache


Xeevis:
1. Fresh install of **1.9-int**
2. Sign out
3. On homepage refresh **CTRL + R** (cached) yields response header
   
   Content-Type:"text/html"
4. **CTRL + Shift + R** (no-cache) yields
   
   Content-Type:"text/html; charset=utf-8"

Xeevis:
I'll make a bold statement here *moment of silence*, this appears to be bug in .NET Framework :scream_cat: *(or not)*

In `Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs` there's method `ServeCachedItem()`

    private void ServeCachedItem(ActionExecutingContext filterContext, CacheItem cacheItem) {
        var response = filterContext.HttpContext.Response;

        // Adds some caching information to the output if requested.
        if (CacheSettings.DebugMode) {
            response.AddHeader("X-Cached-On", cacheItem.CachedOnUtc.ToString("r"));
            response.AddHeader("X-Cached-Until", cacheItem.ValidUntilUtc.ToString("r"));
        }

        // Shorcut action execution.
        filterContext.Result = new FileContentResult(cacheItem.Output, cacheItem.ContentType);

        response.StatusCode = cacheItem.StatusCode;

        ApplyCacheControl(response);
    }

Nothing wrong? WRONG! `filterContext.HttpContext.Charset` string property has value but setter probably has some code that needs to execute so yeah the way to fix this issue is to simply add one fair looking line

    response.Charset = response.Charset;

My best bug fix to date!

PR #5273, opened 2015-05-13
 #5224: Fix for missing charset in response header for cached output

Xeevis:
Honestly I have no idea why this code is needed to get charset back into response headers, but it works ...


sebastienros:
Do you think we should save the Charset as part as the cached item too?

/cc @DanielStolt

DanielStolt:
Yes that actually occurred to me too, when I was working on the module. I was thinking maybe we should save both encoding and charset on the CacheItemRecord?

Xeevis:
I don't see what benefit would come from saving charset as part of the cached item. To my knowledge charset in response header is generally set either through `machine.config` or `web.config` respectively, changing either busts output cache right? So it's virtually impossible to have improper charset on cached item.

I'm sure you noticed I'm not really changing anything, I'm just trying to restore what should be default behavior but for some mysterious reason isn't set properly.

sebastienros:
Let's take it as is right now and refactor if necessary.

Issue #5281, opened 2015-05-15
Menu names showing with encoding when using special chars


Skrypt:
The menu names are displaying with encoding when using special chars

![screencapture-localhost-30321-orchardlocal-admin-navigation-1431702581088](https://cloud.githubusercontent.com/assets/3228637/7655818/3cfef742-faf4-11e4-9c1f-94fe77322b00.png)

PR #5282, opened 2015-05-15
Fixing #5281

Skrypt:
This fix it. Html.ItemDisplayText() always returns an encoded string. So, since we are in an < option > html element it was displaying the HTML encoded string.

Another solution would be to add an optional bool parameter to the Html.ItemDisplayText() function that would say if we return an encoded MvcHtmlString or not (the default as true).

Something like this : 
```C#
        public static MvcHtmlString ItemDisplayText(this HtmlHelper html, IContent content, bool encode = true) {
            var metadata = content.ContentItem.ContentManager.GetItemMetadata(content);
            if (metadata.DisplayText == null)
                return null;
            if (encode) {
                return MvcHtmlString.Create(html.Encode(metadata.DisplayText));
            } else {
                return MvcHtmlString.Create(metadata.DisplayText);
            }
        }
```

Fixing #5281


sebastienros:
Please go with the other solution. Thanks.

Issue #5291, opened 2015-05-16
Placement editor for content types try/catch block breaks HTML


Hazzamanic:
In \Orchard.ContentTypes\Views\Admin\EditPlacement.cshtml there is a try catch block to catch an error on rendering a shape that Orchard may not be able to find. However, the display of this shape also contains a div around it, within the try catch, so when if it fails, the closing div tag is never rendered. This is a problem because the view disables the div in which a shape is rendered, but without the closing div, it means all the form fields get that disabled attribute and won't be sent back to the server. Worst case scenario, it errors when you try to save, best case, it doesn't save everything. Only a minor issue but... yeah

PR #5292, opened 2015-05-16
fixing placementeditor try/catch block

Hazzamanic:
To fix this issue https://github.com/OrchardCMS/Orchard/issues/5291

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditPlacement.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditPlacement.cshtml
@@ -24,11 +24,13 @@
             
             <li data-shape-type="@placement.ShapeType" data-shape-differentiator="@placement.Differentiator" data-shape-zone="Content" data-shape-position="@placement.Position">
                 <div class="shape-type"><h3>@placement.ShapeType @placement.Differentiator</h3></div>
-                @try {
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditPlacement.cshtml
```diff
@@ -24,11 +24,13 @@
...
-                }
-                catch {
-                }
+				<div class="shape-editor">
+					@try {
+						@Display(Model.AllPlacements[i].Shape)
+					}
+					catch {
+					}
+				</div>
```
MpDzik: I think we should avoid tabs and use spaces instead.

sfmskywalker: Yes, that's actually our established standard - all spaces, no tabs.

Hazzamanic: Right, so I need to replace the tab to four spaces? Sorry, it's my first PR. Also, is there somewhere I can see these "standards"? The docs has a very brief section on C# that doesn't mention spaces from what I can see.

sebastienros:
Nit picking, but yes, please fix the indentation. You can also define it as the default (using spacing when you hit TAB) in the Visual Studio options. This is what we usually do.

Issue #5303, opened 2015-05-20
Unable to change edited menu in Navigation


TFleury:
Yet a new french problem : there is a script error if leaveConfirmation message contains a single quote (which is the case with French translation).

Because of this error, auto post-back script don't work, and so, we can't select the menu to edit.

sebastienros:
Fixed by #5282

PR #5304, opened 2015-05-20
#5303 Unable to change edited menu in Navigation

TFleury:
Issue occurs when leaveConfirmation message contains a single quote.


sebastienros:
Duplicates of #4963 #5269

PR #5307, opened 2015-05-20
Feature/async content

aonweb:
This is a modification to the Content Manger, Content Handler, Content Part Driver pipeline to allow asynchronous content drivers and content shape results.

We've been using this basic code on the amctheatres.com website, which makes heavy use of ContentPartDrivers with content sourced from http web service calls on the site for a while now.


DanielStolt:
Wow, this is an impressive contribution!

Just a technicality before we get into reviewing - this should probably be retargeted to `dev` as `1.9.x` is only for very safe, non-breaking fixes and changes.

sfmskywalker:
Awesome. Would you be willing to talk about this coming Tuesday, talk us through?

bleroy:
Nice!

sebastienros:
Agreed, we definitely need to talk about it on a Tuesday meeting.

Quick question, why is AsyncBridge still required? Orchard is requiring 4.5.1. Could you remove that?
What about the improvements, have you been able to measure it?
And it will definitely be a dev branch target.

aonweb:
Thanks. I'll merge in dev and resubmit.

Does 4.5.1 not deadlock when calling .Result or .Wait() on a Task in a synchronous method? If that is the case, then the AsyncBridge could probably be removed. I can also go ahead and make the widget filter render in parallel as well, which I was having problems with previously because of deadlocks .

Regarding performance, I haven't run any numbers verses the synchronous version to see what the async overhead looks like, but that shouldn't be too difficult. In our use case on amctheatres, where at least half the drivers are making an http call, the async overhead is negligible compared to the parallelization benefit. I could have sworn I wrote a unit test that validated that the BuildDisplayAsync was done in parallel for async content part drivers, but I can't seem to located it. Performance on that was what you'd expect; slightly more than the max single driver render time, rather than the sum of all.

aonweb:
Sent an alternate pull request against dev with dependency on AsyncBridge removed.

Issue #5109, opened 2015-03-19
[1.9-int] Email field shouldn't specify type in design mode


orchardbot:
@Xeevis created:
https://orchard.codeplex.com/workitem/21280

I'd like to create contact form using dynamic forms. With that I'd like to use email field to validate user input email address. To make user life simpler, I'd also like to pre-fill this field with {User.Email} token (fills for authenticated, empty for unauthenticated). However I won't let me publish page with this form because {User.Email} is not a valid email address.

Since this is only client-side and form fields don't respect required, length or any other requirements while in design mode, I don't think email element should validate this way either.

It should be enough to remove from EmailField.Design.cshtml this line
tagBuilder.Attributes["type"] = "email";

PR #5321, opened 2015-05-24
#5109: Email field shouldn't specify email type in design mode

Xeevis:
When `email` type is specified on an input it triggers browser validation on that field which prevents administrator from inserting tokens like `{User.Email}`.


sebastienros:
Is this element to be used in the back end or the front end? Dynamic forms elements are supposed to be rendered in the front end, but tokens only in the back end. Also, can it be that the actual solution is to provide an option for that? Let's ask @sfmskywalker ...

Xeevis:
I'm pretty sure `EmailField.Design.cshtml` is only for the backend and it's not doing any server side validation either, it's template that renders in the layout editor.

For frontend it's using `EmailField.cshtml` where email type is set and should definitely remain so.

sfmskywalker:
What Liam said.

PR #5338, opened 2015-05-31
Provide full table name prefix for migrations

qt1:
Sometimes when performing a migration there is work to be done on the database tables.
This PR adds explicit calculation of table name with proper Orchard prefix.


sebastienros:
At the same time doing some SQL in the migration should be used in the last resort.

PR #5344, opened 2015-06-02
Added logging for invalid recipe files

gcsuk:
Currently, the recipe parser throws any recipe errors to the user, causing full page error on the tenant setup screen or ‘Recipes’ tab in the admin area. It would be beneficial for these errors to be trapped and logged, so that users can still work with valid recipes.

If the RecipeParser validated the recipe text, it could return a null for any invalid files. Any null recipes returned by RecipeParser could then be logged, along with the file name. No error would occur for the user.

The manner in which the user is informed is flexible. I would propose a message on-screen informing them of the problem file or files, as a logging-only solution could mask problems.

There are no breaking changes involved in the implementation of the proposed amendments.

While implementing the change, I noticed that when Recipes.cshtml iterates through the Modules list in the ViewModel, it causes deferred execution of a query. This calls RecipeHarvester for a second time, per item in the collection. I have added .ToList() on the initial load to fix this issue - it will prevent the log from being written twice, and also halve the time taken to load & parse recipe files.

### DIFF
--- a/src/Orchard.Tests.Modules/Recipes/Services/RecipeManagerTests.cs
+++ b/src/Orchard.Tests.Modules/Recipes/Services/RecipeManagerTests.cs
@@ -136,8 +136,8 @@ namespace Orchard.Tests.Modules.Recipes.Services {
         }
 
         [Test]
-        public void ParseRecipeReturnsNullOnInvalidXml() {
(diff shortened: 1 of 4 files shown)


sebastienros:
- Throw errors still, to be able to log the exact message
- Please fix formatting (K&R)

Issue #5348, opened 2015-06-04
Latest 1.9.x branch: user.cshtml throws MissingMethodException for ItemDisplayText


netwavebe:
I tried to upgrade a local website (1.9-int branch) to the latest 1.9.x. branch.

I can't run the website at all with these changes. user.cshtml throws an exception at startup:

*System.MissingMethodException: Methode niet gevonden: System.Web.Mvc.MvcHtmlString Orchard.Mvc.Html.ContentItemExtensions.ItemDisplayText(System.Web.Mvc.HtmlHelper, Orchard.ContentManagement.IContent).*

About a week ago, this was merged to the 1.9.x branch:

https://github.com/OrchardCMS/Orchard/pull/5313/files

As you can see, the signature of ItemDisplayText changed from:
*public static MvcHtmlString ItemDisplayText(this HtmlHelper html, IContent content)*
to
*public static MvcHtmlString ItemDisplayText(this HtmlHelper html, IContent content, bool encode = true)*

The call in user.cshtml is:
*@Html.ItemDisplayText(WorkContext.CurrentUser)*

Since 'encode' is an optional parameter, I'm not sure why this isn't working... But it doesn't (or at least not on my machine).

When optimizeCompilations is set to 'false' in web.config, it runs fine. Does razor has issues with optional parameters?

thekaveman:
I can confirm that I'm seeing this error message in my logs on a 1.9.x site.

sebastienros:
Please try again.

thekaveman:
That commit was wrong - you forgot to pass the `HtmlHelper` along. 

```c#
public static MvcHtmlString ItemDisplayText(this HtmlHelper html, IContent content) {
    return ItemDisplayText(html, content, true);
}
```

thekaveman:
New PR #5351 5351

PR #5351, opened 2015-06-04
fixing split extension method for #5348

thekaveman:
c337948 is causing a compiler error in 1.9.x


Skrypt:
faster than the pitcher :)

thekaveman:
:smile:

netwavebe:
Yes, this should fix the problem I mentioned in https://github.com/OrchardCMS/Orchard/issues/5348. Please merge!

Thanks for the follow-up, Kegan!

thekaveman:
Fixed in another commit from @BenedekFarkas

PR #5357, opened 2015-06-07
Read additional module and them folders from web.config appSettings

### LINKED ISSUE #5019: Ability to load extensions from different folder
@Piedone created:
https://orchard.codeplex.com/workitem/21190

It would be great to be able to load modules and themes from other folders than just the default Modules and Themes folder.

There is partial support for this today (IExtensionFolders) but the strings "~/Modules" and "~/Themes" are hard-coded in many places that prevent you from e.g. using templates from modules stored in a different folder than /Modules.

Being able to use arbitrary extension folders (defined in e.g. Web.config, HostComponents.config) would be a step forward. Still, since it's a web app operating with virtual paths you'd have to place your extensions under Orchard.Web (though you could do that into new folders).

Particularly Drupal lets you place your own extensions under a different folder than the built-in ones: https://www.drupal.org/node/2116781

Also see: https://orchard.codeplex.com/discussions/570620

qt1:
Should resolve  #5019,  #5241,  #5354 

Additional comma separated lists can be defined in web.config (example included).
Default is added to the list in any case.
Includes trimming, should ignore empty strings and duplicates.

### DIFF
--- a/src/Orchard.Web/Web.config
+++ b/src/Orchard.Web/Web.config
@@ -15,6 +15,11 @@
         <add key="webpages:Version" value="3.0.0.0"/>
         <add key="log4net.Config" value="Config\log4net.config"/>
         <add key="owin:AppStartup" value="Orchard.Owin.Startup, Orchard.Framework"/>
+    
+<!-- optional additional comma-separated list of folders for Modules or Themes
+        <add key="Modules" value="~/Modules/MyAdditionalModulesFolder, ~/MoreModules"/>
+        <add key="Themes" value="~/Themes/MoreThemes" />
+-->
     </appSettings>
     
     <system.web.webPages.razor>
--- a/src/Orchard/Environment/OrchardStarter.cs
+++ b/src/Orchard/Environment/OrchardStarter.cs
@@ -35,6 +35,8 @@ using Orchard.Mvc.ViewEngines.ThemeAwareness;
 using Orchard.Services;
 using Orchard.WebApi;
 using Orchard.WebApi.Filters;
+using System.Linq;
+using System.Web.Configuration;
 
 namespace Orchard.Environment {
     public static class OrchardStarter {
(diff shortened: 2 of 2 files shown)


Ermesx:
When I read issue #5241 I realize that there is another place where custom folders should be added. I mean RazorViewEngineProvider.cs. There is code which harvest views in modules and if you will not change it I think views in your modules will not work in custom locations.

Look here: https://github.com/OrchardCMS/Orchard/compare/OrchardCMS:dev...paynecrl97:5241/extension_locations?expand=1
This approach looks better because use HostConfiguration and also makes razor engine correct. 
But it need cleanup and more claryfication because e.i. I don't know why exists ModuleFolder class and I think something is missing but maybe I don't understand something :)

qt1:
Hi Ermesx, the harvester is responsible for correct harvesting. IMHO if the harvester is no doing its job properly than a separate issue should be opened and fixed.
So far I didn't notice any problem (but I did not go into edge-cases).

In any case the PR should be safe for existing code.

As for the definitions, the main reason for locating the list of extensions in the web.config is usability.
Anyone that is familiar with adding definitions to web.config can add folders. An alternative such as editing host.config requires understanding of autofac and the internals of the implementation of modules and harvester. i think that whenever possible definition (web.config) should be preferred over implementation (autofac magic).

paynecrl97:
@Ermesx: Author of that change here.

The `ModuleFolder` class was required because the current class (`ModuleFolders`) expects a collection of folder paths, and Autofac's xml config does not support the ability to pass collections as ctor arguments.

@qt1: You definitely will need to edit `RazorViewEngineProvider` similar to the way I have. This is because the locations that it checks for view files is currently hardcoded (ie `~/Themes/` and `~/Modules`).

paynecrl97:
Additionally- the changes found at https://github.com/OrchardCMS/Orchard/compare/OrchardCMS:dev...paynecrl97:5241/extension_locations?expand=1 were just a prototype.

If the community feels like those changes would be welcomed, I'd happily tidy them up and finalise them ready for a PR.

sebastienros:
Chris' changes are closer to the solution, still not complete (ask @Piedone he seems to know why). Zoltan will surely implement it today or tomorrow.

PR #5358, opened 2015-06-07
Feature: PostgreSQL Support

oldrev:
This commit added the PostgreSQL support to Orchard.


Jetski5822:
Cool! - What is the license for the third party libs?

oldrev:
@Jetski5822 
Thanks for the quick reply, there are two third-party libs:

1. Npgsql is licensed under a BSD-like license: 
https://github.com/oldrev/Orchard/blob/feature/postgresql/lib/npgsql/LICENSE.txt

2. Mono.Security.dll is a part of Mono's classes libs, wich are licensed under MIT/X11 license, see also: http://www.mono-project.com/docs/faq/licensing

Jetski5822:
@bleroy Are those licenses compatible with Orchard's?

Issue #5356, opened 2015-06-05
Suggestion: abstract layer evaluation away from the WidgetFilter


paynecrl97:
Currently, the `WidgetFilter` contains the logic to:

 1. Get all the layers from the Content Manager
 1. Iterate through the collection and evaluate via `_ruleManager.Matches`
 1. Build a collection of active layer IDs

I'd like to abstract this logic into a new service (maybe `ILayerEvaluationService`). I believe this would bring several benefits:

 1. Separation of concerns re layer evaluation and widget rendering
 1. Allow for overriding of the default logic when evaluating layers without the need to override the rest of the `WidgetFilter`. Example occasions where this would be useful:
   *  To create an implementation where layer rules are cached in order to bypass the expensive calls to the Content Manager to get each layer
   * To create an implementation where layer evaluation can be logged to Glimpse
   * To create an implementation where each layer is only evaluated once per request (for example we may be interested in which layers are active in both the `WidgetFilter` and when creating a cache key for the Output Cache filter)

I believe this would also simplify the `WidgetFilter`.

Ermesx:
:+1: 
In my Orchard application I have plenty of layers (different menus, many sections of pages where widgets are different, cultures, and so on). When I profiled request execution pipeline then I saw that the 90% calls to data base are retrieving layers data.
Caching layers is good step because of two reasons:
1. If you have enormous site with many widgets there are too many calls to data base in every request.
2. The layers "configuration" is changed relatively rarely
I fully support this improvement. :)

paynecrl97:
Layers is a major factor in the performance of our solution too.

I have already built implementations of all of the things mentioned above, but they are currently built into the widget filter. This work should clean this up a bit.

PR #5363, opened 2015-06-08
Abstract active layer evaluation away from the Widget Filter

paynecrl97:
As it is, the `WidgetFilter` is very bloated.

This pull requests adds no new functionality, but instead abstracts the layer evaluation logic away from the `WidgetFilter`.

This allows developers to override the default functionality for retrieving/ evaluating layers during a request, without the need to override the entire widget filter. 

This would be useful in the following cases:
* To create an implementation that logged to Glimpse
* To create an implementation that took the layer rules from a cache (a major performance gain on a site with several layers)
* To create an implementation that would evaluate each layer only once per request, no matter how many times the call the get the active layers was made (for example to create an output cache key etc...)

Work Item: https://github.com/OrchardCMS/Orchard/issues/5356

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
@@ -21,6 +21,12 @@ namespace Orchard.Widgets.Services{
         public ILogger Logger { get; set; }
         public Localizer T { get; private set; }
 
+        /// <summary>
+        /// Retrieves every Layer from the Content Manager and evaluates each one.
+        /// </summary>
+        /// <returns>
+        /// A collection of integers that represents the Ids of each active Layer
+        /// </returns>
         public IEnumerable<int> GetActiveLayerIds(){
             // Once the Rule Engine is done:
             // Get Layers and filter by zone and rule
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
@@ -21,7 +21,7 @@ namespace Orchard.Widgets.Services{
         public ILogger Logger { get; set; }
         public Localizer T { get; private set; }
 
-        public IEnumerable<int> GetActiveLayerIds()        {
+        public IEnumerable<int> GetActiveLayerIds(){
             // Once the Rule Engine is done:
             // Get Layers and filter by zone and rule
             // NOTE: .ForType("Layer") is faster than .Query<LayerPart, LayerPartRecord>()
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
@@ -4,11 +4,14 @@ using Orchard.Localization;
 using Orchard.Logging;
 using Orchard.Widgets.Models;
 using Orchard.ContentManagement;
+using Orchard.Core.Common.Utilities;
 
 namespace Orchard.Widgets.Services{
     public class DefaultLayerEvaluationService : ILayerEvaluationService {
         private readonly IRuleManager _ruleManager;
         private readonly IOrchardServices _orchardServices;
+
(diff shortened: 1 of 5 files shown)


gcsuk:
+1

Ermesx:
:+1: X :100:

src/Orchard.Web/Modules/Orchard.Widgets/Services/ILayerEvaluationService.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Collections.Generic;
+
+namespace Orchard.Widgets.Services
```
Ermesx: Good idea would be merge interface definition with default implementation in single file as most services in Orchard.

src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Retrieves every Layer from the Content Manager and evaluates each one.
+        /// </summary>
+        /// <returns>
+        /// A collection of integers that represents the Ids of each active Layer
+        /// </returns>
+        public IEnumerable<int> GetActiveLayerIds(){
+            // Once the Rule Engine is done:
+            // Get Layers and filter by zone and rule
+            // NOTE: .ForType("Layer") is faster than .Query<LayerPart, LayerPartRecord>()
+            var activeLayers = _orchardServices.ContentManager.Query<LayerPart>().ForType("Layer").List();
```
Jetski5822: It might be better to put this in to Orchard cache to avoid getting the Layers on every request, and invalidate when the layers are changed. (might be for an improvement later)

Jetski5822:
I like the change

src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
paynecrl97: I purposely have not placed the layers in `ICacheManager` here because deployments into multi node instances will require a distributed cache. If I used a distributed cache (`ICacheService`), then that would make `Orchard.Widgets` become dependant on `Orchard.Caching`, which I think we should avoid.

That said, I have created an implementation of this service in our internal repo that takes the layers from `ICacheService`, and we are seeing relatively large performance gains (especially on sites with many layers). I'd happily contribute this back, but I think it'd be best to create another PR once this one has been merged.

src/Orchard.Web/Modules/Orchard.Widgets/Services/ILayerEvaluationService.cs
paynecrl97: Hmmmm. None of the other interfaces in the `Orchard.Widgets` module contain their default implementation in the same file.

Plus I always like the idea of a single file containing a single class/ interface.

gcsuk: I agree with @paynecrl97 that it is best practice to keep them separate.

paynecrl97:
Question- if the call to `GetActiveLayerIds` is called multiple times during a single request, would it be acceptable to assume that the same Layer IDs would be returned each time?

If so, I should probably update this PR to ensure that the evaluation logic can only be executed once per request.

The only examples I could think of where this would not be the case would be if someone had wrote an `IRuleProvider` that would only return `true` the first time it was executed per request, or an implementation that would differ based on the value of a small measure of time (eg ticks etc...).

Would it be acceptable to make this proposed change to the PR given the caveat above?

src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </returns>
+        public IEnumerable<int> GetActiveLayerIds() {
+            return _activeLayerIDs.Value;
+        }
+
+        private IEnumerable<int> PopulateActiveLayers() {
+            // Once the Rule Engine is done:
+            // Get Layers and filter by zone and rule
+            // NOTE: .ForType("Layer") is faster than .Query<LayerPart, LayerPartRecord>()
+            var activeLayers = _orchardServices.ContentManager.Query<LayerPart>().ForType("Layer").List();
```
sebastienros: TIP: Don't assume a query is faster than another, unless you have tried with a database containing millions of content items.

PR #5371, opened 2015-06-09
Improved Recipe Journal Logging Detail

gcsuk:
Currently, the Recipe Journal reports that a step has begun, and if an error occurs it logs the error message. It does not provide a breakdown of the location in the Step where the problem occurred, which would be useful information when investigating issues.

As an example, the CommandRecipeHandler loops through all commands in the Recipe file and executes them. If a command fails it would be valuable to know which commands succeeded, in order to more easily identify the problem item.

In order to facilitate improved logging, I propose that a string property representing the Execution ID be added to the RecipeContext model object, so that it is available to all RecipeHandlers. Also, IRecipeJournal would be passed into the constructor of each recipe handler, thereby making the WriteJournalEntry method available to the respective classes.

The proposed change would require no breaking changes. There would be a single property added to a POCO class, and some logging code in each handler. I believe that the benefit of having a trace of Commands/Widgets/etc that have been successfully processed makes this a worthwhile undertaking, and our team have already found this to be of value.


src/Orchard.Web/Modules/Orchard.Recipes/RecipeHandlers/CommandRecipeHandler.cs
```diff
@@ -44,6 +46,9 @@ public class CommandRecipeHandler : IRecipeHandler {
 
             foreach (var command in commands) {
                 if (!String.IsNullOrEmpty(command)) {
+                    if (!String.IsNullOrEmpty(recipeContext.ExecutionId)) {
+                        _recipeJournal.WriteJournalEntry(recipeContext.ExecutionId, string.Format("Commands: Executing item {0}.", command));
```
Jetski5822: instead of string.Format, can you use T

src/Orchard.Web/Modules/Orchard.Recipes/RecipeHandlers/CommandRecipeHandler.cs
```diff
@@ -13,9 +13,11 @@ namespace Orchard.Recipes.RecipeHandlers  {
     public class CommandRecipeHandler : IRecipeHandler {
         private readonly ICommandManager _commandManager;
         private readonly CommandParser _commandParser;
+        private readonly IRecipeJournal _recipeJournal;
 
-        public CommandRecipeHandler (ICommandManager commandManager) {
+        public CommandRecipeHandler(ICommandManager commandManager, IRecipeJournal recipeJournal) {
             _commandManager = commandManager;
+            _recipeJournal = recipeJournal;
```
PaulDevenney: do we need to test for null?

gcsuk: I do not believe it is necessary as it is always instantiated and no other parameter is checked. Open to suggestion from collaborators.

paynecrl97:
This would be a pretty good addition. The amount of hours I've lost trying to diagnose issues with recipes makes me sad :cry:

PR #5392, opened 2015-06-17
Abstract active layer evaluation away from the Widget Filter

paynecrl97:
This PR was originally https://github.com/OrchardCMS/Orchard/pull/5363, but I created this new PR in order to merge into `dev` as opposed to `1.9.x`

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Linq;
 using System.Web.Mvc;
 using Orchard.ContentManagement;
 using Orchard.ContentManagement.Aspects;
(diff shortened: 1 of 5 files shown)


src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            Logger = NullLogger.Instance;
+            T = NullLocalizer.Instance;
+
+            _activeLayerIDs = new LazyField<IEnumerable<int>>();
+            _activeLayerIDs.Loader(PopulateActiveLayers);
+        }
+
+        public ILogger Logger { get; set; }
+        public Localizer T { get; private set; }
```
DanielStolt: This one should have a public setter.

src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
```diff
@@ -0,0 +1,62 @@
...
+                    if (_ruleManager.Matches(activeLayer.LayerRule)) {
+                        activeLayerIds.Add(activeLayer.ContentItem.Id);
+                    }
+                }
+                catch (Exception e) {
+                    Logger.Warning(e, T("An error occurred during layer evaluation on: {0}", activeLayer.Name).Text);
+                }
+            }
+
+            return activeLayerIds;
```
DanielStolt: ... and instead do the `.ToArray()` here.

src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
```diff
@@ -51,25 +49,7 @@ public class WidgetFilter : FilterProvider, IResultFilter {
...
-                        activeLayerIds.Add(activeLayer.ContentItem.Id);
-                    }
-                }
-                catch(Exception e) {
-                    Logger.Warning(e, T("An error occured during layer evaluation on: {0}", activeLayer.Name).Text);
-                }
-            }
-
-            IEnumerable<WidgetPart> widgetParts = _widgetsService.GetWidgets(layerIds: activeLayerIds.ToArray());
+            var widgetParts = _widgetsService.GetWidgets(_layerEvaluationService.GetActiveLayerIds().ToArray());
```
DanielStolt: Personal preference maybe but I think it's nicer to maintain a pattern across the codebase where consumers of the API to continue to treat the `IEnumerable<T>` as `IEnumerable<T>` and trust the API to return the simplest underlying form possible. So in this case, trust LayerEvaluationService to return an array... [and see below]

src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
paynecrl97: It should. I'm not sure how I've done that :confused:

PR #5399, opened 2015-06-19
Fix Layout element localization

MpDzik:
The ElementFactory does not always initialize the localizer of the
Element objects which causes the elements to end up with the
NullLocalizer.

This solution still has a major issue - the context of the localized
strings is wrong. It should be the source file in which the element
is defined, not "Orchard.Layouts.Services.ElementFactory".


sebastienros:
Not taking it until Sipke changes the element types to be resolved by autofac.

sfmskywalker:
Elements are now resolved by Autofac as of 4ad806af8ebe316fd50665090a4e3d280d853a56.

@MpDzik Can you please update your PR? Thanks!

sebastienros:
Fixed with sipke's changes

PR #5404, opened 2015-06-20
Fix handling localized strings with quotes

MpDzik:


src/Orchard/Localization/Services/DefaultLocalizedStringManager.cs
```diff
@@ -254,20 +254,28 @@ public class DefaultLocalizedStringManager : ILocalizedStringManager {
             }
         }
 
+        private static string TrimQuote(string str) {
+            if (str.StartsWith("\"") && str.EndsWith("\"")) {
```
sebastienros: Can you add a unit test just to be sure what you are fixing? I'd like to see for instance if it would make sense to to support  just one quote at the beginning or the end. Please see `src/Orchard.Tests/Localization`

MpDzik: Good point. I see that the `DefaultLocalizedStringManager` doesn't have any tests and it is coupled with a lot of interfaces, so I'll have to refactor it a little bit. I'll update the PR soon.

PR #5412, opened 2015-06-22
Fixed Azure Storage Assembly binding version

dcinzona:
Azure Storage assembly binding should point to 4.3.0, not 3.1.0


sebastienros:
Can you target it to 1.9.x ?

dcinzona:
done

Issue #5061, opened 2015-03-01
GetImageProfileUrl reprocess images on every restart


orchardbot:
@jersiovic created:
https://orchard.codeplex.com/workitem/21232

Dynamic ImageProfiles reprocess images after every reboot of the Orchard web server. It has a big impact on performance because all the images that use dynamic profiles are reprocessed when they are requested after a reboot.

I'm using Orchard 1.7 but I have included patches I found in commits related with this issues with no luck:
http://orchard.codeplex.com/workitem/20779
https://orchard.codeplex.com/workitem/20177

Taking a look at the code of the ImageProfileManager.cs I saw clearly the problem. I copy here the current 1.x code that also has the problem:
    public string GetImageProfileUrl(string path, string profileName, FilterRecord customFilter, ContentItem contentItem) {

```
    // path is the publicUrl of the media, so it might contain url-encoded chars

    // try to load the processed filename from cache
    var filePath = _fileNameProvider.GetFileName(profileName, System.Web.HttpUtility.UrlDecode(path));
    bool process = false;

    // if the filename is not cached, process it
    if (string.IsNullOrEmpty(filePath)) {
        Logger.Debug("FilePath is null, processing required, profile {0} for image {1}", profileName, path);

        process = true;
    }
```

...

As you can imagine after rebooting the server,the cache is empty so it won't find the resulting image in the cache so it sets process = true forcing to reprocess the image when indeed it could be available with a valid date in the disk.

A workaround for this problem that I have working without problems in production has been the method resulting something like this:
    public string GetImageProfileUrl(string path, string profileName, FilterRecord customFilter, ContentItem contentItem) {

```
    // try to load the processed filename from cache
    var filePath = _fileNameProvider.GetFileName(profileName, path);
    bool process = false;

    if (string.IsNullOrEmpty(filePath))
    {
        var filterContext = new FilterContext { FilePath = _storageProvider.Combine("_Profiles", _storageProvider.Combine(profileName, CreateDefaultFileName(path))) };
        if (_storageProvider.FileExists(filterContext.FilePath))
        {
            _fileNameProvider.UpdateFileName(profileName, System.Web.HttpUtility.UrlDecode(path), filterContext.FilePath);
            filePath = _fileNameProvider.GetFileName(profileName, path);
        }
    }
```

...

I will try to make my first pull request to Orchard ( I work with Plastic SCM that in theory can sync with Git but not sure on which problems I will find).

jersiovic:
[...]

PR #5425, opened 2015-06-25
Issue #5061 GetImageProfileUrl reprocess images obtained through a dynamic profile on every restart

jersiovic:
Image cache is empty after every reboot of the Orchard app. It makes Dynamic ImageProfiles reprocess images after every reboot. It has a big impact on performance because all the images that use dynamic profiles are reprocessed when they are requested after a reboot. When you open a page with a number of images disk usage increases drammatically and the app answer very slow to other requests.

Related with issue https://github.com/OrchardCMS/Orchard/issues/5061

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
@@ -64,6 +64,18 @@ namespace Orchard.MediaProcessing.Services {
             var filePath = _fileNameProvider.GetFileName(profileName, System.Web.HttpUtility.UrlDecode(path));
             bool process = false;
 
+            //after reboot the app cache is empty so we reload the image in the cache if it exists in the _Profiles folder
+            if (string.IsNullOrEmpty(filePath))
+            {
+                var filterContext = new FilterContext { FilePath = _storageProvider.Combine("_Profiles", FormatProfilePath(profileName, System.Web.HttpUtility.UrlDecode(path))) };
+
+                if (_storageProvider.FileExists(filterContext.FilePath))
+                {
+                    _fileNameProvider.UpdateFileName(profileName, System.Web.HttpUtility.UrlDecode(path), filterContext.FilePath);
+                    filePath = _fileNameProvider.GetFileName(profileName, path);
+                }
+            }
+            
             // if the filename is not cached, process it
             if (string.IsNullOrEmpty(filePath)) {
                 Logger.Debug("FilePath is null, processing required, profile {0} for image {1}", profileName, path);
--- a/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
@@ -65,14 +65,12 @@ namespace Orchard.MediaProcessing.Services {
             bool process = false;
 
             //after reboot the app cache is empty so we reload the image in the cache if it exists in the _Profiles folder
-            if (string.IsNullOrEmpty(filePath))
-            {
-                var filterContext = new FilterContext { FilePath = _storageProvider.Combine("_Profiles", FormatProfilePath(profileName, System.Web.HttpUtility.UrlDecode(path))) };
+            if (string.IsNullOrEmpty(filePath)) {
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
```diff
@@ -64,6 +64,18 @@ public class ImageProfileManager : IImageProfileManager {
             var filePath = _fileNameProvider.GetFileName(profileName, System.Web.HttpUtility.UrlDecode(path));
             bool process = false;
 
+            //after reboot the app cache is empty so we reload the image in the cache if it exists in the _Profiles folder
+            if (string.IsNullOrEmpty(filePath))
+            {
+                var filterContext = new FilterContext { FilePath = _storageProvider.Combine("_Profiles", FormatProfilePath(profileName, System.Web.HttpUtility.UrlDecode(path))) };
```
sebastienros: Why are you creating a filtercontext ?

src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
```diff
@@ -64,6 +64,18 @@ public class ImageProfileManager : IImageProfileManager {
...
 
+            //after reboot the app cache is empty so we reload the image in the cache if it exists in the _Profiles folder
+            if (string.IsNullOrEmpty(filePath))
+            {
+                var filterContext = new FilterContext { FilePath = _storageProvider.Combine("_Profiles", FormatProfilePath(profileName, System.Web.HttpUtility.UrlDecode(path))) };
+
+                if (_storageProvider.FileExists(filterContext.FilePath))
+                {
+                    _fileNameProvider.UpdateFileName(profileName, System.Web.HttpUtility.UrlDecode(path), filterContext.FilePath);
+                    filePath = _fileNameProvider.GetFileName(profileName, path);
```
sebastienros: Then you don't need to read it again, as you just set it.

src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
```diff
@@ -64,6 +64,18 @@ public class ImageProfileManager : IImageProfileManager {
             var filePath = _fileNameProvider.GetFileName(profileName, System.Web.HttpUtility.UrlDecode(path));
             bool process = false;
 
+            //after reboot the app cache is empty so we reload the image in the cache if it exists in the _Profiles folder
+            if (string.IsNullOrEmpty(filePath))
+            {
+                var filterContext = new FilterContext { FilePath = _storageProvider.Combine("_Profiles", FormatProfilePath(profileName, System.Web.HttpUtility.UrlDecode(path))) };
+
+                if (_storageProvider.FileExists(filterContext.FilePath))
+                {
```
sebastienros: Please fix the formatting

src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
```diff
@@ -64,6 +64,18 @@ public class ImageProfileManager : IImageProfileManager {
             var filePath = _fileNameProvider.GetFileName(profileName, System.Web.HttpUtility.UrlDecode(path));
             bool process = false;
 
+            //after reboot the app cache is empty so we reload the image in the cache if it exists in the _Profiles folder
+            if (string.IsNullOrEmpty(filePath))
+            {
```
sebastienros: Formatting

src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
jersiovic: You are right. It was because I was scary of breaking something of a code I didn't know so I copied code from other part without understanding it very well. (Bad decission).

src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
jersiovic: Yes you are right

sebastienros:
Feel free to open another one if it can go in 1.9.x also

PR #5428, opened 2015-06-25
Fix moderator notification

### LINKED ISSUE #4928: Orchard.User notification passes string[] to Orchard.Email
__steve6__ created:
https://orchard.codeplex.com/workitem/21099

Orchard version: 1.8.1

In Orchard.Users.Services.MembershipService, method CreateUser:
If the option of NotifyModeration is true, the service attempts to send an email to a list of recipients.  The problem is that this module is sending the list of recipients as a string[] rather than a comma separated list contained in a single string.

In Orchard.Email.Services.SmtpMessageChannel method Process:
The Email Message only allows the Recipients parameter to be a string and if the parameter is not a string emailMessage.Recipients will be null.  After the creation of the email message the code checks the length of emailMessage.Recipients which can throw a Object Null exception.

Possible solution:
In Orchard.Email.Services.SmtpMessageChannel, Process
Allow Recipients to be a string or a string array and check to see if Recipients is null before using it.  Below is code that I used to fix the issue:
string recipients = null;
if (parameters["Recipients"] is string)
{
    recipients = parameters["Recipients"] as string;
}
else if (parameters["Recipients"] is string[])
{
    var stringArray = parameters["Recipients"] as string[];
    recipients = string.Join(",", stringArray);
}

var emailMessage = new EmailMessage
{
    Body = parameters["Body"] as string,
    Subject = parameters["Subject"] as string,
    Recipients = recipients
};

if (String.IsNullOrEmpty(emailMessage.Recipients))
{
    Logger.Error("Email message doesn't have any recipient");
    return;
}

dalenewman:
`Orchard.Email.Services.SmtpMessageChannel.Process` expects a string for Recipients, not a string array.  This results in an *Object reference not set to an instance of an object.* exception and the moderator(s) do not get notified. Fixes #4928.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Services/MembershipService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Services/MembershipService.cs
@@ -118,7 +118,7 @@ namespace Orchard.Users.Services {
                         var parameters = new Dictionary<string, object> {
                             {"Subject", T("New account").Text},
                             {"Body", _shapeDisplay.Display(template)},
-                            {"Recipients", new [] { recipient.Email }}
+                            {"Recipients", recipient.Email }
                         };
 
                         _messageService.Send("Email", parameters);


sebastienros:
Please target 1.9.x

dalenewman:
Thanks

Issue #5432, opened 2015-06-26
In ShapePlacementStrategy Differentiator is ignored 


jersiovic:
In placement files shapeType uses "__" intead of '-' char for declaring a differenciator. The point is second one is used only in template files as can be seen in documentation http://docs.orchardproject.net/Documentation/Accessing-and-rendering-shapes#NamingShapesandTemplates
However GetShapeType looks for '-' character ignoring placements for a ShapeType with a differentiayor. This produces strange behaviors and a lot of headaches to users because when they set for example: <Place Fields_Common_Text__SeoTitle="/Header:0"/>  in a placement file it forces all Fields_Common_Text shapes within the same Match to be rendered at Header:0 area instead of only affecting those with the SeoTitle differentiator.

PR #5433, opened 2015-06-26
Issue #5432 Differentiator is ignored processing placement

jersiovic:
In placement files shapeType uses "__" intead of '-' char for declaring a differenciator. The point is second one is used only in template files as can be seen in documentation http://docs.orchardproject.net/Documentation/Accessing-and-rendering-shapes#NamingShapesandTemplates
However GetShapeType looks for '-' character ignoring placements for a ShapeType with a differentiator. This produces strange behaviors and a lot of headaches to users because when they set for example: ```<Place Fields_Common_Text__SeoTitle="/Header:0"/>```  in a placement file it forces all Fields_Common_Text shapes within the same Match to be rendered at Header:0 area instead of only affecting those with the SeoTitle differentiator.

Related to this issue https://github.com/OrchardCMS/Orchard/issues/5432

### DIFF
--- a/src/Orchard/DisplayManagement/Descriptors/ShapePlacementStrategy/ShapePlacementParsingStrategy.cs
+++ b/src/Orchard/DisplayManagement/Descriptors/ShapePlacementStrategy/ShapePlacementParsingStrategy.cs
@@ -110,7 +110,6 @@ namespace Orchard.DisplayManagement.Descriptors.ShapePlacementStrategy {
         private void GetShapeType(PlacementShapeLocation shapeLocation, out string shapeType, out string differentiator) {
             differentiator = "";
             shapeType = shapeLocation.ShapeType;
-            var dashIndex = shapeType.LastIndexOf('-');
             var dashIndex = shapeType.LastIndexOf("__");
             if (dashIndex > 0 && dashIndex < shapeType.Length - 2) {
                 differentiator = shapeType.Substring(dashIndex + 2);
--- a/src/Orchard/DisplayManagement/Descriptors/ShapePlacementStrategy/ShapePlacementParsingStrategy.cs
+++ b/src/Orchard/DisplayManagement/Descriptors/ShapePlacementStrategy/ShapePlacementParsingStrategy.cs
@@ -111,8 +111,9 @@ namespace Orchard.DisplayManagement.Descriptors.ShapePlacementStrategy {
             differentiator = "";
             shapeType = shapeLocation.ShapeType;
             var dashIndex = shapeType.LastIndexOf('-');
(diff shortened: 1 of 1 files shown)


src/Orchard/DisplayManagement/Descriptors/ShapePlacementStrategy/ShapePlacementParsingStrategy.cs
```diff
@@ -111,8 +111,9 @@ public class ShapePlacementParsingStrategy : IShapeTableProvider {
             differentiator = "";
             shapeType = shapeLocation.ShapeType;
             var dashIndex = shapeType.LastIndexOf('-');
-            if (dashIndex > 0 && dashIndex < shapeType.Length - 1) {
-                differentiator = shapeType.Substring(dashIndex + 1);
+            var dashIndex = shapeType.LastIndexOf("__");
```
Jetski5822: Will this compile? you have dashIndex specified twice?

            var dashIndex = shapeType.LastIndexOf('-');
            var dashIndex = shapeType.LastIndexOf("__");

jersiovic: You are right, I created the pull request copy pasting my code to the github web editor, but I didn't notice I left original dashIndex declaration in web editor. 
For future pull requests I will configure my PlasticSCM client to connect directly to the git repository to create my pull requests in order to avoid this kind of mistakes..

sebastienros:
Now I understand. Right now a text field would be written using Fields_Common_Text-SeoTitle, and it's used a lot. So this change would break everything.

I suggest you actually get the index of both __ and - and take the biggest one to decide how to extract the differentiator.

sebastienros:
Also please check what the Share Tracing is showing in the alternates.

jersiovic:
I checked Shape Tracing and I found it uses '-' char for alternates.
So, I think that instead of supporting '-' and '__' maybe is better to reject this pull request and instead fix it in orchard documentation to embrace this bug as a feature. http://docs.orchardproject.net/Documentation/Accessing-and-rendering-shapes#NamingShapesandTemplates

sebastienros:
I confirm that we want your fix, but also support for the previous code path. And change the Shape Tracer to suggest using __

jersiovic:
But shape tracing is right. Template files should use '-' is only at placement where documentations says we have to use '__'. So, there is nothing to fix at Shape Tracing, only at ShapePlacementParsingStrategy().

I will make the modification you ask at ShapePlacementParsingStrategy()

jersiovic:
Done

PR #5436, opened 2015-06-26
Fixed [default value] For Mysql

minglei84:
The [default value] can'nt use [modify column] in mysql.


Jetski5822:
Looks cool, but could you fix the formatting? i.e. { on the same line etc.

minglei84:
Thanks for your advice.

PR #5442, opened 2015-06-27
Fixed the items count in media library when modify the item size in css

yw80:
Item shape is square in the media library. If item height in orchard-medialibrary-admin.css modified and smaller than item width, the count of items will be wrong. This work fixed the issue.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library.js
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library.js
@@ -71,11 +71,12 @@ $(function () {
 
         var listWidth = $('#media-library-main-list').width();
         var listHeight = $('#media-library-main-list').height();
-        var itemSize = $('.thumbnail').first().width();
+        var itemWidth = $('.thumbnail').first().width();
+        var itemHeight = $('.thumbnail').first().height();
         var draftText = $("#media-library").data("draft-text");
 
-        var itemsPerRow = Math.floor(listWidth / itemSize);
-        var itemsPerColumn = Math.ceil(listHeight / itemSize);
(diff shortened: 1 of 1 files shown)


yw80:
Media Library can only show file icon. This way is not convenient if someone have a lot of non image files. Maybe someone wants change the item size in orchard-medialibrary-admin.css, like this (only show the different element):
```css
#media-library-main-list li {
    margin-right: 4px;
    margin-bottom: 4px;
    float: left;
    position: relative;
    width: 180px;
    height: 60px;
    border: 4px solid transparent;
}

#media-library-main-list .media-thumbnail {
    width: 58px;
    height: 58px;
    overflow:hidden;
}

.media-library-main-list-overlay {
    position: absolute;
    top: 0;
    bottom: 0;
    left: 65px;
    width: 115px;
    color: 0x808080;
    text-align: left;
    font-size: 1em;
    opacity: 0.6;
}
```
If a folder contains 20 files, originally it can only show 15 files because of the wrong height calculate.


This work fixed the issue.

sebastienros:
I am afraid that mixing small and big squares would feel awkward. There is already an open issue for that which is suggesting to implement two icons on the top right of the window, to toggle between Thumbnails (current) and Details/List view. Could you try that instead ?

yw80:
This work doesn't modify any icon size, bug only corrected the count method. The css code above is only a sample to show the wrong count method. Details/List view is another case which need modify icon size in css and need the right count method. So, the switch view function can base on this work.
The switch view function need more time. I don't know whether I can do it.

sebastienros:
Sorry, too quick.

dcinzona:
I implemented this and got "Infinity" for item count.... which isn't good...

Issue #5424, opened 2015-06-25
Make Output Cache key creation and Response Is Cacheable provider based


paynecrl97:
As mentioned in https://github.com/OrchardCMS/Orchard/issues/5157, it'd be good to allow developers to hook into the Output Cache Filter to allow for custom logic to be executed to:
* Create the cache key
* Decide whether or not the request is cacheable
* Decide whether or not the response is cacheable

This flexibility will allow for a broader use of the Output Cache feature as people who require custom logic such as active layers or user roles to determine the cache key or whether or not the age can be output cached.

PR #5443, opened 2015-06-27
5424 - Adds Output Cache providers to improve extensibility

paynecrl97:
See #5424.

Adds:
* `IRequestIsCacheableProvider` to allow developers to plug in custom logic as to whether or not a request can be cached
* `IResponseIsCacheableProvider` to allow developers to plug in custom logic as to whether or not a response can be cached
* `IOutputCacheKeyCompositeProvider` to allow developers to add custom logic to the creation of the Output Cache key

Most of the logic previously contained in the `OutputCacheFilter` to perform these three tasks has been moved into implementations of their respective providers. This allows developer to suppress and override these providers on a case-by-case basis.

This change will give module developers much more flexibility and extensibility when dealing with the functionality of the Output Cache.


Jetski5822:
@DanielStolt  - this PR need's your keen eye ;)

DanielStolt:
Thanks for the heads-up @Jetski5822 - I'll try to use the keener of the two. ;)

Functionally I'm all for this change - all these tasks should really be externalized from `OutputCacheFilter` and provider based. :+1: 

However, I think this should be implemented differently.

Having 3 separate abstractions for this is "interface overload" IMO. Single responsibility principle sure, love that, but I think the responsibility in this case could and should be defined as "*collaboratively contribute to cache control for the current request*". One reason I think we should collapse this to one abstraction is that, over time, I'd like to add even more colaborative cache control functionality to it (more on that below).

Suggested name for the interface: `ICacheControlProvider`

To help insulate implementations from additions/changes to this interface over time, I think we should do one of the following:

1. Either provide an abstract base class `CacheControlProviderBase` with virtual members which can be overridden in a concrete derived class. This way, an implementation needn't implement the whole interface, just override the members it's interested in.
1. OR provide a fluent API (similar to that of migrations or shape table providers) so that implementations of `ICacheControlProvider` won't need to be updated just because we add more cache control functionality to the mix.

I don't have a strong preference for either one. And I think there are examples of both approaches throughout the Orchard code base.

As for the additional functionality I'd like to see down the road, here's what I'd like to end up with:

- Can current request be served from cache? (You added this)
- Can current response output be stored in cache? (You added this)
- What are the cache key parameters for the current request? (You added this)
- What is the desired cache duration for the output of current request?
- What is the desired cache gracetime for the output of current request?
- *Where* is it OK to cache this output (server, client, proxy, etc, i.e. influence standard HTTP cache control headers)?

Ultimately I'd like to see all of these covered by this abstraction, so that modules can collaboratively have a say in what ends up being the final cache control. The output cache filter should apply approprate logic to determine the "lowest common denominator". For example:

- If **any** implementation says "response is not cachable" it won't get cached.
- The **lowest** cache duration and grace time desired among all implementations determine the values used.
- The most **restrictive** cache control header desired among all implementations determine the rendered cache control header.

Once all this is in place, this will provide the necessary underpinnings for the ultimate improvement: getting rid of configuring these values **per route** (which is almost completely useless) and instead add more granular admin UI allowing configuration of things like duration, grace time and cache storage headers **per thing contributing content to the rendered output**. For example, being able to configure these values per content part, content type, widget, layer, etc. The ultimate cache control for a rendered response would then be based on the appropriate cacheability of the elements of which the response is actually composed, rather than per route.

That last bit is more of a roadmap item of course, but I wanted to start the conversation on these visions now already, as the pieces you are contributing will be laying the groundwork for what is to come, so it's important we get it right.

paynecrl97:
I totally agree with allowing third party modules to dictate the cache duration for the response. In fact, I had already thought of this use case but didn't want too much functionality to be included in a single issue.

As for implementation, I chose three separate interfaces to allow developers to hook into three separate events in the lifecycle of the filter- Can the request be cached? What is the cache key? And can the response be cached? 

This also allows developer to suppress a single implementation without suppressing all the logic. eg someone can suppress `AuthenticatedRequestIsCacheableProvider` without affecting any of the other providers.

That said, I like the idea of a fluent API. If we went down that route, we'd have to be careful to ensure that logic is executed at the correct time. For example- because we check if the request is cacheable in `OnActionExecuting`, if a developer also wanted to dictate whether or not a request is cacheable by adding logic into an `IActionFilter`, we would have to guarantee that their logic would be executed before the `OutputCacheFilter`'s `OnActionExecuting` method is called.

Let me knock up a prototype for the Fluent API over the next few days, and I'll post the link to the diff for discussion.

sebastienros:
I assume we are waiting for some refactoring to only have one interface, right @paynecrl97 ?

sebastienros:
Closing as stale.

PR #5453, opened 2015-06-29
Fix for duplicate version number on creating a new draft

kumards:
The new version should always be contentItemRecord.Versions.Max(x => x.Number) + 1; not latestVersion.Number + 1;

I have a scenario where user can delete the draft content. In this scenario, I set the published version to have Latest=true and the draft version to have Latest=false. Now if I have to create again a new draft(by using ContentManager.Get(id, VersionOptions.DraftRequired);) out of the published content, I end up having two versions of the same content items with the same version number.


sebastienros:
Could you add a unit test for that, in the same set as the existing ones for the Content Manager.
Also, please remove the commented lines. Thanks.

PR #5468, opened 2015-07-02
Redis Module: Added connection string warning to description

dcinzona:
Without the connection string, enabling this module can crash the site.  Added a statement to the description regarding this.


sebastienros:
What about making it not break the website by adding a banner, like we do for SMTP settings? Maybe it's not possible though, haven't checked. This banner could then display what setting is missing.

dcinzona:
Agreed, that was what I was originally hoping for as well, but this seemed like a quick interim solution.

PR #5474, opened 2015-07-05
Add information about super user

### LINKED ISSUE #121: Role checkboxes should not be there when editing the superuser
@bradmi created:
https://orchard.codeplex.com/workitem/16289

REPRO:
1. Manage Users > click to Edit a user

RESULT:
Notice that role checkboxes are not selected according to the roles the user belongs to (e.g. user defined at setup time should be in role Administrators).

Ermesx:
Fix #121 :)

But there is one known issue: 
* When I go to 'Edit User' action (edited superuser) the warning appears after reload page or when I go to another place. 
Any advice ?

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Controllers/AdminController.cs
@@ -103,8 +103,9 @@ namespace Orchard.Users.Controllers {
                 Users = results
                     .Select(x => new UserEntry { User = x.Record })
                     .ToList(),
-                    Options = options,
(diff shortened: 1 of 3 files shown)


sebastienros:
We still want to be able to change roles for the super user account, because you might want to switch it to another account, and keep the roles. 

We are thinking that instead some descriptions could be added on the Settings page to describe what it is, and also in the Roles page for the super as a reminder -> warning notification? 

This way you don't even need to change the View Model, just get the WorkContext from the view.

sebastienros:
https://github.com/OrchardCMS/Orchard/issues/5701

PR #5511, opened 2015-07-14
#5401: Add a NumField DynamicForms Element

### LINKED ISSUE #5401: Dynamic Forms: "Number Field" is needed.
I think a "number field" is needed in Dynamic Forms. Or at least maybe add a regular expression validation to the "Text Field" so it's possible to check that the entered value is a number (regular expression validation is already in the "Password Field").

jtkech:
Fixes #5401

It's a first attempt to add a NumField to the DynamicForms elements, and that needs to be checked because it was not so simple to implement globalization, particularly on client side...

It would be simpler to use an input tag of type number that normally allows, for the same form value, to input a value in different cultures (e.g with Chrome). But this is not well implemented in all browsers (e.g with IE 10)... So, I revert back to an input of type text

Then, because client validation can be enabled and jquery validate isn't culture aware, I needed to add some scripts as globalize.js and cldr.js (Unicode Common Locale Data Repository). But cldr needs many json files in many cultures. So, I've used a dynamic json object with only the necessary data (to parse numbers) grabbed on server side from the current culture

For the validation settings, I also force the user to use the current culture. Not sure it's done in the correct way with this special case of model updating. But maybe, here, it's not necessary to worry about the current culture, let me know...

Note: Beyond my changes, when I built the gulpfile I got this error on client side

    Cannot read property 'adapters' of undefined

In gulpfile.js, I needed to use these script files in this order

    "Scripts/Lib/jquery.validate.unobtrusive.js",
    "Scripts/Lib/jquery.validate.unobtrusive.additional.js",

Best


sebastienros:
@sfmskywalker will work with you on commenting what needs to be changed, or what needs to be added to the module to enable this scenario.

sebastienros:
We should focus on a server side implementation first.

jtkech:
@sebastienros Ok, I understand

If I remember, here there are both server and client implementations. Indeed, client side was more complex, but it was to adapt to a Dynamic form that has the option: `Enable Client Validation`

Best

src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/NumFieldElementDriver.cs
```diff
@@ -0,0 +1,128 @@
...
+using Orchard.Forms.Services;
+using Orchard.Layouts.Framework.Display;
+using Orchard.Layouts.Framework.Drivers;
+using Orchard.Layouts.Helpers;
+using Orchard.Layouts.Services;
+using Orchard.Tokens;
+using DescribeContext = Orchard.Forms.Services.DescribeContext;
+
+namespace Orchard.DynamicForms.Drivers {
+    public class NumFieldElementDriver : FormsElementDriver<NumField>{
```
sfmskywalker: Not very fond of the abbreviation. Maybe use *NumericField* everywhere instead of *NumField*?

sfmskywalker:
I actually like that it supports client side validation as well. Since it's numbers, it makes sense to include the Globalize.js library. @sebastienros will you reconsider, or do you have strong argument against the client libraries and code?

src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/NumFieldElementDriver.cs
jtkech: You're right. Maybe because I had (or I thought I could have) an ambiguous reference with the already existing NumericField class in Orchard.Fields (referenced in Orchard.DynamicForms).

Just seen that seems to be the same with the TextField. So, I think you're right and, if this PR is re-open, I will do this update.

Thanks

bleroy:
Am I the only one to worry about having an element without the corresponding field? Also, it seems like this is also bringing in a large js library. Shouldn't we either use what we already have or make that a reusable resource set?

sfmskywalker:
There's a class called NumField. Or do you mean something else is missing? I thought the same about the JS library, but where would it be hosted? Orchard.jQuery makes most sense currently, except for the fact that it's called jQuery, and Globalize is not a jQuery thing if I'm not mistaken (except for it's jQuery plugin). We did talk several times about introducing a new module containing client libraries that would hold everything we currently use: jQuery, Knockout, etc. I think @danielstolt created an issue for this.

bleroy:
I mean an actual field that you can add to a content type: is there a precedent to having an element that doesn't have an equivalent part and/or field?
About the client library, we're just accumulating more debt. Maybe that other issue should be fixed first? A middleground would be to register those resource from the elements module for now, and move it later.

sfmskywalker:
You mean the `NumericField` in `Orchard.Fields`?

bleroy:
D'oh! Since when do we have that?

sfmskywalker:
:D

Issue #5533, opened 2015-07-17
Extending ProjectionPartDriver and ProjectionManager With part.ContentItem tokens


HermesSbicego-Laser:
I'll try to describe what I want to achieve:
1. I have a Projection with a text field
2. I would like the Query referenced in ProjectionPart to filter Contents by the value stored my projection text field
**but**
1. ProjectionPartDriver (line 113) when it calls projectionManager.GetContentItems doesn't pass part.ContentItem informations
**So the idea is:**
modify line 113 like this
var contentItems = _projectionManager.GetContentItems(query.Id, pager.GetStartIndex() + part.Record.Skip, pager.PageSize, new Dictionary<string, object> { { "Content", part.ContentItem } }).ToList();
full chenages will be there in my next push.

I have a backward compatible solution ready, I will push it soon...
Any

sebastienros:
Thank you, however I don't really like the changes it implies on the code. What you could do is use another existing token which lets you get the currently displayed content item. Then your query could use this token to point the projection, and use a continuation to access its field. Would that work ?

HermesSbicego-Laser:
@sebastienros can you suggest where to find the existing token ({Request.Content} perhaps)? I never used a "continuation"... How should I do that? Can you give me a starting point?
TY

sebastienros:
{Request.Content.Fields.YOURPART.YOURFIELD}

HermesSbicego-Laser:
@sebastienros,  we used proposed solution till now, so, request.content is perfect if i have a request.content. but i cannot use it in other contests, like within widgets for example. Any idea on how to solve it? In general I think that having the ability to pass a dictionary to projectionmanager methods which uses queries should be considered...

sebastienros:
I changed my mind and it actually makes sense based on @HermesSbicego-Laser's feedback.

I suggest to pass the ProjectionPart to the Query api calls.

PR #5534, opened 2015-07-17
Extending ProjectionPartDriver and ProjectionManager With part.Conten…

HermesSbicego-Laser:
…tItem tokens #5533

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/Services/IProjectionManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Services/IProjectionManager.cs
@@ -5,6 +5,7 @@ using Orchard.Projections.Descriptors.Property;
 using Orchard.Projections.Descriptors.Filter;
 using Orchard.Projections.Descriptors.Layout;
 using Orchard.Projections.Descriptors.SortCriterion;
+using System;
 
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.Projections/Services/IProjectionManager.cs
```diff
@@ -17,9 +18,8 @@ public interface IProjectionManager : IDependency {
         SortCriterionDescriptor GetSortCriterion(string category, string type);
         LayoutDescriptor GetLayout(string category, string type);
         PropertyDescriptor GetProperty(string category, string type);
-
-        IEnumerable<ContentItem> GetContentItems(int queryId, int skip = 0, int count = 0);
-        int GetCount(int queryId);
+        IEnumerable<ContentItem> GetContentItems(int queryId, int skip = 0, int count = 0, Dictionary<string, object> tokens = null);
+        int GetCount(int queryId, Dictionary<string, object> tokens = null);
```
sebastienros: Please, no `GetCount` with tokens. Assume it's always `null` if an inner call needs it.

sebastienros:
Closing as the initial issue has a different solution.

PR #5537, opened 2015-07-18
#5512: Incorrect message of DynamicForms StringLength rule if only Maximum is specified

### LINKED ISSUE #5512: Incorrect message of DynamicForms StringLength rule if only Maximum is specified
Can be fixed in StringLength.cs by replacing this

    if (Minimum != null && Maximum != null)
        return T("{0} must be between {1} and {2} characters long.", context.FieldName, Minimum, Maximum);

    return T("{0} must be at least {1} characters long.", context.FieldName, Minimum);

With

    if (Minimum != null && Maximum != null)
        return T("{0} must be between {1} and {2} characters long.", context.FieldName, Minimum, Maximum);
    else if (Minimum != null)
        return T("{0} must be at least {1} characters long.", context.FieldName, Minimum);

    return T("{0} must be at most {1} characters long.", context.FieldName, Maximum);

Best

jtkech:
Fixes #5512


Jetski5822:
@sfmskywalker

PR #5541, opened 2015-07-19
#5538 Urlbuilder Pb in 1.9.X SecureSocketLayer Module

CSurieux:
urlbuilder does not accept urls containing port, so removes port from ctor call, also refactors code for better efficiency work item : 5538

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Services/SecureSocketsLayerService.cs
+++ b/src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Services/SecureSocketsLayerService.cs
@@ -214,11 +214,12 @@ namespace Orchard.SecureSocketsLayer.Services {
             var settings = GetSettings();
             if (settings == null) return path;
             var insecureHostName = settings.InsecureHostName;
-            var builder = new UriBuilder(insecureHostName.Trim('/') + path) {
-                Scheme = Uri.UriSchemeHttp, 
-                Port = 80
+            int port = 80;
+            TryGetPort(insecureHostName, port, out port);
+            var builder = new UriBuilder(insecureHostName.Split(':').First().Trim('/') + path) {
+                Scheme = Uri.UriSchemeHttp,
+                Port = port
             };
-            SetPort(insecureHostName, builder);
             return builder.Uri.ToString();
         }
 
@@ -226,24 +227,28 @@ namespace Orchard.SecureSocketsLayer.Services {
             var settings = GetSettings();
             if (settings == null) return path;
             var secureHostName = settings.SecureHostName;
-            var builder = new UriBuilder(secureHostName.Trim('/') + path) {
+            int port = 443;
+            TryGetPort(secureHostName, port, out port);
+            var builder = new UriBuilder(secureHostName.Split(':').First().Trim('/') + path) {
                 Scheme = Uri.UriSchemeHttps, 
-                Port = 443
+                Port = port
             };
-            SetPort(secureHostName, builder);
             return builder.Uri.ToString();
         }
 
-        private static void SetPort(string hostName, UriBuilder builder) {
-            if (string.IsNullOrWhiteSpace(hostName)) return;
+        private static bool TryGetPort(string hostName, int defaultPort, out int port) {
+            bool ret = false;
+            port = defaultPort;
+            if (string.IsNullOrWhiteSpace(hostName)) return true;
             var splitSecuredHostName = hostName.Split(new[] { ':' }, StringSplitOptions.RemoveEmptyEntries);
             if (splitSecuredHostName.Length == 2) {
-                int port;
                 if (int.TryParse(splitSecuredHostName[1], NumberStyles.Integer, CultureInfo.InvariantCulture,
                     out port)) {
-                    builder.Port = port;
+                    ret = true;
                 }
             }
+            else ret = true;
+            return ret;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Services/SecureSocketsLayerService.cs
```diff
@@ -226,24 +227,28 @@ public class SecureSocketsLayerService : ISecureSocketsLayerService {
...
-                Port = 443
+                Port = port
             };
-            SetPort(secureHostName, builder);
             return builder.Uri.ToString();
         }
 
-        private static void SetPort(string hostName, UriBuilder builder) {
-            if (string.IsNullOrWhiteSpace(hostName)) return;
+        private static bool TryGetPort(string hostName, int defaultPort, out int port) {
```
sebastienros: It's not a TryXXX pattern, it's just a `int GetPort(string hostName)` which returns either 80 or a port if specified.

CSurieux: It also could return 443. Yes. Avoid the SetPort. What is the exact definition of a TryPattern ? Is it something as String/string ?

CSurieux:
Well I changed this name...

sebastienros:
Closing until the code is more professional.

CSurieux:
What is not professional in this code ?

sfmskywalker:
I believe some of your names are too beatiful: `GetPortFomHostNameJoli`?

CSurieux:
Yes I already noticed some clerical near 'sectaire' approach here. Beware :).
But the code is solving an existing bug, what else should we expect from code ?

sfmskywalker:
Oh we beware alright ;)
You're saying you care about fixing bugs but don't care about code quality, proper naming etc?

bleroy:
Just resubmit it with proper naming.

CSurieux:
May be. Let's close this, you are a community. Anybody could solve this bug an write a better code than the existing one. I did it in a way. This is just a facette in the whole thing.

CSurieux:
Back on this.
Orchard is so serious.

I 've just found this in TheThemeMachine layout.cshtml
<div id="layout-tripel" class="group">@* as in beer *@

so you some doors opened for initiates...
:) :) :)

dpomt:
Thanks CSurieux for that fix. I had the issue that SSL Redirect always redirected to some 0.0.x.y page right after upgrade a site from 1.8.x to 1.9.x. 
With this fix SSL works fine for me again!

CSurieux:
You're welcome. :)

dpomt:
Just in case somebody else has this issue:
The fix from CSurieux was not sufficient for me:

Changing from a secured URL, Login, logout and clicking again on an unsecured URL lead to a http://www.somesite.com:443/ call which did not work. 
This just happened on a production site where I could not debug. As ugly workound I Changed MakeInsecure method to set port hard-coded to port 80.


    private string MakeInsecure(string path)
        {
            var settings = GetSettings();
            if (settings == null) return path;
            var insecureHostName = settings.InsecureHostName;
            int port = 80;
            var builder = new UriBuilder(insecureHostName.Split(':').First().Trim('/') + path)
            {
                Scheme = Uri.UriSchemeHttp,
                Port = 80 //GetPortFomHostNameJoli(insecureHostName, port)
            };
            return builder.Uri.ToString();
        }

CSurieux:
This will work when you use Port 80, which is the default implementation.
But with another Binding than 80, you are out.

Have you used the good code for GetPortFomHostNameJoli(insecureHostName, port) ?

 private static int GetPortFomHostNameJoli(string hostName, int defaultPort) {
            int port = defaultPort;
            if (string.IsNullOrWhiteSpace(hostName)) return port;
             var splitSecuredHostName = hostName.Split(new[] { ':' }, StringSplitOptions.RemoveEmptyEntries);
             if (splitSecuredHostName.Length == 2) {
               int.TryParse(splitSecuredHostName[1], NumberStyles.Integer, CultureInfo.InvariantCulture,out port);
             }
            return port;
         }

dpomt:
Yes sure hard-coded port 80 is bad, but for my quickfix for the site good enough.
I have used your GetPortFomHostNameJoli method but it sometimes did not work for me. Don't know why but it just happened.It is not reproducable on my dev machine so at the Moment no Chance to debug it.

jtkech:
@dpomt I don't remember exactly, so not sure it solves all scenarios... Maybe @CSurieux solution is better and needs to be considered...

But, did you try what is already applied to the dev branch (see #5202), but infortunately not applied to 1.9.x...

Best

dpomt:
@jtkech: thanks for the hint. i did not test try that fix but quickly looked at it now. I guess it should work as long as insecure URL runs on port 80? If your insecure port runs on a separate port this will not work?
For now, I will be ok with my fix of  CSurieux+hardcoded port 80.
I have another site on 1.9.x with SSL. I guess I will get back later (January?) on this Topic again.

I am wondering how other guys use SSL with 1.9.x since for me I did not get to work it properly without that fix. 
Some fix should definitely go in 1.9.x

jtkech:
@dpomt not sure because you start with port = 80 but, normally, after the SetPort() will set the right port...

CSurieux:
@dpomt you must input à full URL in thé SSL settings for this to work...

PR #5542, opened 2015-07-20
#5369: Allows simple properties null checking on a dynamic ContentItem or ContentPart

### LINKED ISSUE #5369: To be able to simply test if a property != null on a dynamic ContentItem or ContentPart
Not sure we can do this (needs to be validated) but I've found a way to be able, with a dynamic `ContentItem` / `ContentPart` (e.g got from the `Model` in a razor view) to use the following without throwing an exception

    if (contentItem.SomePart != null)
    if (part.SomeField != null)

By replacing, in `ContentPart.cs` and `ContentItem.cs`, in the overrided `TryGetMember()`

        return false;

With

        result = null;
        return true;

Best

jtkech:
Fixes #5369

Not sure we can do this, but with this changes, with a dynamic ContentItem / ContentPart (e.g got from the Model in a razor view, and not casted to the ContentItem / ContentPart types) I can use the following without throwing an exception

    if (contentItem.SomePart != null)
    if (part.SomeField != null)

Best

### DIFF
--- a/src/Orchard.Tests/ContentManagement/DynamicContentItemTests.cs
+++ b/src/Orchard.Tests/ContentManagement/DynamicContentItemTests.cs
@@ -76,5 +76,39 @@ namespace Orchard.Tests.ContentManagement {
 
             Assert.That((object)testingPartDynamic, Is.AssignableTo<IEnumerable<ContentPart>>());
         }
+
+
+        [Test]
+        public void NullCheckingCanBeDoneOnProperties() {
+            var contentItem = new ContentItem();
+            var contentPart = new ContentPart { TypePartDefinition = new ContentTypePartDefinition(new ContentPartDefinition("FooPart"), new SettingsDictionary()) };
+            var contentField = new ContentField { PartFieldDefinition = new ContentPartFieldDefinition(new ContentFieldDefinition("FooType"), "FooField", new SettingsDictionary()) };
+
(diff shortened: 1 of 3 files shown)


sebastienros:
Could you add a unit test in https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Tests/DisplayManagement/NilTests.cs
or somewhere close if you find a better set of tests.

Thanks

jtkech:
For testing this online patch, I've worked on a quite recent Orchard dev branch. I've run all the hundreds of unit tests and I've got 12 errors. But seems not to be related to my change because I get the same errors with or without my change.

This except my own unit test that needs my change.

If you want, tomorrow I can re-run all the tests on the last 1.9.x branch.

Best

jtkech:
With the last 1.9.x, only 4 errors in Orchard.Profile SpecFlow tests (all NUnit tests was ok)

Due to my environment, in HttpClient.cs, I needed to replace this

    _request = (HttpWebRequest)WebRequest.Create("http://localhost" + url);

With

    _request = (HttpWebRequest)WebRequest.Create("http://localhost:30321/OrchardLocal" + url);

Then, all tests was ok without or with my change and its new NUnit test

Best

jtkech:
@sebastienros
Is it GTG now?

PR #5545, opened 2015-07-20
Adjust the sorting method for taxonomy terms, so that the weight value is smaller in the front

yw80:
Smaller in front is common, such as admin menu.
In the past, the bigger in the front is not convenient, also make trouble for people.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Models/TermPart.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Models/TermPart.cs
@@ -71,7 +71,7 @@ namespace Orchard.Taxonomies.Models {
                 // if two nodes have the same parent, then compare by weight, then by path 
                 // /1/2/3 vs /1/2/4 => 3 vs 4
                 if (x.Path == y.Path) {
-                    var weight = y.Weight.CompareTo(x.Weight);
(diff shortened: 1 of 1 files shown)


sebastienros:
This cannot be done that easily as it would break all existing websites. So this change would require two things;
- we discuss it publicly and agree on the change
- provide some upgrade path to change the weights of all existing terms

yw80:
I think the use of taxonomy is only a few, perhaps because of bugs in the taxonomy, such as the lack of technical name, resulted in many non English users in the use of taxonomy will encounter very serious failure.

Maybe the upgrade process is not complicated, just need to modify the Orchard_Taxonomies_TermPartRecord and Orchard_Framework_ContentItemRecord two tables. 

By the way, I don't know why the Weight value stored in two tables, is it a bit redundant?

yw80:
All other pages are smaller in front, to modify the taxonomy to smaller in front can unified user experience, reducing trouble.
If a larger value in front, then the Weight of the first item how to set it? If it is set to 10, then if terms more than 10, we need to modify all the previous Weight value in order to add a new term, very troublesome.

sebastienros:
I don't disagree with the reason of the change, but lots of websites use it already, and this specific weight property, so it's very risky to change it.

yw80:
ok, there may be a better solution later on.
thank you for your reply.

Issue #5504, opened 2015-07-14
Import creates duplicate homepages


DanielStolt:
When importing a recipe on an existing site that contains an item with alias "/", and the recipe contains an item with that same alias, duplicates are created. For other aliases, it seems this is not the case.

sfmskywalker:
Created a PR here: https://github.com/OrchardCMS/Orchard/pull/5556

sfmskywalker:
Fixed with this PR: https://github.com/OrchardCMS/Orchard/pull/5567

PR #5556, opened 2015-07-21
Issue/homepage alias

sfmskywalker:
This PR changes the way the home page is stored in the system by adding a dedicated alias record that can be updated to point to the route that should be the home page.

This means that the alias of the content item being pointed to by the home alias will not change when making that content item the homepage, which solves an issue described in #5504 as well as simplifies the way that we can set something to be the homepage.

This also handles the case where the user may check a content item to become the home page but saves that item as a draft. Performing this action won't update the home alias record, and yet the updated content item will persist the fact that it is promoted to become the homepage - which will happen as soon as it is published.


src/Orchard.Web/Modules/Orchard.Alias/IAliasService.cs
```diff
@@ -23,7 +23,7 @@ public interface IAliasService : IDependency {
 
         IEnumerable<Tuple<string, RouteValueDictionary>> List();
         IEnumerable<Tuple<string, RouteValueDictionary, string>> List(string sourceStartsWith);
-        IEnumerable<VirtualPathData> LookupVirtualPaths(RouteValueDictionary routeValues, System.Web.HttpContextBase HttpContext);
+        IEnumerable<VirtualPathData> LookupVirtualPaths(RouteValueDictionary routeValues, System.Web.HttpContextBase httpContext);
```
sfmskywalker: Just changed the casing from `HttpContext` to `httpContext`.

src/Orchard.Web/Modules/Orchard.Alias/Implementation/Storage/AliasStorage.cs
```diff
@@ -90,7 +90,7 @@ public class AliasStorage : IAliasStorage {
         }
 
         public void Remove(string path) {
-            Remove(x => x.Path == path && x.Source == path);
+            Remove(x => x.Path == path);
```
sfmskywalker: This fixes the issue where you couldn't delete an alias record from the Alias UI anymore.

PR #5567, opened 2015-07-24
Issue/homepage alias

sfmskywalker:
This fixes issue #5504 and changes the way we store the homepage by creating a dedicated home alias that can point to any route that is supposed to be the home page.


src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -60,55 +57,54 @@ public class AutoroutePartHandler : ContentHandler {
...
-                            _orchardServices.ContentManager.Publish(current.ContentItem);
+            // Should it become the home page?
+            if (part.PromoteToHomePage) {
+                // Get the current homepage an unmark it as the homepage.
+                var currentHomePage = _homeAliasService.GetHomePage(VersionOptions.Latest);
+                if(currentHomePage != null && currentHomePage.Id != part.Id) {
+                    var autoroutePart = currentHomePage.As<AutoroutePart>();
+
+                    if (autoroutePart != null) {
+                        autoroutePart.PromoteToHomePage = false;
```
Jetski5822: Do we need to explicitly set this value here? If we dont, and the default is false, then we dont even need to work with AutoRoutePart, and can work directly with IAliasAspect!

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -60,55 +57,54 @@ public class AutoroutePartHandler : ContentHandler {
...
+            // If the generated alias is empty, compute a new one.
             if (String.IsNullOrWhiteSpace(part.DisplayAlias)) {
                 _autorouteService.Value.ProcessPath(part);
                 _orchardServices.Notifier.Warning(T("The permalink could not be generated, a new slug has been defined: \"{0}\"", part.Path));
                 return;
             }
 
-            // check for permalink conflict, unless we are trying to set the home page
+            // Check for permalink conflict, unless we are trying to set the home page.
             if (part.DisplayAlias != "/") {
```
Jetski5822: Is this rule still correct?

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasExportHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+            // so we can't rely on the route values in that case.
+            if (homePage != null) {
+                var homePageIdentifier = _contentManager.GetItemMetadata(homePage).Identity.ToString();
+                root.Attr("Identifier", homePageIdentifier);
+            }
+            
+            context.Document.Element("Orchard").Add(root);
+        }
+
+        private string Capitalize(string value) {
```
Jetski5822: Is there not an extension method in the StringExtensions for this?

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasImportHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        private readonly IHomeAliasService _homeAliasService;
+        private readonly IContentManager _contentManager;
+
+        public HomeAliasImportHandler(IHomeAliasService homeAliasService, IContentManager contentManager) {
+            _homeAliasService = homeAliasService;
+            _contentManager = contentManager;
+        }
+
+        public void ExecuteRecipeStep(RecipeContext recipeContext) {
+            if (!String.Equals(recipeContext.RecipeStep.Name, "HomeAlias", StringComparison.OrdinalIgnoreCase)) {
```
Jetski5822: lowercase s on String :)

src/Orchard.Web/Modules/Orchard.Autoroute/Services/HomeAliasService.cs
```diff
@@ -0,0 +1,55 @@
...
+                    ? XmlHelper.Parse<int>((string)homePageRoute["id"]) 
+                    : default(int?) 
+                : default(int?);
+
+            return homePageId;
+        }
+
+        public IContent GetHomePage(VersionOptions version = null) {
+            var homePageId = GetHomePageId();
+            var homePage = homePageId != null ? _contentManager.Get(homePageId.Value, version ?? VersionOptions.Published) : default(IContent);
```
Jetski5822: if you pass through null "version ?? VersionOptions.Published" does it default to Published anyways?

Jetski5822:
Were there no changes to the Localization module that needed to be done? i.e. FrontEndCulturePicker.cshtml?

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
sfmskywalker: We do need to reset the precious homepage's `PromoteToHomePage` to `false`, because it is set to `true` by the driver when the user checks the "make this the homepage" checkbox.

sfmskywalker:
Yes you're right, getting the home page can no longer be done by simply querying for the content item where its alias is an empty string. Good catch.

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
sfmskywalker: Actually, we should not even check for the content being the homepage, since we should always check if the alias is in conflict with an existing one. A content item no longer is the home page if its display alias is "", but simply by virtue of the "/" alias record pointing to the content item. Another good catch.

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasExportHandler.cs
sfmskywalker: That's what I thought, but couldn't find any.

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasImportHandler.cs
sfmskywalker: Actually this is till something we need to decide upon, don't we?

sfmskywalker: Do we have a coding convention guide somewhere that we can update when there is consensus?

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasExportHandler.cs
Jetski5822: Could you not just use CamelFriendly?

src/Orchard.Web/Modules/Orchard.Autoroute/Services/HomeAliasService.cs
sfmskywalker: That's what I thought too, but when drilling down into the `Get` method, eventually the `options` argument is accessed without any null checks. There is an overload of `Get` that passes in `VersionOptions.Published`, so 'Published' is the correct default.

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasImportHandler.cs
DanielStolt: Lowercase `string` when specifying a variable type, yes. But not when referring to the `String` class, as is the case here. The following code is correct IMO:

    string x = String.Format("{0}", number);

The first is typing a variable, using a type keyword. The second is referring to a static method on a class. It's very weird IMO to use lowercase in the latter case - you're not calling a static method on a type keyword.

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasExportHandler.cs
sfmskywalker: Nope, `CamelFriendly` simply inserts spaces before capitals.

src/Orchard.Web/Modules/Orchard.Autoroute/Services/HomeAliasService.cs
```diff
@@ -0,0 +1,55 @@
...
+        public RouteValueDictionary GetHomeRoute() {
+            return _aliasService.Get(HomeAlias);
+        }
+
+        public int? GetHomePageId() {
+            var homePageRoute = GetHomeRoute();
+            var homePageId = 
+                homePageRoute != null 
+                ? homePageRoute.ContainsKey("id") 
+                    ? XmlHelper.Parse<int>((string)homePageRoute["id"])
```
Ermesx: Why not ? :)
```c#
int.Parse((string)homePageRoute["id"])
```
Is there any reason ?

sfmskywalker: You mean instead of `XmlHelper.Parse<int>`? No particular reason other than `XmlHelper` performing a nice null check in case the value of that dictionary entry is null (although granted, that's extremely unlikely to happen) and it passes in the invariant culture for me. Using `int.Parse` would be just as fine I suppose.

src/Orchard.Web/Modules/Orchard.Autoroute/ImportExport/HomeAliasImportHandler.cs
Jetski5822: Worth taking a look at this. https://github.com/aspnet/Home/wiki/Engineering-guidelines

sfmskywalker: I think it's pretty clear that opinions on this matter differ. What we need to do is make a decision and stick with it, Personally I am in favor of using the .NET type when invoking static members, but I am even more for sticking with coding guidelines established by the ASP.NET team, as it is likely that others will follow that as well, resulting in a more global standard, which is a good thing. Let's discuss this on tomorrow's call and create a poll or something, then make a final decision.

PR #5568, opened 2015-07-24
Token chaining improvements

TFleury:


sebastienros:
Can you add some unit tests? At least so that we can understand what the change is enabling. Thanks.

TFleury:
I Added tests for existing parameterized token method (`Token(Func<string, string> filter, Func<string, TData, object> tokenValue)`) which was not explicitly tested yet (but implicitly tested through `DateTokens` in `TestChainedTokens`), and for the newly created parameterized chain method.

PR #5571, opened 2015-07-25
Fixed #5140 Recipe from Migrations didn't work

### LINKED ISSUE #5140: After upgrade from Orchard 1.7.1 to Orchard 1.8.1 execute a Recipe from Migrations doesn't work
@jersiovic created:
https://orchard.codeplex.com/workitem/21311

In Orchard 1.7.1 this code works without problems when it is executed within Migrations for running a Module Recipe
    Recipe recipe = _recipeHarvester.HarvestRecipes("PCCom.Plantilla").FirstOrDefault(x => x.Name == "Plantilla Update");

            if (recipe != null)
                _recipeManager.Execute(recipe);                          

However after upgrade to Orchard 1.8.1 same code in the same migrations class doesn't execute the recipe. 
One clue is that I'm still able to execute the Recipe through the Admin Dashboard using the Recipes option within Modules section.
Debugging I have seen problem happens with DefaultProcessingEngine class. Its AddTask method is called and it properly adds an entry to the list of entries to execute. I checked after it is called that _processingEngine.AreTasksPending()==true.
However when this DefaulOrchardHost.EndRequest() method is executed _processingEngine.AreTasksPending() is always false, so _processingEngine.ExecuteNextTask() is never called.
When I debug running the opertation through the Admin Dashboard it works properly: _processingEngine.AreTasksPending() is  true, so _processingEngine.ExecuteNextTask() is called.

jersiovic:
Fixed #5140 Recipe from Migrations didn't work 
Please review the related issue https://github.com/OrchardCMS/Orchard/issues/5140 to have an explanation of the reason why I have chosen this solution.

### DIFF
--- a/src/Orchard/Environment/DefaultOrchardHost.cs
+++ b/src/Orchard/Environment/DefaultOrchardHost.cs
@@ -143,6 +143,10 @@ namespace Orchard.Environment {
                     catch (Exception e) {
                         Logger.Error(e, "A tenant could not be started: " + settings.Name);
                     }
+                    while (_processingEngine.AreTasksPending()) {
+                        Logger.Debug("Processing pending task after activate Shell");
+                        _processingEngine.ExecuteNextTask();
+                    }
                 });
             }
             // no settings, run the Setup


sebastienros:
The solution is the same code, but the in the background task runner. It should be after the loop running all the background tasks.

jersiovic:
Did you mean somthing like this?

```
public class BackgroundService : IBackgroundService {
        private readonly IEnumerable<IBackgroundTask> _tasks;
        private readonly ITransactionManager _transactionManager;
        private readonly string _shellName;
        private readonly IProcessingEngine _processingEngine;

        public BackgroundService(
            IEnumerable<IBackgroundTask> tasks,
            ITransactionManager transactionManager,
            ShellSettings shellSettings,
            IProcessingEngine processingEngine) {

            _tasks = tasks;
            _transactionManager = transactionManager;
            _shellName = shellSettings.Name;
            Logger = NullLogger.Instance;
            _processingEngine = processingEngine;
        }

        public void Sweep() {
            foreach (var task in _tasks) {
                var taskName = task.GetType().FullName;

                try {
                    Logger.Information("Start processing background task \"{0}\" on tenant \"{1}\".", taskName, _shellName);
                    _transactionManager.RequireNew();
                    task.Sweep();
                    Logger.Information("Finished processing background task \"{0}\" on tenant \"{1}\".", taskName, _shellName);
                }
                catch (Exception e) {
                    _transactionManager.Cancel();
                    Logger.Error(e, "Error while processing background task \"{0}\" on tenant \"{1}\".", taskName, _shellName);
                }
            }

            while (_processingEngine.AreTasksPending())
            {
                Logger.Debug("Processing pending task");
                _processingEngine.ExecuteNextTask();
            }
        }
```

This doesn't wortk because this solutions suffers the same problem I explained in the issue https://github.com/OrchardCMS/Orchard/issues/5140

DefaultProcessingEngine uses ContextState to store pending tasks. Problem is the the ContextState implementation when HttpContext.Current == null is based on CallContext.GetData() method. Which stores data at thread scope.
So, the only way of finding an enqueued task pending of execute in this circumstances is to use the same thread which enqueued it for askig for it calling _processingEngine.AreTasksPending(). That's why the only way I found  to execute the recipe was to add it to the code in the lambda of the Parallel.ForEach() execute in the DefaulOrchardHost.CreateAndActivateShells() method. Because there is where the thread who calls migrations is created and destroyed.

This problem doesn't happen when you launch a recipe from the UI because this request has HttpContext.Current != null so ContextState doesn't rely on the thread to store pending Tasks because it uses HttpContext to store pending recipes, so in that way they can be found by other threads based on their HttpContext.

What I have proposed is a patch, a good solutions should store pending task when HttpContext.Current ==null in another place non dependent from the thread. ¿Db?

sebastienros:
Might also be related to ce630f980c7a79ae17b36598f3834c4ffd523739

sebastienros:
Sorry, I thought the PR was about recipes in background tasks, not migrations.

Issue #5572, opened 2015-07-25
Bug fixes for Google ReCaptcha 2.0


emeraldarcher:
Bug fixes for Google ReCaptcha 2.0 merged into dev branch

PR #5573, opened 2015-07-25
Issue #5572 - Bug fixes for Google ReCaptcha 2.0 update

emeraldarcher:
This contains the remaining bug fixes for the update to Google ReCaptcha in Orchard. The update was merged in issues 5331 and 5332.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.DynamicForms/Views/Elements/ReCaptcha.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.DynamicForms/Views/Elements/ReCaptcha.cshtml
@@ -5,7 +5,7 @@
     var element = (ReCaptcha)Model.Element;
     var publicKey = (string)Model.PublicKey;
 }
-@if (!String.IsNullOrWhiteSpace(publicKey)) {
+@if (HasText(publicKey)) {
     @tagBuilder.StartElement
     <script src='https://www.google.com/recaptcha/api.js'></script>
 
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.DynamicForms/Views/Elements/ReCaptcha.cshtml
```diff
@@ -5,7 +5,7 @@
     var element = (ReCaptcha)Model.Element;
     var publicKey = (string)Model.PublicKey;
 }
-@if (!String.IsNullOrWhiteSpace(publicKey)) {
+@if (HasText(publicKey)) {
```
sfmskywalker: I wonder why you're using `HasText` on strings? Isn't that intended to be used on objects other than strings?

emeraldarcher: I was told to change from IsNullOrWhitespace() to HasText() by Jetski5822.
I can change it back though if you'd like.

sfmskywalker: I see. @Jetski5822 what's the thinking behind `HasText`, do you know?

sebastienros: We talked about it during the triage, and I think I commented on the "request" from Nick. @sfmskywalker  is right, you can just use `String.IsNullOrEmptyWhitespace` here. But `HasText` begin technically correct we didn't bother accepting the PR.

Jetski5822: Huh?

sebastienros:
Feel free to create another PR to remove the HasText

PR #5574, opened 2015-07-26
ContainerPart renders ContainableParts using customizable display-mode

siyamandayubi:
Currently, ContainerPart renders its ContainableParts in "Summary" mode and it uses the default "ul", "li" tags to render them. In some scenarios, we need to render Containable items in Detail mode and also using other tags instead of ul-li. 

Using some new properties in the ContainerPartSettings , I make it customizable.

Summary of changes
1) It preserves default behavior (using Summary displayMode and 'ul-li' tags)

2) It adds three new properties (ContainerTag, ItemTag, ItemsDisplayMode) to the ContainerPartSettings and its related classes

3) If the ContainerPart is in detail mode, it tries to use the values in setting for display-mode of the items, and for rendering tags.

4) In case of empty values in the settings, it uses the default values ("Summary", ul, li)

### DIFF
--- a/src/Orchard.Web/Core/Containers/Drivers/ContainerPartDriver.cs
+++ b/src/Orchard.Web/Core/Containers/Drivers/ContainerPartDriver.cs
@@ -16,6 +16,7 @@ using System.Web.Routing;
 using Orchard.Settings;
 using Orchard.Core.Feeds;
 using Orchard.UI.Navigation;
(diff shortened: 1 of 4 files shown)


sebastienros:
Talked with Siyamand, will try to use shape alternates when possible, and keep the display type in the settings, for the items.

PR #5584, opened 2015-07-28
Add: Multiple BlogPost ContentTypes support (http://orchard.uservoice…

ropstah:
For reference: http://orchard.uservoice.com/forums/50435-general/suggestions/9064510-multiple-blogpost-type-support

orchard-blog-example

- BlogPostPart ContentPart is also made Attachable() in migration, while this creates more flexibility in creating custom ContentTypes in the Content Definition admin, it might also create some implications as the Blog module functionality requires ContentTypes which have a BlogPostPart attached to also have TitlePart and BodyPart attached.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Blogs/AdminMenu.cs
+++ b/src/Orchard.Web/Modules/Orchard.Blogs/AdminMenu.cs
@@ -1,16 +1,20 @@
 ﻿using System.Linq;
 using Orchard.Blogs.Services;
+using Orchard.ContentManagement.MetaData.Models;
 using Orchard.Localization;
 using Orchard.Security;
 using Orchard.UI.Navigation;
 
-namespace Orchard.Blogs {
-    public class AdminMenu : INavigationProvider {
+namespace Orchard.Blogs
+{
+    public class AdminMenu : INavigationProvider
+    {
         private readonly IBlogService _blogService;
         private readonly IAuthorizationService _authorizationService;
         private readonly IWorkContextAccessor _workContextAccessor;
 
-        public AdminMenu(IBlogService blogService, IAuthorizationService authorizationService, IWorkContextAccessor workContextAccessor) {
+        public AdminMenu(IBlogService blogService, IAuthorizationService authorizationService, IWorkContextAccessor workContextAccessor)
+        {
             _blogService = blogService;
             _authorizationService = authorizationService;
             _workContextAccessor = workContextAccessor;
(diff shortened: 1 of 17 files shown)


sebastienros:
What is the difference with the List module which can replicate Blogs and support multiple types on contained items?

/cc @sfmskywalker

ropstah:
@sebastienros: From that point of view my guess is that there isn't any, but doesn't that make the Blog module in itself obsolete?

Most of the changes included removing hardcoded calls for the BlogPost contenttype. I added a method to the IBlogService interface to return contenttype definitions with BlogPostPart and changed the "Admin/Blogs/{blogId}/Posts/Create" route to allow for a contenttype to be specified. The rest of the changes are rather minor.

bleroy:
Somebody ought to write a blog post on how to re-build a better blogging system with just the list feature ;)

ropstah:
@bleroy I always avoid using lists because it was mentioned as deprecated here http://docs.orchardproject.net/Documentation/Builtin-features and here http://docs.orchardproject.net/Documentation/Creating-lists.

bleroy:
@ropstah: you should definitely check out the latest then. It has been reprecated.

bleroy:
Check this out, @ropstah : https://weblogs.asp.net/bleroy/many-ways-to-relate-orchard-things-part-3-lists

sfmskywalker:
Someone called me? :)

Sent from my Windows Phone
________________________________

bleroy:
:) Nah.

sfmskywalker:
Oh yeah forgot about that one. The only thing that is still missing is for some way to support RSS for lists and some sort of archive as we have with Blogs.

Sent from my Windows Phone
________________________________

ropstah:
@bleroy that looks nice indeed. Guess it does make my contribution a bit obsolete.

sebastienros:
@ropstah it's always sad to see so much effort wasted, even more because of lack of documentation. Please try to use the List module and tell us if it can fulfil your needs. And if you are willing to, feel free to add the RSS support too, in a good way ... maybe using some kind of optional settings on the lists, triggered by a feature in the same module.

And if you are even more courageous, the List documentation page could be rewritten to reflect the current state of the module. You might want to look at the demos Sipke made when he wrote the module.

ropstah:
@sebastienros the effort wasted wasn't that much. At least now I have my cla signed :). 

I've been looking at the RSS and how it is implemented. As far as I can see it **already supports RSS implicitly** because Lists extend Container functionality. 

I guess you are referring to the BlogPart.FeedProxyUrl which allows for a different url to be registered to the FeedManager? 

I must mention that currently it does not work properly for the Blog module. The FeedProxyUrl does get *registered* and the rendered source uses the 'new' url for the RSS feed, however it isn't *routed* to the FeedController anywhere resulting in a "Page Not Found".

image

Fix/Solution
===
The actual routing of a *custom url for a feed* could be provided in the module which creates the feed (e.g. the Blog module, creating a Blog feed, providing a routing constraint for its FeedProxyUrls and point it to the FeedController in the Core module), or it could be a generic functionality in the Core module which allows for a custom-feed-url-providers solution of some sort?

sebastienros:
Can you open an issue with the feed url bug? Might already be done, I think I saw it.

PR #5592, opened 2015-07-30
Fixes 5591 : ContentType names html encoded twice.

TFleury:
Shows Field's DisplayName instead of technical name in title.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/Edit.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/Edit.cshtml
@@ -2,7 +2,7 @@
 @{
     Style.Require("ContentTypesAdmin");
     Script.Require("jQuery");
-    Layout.Title = T("Edit Content Type - {0}", new HtmlString(Model.DisplayName)).ToString();
+    Layout.Title = T("Edit Content Type - {0}", Model.DisplayName);
 }
 
 <div class="manage">
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/Edit.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/Edit.cshtml
@@ -2,7 +2,7 @@
 @{
     Style.Require("ContentTypesAdmin");
     Script.Require("jQuery");
-    Layout.Title = T("Edit Content Type - {0}", Model.DisplayName).ToString();
+    Layout.Title = T("Edit Content Type - {0}", new HtmlString(Model.DisplayName)).ToString();
 }
 
 <div class="manage">
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
@@ -3,7 +3,7 @@
 @{
     Style.Require("ContentTypesAdmin");
 
-    Layout.Title = T("Edit Field \"{0}\"", Model.Name);
+    Layout.Title = T("Edit Field \"{0}\"", Model.DisplayName);
 }
 
 @using (Html.BeginFormAntiForgeryPost()) {
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
@@ -3,7 +3,7 @@
 @{
     Style.Require("ContentTypesAdmin");
 
-    Layout.Title = T("Edit Field \"{0}\"", new HtmlString(Model.DisplayName)).ToString();
+    Layout.Title = T("Edit Field \"{0}\"", Model.Name);
 }
 
 @using (Html.BeginFormAntiForgeryPost()) {
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
@@ -3,7 +3,7 @@
 @{
     Style.Require("ContentTypesAdmin");
 
-    Layout.Title = T("Edit Field \"{0}\"", Model.Name).ToString();
+    Layout.Title = T("Edit Field \"{0}\"", new HtmlString(Model.DisplayName)).ToString();
 }
 
 @using (Html.BeginFormAntiForgeryPost()) {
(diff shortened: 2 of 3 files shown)


sebastienros:
The actual issue is that `Layout.Title` will be encoded by the main view when it's used. By passing a `string` to `Layout.Title`, the string will be encoded. Here we were doing T().ToString() which is also encoding the translation.

The solution is to remove the `.ToString()` on the `T()` so that an `IHtmlString` is passed to `Layout.Title` and then the encoding will only be applied to the translation once.

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/Edit.cshtml
```diff
@@ -2,7 +2,7 @@
 @{
     Style.Require("ContentTypesAdmin");
     Script.Require("jQuery");
-    Layout.Title = T("Edit Content Type - {0}", Model.DisplayName).ToString();
```
sebastienros: Remove `ToString()`

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/Edit.cshtml
```diff
@@ -2,7 +2,7 @@
 @{
     Style.Require("ContentTypesAdmin");
     Script.Require("jQuery");
-    Layout.Title = T("Edit Content Type - {0}", Model.DisplayName).ToString();
+    Layout.Title = T("Edit Content Type - {0}", new HtmlString(Model.DisplayName)).ToString();
```
sebastienros: Delete this

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
```diff
@@ -3,7 +3,7 @@
 @{
     Style.Require("ContentTypesAdmin");
 
-    Layout.Title = T("Edit Field \"{0}\"", Model.Name).ToString();
```
sebastienros: Remove `ToString()`

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
```diff
@@ -3,7 +3,7 @@
 @{
     Style.Require("ContentTypesAdmin");
 
-    Layout.Title = T("Edit Field \"{0}\"", Model.Name).ToString();
+    Layout.Title = T("Edit Field \"{0}\"", new HtmlString(Model.DisplayName)).ToString();
```
sebastienros: Delete this

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/DisplayTemplates/EditTypeViewModel.cshtml
```diff
@@ -9,7 +9,7 @@
     <div class="properties">
         <h3>@Model.DisplayName</h3> @if (!string.IsNullOrWhiteSpace(stereotype)) { <text><span class="stereotype" title="Stereotype">- @stereotype</span></text>}
         @if (creatable) {
-        <p class="pageStatus">@Html.ActionLink(T("Create New {0}", Model.DisplayName).Text, "Create", new {area = "Contents", id = Model.Name})</p>
+        <p class="pageStatus">@Html.ActionLink(T("Create New {0}", new HtmlString(Model.DisplayName)).Text, "Create", new {area = "Contents", id = Model.Name})</p>
```
sebastienros: Remove `Text` and `new HtmlString`

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
```diff
@@ -3,7 +3,7 @@
 @{
     Style.Require("ContentTypesAdmin");
 
-    Layout.Title = T("Edit Field \"{0}\"", Model.Name).ToString();
+    Layout.Title = T("Edit Field \"{0}\"", Model.Name);
```
TFleury: Should not we show DisplayName instead of Technical name ?

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/DisplayTemplates/EditTypeViewModel.cshtml
TFleury: We can't pass IHtmlString to ActionLink method.

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/EditField.cshtml
sebastienros: Sure

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/DisplayTemplates/EditTypeViewModel.cshtml
sebastienros: That's right, just passe `DisplayName` only. The text will be encoded and that's ok.

TFleury: ActionLink method will always encode linkText parameter, so we need to find a proper way to provide it an unencoded string. Or maybe create a new ActionLink method that takes a LocalizedString as parameter.

sebastienros: Right, what we do everywhere else is to create a link directly using `<a href="Url.Action ...">@T(...)</a>`. You could add the overload with `IHtmlString` but it might too much code. Up to you.

TFleury: If I create the link with your method, it works because browser handles not encoded special characters, but the result is inconsistent.

French translation for "Create New {0}" is "Créer {0}"
My content name is "Représentation"
Generated markup is ``<a href="/Admin/Contents/Create/Representation">Créer Repr&#233;sentation</a>`` when it should be ``<a href="/Admin/Contents/Create/Representation">Cr&#233;er Repr&#233;sentation</a>``.

If we choose to encode translation in po file, it will not work for older versions.

I would want to understand the way we handle translations encoding, I opened a discussion [here](http://orchard.codeplex.com/discussions/642369) to talk about this.

src/Orchard.Web/Modules/Orchard.ContentTypes/Views/Admin/Edit.cshtml
```diff
@@ -2,7 +2,7 @@
 @{
     Style.Require("ContentTypesAdmin");
     Script.Require("jQuery");
-    Layout.Title = T("Edit Content Type - {0}", Model.DisplayName).ToString();
+    Layout.Title = T("Edit Content Type - {0}", Model.DisplayName);
```
TFleury: Displayed title is fixed, but page head's title tag is corrupted : ContentType's DisplayName is encoded twice.

TFleury:
I pushed a new PR #5627 for that.

Issue #2839, opened 2012-09-07
Reversed order of page title and site name


orchardbot:
@Piedone created:
https://orchard.codeplex.com/workitem/19012

I don't see any advantage (neither UX, neither SEO) of the current implementation of IPageTitleBuilder where the site name comes first, then the page name. This really should be reversed.
I've also opened this: http://orchard.codeplex.com/workitem/18179 but in the meantime this would also be good enough.

PR #5597, opened 2015-07-31
#2839 reverse order of page title and site name, SEO improvement

yaplex:
solution for people who don't like to override too much pages from base templates. ticket #2839 

Alex
https://yaplex.com

### DIFF
--- a/src/Orchard.Web/Core/Shapes/Views/Document.cshtml
+++ b/src/Orchard.Web/Core/Shapes/Views/Document.cshtml
@@ -14,7 +14,7 @@
 <html lang="@WorkContext.CurrentCulture" class="static @Html.ClassForPage()" dir="@(isRtl?"rtl":"ltr")"> 
 <head> 
     <meta charset="utf-8" />
-    <title>@Html.Title(title, siteName)</title> 
+    <title>@Html.Title(siteName, title)</title> 
     @Display(Model.Head)
     <script>(function(d){d.className="dyn"+d.className.substring(6,d.className.length);})(document.documentElement);</script> 
     <script>window.isRTL = @(isRtl.ToString().ToLowerInvariant());</script>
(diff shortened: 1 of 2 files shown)


yaplex:
It's fixed the way you suggested in ticket, I fixed "GenerateTitle" method in PageTitleBuilder.

sebastienros:
We can't change this that easily, it would break what people have done so far.
I think I have created an issue explaining what should be done instead. The general idea was that title parts could be added with an optional `Position`, with the same logic as wherever we use positions already (1, 1.2, ...). This way it is not breaking, and you can force a specific one to come first. 

Would you mind trying it?

yaplex:
Where exactly are you using position? I found class Position.cs but don't think it's what you telling, I looked into ShapePosition class, but looks like it's also not what you mean.

sebastienros:
What I mean is that you add an override which takes another string as parameter, called `position`. And the default value is `"5"`. When we render the items, we first order them with  `Orchard.UI.FlatPositionComparer`.

yaplex:
Sorry, but it's not clear for me. 
There is a helper method 
public static MvcHtmlString Title(this HtmlHelper html, params string[] titleParts)
which is called this way  @Html.Title(title, siteName)
I don't think this method can accept another parameter, because of "params" keyword

I looked into details of implementation of FlatPositionComparer, and if I make Title method accept my new type instead of string, I can have Position property there. Soemthing like this:
public static MvcHtmlString Title(this HtmlHelper html, params TitlePart[] titleParts)
where TitlePar is 
class TitlePart{
string Title{get;set}
string Position{get;set}
}

and the using will be 
@Html.Title(new TitlePart{Title="Hello",Position="1"}, 
new TitlePart{Title="Word", Position="2"})

If you talking about something else, could you please specify more details, I am sorry but I start learning Orchard development just a couple of month ago, so some obvious tasks make require additional explanation.

PR #5608, opened 2015-08-04
Fixes #5607 - Assets.json options

### LINKED ISSUE #5607: Assets.json: Differentiate between watch and build files
I'm not sure this will apply to everyone, but we have a main .less file that imports all other less files into it (a master.less, if you will).  With Assets.json, are you specifying the files to watch as well as compile in a single array?  For us, we would want to watch *.less (and **/*.less) and only build master.less.
Is it possible to split up the watch vs build?
Does that make sense?

dcinzona:
Added support for separation of watch vs build .less files.
This required adding a flag for forcing rebuild, because the watch files may have changed, but the build files may not have, causing rebuild to fail.  Setting the rebuildAlways flag to true forces the rebuild to occur, regardless of files changed.  
Also added an option to exclude sourcemaps.  This defaults to false (include them if not set).

All options are ... wait for it ... optional.

Example Assets.json:
```
[
  {
    "inputs": [ "Less/master.less" ],
    "output": "Styles/@.css",
    "watch": ["Less/**/*.less", "Less/*.less"],
    "rebuildAlways": true,
    "excludeSourceMaps": true
  }
]
```

### DIFF
--- a/src/Gulpfile.js
+++ b/src/Gulpfile.js
@@ -18,6 +18,24 @@ var glob = require("glob"),
 
 /*
 ** GULP TASKS
+---------------------------------------
+   Checks Themes, Modules, and Core directories for an Assets.json file. 
+   Assets.json defines what Less, CSS, TypeScript, JS files should be processed by Gulp.
+   
+   When defining your own Assets.json file, it should be saved in the root of module or theme project
+
+   Assets.json example:
+   Saved to /Modules/My.Custom.Module/Assets.json
+   [
+      {
+        "inputs": [ "Less/master.less" ], //Specifies which files to process during the Build task
+        "output": "Styles/@.css", //When @ is specified, each file specified in "inputs" will be converted to [filename].css
+        "watch": ["Less/*.less"], //specifies which files to watch, use this when you have a master Less file that imports other Less files
+        "rebuildAlways": true, //Will force rebuild, defaults to 'true'
+        "generateSourceMaps": true //Will include source maps in unminified version, defaults to 'true'
+      }
+    ]
+    "inputs" and "output" are required.  All other properties are optional.
 */
 
 // Incremental build (each asset group is built only if one or more inputs are newer than the output).
@@ -43,7 +61,8 @@ gulp.task("watch", function () {
     getAssetGroups().forEach(function (assetGroup) {
         gulp.watch(assetGroup.watchPaths != undefined ? assetGroup.watchPaths : assetGroup.inputPaths, function (event) {
             console.log("Asset file '" + event.path + "' was " + event.type + ", rebuilding output '" + assetGroup.outputPath + "'.");
-            var task = createAssetGroupTask(assetGroup, assetGroup.rebuildAlways);
+            var doRebuild = assetGroup.rebuildAlways != undefined ? assetGroup.rebuildAlways : true; //defaults to true
+            var task = createAssetGroupTask(assetGroup, doRebuild);
         });
     });
 });
@@ -67,14 +86,15 @@ function getAssetGroups() {
 
 function resolveAssetGroupPaths(assetGroup, assetManifestPath) {
     assetGroup.basePath = path.dirname(assetManifestPath);
-    if (assetGroup.watch != undefined) {
-        assetGroup.watchPaths = assetGroup.watch.map(function (watchPath) {
-            return path.join(assetGroup.basePath, watchPath);
-        });
-    }
     assetGroup.inputPaths = assetGroup.inputs.map(function (inputPath) {
         return path.join(assetGroup.basePath, inputPath);
     });
+    if (assetGroup.watch != undefined) {
+        assetGroup.watchPaths = assetGroup.watch.map(function (watchPath) {
+            return path.join(assetGroup.basePath, watchPath);
+        })
+        .concat(assetGroup.inputPaths); //include inputs in watch list
+    }
     assetGroup.outputPath = path.join(assetGroup.basePath, assetGroup.output);
     assetGroup.outputDir = path.dirname(assetGroup.outputPath);
     assetGroup.outputFileName = path.basename(assetGroup.output);
@@ -101,16 +121,13 @@ function buildCssPipeline(assetGroup, doRebuild) {
             throw "Input file '" + inputPath + "' is not of a valid type for output file '" + assetGroup.outputPath + "'.";
     });
     var doConcat = path.basename(assetGroup.outputFileName, ".css") !== "@";
-
     if (!doRebuild) {
         console.log("CSS will only rebuild if less files specified in 'inputs' are newer.");
     }
     else {
         console.log("Force Rebuild is enabled, rebuilding all input files.");
     }
-
-    var includeSourcepaths = assetGroup.excludeSourceMaps != undefined ? !assetGroup.excludeSourceMaps : true;
-
+    var generateSourceMaps = assetGroup.generateSourceMaps != undefined ? assetGroup.generateSourceMaps : true;
     return gulp.src(assetGroup.inputPaths)
         .pipe(gulpif(!doRebuild,
             gulpif(doConcat,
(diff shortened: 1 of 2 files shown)


DanielStolt:
I really like what you did here. I have a couple of suggestions for improvement:

1. Why not get rid of the `rebuildAlways` parameter, and in the `watch` task always call `createAssetGroupTask()` with `doRebuild = true` same as the `rebuild` task does? I should have done that from the start, I think I just forgot.
2. The optional `watch` property you introduced should be treated as files to watch **in addition** to the inputs, not instead of the inputs IMO.
3. The `excludeSourcemaps` option is fine but I think we should switch it around so the property name is `generateSourceMaps` and it defaults to `true`.

dcinzona:
Agree and will make the edits when I get into the office (it's 4:45am here :)). 
Regarding 1: if rebuild always is set in the watch task, why use newer at all?
2 is interesting, I'll have to play around with that. The reason this worked for me was that the watch list also includes the inputs file, but I can see where the case might be that it doesn't.  
3. Not a problem, and totally makes sense. Double negatives can be confusing

DanielStolt:
Did my Github notification wake you up?? :)

1. Because the `build` task uses it, and will continue to use it.

dcinzona:
Ha no, I was already awake (back problems).  I'm going to start working on this now.

dcinzona:
Updated the PR with the changes listed in https://github.com/dcinzona/Orchard/commit/0fcf8c740978dfeb17f3ebb05f57e61db363b1c7

I was thinking that it would be nice to also watch Assets.json, and when that changes, to restart the whole watch task (if running). What do you think? I could do a separate PR or implement it into this one.

DanielStolt:
That would be a very nice thing, yes. Separate PR please.

dcinzona:
Sounds good, will do.

DanielStolt:
I can't see that you updated the PR with the changes? #3 is addressed, but not #1 and #2?

dcinzona:
2 is in there - see line 96 on the change. 
```
.concat(assetGroup.inputPaths); //include inputs in watch list
```
For number 1, did you want me to remove the variable altogether or include it as optional?  I included it as optional, but defaulting to true.

DanielStolt:
Ah, I see 2 now. I expected to find it in the ´watch` task function rather than the `resolveAssetGroupPaths` function. I think it would be more semantically correct to combine the two lists in the ´watch` task function.

Re 1, yes by "get rid of the rebuildAlways parameter" I meant actually removing it. :)

A general thing: rather than doing `variable != undefined`  I think it's more robust to do `!!variable` as it handles null, empty string and undefined - right?

dcinzona:
Yup, that works.  I'll update accordingly!

dcinzona:
So, `!!` will return true if the variable value is set to `false`...
```javascript
    var generateSourceMaps = !!assetGroup.generateSourceMaps ? assetGroup.generateSourceMaps : true;
    console.log("source maps: " + generateSourceMaps); // returns true even if explicitly set to false
```
```javascript
    var generateSourceMaps = assetGroup.generateSourceMaps !== undefined ? assetGroup.generateSourceMaps : true;
    console.log("source maps: " + generateSourceMaps); //returns false when explicitly set, otherwise true
```

DanielStolt:
Yes in that particular case, since the default should be `true` you can't use that construct.

dcinzona:
Ok, I think it's good to go now... let me know if you notice any issues.  I switched the `!==` condition to `hasOwnProperty` in the last commit.  May help with legibility.

dcinzona:
FYI, I have it working where if an Asset.json file changes, it reloads the watchers.  You mentioned you wanted this as a separate PR, so that's what I'm going to do.  PR https://github.com/OrchardCMS/Orchard/pull/5613

src/Gulpfile.js
```diff
@@ -13,10 +13,28 @@ var glob = require("glob"),
...
+   Saved to /Modules/My.Custom.Module/Assets.json
+   [
+      {
+        "inputs": [ "Less/master.less" ], //Specifies which files to process during the Build task
+        "output": "Styles/@.css", //When @ is specified, each file specified in "inputs" will be converted to [filename].css
+        "watch": ["Less/*.less"], //specifies which files to watch, use this when you have a master Less file that imports other Less files
+        "generateSourceMaps": true //Will include source maps in unminified version, defaults to 'true'
+      }
+    ]
+    "inputs" and "output" are required.  All other properties are optional.
```
DanielStolt: This comment should go. We're going to write documentation for this, and also include a commented sample `Assets.json` in module scaffolding code-gen.

src/Gulpfile.js
```diff
@@ -13,10 +13,28 @@ var glob = require("glob"),
 	uglify = require("gulp-uglify"),
 	rename = require("gulp-rename"),
     concat = require("gulp-concat"),
-	header = require("gulp-header")
+	header = require("gulp-header"),
+    notify = require("gulp-notify")
```
DanielStolt: Leave out gulp-notify.

src/Gulpfile.js
```diff
@@ -69,6 +88,12 @@ function resolveAssetGroupPaths(assetGroup, assetManifestPath) {
     assetGroup.inputPaths = assetGroup.inputs.map(function (inputPath) {
         return path.join(assetGroup.basePath, inputPath);
     });
+    
+    assetGroup.watchPaths = !!assetGroup.watch ?
+        assetGroup.watch.map(function (watchPath) {
+            return path.join(assetGroup.basePath, watchPath);
+        }).concat(assetGroup.inputPaths) : assetGroup.inputPaths;
```
DanielStolt: I still think this is the wrong place for this - the merging of input paths with watch paths should go in the "watch" task function, not here. The `watchPaths` property should retain its meaning and value, not be "magically" altered here to mean something different.

src/Gulpfile.js
```diff
@@ -95,6 +120,13 @@ function buildCssPipeline(assetGroup, doRebuild) {
             throw "Input file '" + inputPath + "' is not of a valid type for output file '" + assetGroup.outputPath + "'.";
     });
     var doConcat = path.basename(assetGroup.outputFileName, ".css") !== "@";
+    if (!doRebuild) {
+        console.log("CSS will only rebuild if less files specified in 'inputs' are newer.");
+    }
+    else {
+        console.log("Force Rebuild is enabled, rebuilding all input files.");
+    }
```
DanielStolt: Don't understand why you've added this logging. Seems arbitrary - shouldn't be here.

src/Gulpfile.js
```diff
@@ -115,13 +147,14 @@ function buildCssPipeline(assetGroup, doRebuild) {
...
-        .pipe(sourcemaps.write())
+        .pipe(gulpif(generateSourceMaps, sourcemaps.write()))
         .pipe(gulp.dest(assetGroup.outputDir))
         .pipe(minify())
         .pipe(rename({
             suffix: ".min"
         }))
-        .pipe(gulp.dest(assetGroup.outputDir));
+        .pipe(gulp.dest(assetGroup.outputDir))
+        .pipe(gulpif(doRebuild, notify("Rebuild complete"), notify("Build process complete")));
```
DanielStolt: Remove notification.

src/Gulpfile.js
dcinzona: I put it in the "resolveAssetGroupPaths, as that is where you were generating the input paths, but I will move it to the watch function.

dcinzona:
Ok, I've updated patch-5.  Let me know if this works.  I'm still a little unclear as to what you are looking for regarding the watchPaths resolve code, but I moved it to the watch task generation method.  Hopefully that's what you were looking for.  If not, let me know.

DanielStolt:
I probably didn't manage to fully convey what I was after, but I merged this PR now and applied a small refactoring on top of it. Thanks for this contribution @dcinzona - makes the Gulp pipeline even better!

PR #5612, opened 2015-08-05
Add extension point to allow developers to provide custom logic when validating User Data

paynecrl97:
This PR adds an extension point to `FormsAuthenticationService` that allows module developers to place custom values into User Data and then provide custom logic to be able to ensure that this User Data value is still valid during the `GetAuthenticatedUser` method.

The logic to ensure that auth cookies are tenant aware has been abstracted into one of these providers.

Example use cases would be where auth implementation has been switched out to a system with a token that needs to be validated during `GetAuthenticatedUser`.

### DIFF
--- a/src/Orchard/Security/Providers/FormsAuthenticationService.cs
+++ b/src/Orchard/Security/Providers/FormsAuthenticationService.cs
@@ -54,7 +54,7 @@ namespace Orchard.Security.Providers {
             var now = _clock.UtcNow.ToLocalTime();
 
             // The cookie user data is "{userName.Base64};{providerKey|providerValue};{providerKey|providerValue}....".
-            // The UserName is encoded to Base64 to prevent collisions with the ';' and '|' seprarators.
+            // The UserName is encoded to Base64 to prevent collisions with the ';' and '|' seprarators (which do not exist in the Base 64 alphabet).
             var userData = user.UserName.ToBase64();
 
             foreach (var userDataValidationProvider in _userDataValidationProviders) {
@@ -154,8 +154,8 @@ namespace Orchard.Security.Providers {
             }
 
             // Iterate over all but the first user data segment because the first is the username
-            for (var i = 1; i < userDataSegments.Length; i++) {
-                var segmentSplit = userDataSegments[i].Split('|');
+            foreach (var segment in userDataSegments.Skip(1)) {
+                var segmentSplit = segment.Split('|');
 
                 if (segmentSplit.Length < 2) {
                     continue;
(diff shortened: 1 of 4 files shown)


src/Orchard/Security/Providers/FormsAuthenticationService.cs
```diff
@@ -10,7 +12,7 @@
 
 namespace Orchard.Security.Providers {
     public class FormsAuthenticationService : IAuthenticationService {
-        private const int _cookieVersion = 3;
+        private const int _cookieVersion = 4;
```
DanielStolt: Maybe the `IUserDataValidationProvider`s need to contribute to the cookie version number somehow?

src/Orchard/Security/Providers/FormsAuthenticationService.cs
```diff
@@ -48,9 +53,21 @@ public class FormsAuthenticationService : IAuthenticationService {
         public void SignIn(IUser user, bool createPersistentCookie) {
             var now = _clock.UtcNow.ToLocalTime();
 
-            // The cookie user data is "{userName.Base64};{tenant}".
-            // The username is encoded to Base64 to prevent collisions with the ';' seprarator.
-            var userData = String.Concat(user.UserName.ToBase64(), ";", _settings.Name);
+            // The cookie user data is "{userName.Base64};{providerKey|providerValue};{providerKey|providerValue}....".
+            // The UserName is encoded to Base64 to prevent collisions with the ';' and '|' seprarators.
```
DanielStolt: Perhaps also clarify in the comment that the | separator is used to prevent collision with Base64 alphabet.

src/Orchard/Security/Providers/FormsAuthenticationService.cs
```diff
@@ -141,8 +153,30 @@ public class FormsAuthenticationService : IAuthenticationService {
                 return null;
             }
 
-            if (!String.Equals(userDataTenant, _settings.Name, StringComparison.Ordinal)) {
-                return null;
+            // Iterate over all but the first user data segment because the first is the username
+            for (var i = 1; i < userDataSegments.Length; i++) {
```
DanielStolt: Maybe more readable using a `foreach` and a `.Skip(1)`?

src/Orchard/Security/Providers/FormsAuthenticationService.cs
```diff
@@ -141,8 +153,30 @@ public class FormsAuthenticationService : IAuthenticationService {
...
+                var segmentSplit = userDataSegments[i].Split('|');
+
+                if (segmentSplit.Length < 2) {
+                    continue;
+                }
+
+                var providerKey = segmentSplit[0].FromBase64();
+                var providerValue = segmentSplit[1].FromBase64();
+
+                foreach (var userDataValidationProvider in _userDataValidationProviders.Where(p=>p.Key == providerKey)) {
```
DanielStolt: Missing spaces in the lambda.

sebastienros:
I don't think this class needs more extensibility. You can already replace `IAuthenticationService` by your own implementation which would contain the logic. You are adding a way to provide some variance to an implementation detail, i.e. the fact that the cookie contains custom values.

I would suggest that you create a custom service containing the tenant logic.

PR #5613, opened 2015-08-05
Gulpfile auto reload

dcinzona:
Added support for watching and reloading the Asset manifest when Assets.json is changed.


DanielStolt:
Let's finalize and merge #5608 first, and then rebase this one on that work.

dcinzona:
#5608 should be good to go.  I'm going to remove this PR so that we can rebase it to that

PR #5627, opened 2015-08-10
Fixes 5591 : ContentType's name encoding

TFleury:
It's the only workaround if we choose keeping Localizer to default encode arguments.


jtkech:
+1

First, for double encoding issues, I thougt we only need to remove some (string) casting and/or remove some .Text(), .ToString()... But good catch for the double encoding of the head's title tag. Indeed, in Document.cshtml the title is converted to a string in order to be passed to the Html.Title() extension where another encoding is done...

For more details, see my answer on you opened discussion here:
https://orchard.codeplex.com/discussions/642369

Best

PR #5628, opened 2015-08-10
Feature/distributed locking

sfmskywalker:
This PR adds a distributed locking service with a lock provider that uses a database table as the synchronization medium. This database implementation of `IDistributedLock` is provided as part of the `Orchard.TaskLease` module. A default (non-distributed) version of `IDistributedLock` is provided by the framework so that it's safe to use the service even if no real lock provider is enabled.

The `AutomaticDataMigrations` class has been updated to try and acquire a lock using the `IDistributedLockService` to ensure only one node in web farm scenarios execute the migrations after deployment or installing a new module.


src/Orchard/Data/Migration/AutomaticDataMigrations.cs
```diff
@@ -25,30 +29,34 @@ IFeatureManager featureManager
...
-            if (featuresToEnable.Any()) {
-                _featureManager.EnableFeatures(featuresToEnable, true);
-            }
-
-            foreach (var feature in _dataMigrationManager.GetFeaturesThatNeedUpdate()) {
-                try {
-                    _dataMigrationManager.Update(feature);
+            using(var @lock = _distributedLockService.AcquireLock(GetType().FullName, TimeSpan.FromMinutes(30), TimeSpan.FromMilliseconds(250))) {
+                if (@lock == null)
+                    return;
```
DanielStolt: This check is unnecessary.

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Orchard.Tasks.Locking.Migrations {
+    public class FrameworkMigrations : DataMigrationImpl {
+
+        public int Create() {
+            SchemaBuilder.CreateTable("LockRecord", table => table
+                .Column<int>("Id", column => column.PrimaryKey().Identity())
+                .Column<string>("Name", column => column.NotNull().WithLength(256))
+                .Column<string>("Owner", column => column.WithLength(256))
+                .Column<int>("ReferenceCount")
+                .Column<DateTime>("CreatedUtc")
```
DanielStolt: `Count` would be a better name.

src/Orchard/Tasks/Locking/Records/LockRecord.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Orchard.Tasks.Locking.Records {
+    public class LockRecord {
+        public virtual int Id { get; set; }
+        public virtual string Name { get; set; }
+        public virtual string Owner { get; set; }
+        public virtual int ReferenceCount { get; set; }
```
DanielStolt: `Count` would be a better name.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+using Orchard.Services;
+using Orchard.Tasks.Locking.Records;
+
+namespace Orchard.Tasks.Locking.Services {
+
+    public class DistributedLockService : IDistributedLockService {
+        private readonly IMachineNameProvider _machineNameProvider;
+        private readonly ILifetimeScope _lifetimeScope;
+        private readonly IClock _clock;
+        private readonly object _semaphore = new object();
```
DanielStolt: Bit of a misnomer, this is not a semaphore, just a plain monitor. I would remane to `_lock`.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+            _lifetimeScope = lifetimeScope;
+            _clock = clock;
+            Logger = NullLogger.Instance;
+        }
+
+        public ILogger Logger { get; set; }
+
+        public bool TryAcquireLock(string name, TimeSpan maxLifetime, TimeSpan timeout, out IDistributedLock @lock) {
+            lock (_semaphore) {
+                @lock = default(IDistributedLock);
```
DanielStolt: I'm personally not too fond of these `default()` keywords... I think it's more readable to initialize to `null`. What is the point of using `default()` instead when the variable is already typed?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                try {
+                    var waitedTime = TimeSpan.Zero;
+                    var waitTime = TimeSpan.FromMilliseconds(timeout.TotalMilliseconds / 10);
+                    bool acquired;
+
+                    while (!(acquired = TryAcquireLockRecord(name, maxLifetime, out @lock)) && waitedTime < timeout) {
+                        Task.Delay(timeout).ContinueWith(t => {
```
DanielStolt: Shouldn't this be `Task.Delay(waitTime)`?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+                            waitedTime += waitTime;
+                        }).Wait();
+                    }
+
+                    if (acquired) {
+                        Logger.Debug("Successfully acquired a lock named {0}.", name);
```
DanielStolt: I try to always surround arbitrary strings with single quotes to avoid ambiguity in the output message if the string should contain spaces, periods etc.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public ILogger Logger { get; set; }
+
+        public bool TryAcquireLock(string name, TimeSpan maxLifetime, TimeSpan timeout, out IDistributedLock @lock) {
+            lock (_semaphore) {
+                @lock = default(IDistributedLock);
+
+                try {
+                    var waitedTime = TimeSpan.Zero;
+                    var waitTime = TimeSpan.FromMilliseconds(timeout.TotalMilliseconds / 10);
```
DanielStolt: Not sure wait time should be a function of the timeout. What's your reasoning here?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+
+        public IDistributedLock AcquireLock(string name, TimeSpan maxLifetime, TimeSpan timeout) {
+            IDistributedLock lockResult;
+            return TryAcquireLock(name, maxLifetime, timeout, out lockResult) ? lockResult : null;
+        }
+
+        public void DisposeLock(IDistributedLock @lock) {
+            var childLifetimeScope = CreateChildLifetimeScope(@lock.Name);
```
DanielStolt: `ReleaseLock` would be a better name.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+                }
+
+                Logger.Debug("Could not acquire a lock named {0}.", name);
+                return false;
+            }
+        }
+
+        public IDistributedLock AcquireLock(string name, TimeSpan maxLifetime, TimeSpan timeout) {
+            IDistributedLock lockResult;
+            return TryAcquireLock(name, maxLifetime, timeout, out lockResult) ? lockResult : null;
```
DanielStolt: This should throw a `TimeoutException` if the lock could not be acquired, rather than returning `null`.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+            try {
+                var repository = childLifetimeScope.Resolve<IRepository<LockRecord>>();
+                var transactionManager = childLifetimeScope.Resolve<ITransactionManager>();
+                transactionManager.RequireNew(IsolationLevel.ReadCommitted);
+                var record = repository.Get(@lock.Id);
+
+                if (record != null) {
+                    if (record.ReferenceCount > 0)
+                        record.ReferenceCount--;
+                }
```
DanielStolt: I think this should throw if you're trying to release the same lock too many times, or trying to release a lock that doesn't exist. Otherwise might be obscuring bugs in calling code.

src/Orchard/Tasks/Locking/Services/IDistributedLock.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Orchard.Tasks.Locking.Services {
+    /// <summary>
+    /// Represents a distributed lock.
+    /// </summary>
+    public interface IDistributedLock : IDisposable {
```
DanielStolt: This interface should not exist.

src/Orchard/Tasks/Locking/Services/Lock.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Orchard.Logging;
+
+namespace Orchard.Tasks.Locking.Services {
+
+    /// <summary>
+    /// Provides a database driven implementation of <see cref="IDistributedLock" />
+    /// </summary>
+    public class Lock : IDistributedLock {
```
DanielStolt: We should get rid of this interface, and call the class `DistributedLock` I think.

src/Orchard/Tasks/Locking/Services/Lock.cs
```diff
@@ -0,0 +1,31 @@
...
+        private readonly IDistributedLockService _distributedLockService;
+        public string Name { get; set; }
```
DanielStolt: Move this property down to line 21.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                // The current owner name (based on machine name and current thread ID).
+                var currentOwnerName = GetOwnerName();
+                var canAcquireLock = false;
+
+                // Check if there's already an active lock.
+                if (record != null) {
+                    // Check if the owner of the lock is the one trying to acquire the lock.
+                    if (record.Owner == currentOwnerName) {
+                        record.ReferenceCount++;
```
DanielStolt: Don't we need to do a `repository.Flush()` here?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+            try {
+                var transactionManager = childLifetimeScope.Resolve<ITransactionManager>();
+                transactionManager.RequireNew(IsolationLevel.ReadCommitted);
+
+                // This way we can create a nested transaction scope instead of having the unwanted effect
+                // of manipulating the transaction of the caller.
+                var repository = childLifetimeScope.Resolve<IRepository<LockRecord>>();
+
+                // Find an existing, active lock, if any.
+                var record = repository.Table.FirstOrDefault(x => x.Name == name && (x.ValidUntilUtc >= _clock.UtcNow || x.ReferenceCount > 0));
```
DanielStolt: Shouldn't this be an `&&` instead of an `||`?

src/Orchard/Tasks/Locking/Services/IDistributedLockService.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Orchard.Tasks.Locking.Services {
+    /// <summary>
+    /// Provides distributed locking functionality.
+    /// </summary>
+    public interface IDistributedLockService : ISingletonDependency {
+        /// <summary>
+        /// Tries to acquire a lock on the specified name.
+        /// </summary>
+        /// <param name="name">The name to use for the lock.</param>
+        /// <param name="maxLifetime">The maximum amount of time the lock is allowed. This is a safety net in case the caller fails to release the lock.</param>
```
DanielStolt: How about `maxValidFor` as an alternative name? It maps better with the `ValidUntilUtc` database column name.

src/Orchard/Tasks/Locking/Services/IDistributedLockService.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="timeout">The amount of time to wait for the lock to be acquired before timing out.</param>
+        /// <param name="lock">The acquired lock.</param>
+        /// <returns>Returns true if a lock was successfully acquired, false otherwise.</returns>
+        bool TryAcquireLock(string name, TimeSpan maxLifetime, TimeSpan timeout, out IDistributedLock @lock);
```
DanielStolt: There should be an overload that does not accept a `timeout` parameter, that immediately returns.

src/Orchard/Tasks/Locking/Services/IDistributedLockService.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// Acquires a lock with the specified parameters.
+        /// </summary>
+        /// <param name="name">The name to use for the lock.</param>
+        /// <param name="maxLifetime">The maximum amount of time the lock is allowed. This is a safety net in case the caller fails to release the lock.</param>
+        /// <param name="timeout">The amount of time to wait for the lock to be acquired before timing out.</param>
+        /// <returns>Returns a lock if one was successfully acquired, null otherwise.</returns>
+        IDistributedLock AcquireLock(string name, TimeSpan maxLifetime, TimeSpan timeout);
```
DanielStolt: There should be an overload that does not accept a `timeout` parameter, that blocks indefinitely.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Why?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Ok

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
sfmskywalker: Looking at your comment below, why do you think it is better to throw an exception than returning null? Is that a pattern? And if we did change the method to throw, would you surround the `using` statement with a try/catch block, or not use the `using` statement at all?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
DanielStolt: Because not throwing is what the `Try` variant is for.

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
sfmskywalker: Agreed.

src/Orchard/Tasks/Locking/Records/LockRecord.cs
sfmskywalker: Agreed.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Thank you, I'll rename.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Good point, I'll change.

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
DanielStolt: Yes, pattern. And also, otherwise what would be the difference between `Acquire` and `TryAcquire`?

If I wanted to catch, I would get rid of the `using` and dispose in the `finally` block, yes.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: The idea is that if acquiring a lock fails, we want to keep retrying until a *total* of `timeout` time has elapsed. I took this code from someone else and reasoned it would work well, even for small numbers, since worst case the `Task.Delay` would be invoked right after the previous invocation, but only for a short amount of time.

Another function could work too I suppose, but I couldn't come up with anything else that is less arbitrary. I'm open to suggestions though.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Yes absolutely. Thanks for catching that, code runner. ;)

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: I do that from time to time, but never made a decision to stick with it, because I hadn't decided whether it looked good enough. But you just pushed me over the ledge. I'll change.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Much better, yes.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Yes good catch. My original definition of an active (unavailable) lock was as follows:
- One that has not yet expired *or* one whose count is greater than 0.

However, it should of course be:
- One that has not yet expired *and* whose count is greater than 0.

I'll add a unit test to formalize this rule.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: You're right, we need NHibernate to execute the UPDATE statement as soon as possible.

src/Orchard/Tasks/Locking/Services/IDistributedLock.cs
sfmskywalker: I agree, there's no point in keeping the abstraction anymore after this changeset. Removing.

src/Orchard/Tasks/Locking/Services/IDistributedLockService.cs
sfmskywalker: I was actually trying to come up with a better name but failed. I like `maxLifetime`, changing.

src/Orchard/Tasks/Locking/Services/Lock.cs
sfmskywalker: I agree. `Lock` is too generic and doesn't convey the fact that it is a distributed lock. Will update.

src/Orchard/Tasks/Locking/Services/Lock.cs
sfmskywalker: Will do.

src/Orchard/Tasks/Locking/Services/IDistributedLockService.cs
sfmskywalker: Good idea. I suspect it would be used that way in the majority of cases anyway, so that would make it simpler to use.

src/Orchard/Tasks/Locking/Services/Lock.cs
```diff
@@ -0,0 +1,31 @@
...
+            Id = id;
+        }
+
+        public ILogger Logger { get; set; }
+        public int Id { get; set; }
+
+        // This will be called at least and at the latest by the IoC container when the request ends.
+        public void Dispose() {
+            if (!_isDisposed) {
+                _isDisposed = true;
```
DanielStolt: Potential race condition here if Dispose() is called multiple times from different threads.

src/Orchard/Tasks/Locking/Services/IDistributedLockService.cs
sfmskywalker: Makes sense. "Indefinitely" until a lock is acquired, of course.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: That's true actually. I was hoping for a way to be able to use the `using` statement fluidly, like:

```C#
using(var @lock = AcquireLock()) { ... }
```

Now one would have to use either `TryAcquireLock` and do the `using` on its `out` parameter, or use `AcquireLock` inside of a `try/catch`.

src/Orchard/Tasks/Locking/Services/Lock.cs
DanielStolt: Nice fix might be to use `Interlocked.CompareExchange<T>`.

sfmskywalker: Which would be fine if `_distributedLockService` would handle already disposed locks, but that's to be changed. Do you think it would suffice to `lock` on `this`?

DanielStolt: Yes that would suffice. But I would go for lock free, since it's so simple in this case.

sfmskywalker: I'll have a look.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,153 @@
...
+                    // No one has an active lock yet, so good to go.
+                    record = new LockRecord {
+                        Name = name,
+                        Owner = currentOwnerName,
+                        ReferenceCount = 1,
+                        CreatedUtc = _clock.UtcNow,
+                        ValidUntilUtc = _clock.UtcNow + maxLifetime
+                    };
+                    repository.Create(record);
+                    repository.Flush();
```
sebastienros: Red flag !!!!

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
Piedone: Hmm, why is it better? It's not a count of the record itself, but rather that of references (or acquisitions) of it.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
Piedone: Have you examined named locks @sfmskywalker ? Because as we discussed having a single lock object like this will queue up all the acquire calls, regardless of context, and now e.g. a background task executing will wait until a foreground content item publish finishes when both use locking through IDistributedLockService.

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
DanielStolt: @Piedone The lock is reentrant, i.e. the same owner can make multiple acquisitions. Look at line 121.

As to why "Count" is better, IMO the record itself represents an acquisition, so "Count" is intrinsic. But now we're really talking semantics.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
DanielStolt: @Piedone you are right but you shouldn't be. :) Let me explain.

Something I missed in my review yesterday: I believe this monitor is not needed at all, and @sfmskywalker also mentioned he thinks it's only needed from a unit test context. But give that we keep this monitor in place, it should not be held for this long. It should only be held for the duration of the atomic operation of trying to acquire the distributed lock, but if the current thread goes back to waiting because the distributed lock could not be acquired, then the monitor should be released while waiting, and then reacquired when the thread wakes up.

This will ensure that this monitor is only ever held extremely briefly I don't think a named lock should be necessary. So given this implementation you are right, but the implementation should not be this way in the first place.

Piedone: I see. Even when the lock is only being held for the duration of acquiring the lock I'm not sure it's good to have a common lock: acquiring a lock should be fast, but due to it involving a lot more than a Monitor.TryEnter() (creating a separate lifetime scope and issuing a DB roundtrip in the case of SQL-based locks) it can easily be orders of magnitude slower (acquiring a .NET lock is in the order of magnitude of nanoseconds, a DB call in itself is measurable in milliseconds when the DB is on a separate server).

Now of course without having a lot of demanding use-cases for distributed locks already (where I use distributed locking from Helpful Libraries this is nowhere a bottleneck) I can't say can be a big issue unless you have thousands of lock acquisitions per second but it would be the best not to build in bottlenecks in the first place. Actually the best would be if we wouldn't need a second layer of locking at all and just the distributed one would work alone well enough.

DanielStolt: @Piedone I totally agree with that - there shouldn't be a need for the monitor at all. I think it's because the `TransactionManager` instance is not thread safe. I spoke to @sfmskywalker offline about decreasing the scope of the monitor to only cover interactions with `TransactionManager`, which I think he's going to try.

sfmskywalker: Yeah the problem is that when you write a unit test that fires off multiple `Task` objects that create and commit transactions simultaneously using `ITransctionManager`, the following interesting thing occurs: 


Synchronizing thread access to the transaction manager from the calling code prevents this. The `IDistributedLockService.AcquireLockRecord` creates and commits and disposes a transaction using `ITransactionManager`. I added a unit test called `MultithreadedAcquisitionsShouldNotCauseTransactionErrors` that will fail if you remove the lock.

src/Orchard/Tasks/Locking/Services/Lock.cs
sfmskywalker: Like this?

```C#
if(Interlocked.CompareExchange(ref _isDisposed, 1, 0) == 0)
    _service.ReleaseLock(this);
```

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Yeah we actually don't need to Flush anymore since the lifetime scope is being disposed of in this very method anyway.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
Piedone: That's kind of interesting because the transaction manager should be thread-local since there is a separate lifetime scope for each lock operation (i.e. each thread) and nothing should prevent multiple transactions to be handled in parallel (since this is also what happens during requests...).

It seems to me that there is some more involved issue here that a synchronization would just sweep under the rug. Maybe the recent changes regarding session management caused this somehow?

sfmskywalker: Your comment seems to have disappeared on GitHub, so replying by email. So you’re saying the transaction manager should be thread local? But how would that be? Do you know?

src/Orchard.Web/Modules/Orchard.TaskLease/Services/ITaskLeaseService.cs
```diff
@@ -7,6 +6,7 @@ namespace Orchard.TaskLease.Services {
     /// Describes a service to save and acquire task leases. A task lease can't be acquired by two different machines,
     /// for a specific amount of time. Optionnally a State can be saved along with the lease.
     /// </summary>
+    [Obsolete("Use Orchard.Tasks.Locking.IDistributedLockService.AcquireForMachine instead.")]
```
DanielStolt: Should be `AcquireLockForMachine`.

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
```diff
@@ -25,30 +29,34 @@ IFeatureManager featureManager
...
-            }
-
-            foreach (var feature in _dataMigrationManager.GetFeaturesThatNeedUpdate()) {
-                try {
-                    _dataMigrationManager.Update(feature);
-                }
-                catch (Exception e) {
-                    Logger.Error("Could not run migrations automatically on " + feature, e);
+            DistributedLock @lock;
+            if(_distributedLockService.TryAcquireLockForMachine(GetType().FullName, TimeSpan.FromMinutes(30), TimeSpan.FromMilliseconds(250), out @lock)) {
```
DanielStolt: Shouldn't we use the per-thread variant in this case?

src/Orchard/Environment/IThreadProvider.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Orchard.Environment {
+
+    /// <summary>
+    /// Describes a service which returns the managed thread ID of the current thread.
+    /// </summary>
+    public interface IThreadProvider {
```
DanielStolt: Really? :) An abstraction for getting the ID of the current thread..??

src/Orchard/Environment/ThreadProvider.cs
```diff
@@ -0,0 +1,9 @@
+using System.Threading;
+
+namespace Orchard.Environment {
+    public class ThreadProvider : IThreadProvider {
```
DanielStolt: You're killing me..!

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        public int Create() {
+            SchemaBuilder.CreateTable("DistributedLockRecord", table => table
+                .Column<int>("Id", column => column.PrimaryKey().Identity())
+                .Column<string>("Name", column => column.NotNull().WithLength(256))
+                .Column<string>("MachineName", column => column.WithLength(256))
+                .Column<int>("ThreadId", column => column.Nullable())
+                .Column<int>("Count")
+                .Column<DateTime>("CreatedUtc")
+                .Column<DateTime>("ValidUntilUtc"));
```
DanielStolt: Should it perhaps be possible to specify NULL for this for an indefinite lock?

src/Orchard/Tasks/Locking/Records/DistributedLockRecord.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Orchard.Tasks.Locking.Records {
+    public class DistributedLockRecord {
+        public virtual int Id { get; set; }
+        public virtual string Name { get; set; }
+        public virtual string MachineName { get; set; }
+        public virtual int? ThreadId { get; set; }
+        public virtual int Count { get; set; }
+        public virtual DateTime CreatedUtc { get; set; }
+        public virtual DateTime ValidUntilUtc { get; set; }
```
DanielStolt: Perhaps nullable?

src/Orchard/Tasks/Locking/Services/DistributedLock.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string MachineName { get; private set; }
+        public int? ThreadId { get; private set; }
+
+        // This will be called at least and at the latest by the IoC container when the request ends.
+        public void Dispose() {
+            if(Interlocked.CompareExchange(ref _isDisposed, 1, 0) == 0)
+                _service.ReleaseLock(this);
+        }
+
+        public static DistributedLock ForMachine(IDistributedLockService service, string name, string machineName, string lockId) {
```
DanielStolt: Maybe move the statics to the top of the class?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,204 @@
...
+            }
+            catch (Exception ex) {
+                Logger.Error(ex, "Error while trying to acquire a lock named '{0}'.", name);
+                throw;
+            }
+
+            return null;
+        }
+
+        private DistributedLockRecord AcquireLockRecord(string name, TimeSpan maxValidFor, string machineName, int? threadId) {
```
DanielStolt: Not sure I like the name `AcquireLockRecord`... could this maybe be considered another overload of `AcquireLockInternal`?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,204 @@
...
+                    return threadId != null
+                        ? DistributedLock.ForThread(this, name, machineName, threadId.Value, record.Id.ToString())
+                        : DistributedLock.ForMachine(this, name, machineName, record.Id.ToString());
+                }
+            }
+            catch (Exception ex) {
+                Logger.Error(ex, "Error while trying to acquire a lock named '{0}'.", name);
+                throw;
+            }
+
```
DanielStolt: Would be good with a `Logger.Debug()` indicating that the timeout expired.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,204 @@
...
+                    Logger.Error(ex, "An non-fatal error occurred while trying to dispose a distributed lock with name '{0}' and ID {1}.", @lock.Name, @lock.Id);
+                }
+                finally {
+                    childLifetimeScope.Dispose();
+                }
+            }
+        }
+
+        private bool TryAcquireLock(string name, TimeSpan maxValidFor, TimeSpan? timeout, string machineName, int? threadId, out DistributedLock @lock) {
+            @lock = AcquireLockInternal(name, maxValidFor, machineName, threadId, timeout.GetValueOrDefault());
```
DanielStolt: `GetValueOrDefault()` here.. I think it would be more readable to be explicit and provide `TimeSpan.Zero` if `timeout` is null.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,204 @@
...
+        public bool TryAcquireLockForThread(string name, TimeSpan maxValidFor, TimeSpan? timeout, out DistributedLock @lock) {
+            return TryAcquireLock(name, maxValidFor, timeout, GetMachineName(), GetThreadId(), out @lock);
+        }
+
+        public DistributedLock AcquireLockForThread(string name, TimeSpan maxValidFor, TimeSpan? timeout) {
+            return AcquireLock(name, maxValidFor, timeout, GetMachineName(), GetThreadId());
+        }
+
+        public void ReleaseLock(DistributedLock @lock) {
+            lock (_transactionManagerLock) {
```
DanielStolt: Can this critical section be made smaller? Looks like it's protecting a lot more than just the accessing of `_transactionManager`.

src/Orchard/Tasks/Locking/Services/DistributedLock.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// Represents a distributed lock. />
+    /// </summary>
+    public class DistributedLock : IDisposable {
+        private IDistributedLockService _service;
+        private int _isDisposed;
+
+        private DistributedLock() {
+        }
+
+        public string Id { get; private set; }
```
DanielStolt: Why is this a string, shouldn't it be an int? The code on line 53 seems to be assuming so.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,204 @@
...
+
+            throw new TimeoutException("Could not acquire a lock within the specified amount of time.");
+        }
+
+        private DistributedLock AcquireLockInternal(string name, TimeSpan maxValidFor, string machineName, int? threadId, TimeSpan? timeout = null) {
+            try {
+                DistributedLockRecord record = null;
+                var acquired = Poll(() => (record = AcquireLockRecord(name, maxValidFor, machineName, threadId)) != null, timeout);
+
+                if (acquired) {
```
DanielStolt: If any code in this block throws, we'll be left with an orphan lock in the DB which nobody knows about... not sure if we need to mitigate, but might be worth a thought..

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,204 @@
...
+            catch (Exception ex) {
+                Logger.Error(ex, "Error while trying to acquire a lock named '{0}'.", name);
+                throw;
+            }
+
+            return null;
+        }
+
+        private DistributedLockRecord AcquireLockRecord(string name, TimeSpan maxValidFor, string machineName, int? threadId) {
+            lock (_transactionManagerLock) {
```
DanielStolt: Again, can this critical section be made smaller, if it only needs to protect against concurrent access of `ITransactionManager`?

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
sfmskywalker: If multiple threads would potentially access this, then probably so. But I don't know that is the case, since this does not occur from a background task, but in response to the `IOrchardShellEvents.Activated` event. That being said, perhaps it wouldn't hurt to use the thread scoped variant instead.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -0,0 +1,204 @@
...
+            if (@lock != null)
+                return @lock;
+
+            throw new TimeoutException("Could not acquire a lock within the specified amount of time.");
+        }
+
+        private DistributedLock AcquireLockInternal(string name, TimeSpan maxValidFor, string machineName, int? threadId, TimeSpan? timeout = null) {
+            try {
+                DistributedLockRecord record = null;
+                var acquired = Poll(() => (record = AcquireLockRecord(name, maxValidFor, machineName, threadId)) != null, timeout);
```
DanielStolt: Very elegantly coded!

src/Orchard/Environment/IThreadProvider.cs
sfmskywalker: Yes, we need to be able to fake the Thread ID for unit testing.

src/Orchard/Environment/ThreadProvider.cs
sfmskywalker: Why, because I made a choice you would have made differently? Perhaps it seems overkill to provide an abstraction for just accessing a managed thread ID, but it does greatly simplify testing code if one can control the thread ID using a mock. Now, I haven't written a unit test to test logic used with the thread ID, but I will, and then I will be happy I did this. :)

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
sfmskywalker: Good question, I don't know. On the one hand, it would allow even greater control over the duration of a lock, but at the same time it would be risky when acquired for a thread that is relatively short-lived (as opposed to a machine name, in which case it's not as devastating if code fails to release the lock, since the locking code is re-entrant). On the other hand, anyone can shoot themselves in the foot by using a far away date anyway, and if the intention is to have an eternal lock until explicitly released, then it may be nicer to be explicit about it by passing in NULL for `ValidUntilUtc`. If you agree, I'll change.

src/Orchard/Tasks/Locking/Records/DistributedLockRecord.cs
sfmskywalker: Yes, see former note and let me know.

src/Orchard/Tasks/Locking/Services/DistributedLock.cs
sfmskywalker: The idea behind this is that although we are currently storing an ID value of the `DistributedLockRecord` record (which happens to be an `Int32`), the `DistributedLock` itself is agnostic of the actual storage device. Should that ever change, we would not be tied to an Int32 (because maybe other storage implementations want to use GUIDs for example. Anyway, that is all theoretic. In practice I doubt it will ever change, but I still think it's good that the ID property of the `DistributedLock` is agnostic of `DistributedLockRecord`. That being said, I don;t really care, so if you feel strongly about using an `Int32` I would be happy to change it

src/Orchard/Tasks/Locking/Services/DistributedLock.cs
sfmskywalker: Done

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Yea will check. In theory each new lifetime scope should create a new transaction manager, so we are not sure if we need the lock at all. Perhaps it's the way the services are registered in the testing environment. Will investigate (as discussed offline).

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: That's true actually. It is a very subtle control mechanism, so being explicit may help a bit.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Thanks :)

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Looking at the code, it seems extremely unlikely that at this point something will throw (the Logger, maybe?), so nesting another `try/catch` seems a bit overkill. Unless you deem an exception more probably than I think?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Yes, and remove the `Logger.Debug` call on line 79 as that would become redundant.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Same here, but I don't like an overloaded version either since this one does something different - it is focused on retrieving, creating and updating the lock record. So it (subjectively) made more sense to use the word `Record`. I also considered naming it something like `GetOrCreateOrUpdateLockRecord`, but that looks even worse. I was also considering a potential future update where this method becomes protected virtual for sub classes to use a different means of storage (maybe some shared cache or file system), so in that light the word 'Internal' didn't seem right either. That being said, in such scenarios the 'Record' makes even less sense since sub classes wouldn't even be bothered with DB related entities.

src/Orchard/Environment/ThreadProvider.cs
DanielStolt: Gah... I hate complicating design just because you need something to unit test.. but I don't have a better suggestion, so, fair point.

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
DanielStolt: I see your reasoning, and I agree.

src/Orchard/Tasks/Locking/Services/DistributedLock.cs
DanielStolt: A very good point.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
DanielStolt: Nah maybe not worth it.

src/Orchard/Environment/OrchardStarter.cs
```diff
@@ -67,6 +67,7 @@ public static class OrchardStarter {
             builder.RegisterType<ViewsBackgroundCompilation>().As<IViewsBackgroundCompilation>().SingleInstance();
             builder.RegisterType<DefaultExceptionPolicy>().As<IExceptionPolicy>().SingleInstance();
             builder.RegisterType<DefaultCriticalErrorProvider>().As<ICriticalErrorProvider>().SingleInstance();
+            builder.RegisterType<ThreadProvider>().As<IThreadProvider>().SingleInstance();
```
Jetski5822: Is this right to be SingleInstance?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
DanielStolt: But that's how many overloads work. One does the main work, while others decorate and defer to that one. Would make perfect sense in this case wouldn't it?

DanielStolt: Or if we don't want to consider them overloads, maybe `WriteLockRecord` would be sufficiently descriptive even though technically it can both read, insert or update a row depending on state?

src/Orchard/Tasks/Locking/Migrations/FrameworkMigrations.cs
sfmskywalker: Cool, will update.

src/Orchard/Environment/ThreadProvider.cs
sfmskywalker: Good man.

src/Orchard/Environment/OrchardStarter.cs
sfmskywalker: Technically it is correct, since the service does not hold on to any data - it merely provides access to the static `Thread.CurrentThread` object:

```C#
public class ThreadProvider : IThreadProvider {
    public int GetCurrentThreadId() {
        return Thread.CurrentThread.ManagedThreadId;
    }
}
```

If it's ethical, that's up for debate I suppose.

src/Orchard/Environment/ThreadProvider.cs
Piedone: I'd rather call it ThreadIdAccessor. It's not providing anything (other than access to the ID :-)).

Though I'm not sure it's too useful to keep track of the thread ID: in ASP.NET if the code is async multiple threads can handle the same request even.

DanielStolt:
Oh no you DIDN'T! ;)

sfmskywalker:
:see_no_evil:

sebastienros:
You don't need the Thread id but the process id, right ?

src/Orchard.Web/Modules/Orchard.TaskLease/Services/TaskLeaseService.cs
```diff
@@ -9,6 +9,7 @@ namespace Orchard.TaskLease.Services {
     /// <summary>
     /// Provides a database driven implementation of <see cref="ITaskLeaseService" />
     /// </summary>
+    [Obsolete("Use Orchard.Tasks.Locking.DistributedLockService and the AcquireLockForMachine/TryAcquireLockForMachine methods instead.")]
```
sebastienros: Not necessary IMO

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
```diff
@@ -25,30 +29,34 @@ IFeatureManager featureManager
...
-            }
-
-            foreach (var feature in _dataMigrationManager.GetFeaturesThatNeedUpdate()) {
-                try {
-                    _dataMigrationManager.Update(feature);
-                }
-                catch (Exception e) {
-                    Logger.Error("Could not run migrations automatically on " + feature, e);
+            DistributedLock @lock;
+            if(_distributedLockService.TryAcquireLockForThread(GetType().FullName, TimeSpan.FromMinutes(30), TimeSpan.FromMilliseconds(250), out @lock)) {
```
sebastienros: This kind of delays should be available as public properties so that they can be reconfigured from host settings.

sebastienros:
I read why you need the service, for testing. But let me disagree with the fact that you need a thread id in the first place. Please convince me if I miss something, I probably do. Isn't the idea of a distributed lock that a single thread will be able to execute a task across multiple machine? Don't answer this stupid question ;) So let's say I have multiple tenants, on multiple machines. You can prevent on a single machine that the two tasks won't collide, for the same tenant, with a lock on the tenant name + task name. If there is not two threads for the same tenant then it goes on an create a record for the multi-machine scenario, and the tenant should be in the record because two different tenants should not block each other.

So the two asks I would have here are:
- add a Tenant column to the record
- remove the ThreadId column, and use a lock on the tenant name + task name.

src/Orchard.Web/Modules/Orchard.TaskLease/Services/TaskLeaseService.cs
sfmskywalker: So leave it on the interface but remove it from the implementing class here?

PR #5629, opened 2015-08-10
Fixes #5616 : Unable to translate content 

### LINKED ISSUE #5616: Unable to translate content
branch: 1.9.x
When trying to translate a content, I can't select translation localization.
Localization is forced to original's one.

Issue introduced by fa09e2e83a95d99eafb5afaf557eab3f74e86e9f.

TFleury:
Use a flag to allow selecting translation localization.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
@@ -50,9 +50,6 @@ namespace Orchard.Localization.Controllers {
                     existingTranslationMetadata.EditorRouteValues);
             }
 
-            var contentItemTranslation = _contentManager.New<LocalizationPart>(masterContentItem.ContentType);
-            contentItemTranslation.MasterContentItem = masterContentItem;
-
             // build the editor using the master content item so that
             // the form is pre-populated with the original values
             
--- a/src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
@@ -55,8 +55,9 @@ namespace Orchard.Localization.Controllers {
 
             // build the editor using the master content item so that
             // the form is pre-populated with the original values
-
+            
             var content = _contentManager.BuildEditor(masterContentItem);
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Localization/Views/EditorTemplates/Parts/Localization.ContentTranslations.Edit.cshtml
```diff
@@ -7,7 +7,7 @@
     <label for="@Html.FieldIdFor(m => m.SelectedCulture)">@T("Content Localization")</label>
     <div>
         @*Brand new content item*@
-        @if (Model.ContentItem.ContentItem.Id == 0) {
+        @if (Model.ContentItem.ContentItem.Id == 0 || (bool)(ViewData["Translating"] ?? false)) {
```
sebastienros: I think you need to remove this

src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
```diff
@@ -50,13 +50,11 @@ public class AdminController : Controller, IUpdateModel {
...
 
-            var contentItemTranslation = _contentManager.New<LocalizationPart>(masterContentItem.ContentType);
-            contentItemTranslation.MasterContentItem = masterContentItem;
-
             // build the editor using the master content item so that
             // the form is pre-populated with the original values
-
+            
             var content = _contentManager.BuildEditor(masterContentItem);
+            ViewData["Translating"] = true;
```
sebastienros: I think you can remove this

TFleury:
I agree on the fact that it's not very clean, but I didn't found a better way to know if we are editing or translating the original content item.

We could also look at RouteValues...

TFleury:
Another way should be to *Clone* MasterContentItem and update LocalizationPart (reset culture and set MasterContentItem).

The main drawback with this method is that a new ContentItem is created in DB even if you don't save your changes.

urbanit:
You are a life saver @TFleury ;-)

sebastienros:
See issue for more details on how to actually fix the issue. I actually ended up with the same conclusion as you. Thanks.

PR #5659, opened 2015-08-19
Show Default layer by default in Widgets admin

### LINKED ISSUE #5658: Default widget layer is Anonymous
Since ordering the widget layers in alpha order, the default layer when loading Widgets is now Anonymous rather than Default, causing confusion when adding a widget and expecting to see it on the front end.

planetClaire:
Resolves #5658


DanielStolt:
Good catch and neat fix - thanks.

PR #5664, opened 2015-08-20
Fix for Lucene throwing exception when maxClauseCount exceeds 1024

kumards:
When search index gets built with more than 1024 contents, Lucene thrown an exception "maxClauseCount exceeds 1024....."

I thought we had max record of 50 in the IndexingTaskExecutor class implementation.

### DIFF
--- a/src/Orchard.Web/Modules/Lucene/Services/LuceneIndexProvider.cs
+++ b/src/Orchard.Web/Modules/Lucene/Services/LuceneIndexProvider.cs
@@ -27,6 +27,8 @@ namespace Lucene.Services {
         private readonly string _basePath;
         private ILuceneAnalyzerProvider _analyzerProvider;
 
+        private static int MAX_DOC_COUNT = 1000;
+
         public static readonly Version LuceneVersion = Version.LUCENE_29;
         public static readonly DateTime DefaultMinDateTime = new DateTime(1980, 1, 1);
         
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Lucene/Services/LuceneIndexProvider.cs
```diff
@@ -131,11 +133,17 @@ public class LuceneIndexProvider : IIndexProvider {
             Delete(indexName, indexDocuments.Select(i => i.ContentItemId));
 
             using (var writer = new IndexWriter(GetDirectory(indexName), _analyzerProvider.GetAnalyzer(indexName), false, IndexWriter.MaxFieldLength.UNLIMITED)) {
-                foreach (var indexDocument in indexDocuments) {
-                    var doc = CreateDocument(indexDocument);
-
-                    writer.AddDocument(doc);
-                    Logger.Debug("Document [{0}] indexed", indexDocument.ContentItemId);
+                var pageCount = (int)Math.Ceiling((decimal)indexDocuments.Count() / MAX_DOC_COUNT);
+                for (int i = 0; i < pageCount; i++)
```
sebastienros: Please fix the formatting, K&R.
If you are using Visual Studio, you can install Rebracer which will do it automatically when loading Orchard.

sebastienros:
Would you mind adding a unit test which would make the issue obvious, and also ensure that it's not broken again.

Thanks.

sebastienros:
Thanks, I actually hit the issue today and decided to fix it.
https://github.com/OrchardCMS/Orchard/commit/b52a846c5fd91ee54240c4e532c857606409692f

PR #5671, opened 2015-08-21
fixes #4352 : orphaned awaitingactivities

### LINKED ISSUE #4352: Editing/Saving a workflow re-creates activity records and breaks existing workflow records
__BuzzKeith__ created:
https://orchard.codeplex.com/workitem/20523

When you edit a workflow, the existing activityrecords are cleared, and then re-created.  This leaves any existing workflowrecords stuck with their awaitingactivityrecords pointing at deleted activityrecords.

This is made particularly noticeable due to the fact that to save settings for any particular activity (in this instance an email template - which would be common to update) you must save the workflow.

Where timers are not used I'm guessing that this is far less likely to happen (if at all), however where timers are used this is a problem.

I have been testing in 1.7.2 however looking at the source it appears to be the same in 1.x

j3ffb:
Update AwaitingActivities after saving a Workflow definition.

Fix for #4352

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
@@ -23,6 +23,7 @@ using Orchard.UI.Notify;
 using Orchard.Workflows.Models;
 using Orchard.Workflows.Services;
 using Orchard.Workflows.ViewModels;
+using Orchard.Workflows.Helpers;
 
 namespace Orchard.Workflows.Controllers {
     [ValidateInput(false)]
@@ -234,7 +235,7 @@ namespace Orchard.Workflows.Controllers {
                 dynamic activity = new JObject();
                 activity.Name = x.Name;
                 activity.Id = x.Id;
-                activity.ClientId = x.Name + "_" + x.Id;
+                activity.ClientId = x.GetClientId();
                 activity.Left = x.X;
                 activity.Top = x.Y;
                 activity.Start = x.Start;
@@ -259,7 +260,7 @@ namespace Orchard.Workflows.Controllers {
 
         [HttpPost, ActionName("Edit")]
         [FormValueRequired("submit.Save")]
-        public ActionResult EditPost(int id, string localId, string data) {
+        public ActionResult EditPost(int id, string localId, string data, bool clearWorkflows) {
             if (!Services.Authorizer.Authorize(StandardPermissions.SiteOwner, T("Not authorized to edit workflows")))
                 return new HttpUnauthorizedResult();
 
@@ -275,7 +276,6 @@ namespace Orchard.Workflows.Controllers {
             var activitiesIndex = new Dictionary<string, ActivityRecord>();
 
             workflowDefinitionRecord.ActivityRecords.Clear();
-            workflowDefinitionRecord.WorkflowRecords.Clear();
 
             foreach (var activity in state.Activities) {
                 ActivityRecord activityRecord;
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
```diff
@@ -303,6 +303,19 @@ IFormManager formManager
                 });
             }
 
+            foreach (var workflowRecord in workflowDefinitionRecord.WorkflowRecords) {
```
jtkech: @j3ffb just to say that now, just before, the workflowDefinitionRecord.WorkflowRecords collection is already cleared. This since this commit: https://github.com/OrchardCMS/Orchard/commit/1b8e00a9c60cddc0d1e69d867abb44bac1526e9f

Now, when you edit and save a workflow, if they are some running instances, a confirm dialog is displayed. If confirmed, all running instances are deleted, and their related awaiting activities...

Best

j3ffb: @jtkech ok I see. Perhaps the dialog should be asking 'You have running instances of this workflow, do you want to stop them?' and then pass their response. For a small change, such as a change of email content, you do not want your long running workflows to be deleted.

jtkech: @j3ffb I agree, it would be better

Moreover, if I well understood, here it seems that you can keep running instances even if an activity has been deleted in the definition. I did not analyze possible particular cases, but it would be nice. Maybe too complex to do the same even if an activity has been added :). Anyway, your implementation seems to be better

Best

j3ffb:
Updated to give user a choice to stop running workflows.

src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
```diff
@@ -303,9 +303,27 @@ IFormManager formManager
                 });
             }
 
+            if (clearWorkflows) 
+                workflowDefinitionRecord.WorkflowRecords.Clear();
```
sebastienros: Is it really deleting the workflow records or just removing them from the list instead?

src/Orchard.Web/Modules/Orchard.Workflows/Helpers/ActivityRecordExtensions.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Orchard.Workflows.Models;
+
+namespace Orchard.Workflows.Helpers {
```
sebastienros: Can you add a method `GetClientId()` to the record itself?

src/Orchard.Web/Modules/Orchard.Workflows/Views/Admin/Edit.cshtml
```diff
@@ -74,23 +75,41 @@
...
         $.ajax({
             url: stateUrl + "/" + $("#id").val(),
             async: false,
             success: function(state) {
-                if(state.isRunning && !confirm($("#confirm-delete-instances").val())) {
-                    e.preventDefault();
+                if (state.isRunning) {
+                    var dialog = $('<p>' + $("#confirm-delete-instances").val() + '</p>').dialog({
+                        buttons: {
+                            "Yes": function() { $('form').trigger('submit', [true, true]); },
```
sebastienros: Needs to be translated

src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
```diff
@@ -303,9 +303,27 @@ IFormManager formManager
...
             }
 
+            if (clearWorkflows) 
+                workflowDefinitionRecord.WorkflowRecords.Clear();
+            else {
+                foreach (var workflowRecord in workflowDefinitionRecord.WorkflowRecords) {
+                    // Update any awaiting activity records with the new activity record.
+                    foreach (var awaitingActivityRecord in workflowRecord.AwaitingActivities) {
+                        var activityRecord = awaitingActivityRecord.ActivityRecord;
+                        if (activitiesIndex.ContainsKey(activityRecord.ClientId())) awaitingActivityRecord.ActivityRecord = activitiesIndex[activityRecord.ClientId()];
```
sebastienros: Always include the curly braces on `if/else`

sebastienros: Please introduce a local `var clientId`

src/Orchard.Web/Modules/Orchard.Workflows/Controllers/AdminController.cs
j3ffb: Yes as it has a cascade CascadeAllDeleteOrphan on the list.

src/Orchard.Web/Modules/Orchard.Workflows/Views/Admin/Edit.cshtml
```diff
@@ -74,23 +75,41 @@
...
         $.ajax({
             url: stateUrl + "/" + $("#id").val(),
             async: false,
             success: function(state) {
-                if(state.isRunning && !confirm($("#confirm-delete-instances").val())) {
-                    e.preventDefault();
+                if (state.isRunning) {
+                    var dialog = $('<p>' + $("#confirm-delete-instances").val() + '</p>').dialog({
+                        buttons: {
+                            '@T("Yes")': function() { $('form').trigger('submit', [true, true]); },
```
sebastienros: This also needs to be JavascriptEncoded, but that's fine for now. Can be done in another PR.

Issue #4409, opened 2014-03-23
Taxonomy Autocomplete - load terms via service and maintain ordering


orchardbot:
@j3ffb created:
https://orchard.codeplex.com/workitem/20580

Currently the taxonomy autocomplete is loaded via a list on the page, this can cause the page to load slowly if there are a high number of terms.  The order of the terms is also not maintained.

I have changed the module to load the terms via a new service and build the hidden input fields to maintain the order of the terms.

I will submit a pull request with my changes.

PR #5674, opened 2015-08-21
#4409: Use service for autocomplete enabled taxonomy field and maintain ordering

j3ffb:
For Issue #4409.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TagsController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TagsController.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Web.Http;
 using Orchard.ContentManagement;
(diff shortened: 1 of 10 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
```diff
@@ -185,6 +185,12 @@ public class TaxonomyService : ITaxonomyService {
             return TermPart.Sort(result);
         }
 
+        public int GetTermsCount(int taxonomyId) {
```
sebastienros: I don't think we need that with the feature. The only location where it's used we already have the full list loaded.

src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TaxonomyFieldDriver.cs
```diff
@@ -70,9 +70,11 @@ public class TaxonomyFieldDriver : ContentFieldDriver<TaxonomyField> {
                     DisplayName = field.DisplayName,
                     Name = field.Name,
                     Terms = terms,
+                    SelectedTerms = appliedTerms.Select(t => t.Value),
                     Settings = settings,
-                    SingleTermId = terms.Where(t => t.IsChecked).Select(t => t.Id).FirstOrDefault(),
-                    TaxonomyId = taxonomy != null ? taxonomy.Id : 0
+                    SingleTermId = appliedTerms.Select(t => t.Key).FirstOrDefault(),
+                    TaxonomyId = taxonomy != null ? taxonomy.Id : 0,
+                    HasTerms = taxonomy != null && _taxonomyService.GetTermsCount(taxonomy.Id) > 0
```
sebastienros: Can be `terms.Any()`

src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TagsController.cs
```diff
@@ -0,0 +1,60 @@
...
+        private static TagDto CreateTagDto(TermPart term, bool leavesOnly, IEnumerable<TermPart> terms) {
+            return new TagDto {
+                value = term.Id,
+                label = term.Name,
+                disabled = !term.Selectable || (leavesOnly && terms.Any(t => t.Path.Contains(term.Path + term.Id))),
+                levels = term.GetLevels()
+            };
+        }
+    }
+    public class TagDto {
```
sebastienros: Separate file

src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TagsController.cs
```diff
@@ -0,0 +1,60 @@
...
+        public string label { get; set; }
```
sebastienros: Can you use attributes to define the serialization format (lower case)

src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TagsController.cs
```diff
@@ -0,0 +1,60 @@
...
+            ITaxonomyService taxonomyService,
+            IContentManager contentManager) {
+            _taxonomyService = taxonomyService;
+            T = NullLocalizer.Instance;
+            _contentManager = contentManager;
+            Logger = NullLogger.Instance;
+        }
+
+        public IEnumerable<TagDto> Get(int taxonomyId, bool leavesOnly, string query) {
+            if (string.IsNullOrEmpty(query)) return new List<TagDto>();
```
sebastienros: There should be some security check here, at least the same as what the content editor is requiring to list taxonomy terms. Which means at least Admin access, plus other taxonomy specific ones.

src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TaxonomyFieldDriver.cs
j3ffb: `terms` will only be populated when not using autocomplete, otherwise we don't bother loading all the terms

src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TagsController.cs
j3ffb: ok, will check for admin access

src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
j3ffb: We need this for autocomplete as we do not load the full list of terms bc they are not required for building the shape

PR #5677, opened 2015-08-22
Enables gulp watch of Assets.json

dcinzona:
Will restart the watch task when Assets.json is changed.

### DIFF
--- a/src/Gulpfile.js
+++ b/src/Gulpfile.js
@@ -13,7 +13,8 @@ var glob = require("glob"),
     uglify = require("gulp-uglify"),
     rename = require("gulp-rename"),
     concat = require("gulp-concat"),
-    header = require("gulp-header");
+    header = require("gulp-header"),
+    fs = require("fs");
 
 /*
 ** GULP TASKS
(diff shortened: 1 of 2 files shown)


src/Gulpfile.js
```diff
@@ -37,32 +38,53 @@ gulp.task("rebuild", function () {
...
 /*
 ** ASSET GROUPS
 */
-
 function getAssetGroups() {
-    var assetManifestPaths = glob.sync("Orchard.Web/{Core,Modules,Themes}/*/Assets.json");
+    var assetManifestPaths = glob.sync("Orchard.Web/{Core,Modules,Themes}/*/Assets.json", {});
     var assetGroups = [];
     assetManifestPaths.forEach(function (assetManifestPath) {
-        var assetManifest = require("./" + assetManifestPath);
```
dcinzona: Can't use require because that loads the manifest into memory and will not update even if the watch task detects that the manifest has changed.

src/Gulpfile.js
```diff
@@ -37,32 +38,53 @@ gulp.task("rebuild", function () {
...
+        var file = './' + assetManifestPath;
+        var json = fs.readFileSync(file, 'utf8');
+        assetManifest = eval(json);
```
dcinzona: Using `eval` is not my preferred way of doing this but it seems as though this was the only way to get JS to convert the file contents to a valid JSON object.

PR #5680, opened 2015-08-23
Implemented Initializing status.

sfmskywalker:
Issue: #5650

This PR adds an additional `Initializing` status to the `TenantState` enum and provides an appropriate response to the browser in case a request is made while the tenant is in the `Initializing` state.


Jetski5822:
Ill need to remember to bring this across to Brochard!

sfmskywalker:
Yeah, and not just this I suppose. I imagine there'll be some other things that change over time.

src/Orchard/Environment/ShellBuilders/ShellContextFactory.cs
```diff
@@ -103,7 +103,8 @@ public class ShellContextFactory : IShellContextFactory {
                 Features = new[] {
                     new ShellFeature { Name = "Orchard.Setup" },
                     new ShellFeature { Name = "Shapes" },
-                    new ShellFeature { Name = "Orchard.jQuery" },
+                    new ShellFeature { Name = "Orchard.Recipes" },
+                    new ShellFeature { Name = "Orchard.jQuery" }
```
sebastienros: How is it related to the PR ?

sfmskywalker: It's not actually, good catch. It's a leftover of some of the refactoring I did but forgot to undo this. I'll get rid of it.

PR #5687, opened 2015-08-24
Fix user defined outcome connections in Branching activities

### LINKED ISSUE #5686: Workflow connections broken for Branch, Exclusive Branch etc #dev branch
Since changing the way outcome connections are made to allow for localization, activities that accept outcomes from the user are broken.  Ref 2a57a153632d31a8692f2ad437e6d5e897e33c79

planetClaire:
Fixes #5686


sebastienros:
Shouldn't it target 1.9.x if this is a bug fix?

planetClaire:
1.9.x is ok - but in the dev branch, branch outcome labels are broken after this commit https://github.com/OrchardCMS/Orchard/commit/2a57a153632d31a8692f2ad437e6d5e897e33c79 which I'm guessing is necessary for localized sites.  I don't have a localized site, but after the commit mentioned, the branch labels are broken.  See the original issue: https://orchard.codeplex.com/workitem/20834

Issue #3402, opened 2013-04-04
Tokens do not handle collections


orchardbot:
@j3ffb created:
https://orchard.codeplex.com/workitem/19573

There is no way to access content within a collection using tokens.

I have created a Collection Token that will handle a collection of IContent and apply a tokenized string to each element.

Collection.ForEach:<Tokenised string to apply to each element>

http://orchard.codeplex.com/SourceControl/network/forks/j3ffb/CollectionTokens/contribution/3952

j3ffb:
@pszmyd Following on from your comment, if I understood it correctly, I have used `List` instead of Collection as the name and also implemented a few of the tokens you requested. I could not implement `Skip` and `Take` as I couldn't see how to chain on a token that takes parameters. I wanted to chain back onto the List with the reduced list, if that would work??

pszmyd:
`Chain` method does not allow lambda-based token handling, like `Token` method does, unfortunately. This is something already on the feature list. I might come with an update soon.

PR #5691, opened 2015-08-25
#3402: Implemented List tokens.

j3ffb:
Implementation for #3402

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Tokens/Providers/ListTokens.cs
+++ b/src/Orchard.Web/Modules/Orchard.Tokens/Providers/ListTokens.cs
@@ -23,85 +23,96 @@ namespace Orchard.Tokens.Providers {
                 .Token("Sum:*", T("Sum:<Tokenized text>"), T("Sum each element in the list by applying the tokenized text."))
                 .Token("First:*", T("First:<Tokenized text>"), T("Apply the tokenized text to the first element."))
                 .Token("Last:*", T("Last:<Tokenized text>"), T("Apply the tokenized text to the last element."))
-                .Token("Count", T("Count"), T("Get the list count."))
-                ;
+				.Token("Count", T("Count"), T("Get the list count."))
+				.Token("ElementAt:*", T("ElementAt:<index>,<Tokenized text>"), T("Apply the tokenized text to the element at its index."))
+				;
 	    }
 	 
 	    public void Evaluate(EvaluateContext context) {
-			Func<string, IContent, string> getValue = (t, i) => _tokenizer().Replace(t, new { Content = i }, new ReplaceOptions { Encoding = ReplaceOptions.NoEncode });
+			Func<string, IContent, string> tokenValue = (t, i) => _tokenizer().Replace(t, new { Content = i }, new ReplaceOptions { Encoding = ReplaceOptions.NoEncode });
 
 		    context.For<IList<IContent>>("List", () => new List<IContent>())
 			    .Token( // {List.Join:<string>[,<separator>]}
-				    token => {
-					    if (token.StartsWith("Join:", StringComparison.OrdinalIgnoreCase)) {
-						    // html decode to stop double encoding.
-						    return HttpUtility.HtmlDecode(token.Substring("Join:".Length));
-					    }
-					    return null;
-				    },
-				    (token, collection) => {
-					    if (String.IsNullOrEmpty(token)) {
+				   token => FilterToken(token, "Join:"),
+					(token, list) => {
+					    if (String.IsNullOrEmpty(token) || !list.Any()) {
 						    return String.Empty;
 					    }
-					    // Split the params to get the tokenized text and optional separator.
+					    // Split the token to get the tokenized text and optional separator.
 					    var index = token.IndexOf(',');
 					    var text = index == -1 ? token : token.Substring(0, index);
 					    if (String.IsNullOrEmpty(text)) {
 						    return String.Empty;
 					    }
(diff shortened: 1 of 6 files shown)


src/Orchard.Web/Modules/Orchard.Tokens/Tests/ListTokenTests.cs
```diff
@@ -0,0 +1,67 @@
...
+	        builder.RegisterType<TestTokenProvider>().As<ITokenProvider>();
+	        //builder.RegisterType<Work<Tokenizer>>().As<Work<ITokenizer>>();
+	        _container = builder.Build();
+	        _tokenizer = _container.Resolve<ITokenizer>();
+	    }
+	 
+	    [Test]
+	    public void TestListForEachTokens()
+	    {
+            Assert.That(_tokenizer.Replace("{List.ForEach:{Content.Id} }", new { List = new List<IContent> { new TestUser { Id = 5 }, new TestUser { Id = 10 } } }), Is.EqualTo("5 10 "));
```
sebastienros: This looks more like a `Join` than a foreach, and then it could use a separator parameter. Empty being the default.

j3ffb: Agree, updated.

src/Orchard.Web/Modules/Orchard.Tokens/Providers/ListTokens.cs
```diff
@@ -0,0 +1,106 @@
...
+	    }
+	 
+	    public Localizer T { get; set; }
+	 
+	    public void Describe(DescribeContext context) {
+	        context.For("List", T("Lists"), T("Handles lists of Content Items"))
+                .Token("Join:*", T("Join:<Tokenized text>[,<separator>]"), T("Join each element in the list by applying the tokenized text and concatenating the output with the optional separator."))
+                .Token("SumInt:*", T("SumInt:<Tokenized text>"), T("Sum each element in the list by applying the tokenized text."))
+                .Token("SumFloat:*", T("SumFloat:<Tokenized text>"), T("Sum each element in the list by applying the tokenized text."))
+                .Token("SumDecimal:*", T("SumDecimal:<Tokenized text>"), T("Sum each element in the list by applying the tokenized text."))
```
sebastienros: I would rather see a single Sum which handles decimal, then use a different token if it needs to be rounded or formatted.

src/Orchard.Web/Modules/Orchard.Tokens/Providers/ListTokens.cs
```diff
@@ -0,0 +1,106 @@
...
+	    public ListTokens(Func<ITokenizer> tokenizer) {
+	        _tokenizer = tokenizer;
+	        T = NullLocalizer.Instance;
+	    }
+	 
+	    public Localizer T { get; set; }
+	 
+	    public void Describe(DescribeContext context) {
+	        context.For("List", T("Lists"), T("Handles lists of Content Items"))
+                .Token("Join:*", T("Join:<Tokenized text>[,<separator>]"), T("Join each element in the list by applying the tokenized text and concatenating the output with the optional separator."))
```
sebastienros: I don't really like the syntax, will think about it, or let others suggest something better.

sebastienros:
@j3ffb https://cla2.dotnetfoundation.org

sebastienros:
Wouldn't it be easier for users to define lists of values as a token type like Content, Text, User are and then implement Join and other methods you made as operations ? 
Fake example is `Roles` existed `{Content.Owner.Roles.Join(', ')}`

src/Orchard.Web/Modules/Orchard.Tokens/Providers/ListTokens.cs
j3ffb: Ok I have merged them.

j3ffb:
Yes, so users would have to define their own token to chain to `List`.

    context.For<IOrder>("Order")
           .Token("Items", o => string.Join(", ", o.Items.Select(i => i.As<ITitleAspect>().Title).ToArray()))
           .Chain("List", "List", o => o.Items);

The user is then free to specify what they would like to output for each element in the list. So for example they could build up an email something like:

    Order ID: {Content.Order.Id}<br/>
    Thank you for placing the following order:<br/>
    <table><tr><th>Item</th><th>Qty</th><th>Price</th></tr>
    {Content.Order.Items.Join: <tr><td>{Content.OrderItem.Title}</td><td>{Content.OrderItem.Quantity}</td><td>{Content.OrderItem.Price}</td></tr>}
    </table>

sebastienros:
Right, can you make it work this way or is it already the case?

j3ffb:
Yes, this is how it already works.

sebastienros:
Is the result of Join chainable with Text ?

j3ffb:
Yes, it could be, if you want?

sebastienros:
Do it and I merge ;)

j3ffb:
I have taken a look into this and it looks like it could be tricky as we would have to use a `.` as a delimiter, which could also be in the `Join` text. So I'm not sure how we would know if the user is attempting to chain the result?
On reflection I'm not sure chaining the full output of `Join` to `Text` would be that beneficial, perhaps there could be a use case for `Limit`??

On the plus side, I have, however, implemented `ElementAt:<index>,<Tokenized text>` which I think could be useful.

j3ffb:
An immediate use case for the `List` token would be to chain to it from `TaxonomyField.Terms`.

Issue #2657, opened 2012-07-17
TinyMCE configuration extension point with jQuery events


orchardbot:
@Piedone created:
https://orchard.codeplex.com/workitem/18830

I've created a simple extension mechanism for TinyMCE that could be used so custom modules or themes could alter the editor's configuration. For legibility, the code is included in this discussion: http://orchard.codeplex.com/discussions/371739

Piedone:
@Jetski5822 you worked on something like this, any news?

dcinzona:
Does this conflict with the TinyMce requirement in the Layouts module?

Piedone:
What do you mean? But probably @sfmskywalker or @DanielStolt can already answer.

dcinzona:
The layouts module requires the TinyMCE module (at least it did last time I checked).  We were using CKEditor instead of TinyMCE and could not enable the Layouts module with the TinyMCE module disabled.

dcinzona:
It's in the Module.txt
Orchard.Layouts:
        Name: Layouts
        Description: Provides tools to create layouts.
        Dependencies: Common, Orchard.jQuery, Orchard.Forms, Orchard.Tokens, Orchard.MediaLibrary, TinyMCE
        Category: Layout

https://github.com/OrchardCMS/Orchard/blob/master/src/Orchard.Web/Modules/Orchard.Layouts/Module.txt

Piedone:
Ah. No I don't think it would conflict. This is about adding extension points to the TinyMCE module's editor. So if you use TinyMCE you'd be able to e.g. register new plugins, change the toolbars... If you don't use TinyMCE then nothing changes.
And of course it wouldn't impact if a module depends on TinyMCE.

dcinzona:
Hmm... interesting.  We should test.  The readme for deluxe states to disable TinyMCE and that's what caused a problem for us with Layouts.  Granted, deluxe uses TinyMCE as well, so they should be compatible, but because TinyMCE is a dependency, it may disable the layouts module as well... not sure - again, need to test.

sfmskywalker:
Yes, if you disable TinyMCE, Layouts will be disabled. Perhaps we can safely remove that dependency from module.txt in Layouts - I don't think Layouts is using any container-registered service from TinyMce.

dcinzona:
I thought @DanielStolt mentioned at one point that there was a dependency there for a reason...but I can't remember how that came up in conversation...

sfmskywalker:
[...]

PR #5692, opened 2015-08-25
Fixes #5518: Add TinyMce support for inline styles

### LINKED ISSUE #5518: TinyMCE editor strips out inline style code
We were playing around with the layouts module, which requires TinyMCE, and noticed that the editor (in any situation) will strip out inline `<style></style>` tags.  This became problematic as we override some stiles in certain instances on the site.
As such, we can't use TinyMCE and Layouts until this is fixed by the editor.  I've looked at the configs and it looks like it's set to allow all tags, but for some reason, it still gets stripped.

We are using CKEditor instead of TinyMCE.  I wish the layouts module was editor agnostic.
See issue https://github.com/OrchardCMS/Orchard/issues/5112

dcinzona:
TinyMCE will strip out anything with the `<style>` tag by default.  This enables inline style tags within the src editor

### DIFF
--- a/src/Orchard.Web/Modules/TinyMce/Scripts/orchard-tinymce.js
+++ b/src/Orchard.Web/Modules/TinyMce/Scripts/orchard-tinymce.js
@@ -24,6 +24,8 @@ tinyMCE.init({
     valid_elements: "*[*]",
     // Shouldn't be needed due to the valid_elements setting, but TinyMCE would strip script.src without it.
     extended_valid_elements: "script[type|defer|src|language]",
+    //allow inline styles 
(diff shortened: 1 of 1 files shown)


thekaveman:
I wonder if this should (can?) be a configurable setting? I have scenarios where I don't necessarily want the users introducing inline style.

dcinzona:
Not at the moment - right now TinyMCE loads the orchard specific configuration via the JS file referenced in the change. There may be an option to specify your own JS file....

dcinzona:
This fixes issue #5518

sebastienros:
Duplicate of #2657

Issue #5695, opened 2015-08-26
Feature: Check permissions on Projection queries / filters


dcinzona:
I’m going to work on adding permission support to projections.  Currently, I’m thinking there are three possible locations on where to add permissions support, at the query level, at the filter lever, or at the projection level.  Any preference?  It’s basically just a boolean value called `RespectPermissions` or maybe `CheckPermissions`…

My first thought is to put it at the Query level and edit the ProjectionManager.cs to check permissions when it gets the content items.  Content Items are pulled using 
```cs
            foreach(var contentQuery in GetContentQueries(queryRecord, queryRecord.SortCriteria.OrderBy(sc => sc.Position))) {
                contentItems.AddRange(contentQuery.Slice(skip, count));
            }
```

So perhaps adding the boolean at the Filter level may be the best option here (for customizability)? 

The other question is, this would require adding a reference to Orchard.Roles and the appropriate dependency in Module.txt.  My goal is to make this not be a breaking change, obviously... so is this even an option?  On upgrading, will it enable the dependency if required?  Perhaps adding this as a feature, where the feature requires the dependency, and then the site admin can enable the feature - triggering the dependency....Unfortunately, my experience with creating features within modules is minimal (as in 0) so this would be a good way to learn how to do that, but will also take some time and possibly lead to less than optimal code.

Is this something that the Orchard community would even entertain?

Jetski5822:
https://github.com/Jetski5822/Orchard.Projection.RoleSecurity :)

dcinzona:
omg I want to cry

dcinzona:
@Jetski5822 any reason why this is not included in Orchard by default?

dcinzona:
@Jetski5822 just tried out your module and it doesn't seem to filter projection content items based on the permissions assigned to the CI.  I've re-plublished all content items but the projection still displays every content item, event the protected ones.

Jetski5822:
Hmmm... I wrote it for version 1.7 a while ago... Might need some tweaking.

dcinzona:
I tested with 1.9.x after changing the framework to 4.5.1...but that was the only tweaking I did.

dcinzona:
Heres the branch with the module added to an otherwise vanilla 1.9.x https://github.com/dcinzona/Orchard/tree/Module/Orchard.Projections.RoleSecurity

Jetski5822:
Cool. I will check when I'm home, just on the train!

-----Original Message-----

[...]

PR #5696, opened 2015-08-26
Feature / Projection content item security

dcinzona:
Adding preliminary support for content item security when using projections.  Enabling this option will degrade performance if there are a lot of content items returned in the projection query.  This option is enabled at the Filter level:
screen shot 2015-08-26 at 11 04 46 am

Related to issue: https://github.com/OrchardCMS/Orchard/issues/5695

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
@@ -144,6 +144,9 @@ namespace Orchard.Projections.Services {
                     var allCiFiltered = contentQuery.List().FilterContentItems(_services.WorkContext.CurrentUser);
                     contentItems.AddRange(allCiFiltered.Skip(skip).Take(count));
                 }
+                else {
+                    contentItems.AddRange(contentQuery.Slice(skip, count));
+                }
             }
 
             if (queryRecord.FilterGroups.Count <= 1) {
--- a/src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
@@ -12,9 +12,10 @@ using Orchard.Projections.Descriptors.Layout;
 using Orchard.Projections.Descriptors.SortCriterion;
 using Orchard.Projections.Models;
 using Orchard.Tokens;
+using Orchard.Projections.Extensions;
 
 namespace Orchard.Projections.Services {
-    public class ProjectionManager : IProjectionManager{
+    public class ProjectionManager : IProjectionManager {
         private readonly ITokenizer _tokenizer;
         private readonly IEnumerable<IFilterProvider> _filterProviders;
         private readonly IEnumerable<ISortCriterionProvider> _sortCriterionProviders;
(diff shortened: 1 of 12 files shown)


sebastienros:
Should it be just about checking for the core `ViewContent` permission directly? And the authorization pipeline would automatically ensure roles and ownership.

But you understand that this dynamic way of doing it will break paging? 
Another approach would be to create a distinct filter which would check for the permission at the database level, still not perfect.
The ultimate solution is to __index__ the roles who can view the item, by applying the permission whenever the item is updated. Then the filter would query based on `ViewContent` and `ViewOwnContent` roles.

dcinzona:
This doesn't break paging.  The count method also includes the security check.  I've tested paging and it works as expected, page counts are updated accordingly.  My only concern with the above implemented method is performance when the total content item set is very large.

I agree, the optimal would be to index the permissions on the items and then query those in the projection as a filter.
I believe that is what Nick was doing here https://github.com/Jetski5822/Orchard.Projection.RoleSecurity but it doesn't seem to be working correctly on the 1.9.x (see branch: https://github.com/dcinzona/Orchard/tree/Module/Orchard.Projections.RoleSecurity )

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
```diff
@@ -112,9 +116,15 @@ public class ProjectionManager : IProjectionManager{
             }
 
             // aggregate the result for each group query
+            var checkPermissions = queryRecord.FilterGroups.Any(x => x.Filters.Any(y => y.CheckPermissions));
 
-            return GetContentQueries(queryRecord, Enumerable.Empty<SortCriterionRecord>())
-                .Sum(contentQuery => contentQuery.Count());
+            var count = GetContentQueries(queryRecord, Enumerable.Empty<SortCriterionRecord>())
+                .Sum(contentQuery => checkPermissions ? 
+                                contentQuery.List().FilterContentItems(_services.WorkContext.CurrentUser).Count() :
```
dcinzona: This line enables correct paging

sebastienros: What about pages > 1 when the content item which is filtered is on page 1. If it was supported then it would mean checking for ALL items up to the current page. Is that your concern?

dcinzona: So, in my tests, I set the page size to 5 and had 11 content items.  5 of which were private with different sets of permissions.  When I was not logged in, it successfully displayed 2 pages (6 content items for anonymous access).  When I was logged in, it displayed all 11 in 3 pages.

So yes, my concern is that it is checking all items, not just up to the current page... because the page is dynamic now, depending on the number of items the user is authorized to see.

If I have the following:
* contentItem1 public
* contentItem2 public
* contentItem3 private
* contentItem4 public
* contentItem5 private
* contentItem6 public
* contentItem7 private
* contentItem8 public
* contentItem9 private
* contentItem10 public
* contentItem11 private

Logged in, I would see all of them (with the proper role, obviously) split across 3 pages of 5, 5, 1 items

Logged out, I would see:
Page 1:
* contentItem1 public
* contentItem2 public
* contentItem4 public
* contentItem6 public
* contentItem8 public

Page 2:
* contentItem10 public

dcinzona:
I can try to demo this on tuesday (I say try because our firewall is pretty hard core)

dcinzona:
I have a couple of other concerns, including requiring the roles and contentpermissions modules to be enabled with the projections modules.  I may switch this up to a hybrid setup from what Nick was doing ( making it it's own module that requires roles and contentpermissions).

sebastienros:
Design explained and understood by @dcinzona, feature to come.

Issue #5647, opened 2015-08-16
The Orchard.DesignerTools does'n work


neTp9c:
WorkContext is always null, because one getting in constructor.

https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.DesignerTools/Services/ShapeTracingFactory.cs

if change class variable from WorkContext to IWorkContextAccessor all works fine.

Piedone:
Yes, the WC should be resolves from the WCA as late as possible. Just wondering, when exactly does this cause an error for you?

neTp9c:
I don't know. But it occurs for a standard orchard installation.

Clone repository (then use dev branch).
Enable Shape Tracing.
You get this error: "Object reference not set to an instance of an object.".

jtkech:
@Piedone, @sebastienros, @neTp9c,

For infos, just found that this occurs after this commit: https://github.com/OrchardCMS/Orchard/commit/83cae5f0a0d6a3000be2e52e2fcf8e80966005b4 (just before it works) where `FormsAuthenticationService` uses an `IMembershipService` in place of an `IContentManager`.

Maybe because the work context scope is created later with this new dependency injection. This because the `FormsAuthenticationService.GetAuthenticatedUser()` method is used in one of the `IWorkContextStateProvider` that are resolved in the `WorkContextImplementation` constructor...

Not sure but if I use a `Lazy<IMembershipService>` it works, in this case the `ShapeTracingFactory _workContext` is correctly initialized in the constructor, a work context object has already been set in the `httpContext.Items[_workContextKey]`, as done in `CreateWorkContextScope()`...

All of this to say from when this happens and try to understand why. But maybe we need at least to do what @neTp9c is suggesting. Using `_workContextAccessor.GetContext()` later in `ShapeTracingFactory` also solves the issue. Using a `Lazy<WorkContext>` or `Work<WorkContext>` also works...

Notice that there are many other places where a work context is intialized in a class constructor. Some times by injecting directly a `WorkContext` but without using `Lazy<> or Work<>`, other times by using a `wca.GetContext()` but in the constructor...

Best

sebastienros:
Works fine on 1.9.x so I will assume that's a dev branch issue as it is mentioned. Right now I can't make the dev branch to work for me but that's another problem.

PR #5723, opened 2015-08-31
#5647: Fixing WorkContext usage in other places

jtkech:
Same as #5647 already closed but in other places. Seems to be related to all classes that init a `WorkContext` in their constructor, and that inherit from `ShapeDisplayEvents` or `IShapeDisplayEvents`.

Here, to prevent from getting errors, in the last dev branch, when enabling `Culture Picker` and when using a `Body Editor`. This concerns `AdminDirectionalityFactory`, `AdminCultureSelectorFactory` and `TinyMceShapeDisplayEvent`

As I've mentioned in #5647, this occurs after this commit: https://github.com/OrchardCMS/Orchard/commit/83cae5f0a0d6a3000be2e52e2fcf8e80966005b4. Really, if in `FormsAuthenticationService` I use a `Lazy<IMembershipService>`, it works without changing the `WorkContext` usage (see more details in #5647)

So, even `FormsAuthenticationService` is well implemented, let me know if you want to use a `Lazy<IMembershipService>`, this not to break some other third party modules

Best

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Localization/Services/AdminCultureSelectorFactory.cs
+++ b/src/Orchard.Web/Modules/Orchard.Localization/Services/AdminCultureSelectorFactory.cs
@@ -9,22 +9,24 @@ namespace Orchard.Localization.Services {
     [OrchardFeature("Orchard.Localization.CultureSelector")]
     public class AdminCultureSelectorFactory : ShapeDisplayEvents {
         private readonly ICultureManager _cultureManager;
-        private readonly WorkContext _workContext;
+        private readonly IWorkContextAccessor _workContextAccessor;
 
         public AdminCultureSelectorFactory(
             IWorkContextAccessor workContextAccessor, 
             IShapeFactory shapeFactory,
             ICultureManager cultureManager) {
             _cultureManager = cultureManager;
-            _workContext = workContextAccessor.GetContext();
+            _workContextAccessor = workContextAccessor;
             Shape = shapeFactory;
         }
 
         dynamic Shape { get; set; }
 
         private bool IsActivable() {
+            var workContext = _workContextAccessor.GetContext();
+
             // activate on admin screen only
-            if (AdminFilter.IsApplied(new RequestContext(_workContext.HttpContext, new RouteData())))
+            if (AdminFilter.IsApplied(new RequestContext(workContext.HttpContext, new RouteData())))
                 return true;
 
             return false;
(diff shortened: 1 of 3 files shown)


sebastienros:
Can you check if https://github.com/OrchardCMS/Orchard/pull/5738 is also fixing these issues? That would be the ultimate fix then.

harmony7:
Oh great, will check these too.

jtkech:
I will try it asap

harmony7:
@sebastienros The changes made in this pull request look like they are about injecting IWorkContextAccessor instead of WorkContext in the constructor ... a better practice anyway. Personally, I don't think these should be reverted.

sebastienros:
Right but in the case of IDependency the end result is exactly the same. And I don't like the fact that there is a difference in behavior so I am expecting your change to fix the issue, just to be sure. then we can talk about what should be the recomendation about IWorkcontextAccessor vs. WorkContext.

jtkech:
@sebastienros just tried by reverting the changes in TinyMceShapeDisplayEvent, and applying @harmony7 PR. This doesn't solves the issue, I got same errors in the log file

harmony7:
I kind of felt it would be unrelated, my changes only affect components that are used during IOrchardShellEvents

jtkech:
@harmony7 I think you're right, it's just because I was notified that @sebastienros wanted to test it

For this issue (in the dev branch), I've done some tests under debugger and I've seen that the constructor of `TinyMceShapeDisplayEvent`, where the workcontext is initialized, is called as soon as this line, in `WorkContextAccessor.cs`, is executed

    _workContext = lifetimeScope.Resolve<WorkContext>();

And before

    httpContext.Items[workContextKey] = _workContext;

So, `the WorkContextAccessor GetContext()` used in `TinyMceShapeDisplayEvent` return null

As I've mentioned in #5647, this occurs after this commit: 83cae5f. Just for testing, if in `FormsAuthenticationService` I use a `Lazy<IMembershipService>`, it works...

I can't right now, but I will continue my searching this night

Best

jtkech:
@sebastienros I've found a more global solution that seems to work. Need to be validated to see if no other side effects...

Because after this commit: 83cae5f, when the workcontext and its workcontext state providers are  resolved, more class constructors are called and some of them use a workcontext accessor...

I've tried the following that seems to work. In the `WorkContextImplementation constructor`, I've used a collection of **Lazy** `IWorkContextStateProvider`

    _workContextStateProviders = componentContext.Resolve<IEnumerable<Lazy<IWorkContextStateProvider>>>();

And further

    var resolver = _workContextStateProviders.Select(wcsp => wcsp.Value.Get<T>(name)).FirstOrDefault(value => !Equals(value, default(T)));

Let me know if you agree, and if I can do the update

Best

sebastienros:
> Thanks for signing the contribution license agreement so quickly
We recognize the advance IA behind this robot

sebastienros:
I applied your changes and they worked fine, and the explanation makes sense too. This is a workaround but not a complex one so I am fine with it, Thanks.

PR #5734, opened 2015-09-03
Refactoring locking strategy

sebastienros:
Adding support for tenants
Removed ThreadId field
Removed Count field
Optimized db access by using Monitor for the local machine/tenant checks
Removed Machine/Thread scopes


src/Orchard.Tests/Tasks/DistributedLockServiceTests.cs
```diff
@@ -36,74 +36,64 @@ public class DistributedLockServiceTests : DatabaseEnabledTestsBase {
...
+            _distributedLockService.TryAcquireLock(LockName, TimeSpan.FromSeconds(60), out lock1);
 
-            var lockId = Int32.Parse(@lock.Id);
-            _distributedLockService.ReleaseLock(@lock);
-            var lockRecord = _distributedLockRepository.Get(lockId);
+            lock1.Dispose();
+            var lockRecord = _distributedLockRepository.Table.FirstOrDefault();
 
-            Assert.That(lockRecord, Is.Null);
+            Assert.IsNull(lockRecord);
```
sfmskywalker: Nice to know.

src/Orchard/Tasks/Locking/Services/DistributedLock.cs
```diff
@@ -3,44 +3,34 @@
...
+        public void IncreaseReferenceCount() {
+            _count++;
+        }
 
-        // This will be called at least and at the latest by the IoC container when the request ends.
         public void Dispose() {
-            if(Interlocked.CompareExchange(ref _isDisposed, 1, 0) == 0)
-                _service.ReleaseLock(this);
+            _count--;
+            if (_count == 0) {
```
DanielStolt: I think we need atomicity here. `Dispose()` should be thread safe.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -13,128 +14,120 @@
...
+                return l != null;
+            }
+            catch {
+                l = null;
+                return false;
+            }
         }
 
-        public void ReleaseLock(DistributedLock @lock) {
-            var childLifetimeScope = CreateChildLifetimeScope(@lock.Name);
```
sfmskywalker: The reason we created a child scope is so that we could create a new transaction with a `ReadCommitted` isolation level (without affecting the isolation level of the current transaction).
Do you think that was unnecessary? What if two machines try and acquire a lock at the same time and the system is configured to use `ReadUncommitted`?

src/Orchard/Tasks/Locking/Services/DistributedLock.cs
sebastienros: Why? How could two thread use the same lock object? They are created per dependency, so per request, so two threads can't call dispose on the same lock.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sebastienros: OK, I assumed that you created a new scope because the service was a singleton initially, so by making it an `IDependency` I could remove this. Because we can still create a new transaction in the current scope, so I think the real comment here is about creating a new transaction or not.

I think that by making the `Name` a unique constraint it would solve everything, as the unique constraint would prevent two records to be created. And actually this is what should be done in the first place. And a try/catch on the creation and commit on the Record to be sure it was correctly created and that we have the lock. Right ?

src/Orchard.Tests/Tasks/DistributedLockServiceTests.cs
sebastienros: Was not even intentional ;) Same.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sfmskywalker: Yes that makes sense, because even if the one machine got a null row back in the case of a race condition and attempts to store the record while another machine beat them to it, an exception will occur and thus the acquisition will have failed. @DanielStolt unless I'm missing something?

sebastienros: Sorry, didn't see it was in the release part. Well I assume only one machine can release a lock as a lock can only be acquired by one machine. My previous statement is still valid though.

src/Orchard.Tests/Tasks/DistributedLockServiceTests.cs
Jetski5822: I prefer Assert.That.... But that's just me.

-----Original Message-----

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
sebastienros: GTG on both Creation and Release

PR #5736, opened 2015-09-03
Fixes #5735 : ContentType name encoding in Lists

### LINKED ISSUE #5735: ContentType name encoding in Lists
ContentType's name is encoded twice on "New *ContentType*" link in *Parts.Container.Manage* view.

TFleury:


sebastienros:
Do you think we could have an `ActionLink` taking an `IHtmlString` and would not encode it when rendered?

sebastienros:
Either this or not use `ActionLink` when the text is localized, but render a link and use `Url.Action`

TFleury:
I implemented `ActionLink` with `IHtmlString` linkText.
Also done the same with `RouteLink`

sebastienros:
Much better, right ?

PR #5528, opened 2015-07-17
Fixed #5490 Exception when you get httpContext from Migrations

jersiovic:
fixes https://github.com/OrchardCMS/Orchard/issues/5490

Issue #4852, opened 2014-10-22
Wrap IOrchardShellEvents into WorkContext


orchardbot:
@Piedone created:
https://orchard.codeplex.com/workitem/21023

Currently there is no active WorkContext when IOrchardShellEvents events are fired. Thus if you implement this interface and try to use any injectable dependency in it that depends on a WorkContext being present (this could be the case with a simple Work<T> injection) then you'll get errors like this one.

To mitigate this IOrchardShellEvents could be executed in the scope of a short-lived WorkContext.

PR #5738, opened 2015-09-03
#4852: Wrap IOrchardShellEvents in a work context

harmony7:
This is a fix for #4852

This fixes some of the problems we are seeing relating to being unable to inject HttpContextBase during IOrchardShellEvents.

This has mainly been a problem when attempting to access ContentManager or site settings in Data Migrations, such as seen in #5305 #5490 #5528.

We seem to only be hitting the problem when Orchard is compiled under VS2015, but this is probably the correct behavior even when building under previous environments.

### DIFF
--- a/src/Orchard/Environment/DefaultOrchardShell.cs
+++ b/src/Orchard/Environment/DefaultOrchardShell.cs
@@ -17,7 +17,7 @@ using IModelBinderProvider = Orchard.Mvc.ModelBinders.IModelBinderProvider;
 
 namespace Orchard.Environment {
     public class DefaultOrchardShell : IOrchardShell {
-        private readonly Func<Owned<IOrchardShellEvents>> _eventsFactory;
+        private readonly IWorkContextAccessor _workContextAccessor;
         private readonly IEnumerable<IRouteProvider> _routeProviders;
         private readonly IEnumerable<IHttpRouteProvider> _httpRouteProviders;
         private readonly IRoutePublisher _routePublisher;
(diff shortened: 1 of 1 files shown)


jersiovic:
It works perfect, the problems has disappeared. 
Thank you so much!!!

sebastienros:
I think we should try your change and revert this: https://github.com/OrchardCMS/Orchard/commit/ce630f980c7a79ae17b36598f3834c4ffd523739

And also revert all the other changes related to HttpContext resolution, including the Shape Tracing one in the dev branch.

harmony7:
@sebastienros  is there anything else i need to do, or are you guys going to be doing the reverting ?

sebastienros:
If you want to try to revert all the changes and ensure it still fixes the reported changes then I would be more than happy. On the 1.9.x branch.

harmony7:
I can try reverting ce630f9.  There is also 4b59569.  Are there any other change sets?

harmony7:
Actually, though, this problem (injecting ContentManager in Migrations) occurs in 1.8 on VS2015 too.  My change fixes it on that as well, and since 1.8 obviously doesn't have any of the recent HttpContext changes, I'm pretty sure this is all we need.

sebastienros:
Would you be able to revert all these unnecessary changes then? //cc @jtkech

sebastienros:
I will try to add your change here, and also revert the fixes for the 3 issues you mentioned if I can't repro them.

PR #5746, opened 2015-09-04
Add iframe element to Layouts

### LINKED ISSUE #5740: Add iframe element to Layouts
Adding iframe element to layouts should be usefull for embedding video, maps or anything else without using HTML element and TinyMce which surrounds anything in "p" tags.

TFleury:
Closes #5740


sebastienros:
Looks good. Waiting just to see if it should be in 1.9.2 or next one.

bleroy:
Not a mainline scenario. Can be done in a gallery module.

TFleury:
A Google Maps or YouTube video element are specific scenarios, but iframe is just a standard  HTML element like break, heading or paragraph.

bleroy:
Google Maps and YouTube are not good scenarios for this: both of those should have their own elements. Maps should be available as a gallery module, and YouTube is already handled by the media library's media element. If those sites change the way they are embedded, you would have to change your site in every single place you've used an iframe, whereas if they are specialized elements, it's just a template change. It's also much easier to manage this way. iFrame is a standard element, yes, but one that is problematic in many ways, and that is often actively avoided. There are plenty of elements that don't have a corresponding element, that would be more important to include, as a matter of facts, such as UL/LI.

PR #5752, opened 2015-09-04
Refactored distributed locks to support existing installations.

sfmskywalker:
This ensures that the required table exists on existing Orchard installations.

The table-creation code is provided by a class so that the code can be used from both the setup service as well as the automatic data migrations class.

The FrameworkMigrations class has been removed to avoid code duplication.


src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
```diff
@@ -101,6 +102,10 @@ public class SetupService : Component, ISetupService {
                     schemaBuilder.AlterTable("Orchard_Framework_DataMigrationRecord",
                         table => table.AddUniqueConstraint("UC_DMR_DataMigrationClass_Version", "DataMigrationClass", "Version"));
```
DanielStolt: I am becoming less and less fond of this duplicated `SetupService`. :( Why is that necessary again??

In any case, since we allow the other schema builder operations directly in this code, I think we should either do the same with the distributed lock record, or refactor **all** of them into some kind of `CoreSchemaBuilder` responsible for creating all these low level tables.

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
```diff
@@ -30,6 +39,7 @@ public class AutomaticDataMigrations : IOrchardShellEvents {
 
         public void Activated() {
             IDistributedLock @lock;
+            EnsureDistributedLockSchema();
```
DanielStolt: Maybe name this one `EnsureDistributedLockSchemaExists()` and move it up one line?

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
```diff
@@ -58,5 +68,15 @@ public class AutomaticDataMigrations : IOrchardShellEvents {
...
         }
+
+        /// <summary>
+        /// This ensures that the framework migrations have run for the distributed locking feature, as existing Orchard installations will not have the required tables when upgrading.
+        /// </summary>
+        private void EnsureDistributedLockSchema() {
+            // Ensure the distributed lock record schema exists.
+            var schemaBuilder = new SchemaBuilder(_dataMigrationInterpreter);
+            var distributedLockSchemaBuilder = new DistributedLockSchemaBuilder(_shellSettings, schemaBuilder);
+            distributedLockSchemaBuilder.EnsureSchema();
```
DanielStolt: This code would then call a method on that `CoreSchemaBuilder` I suggested in my earlier comment.

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
sfmskywalker: It's necessary for the Import screen where the user has the option to blow away the entire database and setup the site using the uploaded recipe file. The fundamental difference between this SetupService and the one from Orchard.Setup is the way the recipe is provided I believe, so it should in principle not be too difficult to refactor some bits and bobs. I'll add a card for it.

I'm not 100% sure what you mean with your second part of your comment. Do you mean we should have a single `CoreSchemaBuilder` class that handles both the creation of the DataMigrationRecord table as well as the DistributedLockRecord table (including related key constraints)?

src/Orchard/Data/Migration/AutomaticDataMigrations.cs
sfmskywalker: Sure thing. And yes, the tidier the better.

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
DanielStolt: Yes, that's what I mean. A single class whose responsibility is to create all the schema stuff that for one reason or another can't be created through migrations.

And btw, you are not @paynecrl97... :)

sfmskywalker: Thank you for clarifying (on both accounts ;))

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
```diff
@@ -101,6 +102,10 @@ public class SetupService : Component, ISetupService {
...
 
+                    // Create the distributed lock record schema.
```
sebastienros: Is it necessary as the AutomaticDataMigration class will handle it?

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
sebastienros: Yeah, what is this new SetupService?

src/Orchard.Web/Modules/Orchard.Setup/Services/SetupService.cs
```diff
@@ -154,6 +155,10 @@ public class SetupService : ISetupService {
                     schemaBuilder.AlterTable("Orchard_Framework_DataMigrationRecord",
                         table => table.AddUniqueConstraint("UC_DMR_DataMigrationClass_Version", "DataMigrationClass", "Version"));
 
+                    // Create the distributed lock record schema.
+                    var distributedLockSchemaBuilder = new DistributedLockSchemaBuilder(_shellSettings, schemaBuilder);
```
sebastienros: Unnecessary as the automatic migration will create it.

sfmskywalker: True, but I removed that migration so we don't have two places to maintain the schema. Do you think it's better to have both a migration _and_ another class that creates the same table?

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
sfmskywalker: Not per se if we keep the migration class. However that means we would have bit of duplicated code.

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
paynecrl97: He's not me yet, but I recognize a few bits and bobs that remind me of myself :)

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
sebastienros: I insist! This table is created in the automatic migration event handler. And the setup will also call the automatic migration event handler, so you don't need to call the schema migration from the setup services. Or is it not called during setups?

src/Orchard.Web/Modules/Orchard.Setup/Services/SetupService.cs
sebastienros: I am talking about the automatic migration class itself. Which do create the schema then try to acquire the lock.

src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupService.cs
sfmskywalker: Yes absolutely, if I add back the `Migrations` class that will work fine as you say. If you don't care about the bit of code duplication (schema creation in `Migrations` as well as in `DistributedLockSchemaBuilder`) then I'd be happy to add it back and remove the call here. Just say the word.

src/Orchard.Web/Modules/Orchard.Setup/Services/SetupService.cs
sfmskywalker: I think we're talking about the same thing. It will work the way you say. My only concern is code duplication (a small amount of it). But if you think that's less of an evil than using `DistributedLockSchemaBuilder` here then I'll remove this line here and add back the `Migrations` class.

sebastienros: Look at this line:
https://github.com/OrchardCMS/Orchard/pull/5752/files#diff-c4d675a4ef668ca2d924df563f3d2548R41

Then if you see it you understand you can __remove__ the calls to the same thing from both Setup classes.

sfmskywalker: I see what you're saying now. Since we're note using the distributed lock service during setup anyway, we don;t need the lock record table to exist in the first place. After setup has finished, automatic migrations kick in and will take care of creating the required table. No invocation needed from Setup, no need for the `Migrations` class. Will apply the suggestion and test.

sebastienros:
@sfmskywalker  Can you please fix the merge and finish it?

Issue #5754, opened 2015-09-04
compiling dev (1c3e253) needs 3 rounds to work


qt1:
Current dev branch (1c3e253edb2e6278f9deacc18cab4a97fa31f563 )

Rebuild solution -> about 390 error similar to this:
```
Severity	Code	Description	Project	File	Line	Source
Error	CS0234	The type or namespace name 'Environment' does not exist in the namespace 'Orchard' (are you missing an assembly reference?)	Orchard.JobsQueue	F:\git_managed\github\Orchard_Ibn\src\Orchard.Web\Modules\Orchard.JobsQueue\AdminMenu.cs	1	Build
```

Takes another 2-3 "Build Solution" to get the solution built.

Probably some project dependency definition needed.

Tested with VS2015

sebastienros:
I confirm, and some others can too. I think all the references are OK, otherwise it would not compile at all. Maybe a bug in how the parallel msbuild processes are launched?

sfmskywalker:
I'm seeing it too. One time even incremental builds didn't work and has to restart VS2015. But that is a one time thing so far.

sebastienros:
I am closing it until we have a repro. But I agree we've seen it too sometimes.

qt1:
Please check out PR  #5766 to see if makes any difference.

PR #5766, opened 2015-09-09
Possible fix for #5754

qt1:
This fixed the build on my configuration, possibly related to additional tools like rebraces - which are not installed in my VS.

Order of projects change in sln - csprojs come before rebracer and gulp


sebastienros:
It doesn't hurt to merge it.

PR #5770, opened 2015-09-10
Changed the handling of the Initializing shell status.

sfmskywalker:
This improves the responsiveness of subsequent requests coming in dramatically.


src/Orchard.Web/Modules/Orchard.Setup/Services/SetupService.cs
```diff
@@ -174,7 +177,7 @@ public class SetupService : Component, ISetupService {
                 }
             }
 
-            // Set shell state to "Initializing" so that subsequent HTTP requests are responded to with "Service Unavailable" while Orchard is setting up.
+            // Reset shell state to "Initializing" so that subsequent HTTP requests are responded to with "Service Unavailable" while Orchard is setting up.
```
DanielStolt: On line 169 the state is set to `Running`. This seems flakey. What if a request comes in while the state is `Running`? Wouldn't it be better to modify the code in `CreateTenantData` so that it functions for both `Running` and `Initializing`?

src/Orchard/Environment/DefaultOrchardHost.cs
```diff
@@ -343,6 +348,20 @@ public class DefaultOrchardHost : IOrchardHost, IShellSettingsManagerEventHandle
             _tenantsToRestart.GetState().Add(context.Settings);
         }
 
+        private void BlockRequestsDuringSetup() {
+            if (_shellContexts == null)
+                return;
+
+            // If there's only one tenant and it's initializing, return a Service Unavailable HTTP status code.
```
DanielStolt: OK this is fine for the "only one tenant" case, but what happens for other tenants while they are initializing?

src/Orchard/Environment/DefaultOrchardHost.cs
```diff
@@ -343,6 +348,20 @@ public class DefaultOrchardHost : IOrchardHost, IShellSettingsManagerEventHandle
...
+            if (_shellContexts == null)
+                return;
+
+            // If there's only one tenant and it's initializing, return a Service Unavailable HTTP status code.
+            var shellContexts = _shellContexts.ToList();
+            if (shellContexts.Count == 1 && shellContexts[0].Settings.State == TenantState.Initializing) {
+                var response = _httpContextAccessor.Current().Response;
+                response.StatusCode = 503;
+                response.StatusDescription = "Orchard is currently setting up. Please check back in a few moments.";
+                response.Write("Orchard is currently setting up. Please check back in a few moments.");
```
DanielStolt: I would suggest a more techno-formal wording:
"This tenant is currently initializing. Please try again later."

DanielStolt:
As an additional general comment, have you considered and tested error cases? What happens if an exception is thrown in the POST request, but after the state has been set to `Initializing`? The next attempt would be greeted with a HTTP 503 wouldn't it? And then you're stuck, can't retry without going into Settings.txt and manually reverting the state.

I think we need to catch errors and reset the state to `Uninitialized` if an unexpected error occurs, so that the flow remains the same as before we introduced this new state.

src/Orchard.Web/Modules/Orchard.Setup/Services/SetupService.cs
sfmskywalker: Yes exactly, I thought the exact same thing. It would break implementations if `IOrchardHost`, which most likely means we wouldn't be breaking anyone's code.

src/Orchard/Environment/DefaultOrchardHost.cs
sfmskywalker: Yeah I thought of that. If a request comes in for another tenant, then with this code it would just behave is before the Initializing status code changes (meaning clients would see the Setup screen even if the tenant is currently cooking).

To solve for that, I think we'd have to know what shell the current request matches. Once we know that, it should be simple to update this code to check for any shell's state rather than the first and only shell.
Can you think of any issues with such an implementation?

src/Orchard/Environment/DefaultOrchardHost.cs
sfmskywalker: Works for me.

sfmskywalker:
Very good catch. I tried by simulating an exception, and indeed the setup page would get stuck in the 503 status (when thrown from the correct location).

PR #5785, opened 2015-09-11
Issue 21244 - Display warning when insecure custom forms are created

### LINKED ISSUE #5073: Display warning when insecure custom forms are created
@mjy78 created:
https://orchard.codeplex.com/workitem/21244

Display a warning/info message in the admin if a user creates a custom form set to...

a) save a content item when the form is submitted; and 
b) where the content type used by the form is NOT DRAFTABLE (so it gets published when the form is submitted); and 
c) where the content type has no permissions preventing it from being viewed publicly. 

My reasoning here is that I don't believe it's immediately obvious that when you create a custom form in Orchard that you may be allowing the submitted form data to be publicly accessible. In the majority of use cases for forms, I suspect the intention would not be for the submitted information to be publicly accessible (think enquiry forms, contact forms, etc with user's email addresses and other private info).

mjy78:
Fix for issue raised on codeplex https://orchard.codeplex.com/workitem/21244

Ensure when creating custom form widgets or pages if the underlying content type that the form uses is not draftable and has security to allow it to be anonymously viewed, that a security warning is displayed.

Fixes #5073

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.CustomForms/Drivers/CustomFormPartDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.CustomForms/Drivers/CustomFormPartDriver.cs
@@ -8,18 +8,27 @@ using Orchard.CustomForms.Models;
 using Orchard.CustomForms.ViewModels;
 using Orchard.ContentManagement.Handlers;
 using System;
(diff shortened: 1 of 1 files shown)


sebastienros:
This module is deprecated in 1.9.

Issue #5791, opened 2015-09-11
HomeAlias causes content to have two paths


sfmskywalker:
Now that a content item becomes the homepage no longer because it has an empty alias, but by virtue of an alias record pointing to the content item, that content item now has two URLS:

- /
- my-page

Even if you consider that to be a non-issue, then the actual issue is exposed by the following spec flow test:

```
Scenario: I set my blog to be the content for the home page and the posts for the blog be rooted to the app
    Given I have installed Orchard
    When I go to "admin/blogs/create"
        And I fill in
            | name | value |
            | Title.Title | My Blog |
            | AutoroutePart.PromoteToHomePage | true |
        And I hit "Save"
        And I go to "admin/blogs"
        And I follow "My Blog"
        And I follow "New Post" where class name has "primaryAction"
        And I fill in
            | name | value |
            | Title.Title | My Post |
            | Body.Text | Hi there. |
        And I hit "Publish Now"
        And I am redirected
        And I go to "/"
    Then I should see "<h1>My Blog</h1>"
    When I go to "/"
    Then the status should be 200 "OK"
    When I go to "/my-post"
    Then I should see "<h1>My Post</h1>"
```

To summarize, this is what the test is trying to assert:

1. Create a Blog (my-blog) and make it the homepage.
2. Create a Blog Post (my-post).
3. Check that the Blog is served when navigating to "/".
4. Check that the Blog Post is served when navigating to "/my-post".

Check 4 will fail, because when generating the alias for the blog post, it uses the {Content.Container.Path} token, which evaluates to "my-blog" instead of "".

One solution to restore this expected behavior of homepages is to update the Path token implementation to see if the content item is actually the homepage, and return an empty string if it is.

sebastienros:
Right, assuming the token is the only issue. Let's also see where the same property is used anywhere else. And just to be sure, that `@Url.ItemDisplayUrl` and `@Html.ItemDisplayLink` work too.

sfmskywalker:
Good point. Probably need to update the content handler for the autoroutepart to set the proper route values.

sfmskywalker:
Created a PR that fixes the issue here: https://github.com/OrchardCMS/Orchard/pull/5796

This maintains proper URL generation so that `Url.ItemDisplayUrl` and `Html.ItemDisplayLink` return the homepage alias. That's achieved by making sure that of the two available aliases based on the route values, the alias with the empty string is taken to generate the virtual path.

PR #5796, opened 2015-09-12
Home Alias fixes for issue #5791

sfmskywalker:
- Fixes a bug with the HomeAlias service
- Fixes Pages specflow test (by changing the specification)

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Alias/Implementation/Map/AliasMap.cs
+++ b/src/Orchard.Web/Modules/Orchard.Alias/Implementation/Map/AliasMap.cs
@@ -171,7 +171,8 @@ namespace Orchard.Alias.Implementation.Map {
             }
 
             if (match == null) {
-                var foundPath = focus.Paths.Keys.FirstOrDefault();
+                // Get the shortest path available to ensure the "home" alias is always taken if present.
+                var foundPath = focus.Paths.Keys.OrderBy(x => x).FirstOrDefault();
                 if (foundPath != null) {
                     // Here the deepest match is being created, which will be populated as it rises back up the stack, but save the path here.
                     // Within this function it's used to count how many items match so we get the best one; but when it's returned
--- a/src/Orchard.Web/Modules/Orchard.Alias/Implementation/Map/AliasMap.cs
+++ b/src/Orchard.Web/Modules/Orchard.Alias/Implementation/Map/AliasMap.cs
@@ -172,7 +172,7 @@ namespace Orchard.Alias.Implementation.Map {
 
             if (match == null) {
                 // Get the shortest path available to ensure the "home" alias is always taken if present.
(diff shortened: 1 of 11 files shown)


src/Orchard.Web/Modules/Orchard.Alias/Implementation/Map/AliasMap.cs
```diff
@@ -171,7 +171,8 @@ public class AliasMap {
             }
 
             if (match == null) {
-                var foundPath = focus.Paths.Keys.FirstOrDefault();
+                // Get the shortest path available to ensure the "home" alias is always taken if present.
+                var foundPath = focus.Paths.Keys.OrderBy(x => x).FirstOrDefault();
```
Jetski5822: I feel like OrderBy Key isent the right thing. what if my alias a length of 1 and wasent / ?

src/Orchard.Web/Modules/Orchard.Autoroute/Services/HomeAliasService.cs
```diff
@@ -19,21 +24,27 @@ public class HomeAliasService : IHomeAliasService {
...
 
-        public int? GetHomePageId() {
-            var homePageRoute = GetHomeRoute();
-            var homePageIdValue = homePageRoute != null && homePageRoute.ContainsKey("id") ? (string)homePageRoute["id"] : default(string);
-            var homePageId = TryParseInt32(homePageIdValue);
-
-            return homePageId;
+        public int? GetHomePageId(VersionOptions version = null) {
+            var homePage = GetHomePage(version);
+            return homePage != null ? homePage.Id : default(int?);
```
Jetski5822: much cleaner

src/Orchard.Web/Modules/Orchard.Alias/Implementation/Map/AliasMap.cs
sfmskywalker: Good catch. It needs to order by the length of the key, not the value of the key.

src/Orchard.Web/Modules/Orchard.Autoroute/Services/HomeAliasService.cs
sfmskywalker: Thanks. And equally important, it fixes the issue for content items that use a custom route that doesn't involve the "id" route value (such as Blogs which uses "blogId").

src/Orchard.Web/Modules/Orchard.Autoroute/Services/IHomeAliasService.cs
```diff
@@ -5,8 +5,9 @@ namespace Orchard.Autoroute.Services {
     
     public interface IHomeAliasService : IDependency {
         RouteValueDictionary GetHomeRoute();
-        int? GetHomePageId();
+        int? GetHomePageId(VersionOptions version = null);
```
sebastienros: In case we keep the class, we can remove it.

PR #5803, opened 2015-09-14
Fixes #2657 - TinyMCE settings

### LINKED ISSUE #2657: TinyMCE configuration extension point with jQuery events

dcinzona:
Adds support for modifying the TinyMCE initialization script within the Admin UI.
When the feature TinyMCE Settings is enabled, a new setting navigation item is added to the Settings section in the admin portal.
By default, this will load the script data stored in the orchard-tinymce.js file.  Once edited, the view will load the contents of the TinyMceSettingsPart.

Fixes issue #2657


sebastienros:
Change the main TinyMCe module to add some events to modify the script. And the main module is responsible for writing the script in full, only once, in the html. It should not be an event, The interface should have a `Priority` property to order the providers. The events will not no about each other so we can't use the dependencies order for that.

Your feature will then use these events to let users modify it in full, while allowing other module to alter it too.

Issue #5663, opened 2015-08-20
Call to OrchardAPIController method does not return to callee


dpomt:
Hi,

I have a custom module implementing an ApiController for misc. functionality.
One one of the methods executes an Import recipe that updates multiple Content items (that heavily use taxonomies).
The issue I face is that when Import and hence the ApiController method is done callee does not receive response. If I comment out the Import stuff in the APIController method callee does receive response).
I have similar methods importing/updating items that work fine and callee gets response for them.

I am calling the ApiController method using WebClient.UploadString. In Fiddler I see my application is still waiting for response after Orchard did its work in the APIController method. If I break the Orchard w3wp.exe process I do not see anything Happening/still working.

Any ideas what the issue might be? The timeouts are large enough.

Thanks!

dpomt:
I figured out why the APIController method call does not return to caller:
due to the massive taxonomy changes for the imported content items there are lots of ITermCountProcessor.Process tasks that prevent to return to callee prior not having finished those tasks.
Is this behavior by design? Would be better if tasks would be executed asynchronously and return to caller immedieatly.

As workaround I have commented out the the two calls to
   _processingEngine.AddTask(_shellSettings, _shellDescriptorManager.GetShellDescriptor(), "ITermCountProcessor.Process", ...

Is this ok if I am not interessed in terms count or is this term count used for anything important?


If I am not intereste

mvarblow:
I'm running into a similar problem. When I try to import about 24 content items with a taxonomy field and a fairly simply taxonomy (about 16 terms) the import takes several minutes to complete. A majority of the time is spent executing the 60 ITermCountProcessor.Process tasks that are generated during the import.  About half of the ITermCountProcessor.Process tasks complete in about 1ms.  The others run for 500 to 3,000+ ms each, many of them in the 2-3 second range.  That seems like an awful lot of time for counting terms with my simple taxonomy and a couple dozen content items.  It seems that there must be some way to optimize the Process method.  

Perhaps when importing many content items we could skip adding the tasks as we go and simply reprocess all the terms with a single task at the end.  I suspect that the overhead here is all the time it takes to spin the environment up and down for each task.  So executing a single large task would be much faster than the 60 small tasks.  Maybe a command to suspend term count updating and a command to update all term counts would do the trick?

[...]

PR #5804, opened 2015-09-14
Prevent no-op tasks for taxonomy field updates.

mvarblow:
Improves, but does not fully resolve, the performance problems described in issue #5663.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs
@@ -92,8 +92,9 @@ namespace Orchard.Taxonomies.Handlers {
         // Fires off a processing engine task to run the count processing after the request so it's non-blocking.
         private void RecalculateCount(IProcessingEngine processingEngine, ShellSettings shellSettings, IShellDescriptorManager shellDescriptorManager, TermsPart part) {
             var termPartRecordIds = part.Terms.Select(t => t.TermRecord.Id).ToArray();
-            processingEngine.AddTask(shellSettings, shellDescriptorManager.GetShellDescriptor(), "ITermCountProcessor.Process", new Dictionary<string, object> { { "termPartRecordIds", termPartRecordIds } });
-
+            if (termPartRecordIds.Length > 0) {
+                processingEngine.AddTask(shellSettings, shellDescriptorManager.GetShellDescriptor(), "ITermCountProcessor.Process", new Dictionary<string, object> { { "termPartRecordIds", termPartRecordIds } });
+            }
         }
 
         protected override void Activating(ActivatingContentContext context) {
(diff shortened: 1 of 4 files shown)


sebastienros:
The change implies that if all terms have been removed, the count won't be processed. Which is wrong. Is it what you mean by "not fully resolve" ?

mvarblow:
Actually, I think that's wrong.  If you look at TermCountProcessor.Process it does nothing if the termPartRecordIds is empty.  So we should lose nothing by not scheduling the task.  

This does not fully resolve the issue because it only makes a moderate dent in the number of tasks that get scheduled when importing content items with taxonomy fields.  The performance problem is the result of (1) the overhead of running a single task and (2) the number of tasks that are generated.  I was seeing two tasks for each taxonomy field for each imported content item.  And on quite good hardware it takes about 2 seconds of spin-up/down time (plus the time to execute the task) for each task.

mvarblow:
The performance problem is also compounded by the fact that multiple content items may use the same taxonomy terms.  But for each taxonomy field that is updated you get a task scheduled to run on the relevant terms.  You may end up processing the same term many times.

Ideally there would be a single task to run at the end of the request.  Instead of passing the terms as a parameter to the task, the list of terms would have to be queued up somewhere.  The task would eliminate duplicates and process the queue.  An approach like this would cut the content import time by about 80% for my simple example.

sebastienros:
@mvarblow yes, a single task. Actually that's what I am doing for blog posts if I remember correctly. 

```
        public void ProcessBlogPostsCount(int blogPartId) {
            if (!_processedBlogParts.Contains(blogPartId)) {
                _processedBlogParts.Add(blogPartId);
                _processingEngine.AddTask(_shellSettings, _shellDescriptorManager.GetShellDescriptor(), "IBlogPostsCountProcessor.Process", new Dictionary<string, object> { { "blogPartId", blogPartId } });
            }
        }
```

So this method is called every time there is a change, but itself will call the processing task accordingly. This is how we should do it, adding a level of indirection.

mvarblow:
I just pushed some changes along those lines.  The TaxonomyService now queues up the updated terms in a HashSet and fires just a single task.  The TermsPartHandler does the same.  In addition, TermCountProcessor now checks to make sure it's not processing the same parent elements multiple times.

(I'd like to consolidate the code in TaxonomyService and TermsPartHandler into a single RecalculateCount method in TaxonomyService.  That would get us down to just a single task.  But then I'd have to change the ITaxonomyService interface.  I'd like to fix this in 1.9.x.  So a change to ITaxonomyService seems out of the question.)

mvarblow:
@sebastienros I just pushed some changes along those lines. The TaxonomyService now queues up the updated terms in a HashSet and fires just a single task. The TermsPartHandler does the same. In addition, TermCountProcessor now checks to make sure it's not processing the same parent elements multiple times.

(I'd like to consolidate the code in TaxonomyService and TermsPartHandler into a single RecalculateCount method in TaxonomyService. That would get us down to just a single task. But then I'd have to change the ITaxonomyService interface. I'd like to fix this in 1.9.x. So a change to ITaxonomyService seems out of the question.)

PR #5813, opened 2015-09-15
Refactored IMachineNameProvider

sfmskywalker:
* Renamed to IApplicationEnvironment.
* Added ScriptTimeout property.


sebastienros:
Please target __dev__. 
Please rename GetMAchineName to GetEnvironmentIdentifier and change the default implementation to using the process id. The machine name per se hardly makes sense, and users can still get Environment.MachineName if they really want it. We actually don't want to abstract how to get the machine name, but what physical process is running the code.

PR #5814, opened 2015-09-15
Refactor renamed IMachineNameProvider.

sfmskywalker:


src/Orchard.Web/Modules/Orchard.Azure/Services/TaskLease/AzureApplicationEnvironment.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Microsoft.WindowsAzure.ServiceRuntime;
+using Orchard.Environment;
+
+namespace Orchard.Azure.Services.TaskLease {
```
sebastienros: why is this in TaskLease module ?

sebastienros: The namespace is wrong, it should not mention TaskLease anymore but maybe Environment.

sebastienros:
GTG once the namespace has been fixed.

src/Orchard.Web/Modules/Orchard.Azure/Services/TaskLease/AzureApplicationEnvironment.cs
sfmskywalker: It was generated as such by R# because of its folder structure. OK if I change that too?

sebastienros:
Sure

PR #5818, opened 2015-09-16
Fix translating a content item

neTp9c:
When a specific content item is being translated that a page with specific content item values is displayed witch also includes localization values. So, when a new localization is being created that one is already has assigned culture and a culture picker isn't displayed.

It's my first pull request on GitHub sorry that is untidy.


neTp9c:
I move flagging page as "It is being translated" before calling handlers and drivers. So that flag can be used by other parts. E.g. taxonomy part could check this and don't create new content item type for terms (but use already existing).

And may be Import-Export should be used instead of displaying existing content item. (like ContentManager.Clone but without calling create content item).

sebastienros:
See issue for more details on how to actually fix the issue. Thanks.

PR #5827, opened 2015-09-17
Restore placements confirmation message fix

### LINKED ISSUE #5826: Orchard.ContentTypes - Edit Placement Restore Button Shows Wrong Confirmation Wording
Fresh install of Orchard 1.9.x branch

When you go to the Edit Placement screen, for example:

    http://localhost:30321/OrchardLocal/Admin/ContentTypes/EditPlacement/BlogPost

And click the `Restore` button at the end you are shown this message:

image

    "Are you sure you want to remove this element?"

It should say something like:

    "Are you sure you want to restore these placements?"

rtpHarry:
Fixes incorrect message with restore placements feature:

image

Closes #5826

### DIFF
--- a/src/Orchard.Web/Core/Shapes/Scripts/base.js
+++ b/src/Orchard.Web/Core/Shapes/Scripts/base.js
@@ -236,7 +236,13 @@
             }
 
             if (_this.filter("[itemprop~='RemoveUrl']").length == 1) {
-                if (!confirm(confirmRemoveMessage)) {
+            	// use a custom message if its set in data-message
+            	var dataMessage = _this.data('message');
+            	if (dataMessage === undefined) {
+            		dataMessage = confirmRemoveMessage;
+            	}
+
+                if (!confirm(dataMessage)) {
                     return false;
                 }
             }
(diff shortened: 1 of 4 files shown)


sebastienros:
The actual solution is to improve the `RemoveUrl` handler to check if an attribute `data-message` is present, and if so use this instead of the default one.

```
var confirmRemoveMessage = $(this).data('message') || confirmRemoveMessage;
```

rtpHarry:
I've fixed the issue as you described. I used a slightly different piece of javascript code:

    	// use a custom message if its set in data-message
        var dataMessage = $(this).data('message');
        if (dataMessage !== undefined) {
        	confirmRemoveMessage = dataMessage;
        }

The reason behind this is that although it's a commonly used pattern to do something like `var someValue = someTest || someValue` it isn't correct. 

What you want to test is if it's undefined but the short version of the code can introduce strange edge cases in to the code. If for example a developer is working on some code and puts either of these messages in:

   itemprop="RemoveUrl" data-message="@T("false")">@T("Restore")</button>
   itemprop="RemoveUrl" data-message="@T("0")">@T("Restore")</button>

Then they would get the original message not the `false` or `0`.

Granted it's not likely to make it into production but it could save a dev scratching their head wondering why their test message isn't appearing.

sebastienros:
Actually it's not sufficient, you need to use `dataMessage` all along otherwise it overwrites the default value. For instance if you use the pattern twice in the same page.

rtpHarry:
You're right, good spot. I've switched it around and I also updated a second version of it in base.js

rtpHarry:
A better variable name for it would be `defaultConfirmRemoveMessage` but I didn't refactor as it's in the global namespace, potentially some module in the gallery may rely on it.

PR #5828, opened 2015-09-17
Adding ManageLayout permission

sebastienros:
New Manage Layout permission which displays the Layout menu and grants access to Elements creation.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Controllers/BlueprintAdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Controllers/BlueprintAdminController.cs
@@ -34,7 +34,8 @@ namespace Orchard.Layouts.Controllers {
             ICultureAccessor cultureAccessor, 
             IShapeFactory shapeFactory, 
             ITransactionManager transactionManager, 
-            ISignals signals) {
+            ISignals signals,
+            IOrchardServices orchardServices) {
 
             _elementBlueprintService = elementBlueprintService;
             _notifier = notifier;
@@ -43,12 +44,19 @@ namespace Orchard.Layouts.Controllers {
             _shapeFactory = shapeFactory;
             _transactionManager = transactionManager;
             _signals = signals;
+            Services = orchardServices;
+
             T = NullLocalizer.Instance;
         }
 
+        public IOrchardServices Services { get; set; }
         public Localizer T { get; set; }
 
         public ActionResult Index() {
+            if (!Services.Authorizer.Authorize(Permissions.ManageLayouts, T("Not authorized to manage layouts."))) {
+                return new HttpUnauthorizedResult();
+            }
+
             var blueprints = _elementBlueprintService.GetBlueprints().ToArray();
             var viewModel = new BlueprintsIndexViewModel {
                 Blueprints = blueprints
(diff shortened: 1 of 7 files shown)


src/Orchard.Web/Modules/Orchard.Layouts/Controllers/BlueprintAdminController.cs
```diff
@@ -43,12 +44,19 @@ public class BlueprintAdminController : Controller, IUpdateModel {
             _shapeFactory = shapeFactory;
             _transactionManager = transactionManager;
             _signals = signals;
+            Services = orchardServices;
+
             T = NullLocalizer.Instance;
         }
 
+        public IOrchardServices Services { get; set; }
```
sfmskywalker: Out of curiosity, why is this a public property instead of a private field? I've seen this pattern elsewhere, but never understood why.

sebastienros: Just a pattern ... sticking with it for the sake of consistency. I'd personally like to have this plugged to autofac like we do for Localizer for instance. Could also be something we add to the `Component` class.

src/Orchard.Web/Modules/Orchard.Layouts/Controllers/LayoutController.cs
```diff
@@ -15,15 +16,29 @@ public class LayoutController : Controller {
...
+            T = NullLocalizer.Instance;
         }
 
+        public IOrchardServices Services { get; set; }
+        public Localizer T { get; set; }
+
         [HttpPost, ValidateInput(enableValidation: false)]
-        public ContentResult ApplyTemplate(int? templateId = null, string layoutData = null, int? contentId = null, string contentType = null) {
+        public ActionResult ApplyTemplate(int? templateId = null, string layoutData = null, int? contentId = null, string contentType = null) {
+            if(!Services.Authorizer.Authorize(Permissions.ManageLayouts, T("Not authorized to manage layouts."))) {
```
sfmskywalker: Managing template layouts or applying them are two different things. Perhaps as a content editor I should be able to use existing templates, but not create or edit them.

sfmskywalker: So I think either this permission check should be removed, or replaced with an additional permission (e.g. `UseLayouts`).

src/Orchard.Web/Modules/Orchard.Layouts/Controllers/BlueprintAdminController.cs
sfmskywalker: I see. So it would be automatically be set using property injection. Which is kind of cool. The downside I can think of is that the dependency on `IOrchardServices` becomes less visible (which is the nice thing about constructor injection). But from an ease-of-use point of view, property injection is nicer.

src/Orchard.Web/Modules/Orchard.Layouts/Controllers/LayoutController.cs
sebastienros: I checked and this one should not have any permission. It's just used to change the local state of the content item, the actual check for the content item is done before and after.

sfmskywalker: GTG then.

PR #5839, opened 2015-09-21
Update AdminController.cs

JRKelso:
The dynamic Shape is implicitly private, so I made it explicitly private and renamed the variable accordingly.

Done on behalf of Radio Systems Corporation


sebastienros:
Thanks Ryan but even though it's not wrong, we don't want to spend too much time reviewing PRs which are just about refactoring. I assume you are trying to help, so can I suggest you to pick some easy bugs from the issues list? There is a tag named EasyPick that you can filter on.

PR #5844, opened 2015-09-23
Enhancement to allow multi-select for the content picker

escofieldnaxos:
We needed the ability to create a custom tab on the contentPicker for the new layouts to allow our clients the ability to paste a list of ID's into.  In doing so, I found I needed to modify the contentPicker.js callback to handle an array of items.  I would love for contentPicker.js to be submitted on it's own but felt that a more complete solution would be more beneficial.  With this addition I have added new checkbox's for each content item in addition to the select button.  I added a new button at the bottom of the list to add the checked content items.  Once the add items button is click the window closes as does the current behavior for select buttons.  If the user experience of this new behavior does not pass review, I ask that you consider keeping the contentPicker.js changes as this would allow me to maintain my custom tab functionality.

Thank you for your time in review of this submission.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentPicker/Orchard.ContentPicker.csproj
+++ b/src/Orchard.Web/Modules/Orchard.ContentPicker/Orchard.ContentPicker.csproj
@@ -72,6 +72,7 @@
   <ItemGroup>
     <Compile Include="ViewModels\NavigationPartViewModel.cs" />
     <Content Include="Scripts\ContentPicker.js" />
+    <Content Include="Scripts\RecentContentTab.js" />
     <Content Include="Styles\content-picker-admin.css" />
     <Content Include="Styles\ContentPicker.css" />
     <Content Include="Styles\Images\move.gif" />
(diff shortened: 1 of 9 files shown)


src/Orchard.Web/Modules/Orchard.ContentPicker/Views/RecentContentTab.cshtml
```diff
@@ -1,5 +1,7 @@
 ﻿@using Orchard.Core.Contents.ViewModels;
 @{
+    Script.Include("RecentContentTab.js");
```
danoliver1: This file is in the resource manifest so should this be `Script.Require("RecentContentTab");`?

escofieldnaxos: wow, First yes,  Second, wow.  Been completely overloaded. Looking into
this now.

src/Orchard.Web/Modules/Orchard.ContentPicker/Orchard.ContentPicker.csproj
```diff
@@ -72,6 +72,7 @@
   <ItemGroup>
     <Compile Include="ViewModels\NavigationPartViewModel.cs" />
     <Content Include="Scripts\ContentPicker.js" />
+    <Content Include="Scripts\RecentContentTab.js" />
```
sebastienros: Should it be `Selectcontenttab.js` ?

escofieldnaxos:
Yes fixed the project reference in latest commit.

PR #5864, opened 2015-09-26
Serialization error is thrown when creating layout or page in the dashboard when using Layouts in a Web Farm.

Garpo001:
Make Model Serializable for use in Web Farms with session state stored in a database. Currently a Serialization error is thrown when creating layout or page in the dashboard.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Models/ElementSessionState.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Models/ElementSessionState.cs
@@ -1,5 +1,8 @@
+using System;
+
 ﻿namespace Orchard.Layouts.Models {
-    public class ElementSessionState {
+﻿        [Serializable]
+        public class ElementSessionState {
         public string TypeName { get; set; }
         public string ElementData { get; set; }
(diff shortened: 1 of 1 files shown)


sebastienros:
The fix is good, but it looks like something that could not use sessions at all. Could be stored in local storage and sent in the POST when necessary for the server.

sfmskywalker:
The reason Element state is stored in Session is to support the following scenario:

1. User opens Editor Dialog for a specific element.
2. Element state is posted to the server (no need for session at this point).
3. The Element Editor page (loaded in an iframe inside the dialog) could provide multiple screens for the element editor experience. For example when adding multiple Slides to a SlideShow element, involving multiple redirects to various pages.

Although this could be done using post backs only, it would require the element developer to always post back the element state, as subsequent screens may need this information.
It would work, but it would make the implementation for element developers more complex I think.

sebastienros:
I think it would be a breeze to inject a string (the json) in a hidden input in the form. Also, how many elements need multiple pages so far?  
In MVC we tend to prevent any usage of Sessions, the only one so far has been for notifications (TempData).

PR #5872, opened 2015-09-29
Refactored Import process.

### LINKED ISSUE #5848: Invoke Imported on all imported items AFTER Importing has been invoked on ALL content items being imported.
This allows content items to properly configure related information depending on other content items.

For example, the ProjectionElement depends on the selected Query and Layout index. However, if the Query occurs after the content item with the ProjectionElement, the ProjectionElement import code will fail to properly set the selected query layout index since the Query has not yet been fully imported (there's just an on-the-fly- created Query item with zero Layouts).

sfmskywalker:
* Added Imported event to Elements.
* Invoking Imported event at the end after all content items have been imported.

Fixes #5848.


src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
```diff
@@ -262,16 +266,26 @@ public class Migrations : DataMigrationImpl {
...
+            SchemaBuilder.AlterTable("ProjectionPartRecord", table => table
+                .AlterColumn("PagerSuffix", c => c.WithType(DbType.String).WithLength(255))
+            );
 
             return 3;
         }
+
+        public int UpdateFrom3() {
+            SchemaBuilder.AlterTable("LayoutRecord", table => table
+                .AddColumn<string>("Guid", c => c.WithLength(64))
```
sebastienros: I think we added support for the Guid type directly in SchemaBuilder. And a Guid is not 64, nit picking ;)

sfmskywalker: Oh good! I'll change then. I originally wanted to specify 32 characters, but that isn't enough because of the hyphens. I either had to skip the hyphens or count the actual number of characters, so I decided to be lazy and just go with 64. I love nitpicking!

src/Orchard/ContentManagement/DefaultContentManager.cs
```diff
@@ -722,7 +722,7 @@ public class DefaultContentManager : IContentManager {
 
         // Insert or Update imported data into the content manager.
         // Call content item handlers.
-        public void Import(XElement element, ImportContentSession importContentSession) {
+        public void Import(XElement element, ImportContentSession importContentSession, Action<ContentItem> importItemCallback = null) {
```
sebastienros: Nope. Please create a new `IEventHandler` for this event. There might be others.

PR #5884, opened 2015-10-01
Resolves #5878 : URL Field for Dynamic Forms

### LINKED ISSUE #5878: Dynamic Forms: Web Address Field
All,

I have created a Web Address Field that validates the entered URL to make sure it is HTTP(S).  I was binding a form to a content type that uses URL and wanted to make sure it validated before hand.  I have this code ready to commit if it is wanted.  If so, which branch?  1.9.x?

Thanks,

Jamie

phillipsj:
Created this because I was binding to a content type that had a URL field. Needed to do this validation when submitting the dynamic form.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
+++ b/src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
@@ -34,7 +34,9 @@ namespace Orchard.DynamicForms.ValidationRules
 
         private LocalizedString GetValidationMessage(ValidationContext context)
         {
-            return T(Tokenize(ErrorMessage.WithDefault(String.Format("{0} is not a valid url.", context.FieldName)), context));
+            return String.IsNullOrWhiteSpace(ErrorMessage)
+                ? T("{0} is not a valid URL.", context.FieldName)
+                : new LocalizedString(Tokenize(ErrorMessage, context));
         }
     }
 }
--- a/src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
+++ b/src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Text.RegularExpressions;
+using Orchard.DynamicForms.Helpers;
+using Orchard.DynamicForms.Services;
+using Orchard.DynamicForms.Services.Models;
+using Orchard.Localization;
+
+namespace Orchard.DynamicForms.ValidationRules
+{
+    public class UrlAddress : ValidationRule
+    {
+        public UrlAddress() {
+            RegexOptions = RegexOptions.Singleline | RegexOptions.IgnoreCase;
+            Pattern = @"^http(s)?://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$";
+        }
+
+        public string Pattern { get; set; }
+        public RegexOptions RegexOptions { get; set; }
+        
+        public override void Validate(ValidateInputContext context)
+        {
+            if (!Regex.IsMatch(context.AttemptedValue, Pattern, RegexOptions))
+            {
+                var message = GetValidationMessage(context);
+                context.ModelState.AddModelError(context.FieldName, message.Text);
+            }
+        }
+
+        public override void RegisterClientAttributes(RegisterClientValidationAttributesContext context)
+        {
+            context.ClientAttributes["data-val-regex"] = GetValidationMessage(context).Text;
+            context.ClientAttributes["data-val-regex-pattern"] = Pattern;
+        }
+
+        private LocalizedString GetValidationMessage(ValidationContext context)
+        {
+            return T(Tokenize(ErrorMessage.WithDefault(String.Format("{0} is not a valid url.", context.FieldName)), context));
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public override void RegisterClientAttributes(RegisterClientValidationAttributesContext context)
+        {
+            context.ClientAttributes["data-val-regex"] = GetValidationMessage(context).Text;
+            context.ClientAttributes["data-val-regex-pattern"] = Pattern;
+        }
+
+        private LocalizedString GetValidationMessage(ValidationContext context)
+        {
+            return T(Tokenize(ErrorMessage.WithDefault(String.Format("{0} is not a valid url.", context.FieldName)), context));
```
sebastienros: This is wrong. If you got the code from another existing class then this is also wrong and needs to be fixed. You can't translate a dynamic string.

phillipsj: Yep I pulled it from the email field class.  Most of them are that way.

phillipsj: I will fix it tonight.

sebastienros: We'll blame @sfmskywalker , he agrees

sfmskywalker: So this should be the correct way, agree?

```C#
return new LocalizedString((Tokenize(ErrorMessage.WithDefault(T("{0} must match the value of {1}.", context.FieldName, TargetName).ToString()), context)));
```

sebastienros: I still don't agree. If you call `T()` on something that is variable it will never work. Here `T` is called on `ErrorMessage` which comes from user input, or from `String.Format("{0} ...", fieldName)` with fieldName also coming from user input. So T can only be used on `"{0} ...", fieldName`.

```
    return String.IsNullOrEmpty(ErrorMessage) 
        ? T("{0} must match the value of {1}.", context.FieldName, TargetName)
        : new LocalizedString(Tokenize(ErrorMessage))
        ;
```

sfmskywalker: Yeah but even though the ErrorMessage is user input, this provides a way to localize that string using PO files, should the user really need to, as a temporary workaround. But that of course is an ugly and brittle workaround, so best would be to come with a real localization solution for user input strings. I'll make another change as per your suggestion.

sebastienros: No, for this we agreed during a meeting, and mentioned at harvest, that we should have a module that harvests dynamic values like content types and messages to let the user translate it from the admin.

sebastienros: Ok, I think we need a round three. If we want to support error message localization like I am suggesting, then the `ErrorMessage` variable can be translated directly.

return String.IsNullOrEmpty(ErrorMessage) 
        ? T("{0} must match the value of {1}.", context.FieldName, TargetName)
        : Tokenize(T(ErrorMessage))
        ;

And we can change Tokenize to accept a LocalizedString and return a LocalizedString too, but it should use the `EncodeHtml` version of tokenize.

sebastienros: At the same time, do we really want to tokenize this property?

phillipsj:
@sebastienros  @sfmskywalker  I am ready to roll when it is decided.  Thanks for the awesome discussion, great to see.

src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
sfmskywalker: Exactly.
Regarding whether or not we want to tokenize `ErrorMessage`, if I had to guess I'd say the majority will probably never use it, but I added it nonetheless so others wouldn't be limited by my lack of imagination.
What if we removed it for now and consider adding it back if people actually ask for it?

phillipsj:
@sebastienros @sfmskywalker I made the changes.

src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public override void RegisterClientAttributes(RegisterClientValidationAttributesContext context)
+        {
+            context.ClientAttributes["data-val-regex"] = GetValidationMessage(context).Text;
+            context.ClientAttributes["data-val-regex-pattern"] = Pattern;
+        }
+
+        private LocalizedString GetValidationMessage(ValidationContext context)
+        {
+            return String.IsNullOrWhiteSpace(ErrorMessage)
```
sebastienros: Agreed with @sfmskywalker to remove Tokenization completely from here, so you can just return `T(ErrorMessage)` in the second case

src/Orchard.Web/Modules/Orchard.DynamicForms/ValidationRules/UrlAddress.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Text.RegularExpressions;
+using Orchard.DynamicForms.Helpers;
+using Orchard.DynamicForms.Services;
+using Orchard.DynamicForms.Services.Models;
+using Orchard.Localization;
+
+namespace Orchard.DynamicForms.ValidationRules
+{
```
sebastienros: Please format this file accordingly.

phillipsj:
@sebastienros  @sfmskywalker  I don't want let this die, is there anything else I need to do.

PR #5887, opened 2015-10-01
Fixes #5879: Layout part OnIndexing fails when called in a background task

### LINKED ISSUE #5879: Layout part OnIndexing fails if in a background task
**Repro**: With a valid search index on layout pages, edit a layout page, do a `Publish Later`, and wait for the background task execution. Then it fails in the `WebViewPage class` with a `null WorkContext`.

Due to the quite recent updates around workcontext and httpcontext in background tasks. E.g in the `BuildDisplayContext()` method, I can see that the `_requestContext.HttpContext`, which is an `HttpContextPlaceHolder`, has an empty `Items Dictionary without WorkContext`.

Currently I do some investigation with the help of @Piedone (see last posts of #4338) to use the `CallContext Logical Data` to carry the `WorkContext` in bkg tasks with async code. Very simple code that seems to work but needs to be seriously evaluated. If it could be reliable, maybe a solution to also carry a fake `HttpContext`...

Best

jtkech:
Fixes #5879

And maybe other issues, when the `GetContext(httpContext) method`, the one that has an `HttpContextBase parameter`, is called in the context of a `BackgroundTask`.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Services/DefaultCacheStorageProvider.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Services/DefaultCacheStorageProvider.cs
@@ -27,7 +27,8 @@ namespace Orchard.OutputCache.Services {
         }
 
         public void Remove(string key) {
-            _workContext.HttpContext.Cache.Remove(key);
+            if (_workContext.HttpContext != null)
+                _workContext.HttpContext.Cache.Remove(key);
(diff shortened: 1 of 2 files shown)


jtkech:
Same kind of issue in DefaultCacheStorageProvider.cs

When a content item is published it is removed from the httpcontext cache. But, through the Publish Later command, this is done in the context of a backgroud task...

PR #5897, opened 2015-10-05
Fixes 1+n db call issue

paynecrl97:
Adds

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Services/DefaultLayerEvaluationService.cs
@@ -41,7 +41,7 @@ namespace Orchard.Widgets.Services{
             // Once the Rule Engine is done:
             // Get Layers and filter by zone and rule
             // NOTE: .ForType("Layer") is faster than .Query<LayerPart, LayerPartRecord>()
-            var activeLayers = _orchardServices.ContentManager.Query<LayerPart>().ForType("Layer").List();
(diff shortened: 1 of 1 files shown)


agriffard:
Fastest merge ever.

sebastienros:
What seemed weird at first when I saw the code is that the LayoutPart is backed by InfoSet properties so the Record should not be hit. In your slides you explicitly show  `.Record` access that I don't see in the actual code. Did you simplify the issue to make it clearer for the audience what the problem is?

paynecrl97:
No, the code I demoed was exactly as it was in the dev branch at the time.

Although this fix was for Layers, and not Layouts, so maybe that's where this confusion comes from?

dcinzona:
Can this be merged with 1.9.x as well (and manually patched into my 1.9.2 source) or does it require something else?  Basically, is this a monolithic change, not requiring any other source code changes?

paynecrl97:
This layer evaluation service does not exist in 1.9.x (layers are evaluated in the `WidgetFilter`) instead.

However, the code in `WidgetFilter` is exactly the same as the code in this service, so you should be able to manually patch in your source (or create a PR that contains the fix).

Offending line is: https://github.com/OrchardCMS/Orchard/blob/1.9.x/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs#L57. Just apply the same query hints.

sebastienros:
Isn't changing `activeLayerIds.Add(activeLayer.ContentItem.Id);` to `activeLayerIds.Add(activeLayer.Id);` actually better, as I assume this is where the n+1 is made?

Otherwise @paynecrl97 would you mind pointing the line which is accessing the `.Record` as you slide was showing?

PR #5902, opened 2015-10-07
Naxos - Select-able content picker

escofieldnaxos:
I have fixed the resource include,
I have also added the options to the Search Content Picker as well, In doing so I have renamed the script include from RecentContentTab to SelectableContentTab
I have also added a select all on the search content picker for this purpose.


.gitignore
```diff
@@ -188,3 +188,7 @@ src/Orchard.Azure/Orchard.Azure.CloudService/Staging/
 #enable all /lib artifacts
 !lib/*/*.*
 */.vs/*
+src/Orchard.Web/Modules/Naxos*
```
sebastienros: Bad

sebastienros:
How is it related to the other PR?

Would you like to give a demo during the Tuesday meeting so we can see how it works before we pull them?

Thanks

escofieldnaxos:
I think this is acceptable. I've never done this before, but I would love
to demo it.

-Erik

.gitignore
escofieldnaxos: I don't even see this in my commits..  How do I tell what commit hash this is?

escofieldnaxos: I think I see the problem and it's my in-experience with using GitHub and pull requests.  The branch I would like a pull request from is feature/Content-Picker-MultiSelect not the Naxos branch as the Naxos branch has our specific code changes and as you see git ignore so I can separate our code from the orchard code.

escofieldnaxos:
I thought a pull request was on a commit hash, so I thought i was moving it forward by creating the next pull request.

PR #5915, opened 2015-10-08
fix css issue

### LINKED ISSUE #5912: Orchard.MultiTenant - Wrong CSS class in admin view
Branch: 1.9.2

I first mentioned this as an aside in #5906 but now I've found the solution in Orchard.Users.

image

This should have the class `message message-Warning` instead of just `message warning`.

Then it will be formatted correctly like Orchard.Users does:

image

rtpHarry:
closes #5912


sebastienros:
Please fix the merge issue

rtpHarry:
@sebastienros Can I do that without write access? I don't see any way to see the details of the actual merge issue?

image

sebastienros:
If you try to merge your branch into 1.9.x, from your local repository, then you will see then.

rtpHarry:
Just when I think that me and Git are friends, it turns around and takes an hour of my life away from me and seems to enjoy trying to embarrass me. All I did was edit one class name via the GitHub website editor and it's had a meltdown over it.

I think this will work now?

PR #5922, opened 2015-10-09
make orchard help command more intuitive - closes #422

### LINKED ISSUE #422: Orchard.exe help command doesn't filter by feature
__omidkrad__ created:
https://orchard.codeplex.com/workitem/16589

The help command in Orchard.exe tool currently works either like "help commands" that lists all the full commands or "help <command-name>" that lists help for only the specified command. I'd like to also be able to get help only for a specific feature, for example "help scaffolding" should show the command help for all "scaffolding *" commands.

---- STEPS ----
From the Orchard installation path run bin\orchard.exe
run: help scaffolding

---- ACTUAL ----
Output is: Command scaffolding doesn't exist

---- EXPECTED ----
Output should be:
scaffolding create controller <module-name> <controller-name>
        Create a new Orchard controller in a module

scaffolding create datamigration <feature-name>
        Create a new Data Migration class

scaffolding create module <module-name> [/IncludeInSolution:true|false]
        Create a new Orchard module

rtpHarry:
closes #422

This makes the orchard.exe `help` command more intuitive. 

Currently it makes you type out the full command to get help on it:

image

From the way command line's work I think most users would expect `user` to be the command and the rest of the elements are arguments.

This small change filters the help so it just lists all the help it can based on the input:

`> help user`

image

I decided not to change the help command description:

    [CommandHelp("help <command>\r\n\t" + "Display help text for <command>")]

This is because:

  - Like I said, users won't realise that orchard considers the actual command to be several words long
  - The new filtered help just acts like autocomplete

Reminder note: If help text is deemed to need changing we also need to update text in [OrchardHost.cs](https://github.com/OrchardCMS/Orchard/blob/6763bd8bb85a8f226955f4331dfa830be05e0666/src/Tools/Orchard/OrchardHost.cs)

### DIFF
--- a/src/Orchard/Commands/Builtin/HelpCommand.cs
+++ b/src/Orchard/Commands/Builtin/HelpCommand.cs
@@ -37,12 +37,15 @@ namespace Orchard.Commands.Builtin {
         [CommandHelp("help <command>\r\n\t" + "Display help text for <command>")]
         public void SingleCommand(string[] commandNameStrings) {
             string command = string.Join(" ", commandNameStrings);
-            var descriptor = _commandManager.GetCommandDescriptors().SingleOrDefault(d => string.Equals(command, d.Name, StringComparison.OrdinalIgnoreCase));
-            if (descriptor == null) {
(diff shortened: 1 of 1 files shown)


DanielStolt:
Nice change! :+1:

PR #5930, opened 2015-10-11
upgrade the WebApi dlls in lib to version 2.2.29 fixes #5873

### LINKED ISSUE #5873: Could we upgrade the WebApi dlls in lib to Microsoft HTTP Client Libraries 2.2.29 (package)
It seems to be actually 2.28 and it generates many troubles with nugets packages using 2.2.29.

CSurieux:
Change 2 assemblies in lib\aspnetwebapi from v 2.2.28 to v 2.2.29


Jetski5822:
LGTM

Issue #5936, opened 2015-10-12
1.9.2 - Media Library browser tries to load all media files instead of using infinite scroll


dcinzona:
This is actually a breaking change because the media library.js script was changed to load all media items in a given folder, instead of paginating x at a time.  Specifically, starting at line 74
```javascript
        var itemWidth = $('.thumbnail').first().width();
        var itemHeight = $('.thumbnail').first().height();
        var draftText = $("#media-library").data("draft-text");

        var itemsPerRow = Math.floor(listWidth / itemWidth);
        var itemsPerColumn = Math.ceil(listHeight / itemHeight);
```
The xhr request sets the item count to infinity in this case.

Based on this PR https://github.com/OrchardCMS/Orchard/commit/51a364a1846cb41d7d4ea9baae633817ef794bc1

Issue #5891, opened 2015-10-02
media--null ex in recentmediaitems 


infofromca:
user:admin
go to media, create folder,
click recent , got ex
![recent-null](https://cloud.githubusercontent.com/assets/297675/10256688/d71a140c-6920-11e5-88c0-b74c2ca6f4e2.png)

infofromca:
this should go to 1.9.2

dcinzona:
This is fixed in my PR
https://github.com/OrchardCMS/Orchard/pull/5937

PR #5937, opened 2015-10-12
Fixes issued #5936 #5938 #5941 and #5891

dcinzona:
Added multiple fixes to the media library module:
Fixes infinite count on media library admin view
Fixes issue where Recents link in media library admin view would cause a server error
Fixes minor issue where loading GIF is not really visible when loading more items on scroll

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/AdminController.cs
@@ -156,10 +156,10 @@ namespace Orchard.MediaLibrary.Controllers {
             if (!Services.Authorizer.Authorize(Permissions.ManageOwnMedia, T("Cannot view media")))
                 return new HttpUnauthorizedResult();
 
-            var rootMediaFolder = _mediaLibraryService.GetRootMediaFolder().MediaPath;
-
-            var mediaParts = _mediaLibraryService.GetMediaContentItems(rootMediaFolder, skip, count, order, mediaType);
-            var mediaPartsCount = _mediaLibraryService.GetMediaContentItemsCount(rootMediaFolder, mediaType);
+            var rootMediaFolder = _mediaLibraryService.GetRootMediaFolder();
+            var rootMediaFolderPath = rootMediaFolder == null ? null : rootMediaFolder.MediaPath;
+            var mediaParts = _mediaLibraryService.GetMediaContentItems(rootMediaFolderPath, skip, count, order, mediaType);
+            var mediaPartsCount = _mediaLibraryService.GetMediaContentItemsCount(rootMediaFolderPath, mediaType);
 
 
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library.js
```diff
@@ -71,8 +71,11 @@ $(function () {
 
         var listWidth = $('#media-library-main-list').width();
         var listHeight = $('#media-library-main-list').height();
-        var itemWidth = $('.thumbnail').first().width();
-        var itemHeight = $('.thumbnail').first().height();
+        var itemWidth = $('#media-library-main-list li').first().width();
+        var itemHeight = $('#media-library-main-list li').first().height();
+        var defaultDimension = $(window).width() < 1420 ? 120 : 200;
```
dcinzona: There's a media query that changes the default thumbnail size to 120 on smaller width screens.  This accounts for that.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library.js
```diff
@@ -71,8 +71,11 @@ $(function () {
 
         var listWidth = $('#media-library-main-list').width();
         var listHeight = $('#media-library-main-list').height();
-        var itemWidth = $('.thumbnail').first().width();
-        var itemHeight = $('.thumbnail').first().height();
+        var itemWidth = $('#media-library-main-list li').first().width();
+        var itemHeight = $('#media-library-main-list li').first().height();
```
dcinzona: On load, the .thumbnail class has 0 height. This is what was causing the infinity value for count.
Changing this to the containing li item allows for a set size based on the CSS.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Styles/orchard-medialibrary-admin.css
```diff
@@ -155,6 +155,10 @@
             background-position: center;
         }
 
+        #media-library-main-list.pending li {
+            opacity:.3;
+        }
+
```
dcinzona: I wanted a better UI queue to show that the page was loading more content.  Setting the opacity to semi-transparent on the LI items solves this problem as it now lets the background animated gif become more visible during loading of items

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/AdminController.cs
```diff
@@ -156,10 +156,10 @@ public class AdminController : Controller {
             if (!Services.Authorizer.Authorize(Permissions.ManageOwnMedia, T("Cannot view media")))
                 return new HttpUnauthorizedResult();
 
-            var rootMediaFolder = _mediaLibraryService.GetRootMediaFolder().MediaPath;
```
dcinzona: This would sometimes return null, causing .MediaPath to error out.

PR #5942, opened 2015-10-12
Extends theme.txt to allow different zones for each layer fixes #5933

### LINKED ISSUE #5933: On the Widget Admin page zones should be attached to each layer not global to all layers
For the default Layer, widgets could be displayed if their zone is present on the actual layer ?
This has certainly be tried ?

replace in actual Admin controller
>            IEnumerable<string> allZones = _widgetsService.GetZones();
>            IEnumerable<string> currentThemesZones = _widgetsService.GetZones(currentTheme);
>            string zonePreviewImagePath = string.Format("{0}/{1}/ThemeZonePreview.png", currentTheme.Location, currentTheme.Id);

by

>            IEnumerable<string> allZones = _widgetsService.GetZones(currentLayer);
>            IEnumerable<string> currentThemesZones = _widgetsService.GetZones(currentTheme,currentLayer);
>             string zonePreviewImagePath = string.Format("{0}/{1}/{2}ThemeZonePreview.png", currentTheme.Location, currentTheme.Id, currentLayer);

etc....

CSurieux:
Just add 2 more lines in your theme.txt to create layers and their corresponding zone.
>Layers: TheNextHomepage, Calendar
>LayerZones: {"TheNextHomepage":"Messages, Headerhome, Edito, Agenda, Actus, Catalogue","Calendar":"Header, Navigation, AsideFirst, Messages, BeforeContent, Content, Footer"}

Then in Widgets menu, add the 2 layers TheNextHomepage and Calendar.
This also relies on alternates to create the corresponding layout-url-xxx.cshtml with code for your new zones as
> @if (Model.Edito != null) {
>           @Zone(Model.Edito)
>               }

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Services/PlacementService.cs
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Services/PlacementService.cs
@@ -10,8 +10,10 @@ using Orchard.ContentTypes.Settings;
 using Orchard.DisplayManagement;
 using Orchard.DisplayManagement.Descriptors;
 using Orchard.Environment.Extensions;
+using Orchard.Environment.Extensions.Models;
 using Orchard.FileSystems.VirtualPath;
 using Orchard.Logging;
+using Orchard.Services;
 using Orchard.Themes.Services;
 using Orchard.UI.Zones;
 
@@ -34,6 +36,7 @@ namespace Orchard.ContentTypes.Services {
         private readonly IEnumerable<IContentFieldDriver> _contentFieldDrivers;
         private readonly IVirtualPathProvider _virtualPathProvider;
         private readonly IWorkContextAccessor _workContextAccessor;
+        private readonly IJsonConverter _jsonConverter;
 
         public PlacementService(
             IContentManager contentManager,
@@ -45,7 +48,8 @@ namespace Orchard.ContentTypes.Services {
             IEnumerable<IContentPartDriver> contentPartDrivers,
             IEnumerable<IContentFieldDriver> contentFieldDrivers,
             IVirtualPathProvider virtualPathProvider,
-            IWorkContextAccessor workContextAccessor
+            IWorkContextAccessor workContextAccessor,
+            IJsonConverter jsonConverter
             ) 
         {
             _contentManager = contentManager;
@@ -58,6 +62,7 @@ namespace Orchard.ContentTypes.Services {
             _contentFieldDrivers = contentFieldDrivers;
             _virtualPathProvider = virtualPathProvider;
             _workContextAccessor = workContextAccessor;
+            _jsonConverter = jsonConverter;
 
             Logger = NullLogger.Instance;
         }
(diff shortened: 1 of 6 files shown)


Ermesx:
Could you add tests?

CSurieux:
What kind of tests ? mock tests or samples ?
If you need to test it, just add  2 lines to a copy of the Theme machine and look the results in Widgets admin...

sebastienros:
Discussion pending on the issue itself.

CSurieux:
Demo of the features in this PR
https://youtu.be/oXNVY_LFMJc

bleroy:
While I understand the motivation behind the feature,I don't think themes should know anything about layers, or layers about zones. The same result could be obtained in a cleaner way buy having the widget editor be aware of multiple layouts, with different zones, whether those layouts are statically defined in the theme or dynamically built.

CSurieux:
Bertrand zones are already in theme.txt...
But I agree that it would be nice to be able to define dynamically the zones able to host widgets for a layer.
My solution is et easy to implement: 10 lines in the pr and it preserves all the existing features.

Envoyé à partir de mon sanyo phc 25
________________________________
De : Bertrand Le Roy<mailto:notifications@github.com>
Envoyé : ‎19/‎10/‎2015 18:16
À : OrchardCMS/Orchard<mailto:Orchard@noreply.github.com>
Cc : CSurieux<mailto:christian.surieux@hotmail.fr>
Objet : Re: [Orchard] Extends theme.txt to allow different zones for each layer fixes #5933 (#5942)

While I understand the motivation behind the feature,I don't think themes should know anything about layers, or layers about zones. The same result could be obtained in a cleaner way buy having the widget editor be aware of multiple layouts, with different zones, whether those layouts are statically defined in the theme or dynamically built.

---

CSurieux:
I am not totally satisfied with all the layout story: to much fragmentation, too much complicated, no way to changes the files with text editor, not in phase withe zone (everything goes in content). This solution is enough for me, I get the options to put widgets where I want and I can write my theme and layout as I need.

sebastienros:
> My solution is easy to implement: 10 lines in the pr and it preserves all the existing features.

The issue is that it also mixes concepts that shouldn't. As Bertrand said the layers should not know about zones, and themes should not know about layers. What you are trying to do is just limit the list of zones visible for each layers. That's going right against the initial goal of the feature which is to let you define some content for a zone (any zone) based on a rule. And the fact that you don't want to provide all the zones for specific layers is a very special case.

I understand what you are trying to do, and I could see some better ways to do it. Like first define some alternates on Layout per layer. Like Layout-Homepage.Layer.cshtml. And independently extend the Widgets module (not the theme manifest) to limit the names of zones available for a Theme. But this will be even more useful when the LayoutPart editor is integrated in the Widgets editor, so that users can drag and drop widgets, and also optionally define custom theme-level layouts.

CSurieux:
It is clear that what I propose is another definition for layers and its relation with zones.
What should be and what should not, I don't know.
Concerning the all drag and drop concept for orchard, it is strange for such a product and such an audience. May be some retro futurist way when tablets are everywhere.

sebastienros:
I am closing this PR as it's not what we want. We can continue the discussion on the issue itself.

CSurieux:
It seems I missed the discussion ? 
Where does it took place ?

bleroy:
Bring it up at the community meeting if you want.

CSurieux:
No chance to have this adopted in core, even if it preserves some other way - using placement, zones and widgets, not limited to Drag&Drop - to extend Orchard without breaking everything.

PR #5946, opened 2015-10-13
Fixes #5935 Updates Azure.Storage to 5.0.2.0

### LINKED ISSUE #5935: Update WindowsAzure.Storage to 5.0.2.0
When trying to install the azure storage nuget package, 5.0.0.0 is not available.  5.0.2.0 is the closest available version.  The file in the lib folder should probably match this to make it easier to update.  Unfortunately, I recommend this for 1.9.2 instead of 1.9.x but obviously defer to @sebastienros 
capture

dcinzona:
This is to match the Nuget file version


Jetski5822:
:shipit:

Issue #5869, opened 2015-09-28
Cannot resolve parameter 'Orchard.IWorkContextAccessor workContextAccessor' when running Cloud Service project


mattcowen:
I am getting the following when trying to run the Azure Cloud Service project off the "dev" branch.

The type 'Orchard.Azure.Services.TaskLease.ApplicationEnvironment, Orchard.Azure' could not be found. It may require assembly qualification, e.g. "MyType, MyAssembly".

I guess this should be AzureApplicationEnvironment but that doesnt work either. Anyone know or have any ideas why this type isn't being found?

mattcowen:
I fixed this by making the type Orchard.Azure.Services.Environment.AzureApplicationEnvironment, Orchard.Azure.

However now I get....

Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: Autofac.Core.DependencyResolutionException: None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'Orchard.Environment.DefaultOrchardShell' can be invoked with the available services and parameters:
Cannot resolve parameter 'Orchard.IWorkContextAccessor workContextAccessor' of constructor 'Void .ctor(Orchard.IWorkContextAccessor, System.Collections.Generic.IEnumerable`1[Orchard.Mvc.Routes.IRouteProvider], System.Collections.Generic.IEnumerable`1[Orchard.WebApi.Routes.IHttpRouteProvider], Orchard.Mvc.Routes.IRoutePublisher, System.Collections.Generic.IEnumerable`1[Orchard.Mvc.ModelBinders.IModelBinderProvider], Orchard.Mvc.ModelBinders.IModelBinderPublisher, Orchard.Tasks.ISweepGenerator, System.Collections.Generic.IEnumerable`1[Orchard.Owin.IOwinMiddlewareProvider], Orchard.Environment.Configuration.ShellSettings)'.

Source Error: 


Line 111:            var shellScope = _shellContainerFactory.CreateContainer(settings, blueprint);
Line 112:
Line 113:            return new ShellContext {
Line 114:                Settings = settings,
Line 115:                Descriptor = descriptor,

Source File: d:\Projects\Orchard-1.9.x\src\Orchard\Environment\ShellBuilders\ShellContextFactory.cs    Line: 113

sfmskywalker:
[...]

PR #5950, opened 2015-10-14
Fix for orchard issue #5869

mattcowen:
This fixes a problem when using cloud services. The Azure web project had lost the msbuild tasks that copy the core, modules and themes folders from the Orchard.Web project over to the Azure Web project

### DIFF
--- a/src/Orchard.Azure/Orchard.Azure.Web/Orchard.Azure.Web.csproj
+++ b/src/Orchard.Azure/Orchard.Azure.Web/Orchard.Azure.Web.csproj
@@ -561,6 +561,23 @@
   <Target Name="MvcBuildViews" AfterTargets="AfterBuild" Condition="'$(MvcBuildViews)'=='true'">
     <AspNetCompiler VirtualPath="temp" PhysicalPath="$(WebProjectOutputDir)" />
   </Target>
+  <Target Name="DeleteDebugFiles" AfterTargets="AfterBuild">
+    <RemoveDir Directories="Themes;Core;Modules" />
+  </Target>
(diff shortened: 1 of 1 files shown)


sebastienros:
Is this to make Debug work? Because deployment worked just fine for me.

mattcowen:
Yes, this is purely for getting the debugging of your cloud service in the local emulator. The project already contained what was required to deploy to a cloud service.

I'm simply adding what was in there before - not changed anything with what I've added.

sebastienros:
> I'm simply adding what was in there before

Can you point me to the commit where it was removed, just to be sure it was a mistake and not on purpose.

mattcowen:
I've gone back to https://github.com/OrchardCMS/Orchard/commit/3e20412ef011f840a09b24aff62eeaddf42d5ec4 
and can't find it by checking 5 or so commits on the way. There is a reference in this commit's version to a proj file that appears to copy modules and themes. 

I can vouch for it working (if that helps).

dcinzona:
this PR is a confirmed fix for cloud service deployments (confirmed on gitter)

carlwoodhouse:
nah @dcinzona turned out this isn't what was causing my issue, this just fixes debugging in the emulator

dcinzona:
Ah, thanks for clarifying @carlwoodhouse - thought it did fix the issue based on the convo on gitter, but interesting to know it did not.

carlwoodhouse:
@dcinzona at the same so did i, but i also fixed something else which actually fixed it .. :P and that thing was very specific to me being an idiot.

Issue #5958, opened 2015-10-16
OrchardRedis - RedisDatabaseExtensions, use of keys in script.


carlwoodhouse:
The RedisDatabaseExtensions (https://github.com/OrchardCMS/Orchard/blob/1.9.2/src/Orchard.Web/Modules/Orchard.Redis/Extensions/RedisDatabaseExtensions.cs) are using the `KEYS` command during script for counts, and deleting by prefix?

The `KEYS` command is pretty bad, in fact the redis documentation warns explicitly against it in production. (http://redis.io/commands/KEYS) as it can effectively dos your redis server. `SCAN` is recommended over `KEYS`. I also think `KEYS` won't work properly ran like this (from script) in a redis server set up in a cluster (eg. the new azure redis premium tiers), as `KEYS` targets a single server. (though the same can be said of `SCAN`) reference - https://github.com/StackExchange/StackExchange.Redis/blob/master/Docs/KeysScan.md

The RedisOutputCacheProvider (https://github.com/OrchardCMS/Orchard/blob/1.9.2/src/Orchard.Web/Modules/Orchard.Redis/OutputCache/RedisOutputCacheStorageProvider.cs) actually implements this much better but only in one place, see line 125). This uses the .Keys method on the StackExchange.Redis client api, which will use `SCAN` instead of `KEYS` if the server supports it (older versions of redis don't). It also handles the cluster scenario by looping through each endpoint. Unfortunately in other places the provider is using the databaseextensions previously mentioned.

The extensions probably want updating to use the .Keys method on the StackExchange.Redis client api and to loop through the available endpoints.

sebastienros:
Would you mind providing a PR?

carlwoodhouse:
@sebastienros - sure, will try and look over it more this weekend and get a pr over

carlwoodhouse:
Going to test this a bit more tomorrow but i essentially have a pr once i'm happy with it

BenedekFarkas:
This might have been some sort of merge error, but `RedisConnectionExtensions.cs` is still in the repo on `1.10.x`, but not added to `Orchard.Redis.csproj`. The commit fixing this issue actually removed `RedisDatabaseExtensions.cs` (still in the repo and the csproj) in favor of `RedisConnectionExtensions.cs`.
I think this basically means that this bug is NOT fixed on `1.10.x`.

sebastienros:
Please file another issue and tag it 1.10.2

[...]

PR #5967, opened 2015-10-20
Redis extensions use of keys #5958

carlwoodhouse:
* Changed the database extensions to be connection extensions, as these are connection specific commands to accomodate redis in cluster configuration
* Updated the extensions to use the method supplied by StackExchange.Redis for Keys instead of calling script, this will perform scan on redis servers where it is available over keys
* Updated the outputcache provider to only call the scan/keys command once per request and use the updated extensions
* Updated the storage provider to use the updated extensions


sebastienros:
Looks good. Waiting for next triage to pull it.

Issue #5980, opened 2015-10-23
Adding module dependency to Orchard.ImportExport will cause Setup to fail due to Orchard.Setup.Services


sfmskywalker:
If this sounds weird to you, then you're not alone. I wasted a good half day figuring out why Setup suddenly started to fail with a YSOD. Turned out it's because I added a feature dependency on `Orchard.ImportExport`, which in turn depends on a recently added feature called `Orchard.Setup.Services`. Enabling that feature while enabling `Orchard.Setup` (which happens when you setup a new site) will cause a YSOD with a seemingly completely unrelated exception:

![](http://i.imgur.com/rXsBpRO.png)

You can reproduce this by simply adding a dependency to `Orchard.ImportExport` (or `Orchard.Setup.Services`) to any feature of any module, even if that feature is never enabled, and launching setup.

At this point I am clueless as to why this happens, but this is what I do know:

- The `Orchard.Setup` modules has two features: `Orchard.Setup` and `Orchard.Setup.Services`
- The `Orchard.Setup` feature **has a dependency on** `Orchard.Setup.Services`. This has a smell to it, because this is probably the first time you've seen an *implicit/default feature* (`Orchard.Setup`) have a dependency on one of the other features of the same module.
- The only service bound to the `Orchard.Setup.Services` feature is the `SetupService` class.
- The reason that class is bound to a separate feature is because the `Orchard.ImportExport` module uses that class, but the `Orchard.Setup` feature is automatically disabled once the setup process is done. This means `Orchard.ImportExport` cannot inject services bound to the `Orchard.Setup` feature, since it is disabled. Hence the recent introduction of the `Orchard.Setup.Service` feature, which is enabled at all times.

But apparently that was too clever even for Orchard, so now things explode and trees go up in flames before you can even get to Setup and there's a module manifest with a dependency on `Orchard.Setup.Service` (which only happens in custom modules; none of the core modules have a dependency on Orchard.ImportExport).

> Another thing I don't understand is why Setup doesn't fail with `Orchard.ImportExport` depending on `Orchard.Setup.Services`. Seems weird, but that is most likely more evidence of my lack of understanding.

So far the description of the problem. Here are a couple of suggestions to fix this mess (one suggestion even messier than the next, but at least they fix the bug). From bad to worse:

[...]

PR #5981, opened 2015-10-23
Fixes dependency on Orchard.ImportExport bug by removing the Orchard.Setup.Services feature

sfmskywalker:
This solves the issue as described in #5980 where the Setup screen would throw a YSOD if any module would have a dependency on Orchard.ImportExport.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ImportExport/Services/ISetupServiceFactory.cs
+++ b/src/Orchard.Web/Modules/Orchard.ImportExport/Services/ISetupServiceFactory.cs
@@ -1,0 +1,10 @@
+﻿using Orchard.Setup.Services;
+
+namespace Orchard.ImportExport.Services {
+    /// <summary>
+    /// We need to manually instantiate the SetupService class because the Orchard.Setup feature will be disabled after setup completes.
+    /// </summary>
+    public interface ISetupServiceFactory : IDependency {
+        ISetupService CreateSetupService();
+    }
+}
--- a/src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupServiceFactory.cs
+++ b/src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupServiceFactory.cs
@@ -1,0 +1,50 @@
+using Orchard.Environment;
+using Orchard.Environment.Configuration;
+using Orchard.Environment.ShellBuilders;
+using Orchard.Environment.State;
+using Orchard.Recipes.Services;
+using Orchard.Setup.Services;
+
+namespace Orchard.ImportExport.Services {
+    /// <summary>
+    /// We need to manually instantiate the SetupService class because the Orchard.Setup feature will be disabled after setup completes.
+    /// </summary>
+    public class SetupServiceFactory : ISetupServiceFactory {
+        private readonly ShellSettings _shellSettings;
+        private readonly IOrchardHost _orchardHost;
+        private readonly IShellSettingsManager _shellSettingsManager;
+        private readonly IShellContainerFactory _shellContainerFactory;
+        private readonly ICompositionStrategy _compositionStrategy;
+        private readonly IProcessingEngine _processingEngine;
+        private readonly IRecipeHarvester _recipeHarvester;
+
+        public SetupServiceFactory(
+            ShellSettings shellSettings,
+            IOrchardHost orchardHost,
+            IShellSettingsManager shellSettingsManager,
+            IShellContainerFactory shellContainerFactory,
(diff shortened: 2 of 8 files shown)


src/Orchard.Web/Modules/Orchard.ImportExport/Services/SetupServiceFactory.cs
```diff
@@ -0,0 +1,50 @@
...
+using Orchard.Environment.Configuration;
+using Orchard.Environment.ShellBuilders;
+using Orchard.Environment.State;
+using Orchard.Recipes.Services;
+using Orchard.Setup.Services;
+
+namespace Orchard.ImportExport.Services {
+    /// <summary>
+    /// We need to manually instantiate the SetupService class because the Orchard.Setup feature will be disabled after setup completes.
+    /// </summary>
```
DanielStolt: Superfluous to duplicate this comment here I think.

src/Orchard.Web/Modules/Orchard.ImportExport/Services/ISetupServiceFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Orchard.Setup.Services;
+
+namespace Orchard.ImportExport.Services {
+    /// <summary>
+    /// We need to manually instantiate the SetupService class because the Orchard.Setup feature will be disabled after setup completes.
+    /// </summary>
```
DanielStolt: Maybe this comment could include more context, such as:

    This factory service exists because ISetupService cannot be dependency injected due to the fact that the Orchard.Setup feature will be disabled after setup completes. Instead, this module takes only an assembly reference on Orchard.Setup and instantiates the SetupService implementation directly.

Jetski5822:
YES YES YES YES YES YES YES - This was fooling me with Brochard too. Mind if I merge it?

DanielStolt:
I think a general problem with this approach is that `SetupService` cannot be substituted and/or extended by third-party code. This feature is now hard-coded to instantiate the built-in `SetupService` directly which in effect nullifies the whole purpose of the `ISetupService` interface.

IMO it would be better to remove the option to run setup from import. I always felt this feature was rather awkward and wierd in the first place.

DanielStolt:
@Jetski5822 I mind. :)

Jetski5822:
Okay - ill hold off... got a bit excited ;)

sfmskywalker:
Summoning @sebastienros. Do you agree that we should get rid of the *Reset Site* feature (where the user can re-cook the site on the Import screen with an uploaded recipe)?

sfmskywalker:
> I think a general problem with this approach is that SetupService cannot be substituted and/or extended by third-party code. This feature is now hard-coded to instantiate the built-in SetupService directly which in effect nullifies the whole purpose of the ISetupService interface

Not really, because one can provide a different implementation of `ISetupServiceFactory` which can return a different implementation of `ISetupService`.

DanielStolt:
That's true, good point.

Still it feels brittle to depend on an implementation whose containing module you can't depend on. What if the implementation `SetupService` needed to depend on other things in `Orchard.Setup` which *do* require the module to actually be enabled?

jtkech:
I had this kind of issue with a module manifest file. Not sure, but as I remember I had to deal with the module sub features. Here, the idea would be to define the effective setup as a sub featue that depends on the other one, and disable this sub feature at the end of the setup but keeping the module itself enabled. Just an idea, not tested, I will do it asap.

Best

sfmskywalker:
@DanielStolt Yeah it's a strange one.
@jtkech Yes, so then the scenario would be in line with the rest of the modules and their features. Please let me know how that goes.

jtkech:
@sfmskywalker, it was not so easy to figure it out. In different points, the features are ordered by dependencies and priorities, as before and in the shape table manager... Because the `Setup` uses its own `Document.cshtml`, it needs to win over the `Core Shapes` module. If not, the `Core` one is used and fails because e.g `CurrentCulture` is null...

So, it may depend on which module has a new shared dependency with the `Orchard Setup`. But, because the final shapes ordering also depends on the priority of their related feature, by only adding this to the `Setup Module.txt` file, it seems to work fine.

    Priority: 1

Best

sfmskywalker:
@jtkech Wow, that must have been so tricky to figure out. You're right, adding the Priority setting makes the error go away. Awesome find, I didn't even know that existed. Such a simple change!

I'll close this PR and create a new one with an updated `Module.txt` for `Orchard.Setup`.
This is great man, thanks a lot.

Issue #5111, opened 2015-03-20
nonce encoding issue


orchardbot:
**MrGenius** created:
https://orchard.codeplex.com/workitem/21282

Hi,

I recently encountered a problem with the nonce data in the url being modified by the browser. That is, the sent nonce and received nonce wouldn't match due to the url encoding. The browser changes "+", "/", etc characters into something else and hence the DecryptNonce method in UserService fails.

I found this solution. Please validate.
http://stackoverflow.com/questions/1228701/code-for-decoding-encoding-a-modified-base64-url?rq=1

Here is the fix I have got.
 public string CreateNonce(IUser user, TimeSpan delay) {
            var challengeToken = new XElement("n", new XAttribute("un", user.UserName), new XAttribute("utc", _clock.UtcNow.ToUniversalTime().Add(delay).ToString(CultureInfo.InvariantCulture))).ToString();
            var data = Encoding.UTF8.GetBytes(challengeToken);
            //return Convert.ToBase64String(_encryptionService.Encode(data));
            return HttpServerUtility.UrlTokenEncode(_encryptionService.Encode(data));
        }

```
    public bool DecryptNonce(string nonce, out string username, out DateTime validateByUtc) {
        username = null;
        validateByUtc = _clock.UtcNow;

        try {
            //var data = _encryptionService.Decode(Convert.FromBase64String(nonce));
            var data = _encryptionService.Decode(HttpServerUtility.UrlTokenDecode(nonce));
            var xml = Encoding.UTF8.GetString(data);
            var element = XElement.Parse(xml);
            username = element.Attribute("un").Value;
            validateByUtc = DateTime.Parse(element.Attribute("utc").Value, CultureInfo.InvariantCulture);
            return _clock.UtcNow <= validateByUtc;
        }
        catch {
            return false;
        }

    }
```

PR #5985, opened 2015-10-27
Fix for browser unfriendly characters in the nonce Url.

kumards:
At times, nonce generated using Convert.ToBase64String contains spaces. These type of nonces are altered by browser and ended up failing while decrypting nonce.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Services/UserService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Services/UserService.cs
@@ -14,6 +14,7 @@ using System.Globalization;
 using System.Text;
 using Orchard.Messaging.Services;
 using Orchard.Environment.Configuration;
+using System.Web;
 
 namespace Orchard.Users.Services {
     public class UserService : IUserService {
@@ -85,7 +86,7 @@ namespace Orchard.Users.Services {
         public string CreateNonce(IUser user, TimeSpan delay) {
             var challengeToken = new XElement("n", new XAttribute("un", user.UserName), new XAttribute("utc", _clock.UtcNow.ToUniversalTime().Add(delay).ToString(CultureInfo.InvariantCulture))).ToString();
             var data = Encoding.UTF8.GetBytes(challengeToken);
-            return Convert.ToBase64String(_encryptionService.Encode(data));
+            return HttpServerUtility.UrlTokenEncode(_encryptionService.Encode(data));
         }
 
         public bool DecryptNonce(string nonce, out string username, out DateTime validateByUtc) {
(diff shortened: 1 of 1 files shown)


Jetski5822:
is this fixing https://github.com/OrchardCMS/Orchard/issues/5111 ?

kumards:
yeah, actually it was me who created that issue.

src/Orchard.Web/Modules/Orchard.Users/Services/UserService.cs
```diff
@@ -85,7 +86,7 @@ ISiteService siteService
         public string CreateNonce(IUser user, TimeSpan delay) {
             var challengeToken = new XElement("n", new XAttribute("un", user.UserName), new XAttribute("utc", _clock.UtcNow.ToUniversalTime().Add(delay).ToString(CultureInfo.InvariantCulture))).ToString();
             var data = Encoding.UTF8.GetBytes(challengeToken);
-            return Convert.ToBase64String(_encryptionService.Encode(data));
+            return HttpServerUtility.UrlTokenEncode(_encryptionService.Encode(data));
```
sebastienros: I'd rather re-encode the base64 string. This way it's less breaking as currently working nonce won't be affected, and also still hides the content from plain sight.

sebastienros:
Ping? And by the way can you target 1.9.x please?

src/Orchard.Web/Modules/Orchard.Users/Services/UserService.cs
kumards: I don't understand what exactly you mean by "re-encode the base64 string".

sebastienros: I meant `var data = _encryptionService.Decode(Convert.FromBase64String(HttpServerUtility.UrlTokenDecode(nonce)));` and `HttpServerUtility.UrlTokenEncode(Convert.ToBase64String(_encryptionService.Encode(data)))`

kumards: If I make the suggested code compile-able, I would end up with something like this.

``return 
HttpServerUtility.UrlTokenEncode(Convert.FromBase64String(Convert.ToBase64String(_encryptionService.Encode(data))));``
and
``var data = _encryptionService.Decode(Convert.FromBase64String(Convert.ToBase64String(HttpServerUtility.UrlTokenDecode(nonce))));``

That's repetitive conversion string to bytes and vice versa. Please bear with me, I am really missing something here.

sebastienros:
Let's re-start the discussion.

I think the url encoding should be done in the controller, or wherever the nonce is used in a link. The `CreateNonce` and `DecryptNonce` will only manipulate a Base64 encoded value. The controller will `UrlEncode/Decode` it when necessary.

That should make things simpler to understand.

kumards:
That makes lot of sense. Its the concern of the caller!

PR #6006, opened 2015-11-03
Fires Updating and Updated events when menu items are re-ordered

### LINKED ISSUE #6005: Re-ordering menu items within a menu does not trigger the Updating and Updated events for each MenuPart
Because the properties on this part is updated by the Navigation Admin Controller (as opposed to a driver) then these events are not raised.

Suggest injecting the handlers into the controller and invoking those events of the position property changes on the part.

paynecrl97:
Fixes #6005

### DIFF
--- a/src/Orchard.Web/Core/Navigation/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Core/Navigation/Controllers/AdminController.cs
@@ -1,14 +1,10 @@
 ﻿using System.Collections.Generic;
-using System.ComponentModel;
 using System.Linq;
 using System.Web.Mvc;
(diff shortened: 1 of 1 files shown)


sebastienros:
Shouldn't it be the same from the Menu content item BTW?

sfmskywalker:
@sebastienros Are you proposing that whenever a `Menu` content item is deleted, the `Updating` and `Updated` events on its menu items should be invoked as well? I think that would make sense.

paynecrl97:
Agree. I can do that. Leave it with me.

sebastienros:
No I meant calling update on the Menu, which is what is already done, my bad.

PR #6015, opened 2015-11-05
Fixed reply button and comment link and added useful CSS class.

ryandrewburnett:
1. Fixed comment links so that when the user clicks on comment metadata on the front end the desired scrolling effect is achieved.
2. Added a `.has-replies` class to comments that have replies for easier styling.
3. Stopped the comment reply button from showing when comments are disabled. Also needs fixing to not display the reply button if a user doesn't have permission to post comments.
4. Removed span around reply button as it adds no ease to styling.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Comments/Views/Parts.Comment.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Comments/Views/Parts.Comment.cshtml
@@ -8,8 +8,13 @@
     var children = New.List(Items: Model.Items);
     children.Classes.Add("comments");
 
+    if (Model.Items.Count > 0) {
+        Model.Classes.Add("has-replies");
+    }
+    
     Model.Classes.Add("comment");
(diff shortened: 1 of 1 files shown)


ryandrewburnett:
Have now addressed not displaying the reply button if a user doesn't have permission to post comments.

src/Orchard.Web/Modules/Orchard.Comments/Views/Parts.Comment.cshtml
```diff
@@ -1,15 +1,22 @@
...
     children.Classes.Add("comments");
 
+    if (Model.Items.Count > 0) {
+        Model.Classes.Add("has-replies");
+    }
+    
     Model.Classes.Add("comment");
-    Model.Classes.Add("comment-"+comment.Id);
+
+    Model.Id = "comment" + comment.Id;
```
sebastienros: could you add a dash here? as in `comment-123`

Issue #5995, opened 2015-10-31
Layout Editor ElementController behavior appears incorrect


harmony7:
I've noticed that ElementController, in its `Update()` method, goes to `Request.Form.ToDictionary()` to initialize the Data dictionary.

Update():

            var elementData = ElementDataHelper.Deserialize(sessionState.ElementData);
            var data = Request.Form.ToDictionary();
            var element = _elementManager.ActivateElement(descriptor, e => e.Data = data);
            var context = CreateEditorContext(session, describeContext.Content, element, postedElementData: data, elementData: elementData, updater: this);
            var editorResult = _elementManager.UpdateEditor(context);

For comparison, Edit():

            var data = elementData.Combine(ElementDataHelper.Deserialize(sessionState.ElementEditorData));
            var element = _elementManager.ActivateElement(descriptor, e => e.Data = data);
            var context = CreateEditorContext(session, describeContext.Content, element, postedElementData: data, elementData: elementData);
            var editorResult = _elementManager.BuildEditor(context);

I feel that going to `Request.Form` is incorrect, especially for the purpose of simply saving it into the Data dictionary.

1. Updating the element with submitted data is supposed to happen in the Element driver's `UpdateEditor()` method when `context.Updater != null` (and you have many examples of this in your code too). This gives control over which fields to transfer from the submitted data into the element (or viewModel).
2. There is no guarantee that the `Request.Form` is going to have fields that are compatible with the Element. For example, I often write view models to pass to the EditorTemplate, rather than directly pass the Element to the EditorTemplate.

            var viewModel = new MyElementEditorViewModel();
            /* fill in viewModel based on Element data */
            // e.g.
            viewModel.Field1 = element.Value1;

            if (context.Updater != null) {
                context.Updater.TryUpdateModel(viewModel, context.Prefix, null, null);
                /* update Element based on viewModel */
                // e.g.
                element.Value1 = viewModel.Field1;
            }
     
            var editor = context.ShapeFactory.EditorTemplate(TemplateName: "Elements/TemplateName", Model: viewModel);
            return Editor(context, editor);

I want to suggest that the correct behavior in `Update()` is to initialize Data by using `ElementDataViewModel.ElementData` instead of `Request.Form`. In turn, the actual updating of the element will be performed in the `if (context.Updater != null)` branch of the ElementDriver.

What do you guys think?

harmony7:
I found another problem, that's that ElementController modifies ValueProvider when it creates the EditorContext.

[...]

PR #6026, opened 2015-11-11
Fixed behavior of Element Controller in an around editing and updating.

harmony7:
This changeset is mainly changes to `ElementController`, along with some miscellaneous cleanup.

The important changes are in the `Update` method, which was (incorrectly, as far as I can tell) attempting to update the data of the element instead of allowing the `ElementDriver` to run. I wrote a bit about this in #5995.

The other miscellaneous changes are the removal of file `designer.canvas.js` (as it was completely unused) and the `Create` methods of `ElementController`, as well as a few items in the settings of the `LayoutDesignerHost` script.


sebastienros:
Looks good, waiting for the iDeliverables to give it a go.

harmony7:
The impact of this mainly for me was now I may use a View Model class that does not have field names that are necessarily the same as those on the Element (which you should be able to do).

sfmskywalker:
I like the changes, but since I did certain things in a certain way for certain reasons, this change potentially breaks certain scenarios. However, those scenarios don't exist out of box, and I agree that it should be up to the `ElementDriver` to read from the ValueProvider, instead of updating the element's `Data` property directly (which may very well be an artifact from early iterations of the module), so I suggest we take this PR and see what gives.

harmony7:
Thanks for accepting the PR.

@sfmskywalker yeah i figured they were probably done that way for a reason and this PR is not meant to undermine your efforts. If you can remember what those "certain situations" were, maybe we can look at those.

PR #6028, opened 2015-11-11
Update CacheItemRecord.cs

mvarblow:
Fixed database output cache module when using SQL Server database.  This resolves the following exception which results from NHibernate trying to truncate the Output field to 8K:

NHibernate.AssertionFailure: null id in Orchard.OutputCache.Models.CacheItemRecord entry (don't flush the Session after an exception occurs)


DanielStolt:
Excellent - thank you!

Issue #6035, opened 2015-11-15
Shape element of layout problem in latest dev


infofromca:
it happens between    adc92fb1b4fe59d00889f3017763f4bdd2132408 and  4573f65c19ac6aaef1a609c0e2fa38cd954860c6

create new page,
drag the SHAPE from LAYOUT =>SNIPPETS,
in the popup text box , enter the name of shape
save, does not work.
(if you use the same steps to edit, you can not see the shape name which was already there.)

now i have to use version adc92fb1b4fe59d00889f3017763f4bdd2132408

harmony7:
This might be one of things that "broke" due to fixing #5995. I'll take a look when I have some time.

sfmskywalker:
Fixed in changeset 8693b68ac7ed899e1655e488af7e2d32ea7a76bc.

PR #6042, opened 2015-11-17
Update to 5995 Element Controller Behavior

harmony7:
This is an additional fix for #5995.

My fix (PR #6026) "broke" cases such as #6035 of updating elements whose UIs were designed using the Forms API. It seems that these forms were the reason that the code before what I changed in #6026 included all of the acrobatics with the element's data dictionary and messing with `ValueProvider`.

I've made updates to `FormElementDriver` so that this case is handled. The code instantiates the involved form and iterates through their fields, copying the appropriate values out of the value provider and assigning them to the data dictionary. This I believe is a better approach than the previous way of simply copying Request.Form into the dictionary.

The other thing is, I had meant for this to be a fix for 1.9.x, since it doesn't break any interfaces and is what I would consider a "bug fix". If I'd made the previous PR against dev, then that was my mistake. I'm submitting both changesets again, as a pull request into 1.9.x.


sfmskywalker:
This is looking good. Regarding 1.9.x, on the one hand you're right, but due to the nature of the changes, certain scenarios could break in unexpected ways (as @infofromca reported), so it is safer to merge it into dev. Also, despite the fact that your code is definitely an improvement, the current code in 1.9 is not causing issues as far as I know, so I think that is another reason to apply your changes to dev rather than 1.9.x.

Perhaps @sebastienros has a different opinion: 1.9.x or dev?

harmony7:
@sfmskywalker i can see that. but please remember I started going down this path because of a bug: I had a problem writing a driver for my own element whose view model didn't have the same property names as the element. But sure let's wait for @sebastienros 's call

sfmskywalker:
@harmony7 Ok. In any case, we think it's safest to merge it into dev in this case. If you could retarget your PR, that'd be awesome. Thanks!

src/Orchard.Web/Modules/Orchard.Layouts/Framework/Drivers/FormsElementDriver.cs
```diff
@@ -32,8 +33,33 @@ public abstract class FormsElementDriver<TElement> : ElementDriver<TElement>, IF
...
+                        var name = ((dynamic) shape).Name as string;
+                        if (name != null) {
+                            var value = context.ValueProvider.GetValue(prefix + name);
+                            if (value != null) {
+                                context.Element.Data[name] = value.AttemptedValue;
+                            }
+                        }
+                    }));
+                }
+            }
```
jtkech: @harmony7 after a quick try, indeed posted data are taken into account. Not sure but, as seen in my tests, when you re-edit the element, you don't retreive these data. Here, I needed to add this

            else // here when context.Updater == null
                context.ValueProvider = context.Element.Data.ToValueProvider(CultureInfo.InvariantCulture);

Note: maybe here it's better to use `_cultureAccessor.CurrentCulture` but I would have to change the `FormsElementDriver contructor` that is used in many places...

Best

jtkech:
@harmony7 did you see my line note? As seen in my tests, when you re-edit the form element, you don't retreive its data. In `UpdateElementData()`, I needed to add this

        else // (when context.Updater == null)
            context.ValueProvider = context.Element.Data.ToValueProvider(CultureInfo.InvariantCulture);

Note: maybe here it's better to use `_cultureAccessor.CurrentCulture` but I would have to change the `FormsElementDriver contructor` that is used in many places...

Best

harmony7:
Yeah I saw it, haven't had time to think about it yet. Thanks for the input.

sfmskywalker:
I just worked on a fix that handles both updating the ElementData as per @harmony7 's suggestion as well as rebinding the form shape itself with the ElementData if not currently in a postback as per @jtkech 's suggestion. It's indeed best to use `ICultureAccessor`, but does break the base driver class, but that's easy to update, and since we're still on `dev` I think it's fine. To prevent breaking changes like this in the future I added a "services" interface that exposes `IFormManager` as well as `ICultureManager`, so that we can potentially add more properties to it without breaking the base driver class' signature.

Issue #6043, opened 2015-11-17
Error with listable content when the user does not have the EditContent permission


DanielLackenbyBede:
Straight down to brass tacks:
Orchard.Core.Contents.Controllers.AdminController [ln:83](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Core/Contents/Controllers/AdminController.cs#L83)

    var query = _contentManager.Query(versionOptions, GetListableTypes(false).Select(ctd => ctd.Name).ToArray());

This makes a call to the method _GetListableTypes()_ which is as follows:
Orchard.Core.Contents.Controllers.AdminController [ln:150](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Core/Contents/Controllers/AdminController.cs#L150)

        private IEnumerable<ContentTypeDefinition> GetListableTypes(bool andContainable) {
            return _contentDefinitionManager.ListTypeDefinitions().Where(ctd =>
                Services.Authorizer.Authorize(Permissions.EditContent, _contentManager.New(ctd.Name)) &&
                ctd.Settings.GetModel<ContentTypeSettings>().Listable &&
                (!andContainable || ctd.Parts.Any(p => p.PartDefinition.Name == "ContainablePart")));
        }

Now, when the user authorizes against *Permissions.EditContent* then the check goes on to check whether the Content Type is Listable and the whether the Containable condition checks out. This will successfully return a list of all the types that the user should see. If the user does not authorize against *Permissions.EditContent* then _GetListableTypes(false).Select(ctd => ctd.Name).ToArray()_ will return an empty array. So far so good.

Now, that array of string is passed into _\_contentManager.Query(...)_ and eventually comes down to:
Orchard.ContentManager.DefaultContentQuery [ln:94](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard/ContentManagement/DefaultContentQuery.cs#L94)

       private void ForType(params string[] contentTypeNames) {
            if (contentTypeNames != null && contentTypeNames.Length != 0) {
                var contentTypeIds = contentTypeNames.Select(GetContentTypeRecordId).ToArray();
                // don't use the IN operator if not needed for performance reasons
                if (contentTypeNames.Length == 1) {
                    BindItemCriteria().Add(Restrictions.Eq("ContentType.Id", contentTypeIds[0]));
                }
                else {
[...]

PR #6044, opened 2015-11-17
Filter out when no contentTypeNames provided

DanielLackenbyBede:
Suggested solution to [Issue 6043](https://github.com/OrchardCMS/Orchard/issues/6043)

Previously, when no content type names where provided (from a prior conditional statement, so there is an initialized array but it is empty) then the filtering steps are skipped and instead of serving "no content
types", as an empty array would state: all content types are served.
When contentTypeNamed.Length == 0, still .Select() as this is safe for empty collections and will filter out all types in this event.

Any feedback or sidcussion welcome.

### DIFF
--- a/src/Orchard.Web/Core/Contents/AdminMenu.cs
+++ b/src/Orchard.Web/Core/Contents/AdminMenu.cs
@@ -22,6 +22,7 @@ namespace Orchard.Core.Contents {
             var contentTypeDefinitions = _contentDefinitionManager.ListTypeDefinitions().OrderBy(d => d.Name);
             builder.AddImageSet("content")
                 .Add(T("Content"), "1.4", menu => menu
+                    .Permission(Permissions.EditOwnContent)
                     .Add(T("Content Items"), "1", item => item.Action("List", "Admin", new { area = "Contents", id = "" }).LocalNav()));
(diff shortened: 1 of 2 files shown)


DanielLackenbyBede:
What this does.

Originaly: when a user had the EditContent permission then they see the expected listable content under the Content's list.
admin has relevant permissions
content list when admin has relevant permission

If these permissions are taken revoked then we see all content, even those that are none listable like layers:
admin has no relevant permissions
content list when admin has no relevant permission

With the first update, *Fix to DefaultContentQuery*, by Selecting on the empty array created by _GetListableTypes(false).Select(ctd => ctd.Name).ToArray()_ the admin who cannot edit (or create because creating is covered by the edit permissions) then it removes all content from the list seeing as the admin has no permission pertaining to Content.
content list when admin has no relevant permission with update

Following this, the admin can still readily access this page as the link is available in the menu. The second commit, _Limit Admin Menu for Content_, the Content entry on the Admin menu is limited to the EditOwnContent permission. This is because this is the most base permission in Content and any other permission that pertains to the creation of content (EditContent, PublishOwnContent, PublishContent, etc) ultimately imply this permission.
If you do not have this permission then you certainly won't be able to see anything on the Content List so the item is removed from the menu.
admin has no relevant permissions
menu without content item

Granting this permission returns the item to the menu
admin has relevant permissions
menu with content item

PR #6050, opened 2015-11-18
Fixes #6023 : Ellipsize can spilt an HTML encoded character

### LINKED ISSUE #6023: Ellipsize can spilt an HTML special character
In HtmlHelper's Ellipsize extension method, `Ellipsize` is done after `Encode`. In some rare cases, it can cut an HTML special character.
Furthermore, ellipsize counts HTML encoded characters instead of rendered characters.

TFleury:
I fixed `Excerpt` and `Ellipsize` `HtmlHelperExtensions`'s methods.
I also updated  `Ellipsize` `StringExtensions`'s method as we can't assume it deals with HTML encoded strings, and the related test.

### DIFF
--- a/src/Orchard.Tests/Mvc/Html/HtmlHelperExtensionsTests.cs
+++ b/src/Orchard.Tests/Mvc/Html/HtmlHelperExtensionsTests.cs
@@ -217,7 +217,7 @@ namespace Orchard.Tests.Mvc.Html {
             var html = new HtmlHelper(viewContext, viewDataContainer.Object);
 
             //act
-            var result = html.Excerpt("<p>foo &amp; bar</p>", 5);
+            var result = html.Excerpt("<p>foo &amp; bar</p>", 7);
 
             //assert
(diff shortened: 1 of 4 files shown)


sebastienros:
Why did you change the behavior by using a unicode char? That can be an issue, hence this will be for someone. At the same time you should add a test for this specific issue you detected, when the text is cut in the middle of an html entity:

```
const string toEllipsize = "Lorem ipsum &nbsp;";
Assert.That(toEllipsize.Ellipsize(15), Is.StringMatching("Lorem ipsum &#160;&#8230;"));
```

TFleury:
I assume `HtmlHelperExtensions `'s `Ellipsize` method is intended to be used on non HTML encoded strings, am I wrong ?

> Why did you change the behavior by using a unicode char?

If we want to decode, cut and then re-encode,  `StringExtensions`'s `Ellipsize` method should not introduce HTML encoded characters since the result will be re-encoded.

PR #6057, opened 2015-11-20
fixes #5978

### LINKED ISSUE #5978: 1.9.2 Layouts - 'Image not found' (null) when adding Image Element
After dragging an image element to the layout and selecting the image in the library picker, it always displays 'image not found'.

To replicate - after creating a new column, drag an image element to the canvas. Click the 'add' button, select an image and click the 'Select' button. 

On further testing, the image does insert correctly if you double click the image instead of clicking the select button.

I get the same issue on 1.9.x.

mkinawy:
fixes #5978 by moving the select button to the top of the media picker window **when in modal window  only** and hides the save button untill some image is selected

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library-picker.js
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library-picker.js
@@ -14,6 +14,7 @@
         var promptOnNavigate = element.data("prompt-on-navigate");
         var showSaveWarning = element.data("show-save-warning");
         var addButton = element.find(".button.add");
+        var saveButton = $('.button.save');
         var removeAllButton = element.find(".button.remove");
         var template = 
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library-picker.js
```diff
@@ -14,6 +14,7 @@
         var promptOnNavigate = element.data("prompt-on-navigate");
         var showSaveWarning = element.data("show-save-warning");
         var addButton = element.find(".button.add");
+        var saveButton = $('.button.save');
```
sebastienros: would it work with `element.find(".button.save")` just for consistency?

sebastienros:
Great initiative, thanks.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library-picker.js
mkinawy: No it wouldn't work because the `saveButton` is NOT a child of that `element`.

sfmskywalker: It would be better if the media library picker triggers an event when its list of media ids changes. Then some external script can listen for that event and toggle the Save button's visibility accordingly.

PR #6058, opened 2015-11-20
Invalidate new media profile cache

### LINKED ISSUE #6051: Creating a new Image Profile pre-caches a negative result
When a new image profile is created, the ImageProfilePartDriver calls the GetImageProfileByName method in the ImageProfileService to find out if a profile already exists by the given name. GetImageProfileByName uses the cache manager to find a result; if a result isn't cached, the acquire method performs a query to find a matching ImageProfilePartRecord in the database, and, if it doesn't, caches a result of -1 as the ID for the new profile. Since the result is -1, the ImageProfilePart returned is null, and the creation is allowed to continue. However, when trying to use that profile later in Display.MediaProfile, that negative result is still cached, so the profile is not returned from GetImageProfileByName. Eventually, I believe the cache expires, and things start working, but it takes some time.

This is the patch that seems to have caused this issue; it looks like it's been around awhile: https://github.com/OrchardCMS/Orchard/commit/42ab42e6e238d753d8dab5d3cd4667488817f0b8

flew2bits:
Fixes #6051 Added a signal that is triggered when a new profile is published. This trigger causes the cached value for the new profile to be invalidated.

### DIFF
--- a/src/Orchard.Tests/Mvc/Html/HtmlHelperExtensionsTests.cs
+++ b/src/Orchard.Tests/Mvc/Html/HtmlHelperExtensionsTests.cs
@@ -217,7 +217,7 @@ namespace Orchard.Tests.Mvc.Html {
             var html = new HtmlHelper(viewContext, viewDataContainer.Object);
 
             //act
-            var result = html.Excerpt("<p>foo &amp; bar</p>", 5);
+            var result = html.Excerpt("<p>foo &amp; bar</p>", 7);
 
             //assert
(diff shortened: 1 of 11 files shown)


sebastienros:
Bad PR, please try again. You should do a fetch all to get the latest bits first, rebase your work on the tip, and create the PR.

sebastienros:
If you force push your current branch with the steps I explained it will fix this PR automatically.

flew2bits:
I'm unable to get this to rebase properly, so I'm going to create a new pull request.

PR #6069, opened 2015-11-22
Elements can't modify their children from on created element event.

### LINKED ISSUE #6068: Elements can't modify their children on created element event.
When layout module builds elements tree that it execute IElementSerializer.ParseNode for each element.
Parse node invoke some events like creating element and created element - and it's good place to modify element (different properties and also children). But children can't be modified from those event, because of ElementSerializer.ParseNode method.

neTp9c:
fixes #6068


sebastienros:
@sfmskywalker please review and pull if you agree

neTp9c:
I'm sorry, it's not enough to solve this problem. Can you please close the pull request?

PR #6078, opened 2015-11-25
New feature bringing alternates based on current active layers

CSurieux:
It adds a new feature to widgets and, when enabled, new alternates are added to complement existing one, using a process identical to Url Alternates.
It selects the active layers for the requested url and brings new alternates with the -layer-  prefix.
Examples:
Lets suppose that you created a layer named 'blogs first', you could then use
Menu-layer-blogs-first.cshtml
Layout-layer-blogs-first.cshtml

The default layer is no used, but the authenticated or anonymous layer could also be used.
A mix of the different layers is also proposed as in
Menu-layer-anonymous-blogs-first.cshtml

Very useful.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Orchard.DisplayManagement.Implementation;
+using Orchard.Environment.Extensions;
+using Orchard.Localization;
+using Orchard.Logging;
+using Orchard.Widgets.Models;
+using Orchard.ContentManagement;
+
+namespace Orchard.Widgets.Services
+{
+    [OrchardFeature("Orchard.Widgets.LayerAlternates")]
+    public class LayersAlternatesFactory : ShapeDisplayEvents
+    {
+        private readonly ILayerEvaluationService _layerEvaluationService;
+        private readonly Lazy<List<string>> _layersAlternates;
+        private readonly IOrchardServices _orchardServices;
+
+
+        public ILogger Logger { get; set; }
+        public Localizer T { get; set; }
+
+
+        public LayersAlternatesFactory(ILayerEvaluationService layerEvaluationService, IOrchardServices orchardServices)
+        {
+            _layerEvaluationService = layerEvaluationService;
+            _orchardServices = orchardServices;
+            Logger = NullLogger.Instance;
+            T = NullLocalizer.Instance;
+            _layersAlternates = new Lazy<List<string>>(() => {
+                int[] activeLayers = _layerEvaluationService.GetActiveLayerIds();
+
+                if (activeLayers == null || activeLayers.Length == 0)
+                {
+                    return null;
+                }
+
+                // Get Layer names and remove any ' ', '.'
+                var layers = _orchardServices.ContentManager.GetMany<LayerPart>(activeLayers, VersionOptions.Published, QueryHints.Empty)
+                        .Where(l => !l.Name.Equals("default", StringComparison.InvariantCultureIgnoreCase))
+                        .Select(l => l.Name.Replace(" ", "__").Replace(".", "_"))
+                        .ToArray();
+
+                var lyrs = Enumerable.Range(1, layers.Count()).Select(range => String.Join("__", layers.Take(range))).Union(layers).ToList();
+                return lyrs;
+            });
+        }
+
+        public override void Displaying(ShapeDisplayingContext context)
+        {
+
+            context.ShapeMetadata.OnDisplaying(displayedContext => {
+
+                if (_layersAlternates.Value == null || !_layersAlternates.Value.Any())
+                {
+                    return;
+                }
+
+                // prevent applying alternate again, c.f. https://github.com/OrchardCMS/Orchard/issues/2125
+                if (displayedContext.ShapeMetadata.Alternates.Any(x => x.Contains("__layer__")))
+                {
+                    return;
+                }
+
+                // appends Url alternates to current ones
+                displayedContext.ShapeMetadata.Alternates = displayedContext.ShapeMetadata.Alternates.SelectMany(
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    return null;
+                }
+
+                // Get Layer names and remove any ' ', '.'
+                var layers = _orchardServices.ContentManager.GetMany<LayerPart>(activeLayers, VersionOptions.Published, QueryHints.Empty)
+                        .Where(l => !l.Name.Equals("default", StringComparison.InvariantCultureIgnoreCase))
+                        .Select(l => l.Name.Replace(" ", "__").Replace(".", "_"))
+                        .ToArray();
+
```
sebastienros: The list needs to be sorted explicitly even if the service might do that already. The service might change and you don't want the alternates to change with it.

src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
```diff
@@ -0,0 +1,79 @@
...
+                var lyrs = Enumerable.Range(1, layers.Count()).Select(range => String.Join("__", layers.Take(range))).Union(layers).ToList();
```
BenedekFarkas: No need to store this in a variable, just return the value.

src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
```diff
@@ -0,0 +1,79 @@
...
+                int[] activeLayers = _layerEvaluationService.GetActiveLayerIds();
+
+                if (activeLayers == null || activeLayers.Length == 0)
+                {
+                    return null;
+                }
+
+                // Get Layer names and remove any ' ', '.'
+                var layers = _orchardServices.ContentManager.GetMany<LayerPart>(activeLayers, VersionOptions.Published, QueryHints.Empty)
+                        .Where(l => !l.Name.Equals("default", StringComparison.InvariantCultureIgnoreCase))
```
sebastienros: Maybe filter this in memory instead of from the SQL query. I don't trust SQL or the Equals/Culture operator with LINQ.

src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Linq;
+using Orchard.DisplayManagement.Implementation;
+using Orchard.Environment.Extensions;
+using Orchard.Localization;
+using Orchard.Logging;
+using Orchard.Widgets.Models;
+using Orchard.ContentManagement;
+
+namespace Orchard.Widgets.Services
+{
```
sebastienros: Brace styles please.

src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
```diff
@@ -0,0 +1,79 @@
...
+
+                if (activeLayers == null || activeLayers.Length == 0)
+                {
+                    return null;
+                }
+
+                // Get Layer names and remove any ' ', '.'
+                var layers = _orchardServices.ContentManager.GetMany<LayerPart>(activeLayers, VersionOptions.Published, QueryHints.Empty)
+                        .Where(l => !l.Name.Equals("default", StringComparison.InvariantCultureIgnoreCase))
+                        .Select(l => l.Name.Replace(" ", "__").Replace(".", "_"))
```
sebastienros: I think we should remove the spaces altogether, i.e. `Blog Post` becomes `BlogPost` and then apply the standard shape replacement: `-` to `__` and `.` to `_`.

sebastienros:
Please sign the CLA

src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
CSurieux: Yes I agréé, will add an update asap

src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
CSurieux: Suite but it is useful when debuging an hundred mines of Orchard code use variables. Tell me the time gained ?

sebastienros: Please rename to `layers`

sebastienros: can you simplify the code so that simple minds like @BenedekFarkas can understand what you meant.

CSurieux: Not easy but he gained m'y respect so....:)

CSurieux:
Never received any email about this ?? Anybody in charge of this drone ?

src/Orchard.Web/Modules/Orchard.Widgets/Services/LayerAlternatesFactory.cs
CSurieux: It just recalled me that @DisplayChildren(Model) is still not respecting the order of the items in Model if there is one... so we ould have surprised redefining templates for zones.

Piedone:

sebastienros:
Can you rebase your branch to take the new line endings into account?
Thanks

CSurieux:
My yoga 3/pro crashed its ssd and it will take me one week to recover a new env. I will do all this for the end of next week.

Working on Android with ms tools is still not an option...and Datwendo bizspark plus ended on 12/2015 so everything is slower and moving.


Envoyé via Outlook Mobile






On Thu, Jan 7, 2016 at 12:05 PM -0800, "Sébastien Ros" <notifications@github.com> wrote:





Can you rebase your branch to take the new line endings into account?
Thanks

---

jean:
Was this feature abandoned?

PR #6079, opened 2015-11-26
Fix a problem in TagCloudService when maxcount < bucket

CSurieux:
TagClouds bucket is not correctly set by tagcloudservice when the max count for all the tags is < to the number of buckets selected for the tagCloud widget. There was also some difficulties due to the usage of int in some temporary array, they have been replaced by double to solve the issues.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Tags/Services/TagCloudService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Tags/Services/TagCloudService.cs
@@ -81,12 +81,29 @@ namespace Orchard.Tags.Services {
                     }
 
                     // initialize centroids with a linear distribution
-                    var centroids = new int[buckets];
+                    var centroids = new double[buckets];
                     var maxCount = tagCounts.Any() ? tagCounts.Max(tc => tc.Count) : 0;
                     var minCount = tagCounts.Any() ? tagCounts.Min(tc => tc.Count) : 0;
-                    var maxDistance = maxCount - minCount;
-                    for (int i = 0; i < centroids.Length; i++) {
-                        centroids[i] = maxDistance/buckets * (i+1);
+                    double maxDistance = maxCount - minCount;
+                    if ( maxCount < buckets)
+                    {
+                        int i = 0;
+                        for (; i < buckets && i < maxCount;)
+                        {
+                            centroids[i] = minCount + i;
+                            i++;
+                        }
+                        for (; i < buckets; i++)
+                        {
+                            centroids[i] = int.MaxValue;
+                        }
+                    }
+                    else
+                    {
+                        for (int i = 0; i < centroids.Length ; i++)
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Tags/Services/TagCloudService.cs
```diff
@@ -81,12 +81,29 @@ into g
...
                     var maxCount = tagCounts.Any() ? tagCounts.Max(tc => tc.Count) : 0;
                     var minCount = tagCounts.Any() ? tagCounts.Min(tc => tc.Count) : 0;
-                    var maxDistance = maxCount - minCount;
-                    for (int i = 0; i < centroids.Length; i++) {
-                        centroids[i] = maxDistance/buckets * (i+1);
+                    double maxDistance = maxCount - minCount;
+                    if ( maxCount < buckets)
+                    {
+                        int i = 0;
+                        for (; i < buckets && i < maxCount;)
```
sebastienros: I think `i < buckets` is useless here as `maxCount < buckets`

src/Orchard.Web/Modules/Orchard.Tags/Services/TagCloudService.cs
```diff
@@ -81,12 +81,29 @@ into g
...
-                    var centroids = new int[buckets];
+                    var centroids = new double[buckets];
                     var maxCount = tagCounts.Any() ? tagCounts.Max(tc => tc.Count) : 0;
                     var minCount = tagCounts.Any() ? tagCounts.Min(tc => tc.Count) : 0;
-                    var maxDistance = maxCount - minCount;
-                    for (int i = 0; i < centroids.Length; i++) {
-                        centroids[i] = maxDistance/buckets * (i+1);
+                    double maxDistance = maxCount - minCount;
+                    if ( maxCount < buckets)
+                    {
```
sebastienros: Please fix the bracing style

sebastienros:
Can we add a unit test for this method? If you want I can create it first with the common cases that I know so you can show the limitation you are trying to fix.

CSurieux:
ok please.

CSurieux:
Could you help also with this cla-required thing. If I must wait an email, which is what I understood after lot of search, I have never received anything of it got blocked. If it is anything else, thanks explaining how-to.

SzymonSel:
Visit this page https://cla2.dotnetfoundation.org.

CSurieux:
Thanks. Where do you take this info from ? Do I missed some info ?

SzymonSel:
The dotnetbot post a comment with this information upon your first contribution to a .NET open source project. Perhaps you missed it. :)

-----Original Message-----

CSurieux:
Don't remember and usually I care, but it's done now. Thanks.

PR #6109, opened 2015-11-30
Content Permissions Part works incorrect on content item save.

### LINKED ISSUE #6107: Content Permissions Part works incorrect.
Imagine that user should have edit access for concrete content items only 
So only one particular role assigns to the user that has only access admin panel permissions.
In concrete content items all permission options are selected for the particular role.

When that user try to save concrete content item that items is saved but "Enable Content Item access control" (Content Permissions Part option) became unchecked. So the user doesn't have permission to concrete content item anymore.

neTp9c:
fixes #6107


neTp9c:
This pull request for 1.9.x branch (but I mistook and selected dev)

sebastienros:
Recreate it targeting 1.9.x

Issue #6117, opened 2015-12-02
Tokenised text fields on dynamic forms should decode HTML so HTML entities are not displayed


mattcowen:
If you have a form field whose value comes from a token then any HTML entities are displayed e.g. &amp; instead of &. The processed value of the token should be HTML decoded.

I will submit a PR shortly.

PR #6118, opened 2015-12-02
#6117 decode token data

mattcowen:


src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/TextFieldElementDriver.cs
```diff
@@ -79,7 +80,7 @@ public class TextFieldElementDriver : FormsElementDriver<TextField>{
         protected override void OnDisplaying(TextField element, ElementDisplayContext context) {
             context.ElementShape.ProcessedName = _tokenizer.Replace(element.Name, context.GetTokenData());
             context.ElementShape.ProcessedLabel = _tokenizer.Replace(element.Label, context.GetTokenData());
-            context.ElementShape.ProcessedValue = _tokenizer.Replace(element.RuntimeValue, context.GetTokenData());
+            context.ElementShape.ProcessedValue = HttpUtility.HtmlDecode(_tokenizer.Replace(element.RuntimeValue, context.GetTokenData()));
```
sebastienros: The actual fix is to not encode it in the first place, by using the correct `Replace` overload. But please check with @sfmskywalker if this is the correct behavior. Or anyone who could confirm.

sfmskywalker: @sebastienros You were right regarding the security issue. If you put at least two fields on a form, one of which being required and the other one providing a token, that token gets evaluated and displayed (due to the model validation error the form gets displayed and submitted values are shown). Fixing this.

sfmskywalker:
Fixed in db564d1670854459b176459b432833202f98e157.

sebastienros:
Why not fix it on 1.9.x?

sfmskywalker:
No reason. It should have been done on 1.9.x. I'll take care of it.

PR #6132, opened 2015-12-04
Updating Azure Media Player

sebastienros:
Initial commit. Needs to remove previous scripts if not necessary anymore but I need confirmation.


src/Orchard.Web/Modules/Orchard.Azure.MediaServices/Orchard.Azure.MediaServices.csproj
```diff
@@ -147,6 +147,8 @@
     <Reference Include="System.Xml.Linq" />
   </ItemGroup>
   <ItemGroup>
+    <Content Include="Assets\Css\azuremediaplayer.min.css" />
+    <Content Include="Assets\JavaScript\Lib\azuremediaplayer.min.js" />
```
DanielStolt: We shouldn't commit minified resources in the repo. Instead commit dev version and let the gulp pipeline handle the minification, as previously agreed.

sebastienros: Correction: minified versions should be handled by gulp, I agree, though they will still be committed in the repos.
I will try to find the non-minified version for these files.

sebastienros:
- [ ] Add non-minified version of scripts
- [ ] Remove injector pipeline

sebastienros:
Please fix the merge issues.

Jetski5822:
@sebastienros you are the one that needs to fix merge issues :) haha

TFleury:
It seems the module is outdated. I opened an issue to discuss that #7671

sebastienros:
Closing it in favor of https://github.com/OrchardCMS/Orchard/pull/7618

PR #6139, opened 2015-12-05
Import and Export for menu element.

neTp9c:
It process identifier of selected menu in menu element (like query in projection element).


src/Orchard.Web/Modules/Orchard.Layouts/Drivers/MenuElementDriver.cs
```diff
@@ -124,5 +132,23 @@ public class MenuElementDriver : ElementDriver<Menu> {
...
+
+        protected override void OnExporting(Menu element, ExportElementContext context) {
+            var menu = _contentManager.Get(element.MenuContentItemId);
+            var menuIdentity = menu != null ? _contentManager.GetItemMetadata(menu).Identity.ToString() : default(string);
+
+            if (menuIdentity != null)
+                context.ExportableData["MenuId"] = menuIdentity;
+        }
+
+        protected override void OnImportCompleted(Menu element, ImportElementContext context) {
```
sebastienros: Why not `OnImporting` ?

neTp9c:
This was done only for getting identifier of the menu. But know I see that the order of the elements does not matter (first item in recipe can get identifier of second item on OnImporting event). So in this case OnImportCompleted can be replaced by OnImporting without any loss. Should I do to change? I suppose it will be better.

sfmskywalker:
If you could change it to `OnImporting` then that would be great. Thank you.

neTp9c:
I have changed OnImportCompleted to OnImporting.

Issue #6142, opened 2015-12-07
Layout Editor js race condition


connorsmallman:
Seems to be an error on the Contents Edit page `Uncaught Error: No element with type "Canvas" was found.`

This is casued when `window.layoutEditor = new LayoutEditor.Editor(editorConfig, editorCanvasData);` in the LayoutEditor view is called before `registerFactory("Canvas", function (value) { return LayoutEditor.Canvas.from(value); });` in the Models.js.

https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Layouts/Scripts/Models.js

https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Layouts/Views/EditorTemplates/LayoutEditor.cshtml

Fix:

Change anonymous function to JQuery document ready function `JQuery(function() {});`

https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Layouts/Views/EditorTemplates/LayoutEditor.cshtml#L38

connorsmallman:
Still having this issue with the minified Layouts.Models script.

sfmskywalker:
Could you try running gulp from task explorer and see if that fixes it?

connorsmallman:
@sfmskywalker That fixed it, I'll open a pr shortly

connorsmallman:
@sfmskywalker That fixed it thanks!

PR #6144, opened 2015-12-07
Fix for issue #6142

connorsmallman:
Wait for document ready before calling LayoutEditor constructor

Fix for issue https://github.com/OrchardCMS/Orchard/issues/6142


sfmskywalker:
@connorsmallman  Looks like I was a bit too fast accepting the PR. Would you mind creating another one with the same fix but for the 1.9.x branch instead of dev? Thanks!

PR #6159, opened 2015-12-10
Wait for document ready before calling LayoutEditor constructor

connorsmallman:


TFleury:
That created a new issue on 1.9.x preventing any LayoutPart to be saved.
```javascript
window.layoutEditor = new LayoutEditor.Editor(editorConfig, editorCanvasData);
```
from `LayoutEditor.cshtml` is executed **after**
```javascript
var host = new LayoutDesignerHost($(".layout-designer"));
```
from `LayoutDesignerHost.js`.

So, `window.layoutEditor` is not defined when initializing `LayoutDesignerHost`. This throws an exception at save time in `serializeCanvas` function because `self.editor` is undefined.

PR should be reverted or fixed.

connorsmallman:
@sfmskywalker Could you revert this change?

sfmskywalker:
I just pushed a fix. Please try it out and let me know if you're seeing any other issues regarding this.

TFleury:
@sfmskywalker  It seems to work fine now. Thanks
I saw another issue, but only with Edge and not bound to this PR because it appears before : I can only edit one element's content. After I saved an element, it throws `TypeError: Permission denied` when you try to edit another or same element's content.
I have to save the whole content, so it reloads the editor, in order to edit another element, and so on...
Maybe it's more an Edge's issue than a Layout editor issue. Let me know if you want me to open an issue for that.

PR #6164, opened 2015-12-11
To prevent some conflicts if #5966 is committed

jtkech:
@sebastienros 

As you requested, for remembering. This one need to be first applied before giving a try to this experimental PR #5966.

Thanks


jtkech:
Because #5966 is now synchronized, this one is no longer useful

PR #6169, opened 2015-12-12
Update Deploy to Azure link

KexyBiscuit:
The original one is for MS employee and is too old.


sebastienros:
This change should target __master__. Thanks.

PR #6172, opened 2015-12-14
Fixes #6168: Generating TechnicalName for ContenTypes, ContentFields …

### LINKED ISSUE #6168: Generating TechnicalName for ContenTypes, ContentFields creates disallowed names
When creating adding a Field to a ContentPart/ContentType with a DisplayName that already exists, a TechnicalName is generated, which contains a '-'. If we try to save, then a validation error message is display: "The technical name contains invalid characters."

So we have some disambiguity here. This might also cause problems with ContentType names.

Should "-" be allowed in TechnicalNames?

SzymonSel:
…creates disallowed names

Replace hyphen with underscore.
I left the separator - it's just a simpler fix


sebastienros:
Can you actually remove the separator completely? Like 'Blog2'

SzymonSel:
Sure, will do.

-----Original Message-----

SzymonSel:
Done

sebastienros:
Sorry, no Regex, please do a for loop to extract the last digits. Thanks.

SzymonSel:
Sure, will do. But why not Regex?

bleroy:
Too heavy for such as simple check.

SzymonSel:
Ok, thanks. Made the change as desired.

PR #6173, opened 2015-12-14
[Fixes #6154] for Redis Message Bus Shell Restarts on Import

### LINKED ISSUE #6154: Import and Recipe executions cause an error
My guess is that this is related to our implementation of Orchard specifically, however I'm not sure where to start looking.  I'm also guessing that both of the errors below are related (import or recipe execution).

I'm not sure if it's related, but we are overriding the admin theme - however, disabling this override doesn't fix the errors.  Also, I noticed that `ContentPartLayoutSettings.Placable` is spelled "Placable", not "Placeable"...however, changing this to the latter doesn't solve the issue.

### Import Error
When I try to import the following export, I get an error (either the first or the second), __however, the content type is actually created__.
```xml
<!-- Exported from Orchard -->
<Orchard>
<Recipe>
<Name>Generated by Orchard.ImportExport</Name>
<Author>admin</Author>
<ExportUtc>2015-12-08T22:56:28.1235038Z</ExportUtc>
</Recipe>
<Metadata>
<Types>
<Test ContentTypeSettings.Creatable="True" ContentTypeSettings.Draftable="True" ContentTypeSettings.Listable="True" ContentTypeSettings.Securable="True" DisplayName="Test">
<CommonPart/>
<TitlePart/>
</Test>
</Types>
<Parts>
<CommonPart ContentPartSettings.Attachable="True" ContentPartSettings.Description="Provides common information about a content item, such as Owner, Date Created, Date Published and Date Modified." ContentPartLayoutSettings.Placable="True"/>
<TitlePart ContentPartSettings.Attachable="True" ContentPartSettings.Description="Provides a Title for your content item." ContentPartLayoutSettings.Placable="True"/>
</Parts>
</Metadata>
</Orchard>
```
Error 1:
```
Object reference not set to an instance of an object.
Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.NullReferenceException: Object reference not set to an instance of an object.

Source Error: 

Line 33: 
Line 34:         public IShape Create(string shapeType, INamedEnumerable<object> parameters, Func<dynamic> createShape) {
Line 35:             var defaultShapeTable = _shapeTableLocator.Value.Lookup(null);
Line 36:             ShapeDescriptor shapeDescriptor;
Line 37:             defaultShapeTable.Descriptors.TryGetValue(shapeType, out shapeDescriptor);
```

Error 2:
```
Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed.

Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed.

Source Error: 


Line 50:             ShapeDescriptor descriptor) {
Line 51:             return context => {
Line 52:                 var serviceInstance = _componentContext.ResolveComponent(attributeOccurrence.Registration, Enumerable.Empty<Parameter>());
Line 53: 
Line 54:                 // oversimplification for the sake of evolving
```

### Recipe Execution Error
When I try to execute a recipe, I get a different error, __but again, the recipe executes successfully__

Error executing Audit Trail 1.9.2 recipe:
```
Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed.

Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed.

Source Error: 


Line 50:             ShapeDescriptor descriptor) {
Line 51:             return context => {
Line 52:                 var serviceInstance = _componentContext.ResolveComponent(attributeOccurrence.Registration, Enumerable.Empty<Parameter>());
Line 53: 
Line 54:                 // oversimplification for the sake of evolving
```

dcinzona:
This fixes an issue where when using Redis Message Bus and importing a
recipe, multiple shell restarts are triggered on the root host.  This
results in an error being displayed after the import completes.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MessageBus/Services/DistributedShellStarter.cs
+++ b/src/Orchard.Web/Modules/Orchard.MessageBus/Services/DistributedShellStarter.cs
@@ -8,7 +8,7 @@ using Orchard.Environment.Extensions;
 namespace Orchard.MessageBus.Services {
 
     public interface IDistributedShellStarter : ISingletonDependency {
-
+        bool IsStarting();
     }
 
     [OrchardFeature("Orchard.MessageBus.DistributedShellRestart")]
@@ -17,9 +17,13 @@ namespace Orchard.MessageBus.Services {
 
         private readonly IMessageBus _messageBus;
 
-        private static readonly object locker = new object();
+        private object locker = new object();
 
-        public static bool IsStarting { get; private set; }
+        public bool _isStarting { get; private set; }
+        
+        public bool IsStarting() {
+            return _isStarting;
+        }
         
         public readonly static string Channel = "ShellChanged";
 
@@ -44,7 +48,7 @@ namespace Orchard.MessageBus.Services {
                         lock (locker) { 
                             // Set a flag indicating that we are in the process of activating the shell. 
                             // This is used to prevent recursive message bus calls.
-                            IsStarting = true;
+                            _isStarting = true;
                             shellSettingsManagerEventHandler.Saved(shellSettings);
 
                             if(orchardHost != null) {
@@ -52,7 +56,7 @@ namespace Orchard.MessageBus.Services {
                                 startUpdatedShellsMethod.Invoke(orchardHost, null);
                             }
 
-                            IsStarting = false;
+                            _isStarting = false;
                         }
                     }
                 }
--- a/src/Orchard.Web/Modules/Orchard.MessageBus/Services/DistributedShellStarter.cs
+++ b/src/Orchard.Web/Modules/Orchard.MessageBus/Services/DistributedShellStarter.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System.Collections.Concurrent;
+using System.Linq;
 using Orchard.Caching;
 using Orchard.Environment;
 using Orchard.Environment.Configuration;
@@ -16,6 +17,10 @@ namespace Orchard.MessageBus.Services {
 
         private readonly IMessageBus _messageBus;
 
+        private static readonly object locker = new object();
+
+        public static bool IsStarting { get; private set; }
+        
         public readonly static string Channel = "ShellChanged";
 
         public DistributedShellStarter(IMessageBus messageBus, IWorkContextAccessor workContextAccessor) {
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.MessageBus/Services/DistributedShellStarter.cs
```diff
@@ -16,6 +17,10 @@ public class DistributedShellStarter : IDistributedShellStarter, IOrchardShellEv
 
         private readonly IMessageBus _messageBus;
 
+        private static readonly object locker = new object();
+
+        public static bool IsStarting { get; private set; }
```
sebastienros: This is wrong, static values are the same across tenants. It should be some `IDependency` baked state.

src/Orchard.Web/Modules/Orchard.MessageBus/Services/DistributedShellStarter.cs
```diff
@@ -33,12 +38,21 @@ public class DistributedShellStarter : IDistributedShellStarter, IOrchardShellEv
...
                         var shellSettingsManagerEventHandler = scope.Resolve<IShellSettingsManagerEventHandler>();
-                        shellSettingsManagerEventHandler.Saved(shellSettings);
-
                         var orchardHost = scope.Resolve<IOrchardHost>() as DefaultOrchardHost;
-                        if(orchardHost != null) {
-                            var startUpdatedShellsMethod = typeof(DefaultOrchardHost).GetMethod("StartUpdatedShells", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
-                            startUpdatedShellsMethod.Invoke(orchardHost, null);
+
+                        // We only want a single thread setting the "IsStarting" flag
+                        lock (locker) {
```
sebastienros: Do you really mean `lock` where the next thread will also execute this code, or `Monitor.TryEnter` where the first one executes and not the others?

sebastienros:
There is also a `IDistributedLock` in dev that might be a solution here instead of locks if it makes sense. Will take more time to understand the issue and the fix later. Thanks.

sebastienros:
- Check if when a tenant restarts, it invokes 
IShellDescriptorManagerEventHandler.Changed
IShellSettingsManagerEventHandler.Saved

- Check if the subscriptions are lost after a tenant restarts

dcinzona:
I applied this change to my local copy of 1.10 and it still appears to work as expected.

PR #6175, opened 2015-12-14
All /lib dll files are now correctly taken into account

Jwheuvel:
Changed the .gitignore so all /lib dependencies are included


sebastienros:
Could you target 1.9.x instead? It would still go in dev.

Jwheuvel:
Sure.. Just did..

PR #6187, opened 2015-12-16
Fixed bad assumption about HttpContext

### LINKED ISSUE #6181: ImportExport can no longer be run at command line
```
/t:testtenant import file /Filename:C:\temp\recipes\testimport.xml
```

will fail due to `Orchard.ImportExport.Providers.ImportActions.ExecuteRecipeAction` L156.

```
_orchardServices.WorkContext.HttpContext.Server.ScriptTimeout = RecipeExecutionTimeout;
```
assumes an HttpContext, but none is provided when executing from command line (apparently a context used to be provided - I think this is recent refactoring on the develop branch)

PaulDevenney:
Fixes #6181 

When running Import/Export at the console it would fail due to a null HttpContext.

* Added a guard clause as  HttpContext is not available when running at console


sebastienros:
We just merged a PR that should solve the null value. But now I am concerned about the ScriptTimeout itself, can you check with the latest code if this value is correct? As in a command/background task it will use a fake HttpContext.

PR #6191, opened 2015-12-16
Fixes #6189 : TextAreaElement throwing on null ProcessedValue

### LINKED ISSUE #6189: DynamicForms : TextAreaElement throwing on null ProcessedValue
1.9.x branch

Adding a TextArea in a DynamicForm throws `RuntimeBinderException` if no value is specified.
Issue appears after e29fb90298659a67b58711a64f0c23746817cb06

TFleury:


sebastienros:
Does it only applies to TextArea?

TFleury:
Other elements affected by the PR (Checkbox, Email, URL, Hidden and Text) are not concerned by this issue as `ProcessedValue` is applied to an attribute. It doesn't matter if it's null.
TextArea element is the only one using `tagBuilder.SetInnerText(string innerText)`

PR #6195, opened 2015-12-17
Trying to revert a famous PR, sorry

jtkech:
Sorry for trying to revert this PR related to the awesome presentation of Chris Payne at Harvest where a `QueryHints()` with `ExpandParts()` was added to access the layer parts...

But after the query, as mentioned in a meeting by Sébastien, a previous version was used with access to the LayerPartRecord `activeLayer.Record.LayerRule`. But now, because the LayerRule is in the infoset, we only use `activeLayer.LayerRule`.

So, this PR requested by @sebastienros 

Note1: the substance of the Harvest presentation remains fully relevant.
Note2: not tested but validated by Sébastien. If wrong it's not my PR.

Best


sebastienros:
@paynecrl97

PR #6207, opened 2015-12-20
Close #2027 : Add role rule  in Orchard.Widgets module

### LINKED ISSUE #2027: Role based rule in Orchard.Widgets module
__finlan__ created:
https://orchard.codeplex.com/workitem/18200

I implemented this on my badly named fork as a distinct changeset.
I copied (with slight modifications if I recall correctly) from the Role based rule module (and the author is credited in a comment in the class I added.

Here is the changeset:
http://orchard.codeplex.com/SourceControl/network/Forks/finlan/WidgetRoleRuleProvider/changeset/changes/12c98b6a13dc

EXPLANATION:
I added a RoleRuleProvider to the widgets module. For me being able to add a rule that is Role based is so useful that I "REALLY" believe it should be part of the core code and not a module. I took the code from a related module. I think I changed it slightly but I am not sure. There is a comment in the code of who the author is. (I am praying that you include this because at the moment I am manually adding this everytime I get a new version of Orchard and it is the only class that I manually add)

TFleury:


TFleury:
See #6210 for dev branch.

PR #6210, opened 2015-12-21
Move RoleRuleProvider to RoleCondition

TFleury:
See #6207 for 1.9.x branch


src/Orchard.Web/Modules/Orchard.Roles/Conditions/RoleCondition.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Orchard.Conditions.Services;
+using Orchard.ContentManagement;
+using Orchard.Roles.Models;
+using Orchard.Security;
+using System;
+using System.Linq;
+
+namespace Orchard.Roles.Conditions {
+    public class RoleRuleProvider : IConditionProvider {
```
sebastienros: Because `IConditionProvider` is an event handler you should not take a reference on this, just copy the interface in your project, and remove the reference. The implementation will be called because their interface names match. MAGIC !

src/Orchard.Web/Modules/Orchard.Roles/Orchard.Roles.csproj
```diff
@@ -132,6 +133,10 @@
       <Name>Orchard.Core</Name>
       <Private>false</Private>
     </ProjectReference>
+    <ProjectReference Include="..\Orchard.Conditions\Orchard.Conditions.csproj">
```
sebastienros: Remove

sebastienros:
This PR seems wrong, please redo

PR #6213, opened 2015-12-21
Widget: add "Deactivated" flag to prevent widget from being rendered

dpomt:
New feature: Deactivated flags for widgets to prevent widgets from being rendered


dpomt:
there are other possibilities to achieve this - see https://github.com/OrchardCMS/Orchard/issues/6226#issuecomment-166991354

PR #6214, opened 2015-12-21
Fix media library (manage own media) issue with german umlauts in usernames

dpomt:
Hi,

managing media libary has issues e. g. with german umlauts in file names.
Startet getting 
A potentially dangerous Request.QueryString value was detected from the client...
errors. Wrong url encoding was also an issue.
With my suggested fix here meda library management work with german umlauts.


src/Orchard.Web/Modules/Orchard.MediaLibrary/Orchard.MediaLibrary.csproj
```diff
@@ -38,14 +38,15 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <DebugType>pdbonly</DebugType>
-    <Optimize>true</Optimize>
+    <DebugType>full</DebugType>
```
sebastienros: ???

dpomt: sorry, this should not have been part of the fix

sebastienros:
I understand the issue but this PR looks like a big hack. Let's see if it can't be fixed by simply url-encoding the folder path. The decoding should be done automatically when the query string is read from ASP.NET.

dpomt:
You  might be right concerning the "big hack". If somebody finds a better fix I would be happy with that. This was the only solution I found (for a site that was already in production with users facing the issue) and thought it would be good to share.

sebastienros:
Please open a bug with some repro steps, this fix is incorrect, people will be able to comment and suggest a solution, or repro it.

PR #6223, opened 2015-12-23
fix CommonMetadataOwner issue: german umlauts are displayed encoded

dpomt:
simple fix to display german umlauts user name correctly (e. g. ü instead of &#252;)

### DIFF
--- a/src/Orchard.Web/Core/Common/Views/CommonMetadataOwner.cshtml
+++ b/src/Orchard.Web/Core/Common/Views/CommonMetadataOwner.cshtml
@@ -11,4 +11,4 @@
     // owner isn't really who last modified this, is it?
     IUser owner = commonPart.As<CommonPart>() == null ? null : commonPart.As<CommonPart>().Owner;
 }
-@T("By {0}", owner == null ? T("unknown").ToString() : Convert.ToString(Html.ItemDisplayText(owner)))
+@T("By {0}", owner == null ? T("unknown").ToString() : Convert.ToString(Html.ItemDisplayText(owner, false)))


src/Orchard.Web/Core/Common/Views/CommonMetadataOwner.cshtml
```diff
@@ -11,4 +11,4 @@
     // owner isn't really who last modified this, is it?
     IUser owner = commonPart.As<CommonPart>() == null ? null : commonPart.As<CommonPart>().Owner;
 }
-@T("By {0}", owner == null ? T("unknown").ToString() : Convert.ToString(Html.ItemDisplayText(owner)))
\ No newline at end of file
+@T("By {0}", owner == null ? T("unknown").ToString() : Convert.ToString(Html.ItemDisplayText(owner, false)))
```
sebastienros: The solution is to remove the `ToString` calls from this line. No to remove encoding.

```
@T("By {0}", owner == null ? T("unknown") : Html.ItemDisplayText(owner, false)) 

```

dpomt: Your proposol is not working here. I am getting:

CS0173: Type of conditional expression cannot be determined because there is no implicit conversion between 'Orchard.Localization.LocalizedString' and'System.Web.Mvc.MvcHtmlString'

sebastienros:
This file doesn't exist anymore in 1.10, and the issue is fixed for me. If you want to fix it locally you can just render the user name property.

PR #6228, opened 2015-12-23
ContentPickerField enhancement: (for not admins) show only the current users content items

dpomt:


sebastienros:
I think instead of having a setting, let users decide what they want to see by adding a drop down in the Media library UI to filter items only owned by the current user. Like it's done in the Contents page.

dpomt:
In my special case the user should not be able to select Content of another user. No Chance to get this in?

sebastienros:
Have you tried to use the Manage Own Media permission instead, and remove Manage Media for these specific users? It might be new in 1.9.x though, I am using it.

dpomt:
Since I am already using 1.9.x and altough the Manage Own Media Feature so I quickly commented my code out and re-tested it ==>  Still Content items of other users are displayed. 

In my scenario here I am using the ContentPicker select a "standard" Content item, not a media Content item.

sebastienros:
Ah sorry, I read the issue too fast.

dpomt:
You still think that this is not functionality that should be provided?

sebastienros:
I need to think about it. We'll talk about it during our next meeting probably, feel free to join.

dpomt:
This suggestion was discussed in 19th january meeting (see https://www.youtube.com/watch?v=sRmF_fAktds&index=2&list=PLuskKJW0FhJfOAN3dL0Y0KBMdG1pKESVn at 01:01:25
If I understood it correctly the suggestion was accepted? If yes, how will this pull request gets merged (since it is already closed)?

PR #6233, opened 2015-12-26
Adds IDecorator functionality

paynecrl97:
Adds support for decorator functionality.

A developer can decorate any interface that is `IDependency` simply by implementing `IDecorator<>`. For example, to decorate `IContentManager`, you'd do the following:

```
public class MyContentManagerDecorator : IContentManager, IDecorator<IContentManager> {

    public MyContentManagerDecorator(IContentManager decoratedService) {
        // decoratedService will be an implementation of IContentManager other than this one
    }

    ...

}
```

The magic all happens in `ShellContainerFactory`. Changes to the logic in this class:
- `RegisterType` is deferred until all services have been iterated to determine if they have been decorated
- Concrete implementations are then registered as before, but with the addition of being named, and registered `AsSelf()`
- Decorator implementations are not registered until all concrete implementations have been registered (this is to guarantee we know the name given to the concrete implementation at decorator)
- Once all the concrete implementations are registered, the decorators are then all registered with `WithParamater()` to tell Autofac the name of the service that is being decorated
- A list of registered implementations is maintained. This allows us to register multiple decorators on top of multiple concrete implementations

In short, decorators are registered similar to this:

```
// RegisterType as itself
// Name it uniquely as itself
// If its a decorator, do WithParameter and resolve named
// If it is not decorated, as register as the interface


builder.RegisterType<TestDecoratorImpl1>().Named<TestDecoratorImpl1>("Impl1").AsSelf();
builder.RegisterType<TestDecorator1>().Named<TestDecorator1>("Dec1").As<TestDecorator1>().WithParameter((p, c) => p.ParameterType == typeof(ITestDecorator), (p, c) => c.ResolveNamed<TestDecoratorImpl1>("Impl1"));
builder.RegisterType<TestDecorator2>().Named<TestDecorator2>("Dec2").As<TestDecorator2>().WithParameter((p, c) => p.ParameterType == typeof(ITestDecorator), (p, c) => c.ResolveNamed<TestDecorator1>("Dec1"));
builder.RegisterType<TestDecorator3>().Named<TestDecorator3>("Dec3").As<TestDecorator3>().WithParameter((p, c) => p.ParameterType == typeof(ITestDecorator), (p, c) => c.ResolveNamed<TestDecorator2>("Dec2")).As<ITestDecorator>();
```

This functionality is a pre-cursor to the `Orchard.Glimpse` module demoed at Orchard Harvest.

### DIFF
--- a/src/Orchard/Environment/ShellBuilders/ShellContainerFactory.cs
+++ b/src/Orchard/Environment/ShellBuilders/ShellContainerFactory.cs
@@ -151,8 +151,7 @@ namespace Orchard.Environment.ShellBuilders {
                         }
 
                         if (itemToBeRegistered.IsEventHandler) {
-                            // var interfaces = itemToBeRegistered.Item.Type.GetInterfaces();
-                            var interfaces = itemToBeRegistered.InterfaceTypes;
+                            var interfaces = itemToBeRegistered.Item.Type.GetInterfaces();
                             foreach (var interfaceType in interfaces) {
 
                                 // register named instance for each interface, for efficient filtering inside event bus
--- a/src/Orchard/Environment/ShellBuilders/ShellContainerFactory.cs
+++ b/src/Orchard/Environment/ShellBuilders/ShellContainerFactory.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Web.Hosting;
@@ -14,6 +16,8 @@ using Orchard.Environment.AutofacUtil.DynamicProxy2;
 using Orchard.Environment.Configuration;
 using Orchard.Environment.ShellBuilders.Models;
 using Orchard.Events;
+using Orchard.Localization;
+using Orchard.Logging;
 
 namespace Orchard.Environment.ShellBuilders {
 
@@ -24,11 +28,18 @@ namespace Orchard.Environment.ShellBuilders {
     public class ShellContainerFactory : IShellContainerFactory {
         private readonly ILifetimeScope _lifetimeScope;
         private readonly IShellContainerRegistrations _shellContainerRegistrations;
+        private ConcurrentDictionary<Type, ConcurrentBag<NamedRegistration>> _concreteRegistrationNames;
 
         public ShellContainerFactory(ILifetimeScope lifetimeScope, IShellContainerRegistrations shellContainerRegistrations) {
             _lifetimeScope = lifetimeScope;
             _shellContainerRegistrations = shellContainerRegistrations;
+
+            Logger = NullLogger.Instance;
+            T = NullLocalizer.Instance;
         }
+
+        public ILogger Logger { get; set; }
+        public Localizer T { get; set; }
 
         public ILifetimeScope CreateContainer(ShellSettings settings, ShellBlueprint blueprint) {
             var intermediateScope = _lifetimeScope.BeginLifetimeScope(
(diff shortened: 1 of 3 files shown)


src/Orchard/Environment/ShellBuilders/ShellContainerFactory.cs
```diff
@@ -56,50 +67,150 @@ public class ShellContainerFactory : IShellContainerFactory {
...
                         }
                     }
 
+                    foreach (var decorator in decorators) {
+                        // We need to ensure that there is an implementation of this service that can be decorated
+                        if (!_concreteRegistrationNames.ContainsKey(decorator.InterfaceType) || _concreteRegistrationNames[decorator.InterfaceType] == null || !_concreteRegistrationNames[decorator.InterfaceType].Any()) {
+                            var exception = new OrchardFatalException(T("The only registered implementations of `{0}` are decorators. In order to avoid circular dependenices, there must be at least one implementation that is not marked with the `OrchardDecorator` attribute.", decorator.InterfaceType.FullName));
+                            Logger.Fatal(exception, "Could not complete dependency registration as a circular dependency chain has been found.");
+
+                            throw exception;
```
paynecrl97: Is throwing a fatal exception correct in this instance? Or should we just log and move on?

Ermesx:
Decorations should be done only over implementations and Orchard have that functionality by SuppressDependecyAttribute. If you want intercept interfaces it should be done by proxies like DynamicProxy.

sfmskywalker:
`SuppressDependencyAttribute` is not suitable to decorate other services, as it would require you to inherit from the class that you're decorating. It also means you won't be able to use "decorators" independently from one another. With Chris' work, we have complete control over decorating implementations without such limitations. I haven't worked with DynamicProxy myself, but I'm curious as to why you think it's better to use that rather than what's done as part of this PR?

Ermesx:
@sfmskywalker *It would require you to inherit from the class that you're decorating*, I'm not agree.
I can do something like:

```c#
[SuppressDependecy("Orchard.ContentManagement.DefaultContentManager")]
public class TotalReimplementedManager : IContentManager
{
    // new code
}
```

And also
```c#
[SuppressDependecy("Orchard.ContentManagement.DefaultContentManager")]
public class OverridedManager : DefaultContentManager
{
    // override methods
}
```

and in different way if class does not have virtual methods
```c#
[SuppressDependecy("Orchard.ContentManagement.DefaultContentManager")]
public class DecoratedManager : IContentManager
{
    private IContentManager _oldManager;
    public DecoratedManager (...dependencies...)
    {
        _oldManager = new DefaultContentManager(...dependencies...);

    // New code with using internal old implementation
}
```

Maybe `SuppressDependencyAttribute` has incorrect name ?

paynecrl97:
The power of `IDecorator` is that it does not care what it is decorating, nor does it need to know any of the logic contained within the service it is decorating (which can't be said about any of the three examples above).

`IDecorator` is meant to compliment `OrchardSuppressDependency`, and they can both be used to achieve different end results.

Another thing that decorators allow you to do is to stack implementations independently on top of each other. For example, we could decorate `IContentManager` with an implementation that logs, another that provides a layer of caching, and a third that provides performance statistics. And because each decorator is independent of the implementation it is decorating, we can feature switch any combination of those three decorators on or off as we please.

I realise that **all of the points I've raised above are also true of Interceptors**, (and I'll admit that I have no 'real world' experience with interceptors, so my logic may be flawed here), but the reason why I chose the decorator pattern over the interceptor pattern is because decorators provide, in my opinion, a much more robust solution to type safety and changes in the interface.

An interceptor would look something along the lines of this (from http://docs.autofac.org/en/latest/advanced/interceptors.html):

```
  public void Intercept(IInvocation invocation)
  {
    _output.Write("Calling method {0} with parameters {1}... ",
      invocation.Method.Name,
      string.Join(", ", invocation.Arguments.Select(a => (a ?? "").ToString()).ToArray()));

    invocation.Proceed();

    _output.WriteLine("Done: result was {0}.", invocation.ReturnValue);
  }
```

Now imagine if your interceptor needed to use the instances of the arguments passed. You'd have to find the correct argument (eg by index/ name/ type etc...), and then cast to the correct type in order to access the properties. I think this is quite brittle, and can lead to runtime issues. And then if the interface or method signatures change, you'd struggle to realise the issue immediately.

With decorators, if the interface or method signatures changed, you'd get a build time error; which is probably better than a potential runtime error that you don't spot until it's too late.

I think a change to an interface should be considered a breaking change, and so I don't feel like guarding against them should be that persuasive when it comes to design decisions.

I personally see interceptors as a better solution when you need a more generic solution (for example to add logging to the entry and exit of all methods on a series of services etc...), and decorators are a better solution when you need a wrap a specific method with typed access to the arguments and return value. I don't think it's as clear cut as saying "always use x, and never use y".

That said, this PR adds decorator functionality to Orchard. If it is decided that the ability to natively register interceptors would be a good asset to the out of the box functionality of Orchard, then there is no reason why Orchard could not support both decorators and interceptors.

Ermesx:
So after your explanation I agree with you, because now I understand your point of view and reasons.
You are right that better to have compilation errors than runtime errors and I'm happy that you agree with me that "decorators" do almost the same like interceptors. 

To be honest now our discussion comes to decide about correct name for that solution. 
Because when I saw "Decorator" I have in my mind simple definition:
When I want decorate something I know what exactly it is and how it works but I cannot inherit and override or I want to have control over dynamic extending process for specific case than wide cross application operation.

That's why I was afraid of this solution because your decorators better describe interceptors, but now you are right because when I create interceptor I almost don't care about anything and I try inject some additional operations independently from intercepted interface/implementations event don't care about method arguments but in this case we want to take care to it. It helps to retrieve informations about executing method, parameters make compilation error when something is wrong. It gives more data.

But It is something between decorators and interceptors. I thing there should be better name for it with more context what can be done witch that feature.
Maybe proxy is good name ? I want to avoid other misunderstanding because name of this interface is greedy it invites to use it with wrong intentions. 

I propose to rename it with name which better description that it exactly does. Describe cases when It should be used and when use `SuppressDependencyAttribute` because it is complementary.

Best regards.

paynecrl97:
An update on this: we have pulled this commit into our internal Orchard fork and are successfully using it.

Do you require anything else from me in order to get this merged?

sebastienros:
Patience ...

sebastienros:
And I don't forget, I even looked into adding this too in O2.

sfmskywalker:
Cool. Any timeline you have in mind? Are we delaying this in order for the implementation to be inspected and grasped in full before we can take it?

PR #6243, opened 2015-12-28
Adds Glimpse

paynecrl97:
**Note:** This branch includes the changes in #6233; that PR should be reviewed before this one.

Adds Glimpse support to Orchard out of the box.

To enable, pull this branch, build, run, navigate to `~/glimpse.axd` and click to enable. Each Orchard specific inspector (including SQL) is enabled from the Features page in the admin area.

Current inspectors are:
- Authorizer
- Cache Service
- Content Manager
- Layers
- Parts
- SQL
- Shapes
- Widgets

Most inspectors make use of `IDecorator<>`, those that don't can only be changed to decorators if the underlying service that is being inspected is refactored.

Includes an extension point to allow for 3rd party modules to consume Glimpse data. Example usage:

```
public class WidgetMessageInterceptor : GlimpseMessageInterceptor<WidgetMessage> {
    public override void ProcessMessage(WidgetMessage message) {
        _myThirdPartyService.RecordData(message);
    }
}
```

This is only an MVP for this feature, but I'm keen to get it into the dev branch to allow people to try it out, test it, give feedback, and to help contribute.

Possible further improvements:
- Investigate why multiple SQL connections sometimes appear to be reported as opened
- Fields Tab
- Layout Tab
- Token Tab
- Implement [IDocumentation](https://github.com/Glimpse/Glimpse/blob/master/source/Glimpse.Core/Extensibility/IDocumentation.cs) support for each tab
- Allow Glimpse's DI framework to work with Orchard's in order to allow Orchard services to be consumed from Glimpse services etc...


dcinzona:
:clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap:

Ermesx:
Great module and I'm agree in 100% with @dcinzona :)

But I'm afraid of IDecorator<> because when we want to "decorate" something we want do it over **specific implementation** and there is the correct way how to do it in Orchard. It can be done by SuppressDependencyAttribute, use old implementation and add something new or even totally reimplement.

In your case you want to intercept whole **interfaces** with collecting audit/statistics/profiling data. It should be made by AOP mechanism like proxy over interfaces. I think it should be done in different way. Decoration and proxy are different approaches.

In summary you do great work but I think you choose incorrect tool to intercept operations. Maybe better will be use DynamicProxy?

Best regards.

paynecrl97:
For the benefit of people stumbling upon this PR, the use of decorators has been discussed at https://github.com/OrchardCMS/Orchard/pull/6233, which this PR is dependent on.

I originally opened this PR over 6 months ago, so lots has changed since then. I'm especially interested in hearing people's thoughts on the Glimpse integration implementation, and also any feedback on the tabs/ data within each tab.

Once the decorator PR is merged, I'll start to put more effort into tidying this PR up to the point where I'd be happy to put it forwarded as a candidate for merging.

paynecrl97:
Closing this PR while I work on fixing the conflicts etc...

PR #6253, opened 2015-12-31
Fix media-library-picker.js

TFleury:
Stops `media-library-picker.js` starting media ids list with a comma.
There is no issue with `MediaLibraryPickerField` as it uses `StringSplitOptions.RemoveEmptyEntries`, but it's an issue if you use `MediaLibraryPicker` shape and expect a single id.

The fix slightly reduces DOM access, it should not be bad.


sfmskywalker:
I like it, I always hated that the string started with a comma.

Issue #5789, opened 2015-09-11
Orchard does not support cookie domains per tenant.


PaulDevenney:
**Scenario:**
You wish your site to issue auth cookies on the domain, rather than sub domain, for example; instead of issuing the .ASPXAUTH token to `www.testdomain.com` you wish it to be issued to `.testdomain.com` so that your user is signed into other sites on the same domain at the same time

Currently there is only one way to alter the cookie domain, that is to alter the forms section of web.config as below:

    <authentication mode="Forms">
        <forms loginUrl="~/Users/Account/AccessDenied" timeout="2880" domain=".testdomain.com" />
    </authentication>

However, this only works in a single site application, and would not work in a multi-tenant situation (unless you were fortunate enough that all your tenants were sub domains of the same root domain)

There are two routes that seem viable to me to resolve this

1. Cookie domain is a parameter available at tenant setup (along with Url Host and Url Prefix). This would allow you  to setup a site as `Urlhost : www.testdomain.com` with `CookieDomain: .testcomain.com`

2. Expose a public property on `FormsAuthenticationService` to allow extending code to set a value for this property

DanielStolt:
1 would seem very practical and sensible, but does bake in the assumption that cookies are used for authentication, which could be substituted for a different implementation. (On the other hand, Web.config assumes ASP.NET forms authentication is used for the whole site, so...)

2 seems like a partial solution. It adds an extensibility point, but you still have to add some other mechanism to determine the cookie domain.

I would opt for 1, as it would be a turn-key solution. Perhaps we could use a more abstract term than `CookieDomain` such as `AuthenticationScope` or something, so that it still makes sense even if a completely different mechanism is used to scope the authentication session?

sfmskywalker:
Yeah exactly. For 2 one could make the cookie domain a public property on the `FormsAuthenticationService` and configure its value via HostComponents.config (which can be configured per tenant).

DanielStolt:
@sfmskywalker Yeah that's true - but seems a little disjointed maybe?

sfmskywalker:
In what way? Isn't HostComponents.config intended to configure Autofac-registered services with properties? At the same time, HostComponents should probably be used for lower-level, core framework services, that aren't typically replaced with other implementations.

Alternatively, a site-setting could be introduced for the FormsAuthenticationService. Also tenant-specific.

DanielStolt:
[...]

PR #6281, opened 2016-01-15
Get Auth status from WorkContext in OutputCacheFilter

paynecrl97:
Gets auth state from WorkContext as opposed to HttpContext.

Solves a potential issue for people who have swapped out the implementation of `IAuthenticationService.GetAuthenticatedUser()` where the output cache could think that the user was logged in when in fact they were not. For example: `GetAuthenticatedUser()` has been overridden to include additional auth checks to enforce business logic, these additional checks will not be performed without this change.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -386,7 +386,7 @@ namespace Orchard.OutputCache.Filters {
 
             // Vary by authentication state if configured.
             if (CacheSettings.VaryByAuthenticationState) {
-                result["auth"] = result["auth"] = _workContext.CurrentUser != null;
+                result["auth"] = _workContext.CurrentUser != null;
             }
 
             return result;
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -386,7 +386,7 @@ namespace Orchard.OutputCache.Filters {
 
             // Vary by authentication state if configured.
             if (CacheSettings.VaryByAuthenticationState) {
-                result["auth"] = filterContext.HttpContext.User.Identity.IsAuthenticated.ToString().ToLowerInvariant();
+                result["auth"] = result["auth"] = _workContext.CurrentUser != null;
             }
 
             return result;


src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -386,7 +386,7 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
 
             // Vary by authentication state if configured.
             if (CacheSettings.VaryByAuthenticationState) {
-                result["auth"] = filterContext.HttpContext.User.Identity.IsAuthenticated.ToString().ToLowerInvariant();
+                result["auth"] = result["auth"] = _workContext.CurrentUser != null;
```
Ermesx: Why is `result["auth"] ` two times ?

jchenga: there is only one result["auth"] assignment. The line in pink is replaced by the line in green.

Ermesx: Funny :full_moon_with_face: 
zrzut ekranu 2016-01-15 o 22 12 33

jchenga: ah. my bad. :(

paynecrl97: Not sure how that's happened. Good catch though.

DanielStolt:
@paynecrl97 In the previous code `VaryByAuthenticationState` would vary by whether you were authenticated at all or not, regardless of whether that authentication carried through to a valid Orchard user or not. Just to verify, does that still hold true in this proposed change where you instead check `_workContext.CurrentUser`?

paynecrl97:
No, that is now no longer the case. Did it behave like that previously by design?

I was under the impression that we'd only want to vary when authenticated to a valid Orchard user, and so the way it previously behaved was a bug.

sebastienros:
I don't remember it to be a feature (being authenticated and not linked to a user account). Even APIs I assume would in the end be associated to a user account.

DanielStolt:
We have actually built several sites where being authenticated is not the same thing as being an Orchard user. This can make perfect sense when your Orchard site is acting as a front-end to some larger system.

For example, you have an API behind it, and you have several client-side apps hosted by Orchard which call this API from the user's browser. The users authenticates using some method of authentication that is *not* Orchard (such as SSO or smart card or whatever) and can then get access to parts of the sites that are for authenticated users only, such as these client-side apps. And those can now use the authentication session to call back to the backend API.

In this case, we want the Orchard site to render different UI depending on whether you are authenticated or not, even though your authentication does not have anything to do with an Orchard user record, and will not grant you access to the admin UI, for example.

I don't know if this has been a "feature" (I suspect nobody knows, it's just the way it's always worked) but I do know that there are definitely sites out there that depend on this behavior, and those would be broken by this change.

I believe that in general, we should strive to keep the concepts of "authenticated" and "logged in as an Orchard user record" completely separate. There are many scenarios where this is both legitimate and very useful. In fact I think @PaulDevenneyBede also recently opened an issue to this effect, for Bede's scenario, no?

Ah, found it. :)

https://github.com/OrchardCMS/Orchard/issues/5789#issuecomment-141085653

sebastienros:
@Daniel so do you mean the change is wrong?

DanielStolt:
@sebastienros Yes, the change is wrong.

PR #6290, opened 2016-01-19
Updated feed URL for Orchard gallery.

### LINKED ISSUE #6289: Orchard Gallery Url must be changed
Because the Orchard gallery has been changed, the new packages feed url should be https://orchardgallery.azurewebsites.net/api/FeedService 

The url is still http://packages.orchardproject.net/FeedService.svc in DefaultPackagingUpdater.cs and has to be changed.

The attribute "xml:base=" must be changed in the gallery feed XML.

This change should also be mentioned in the next release notes to explain how to point to the correct feed url.

armanforghani:
Fixes #6289


sebastienros:
The correct URL is https://gallery.orchardproject.net/api/FeedService, can you fix it? And also squash the two commits?

armanforghani:
I picked the URL from [here](http://gallery.orchardproject.net/). I do it immediately.

armanforghani:
@sebastienros, I fixed the URL as you mentioned above and squashed the two commits.

PR #6291, opened 2016-01-20
fix #6065

### LINKED ISSUE #6065: image size is too big for layout imge element
when i used media picker field to upload image, it gave me 20k to display.
but when i go to layout=>image to upload, it gave me original size of the image (2,7M) which is obviously  too big for website

infofromca:
if img is too big size, make it smaller


sebastienros:
We can't decide what is good for mobile like this. It should at least be in a setting and not hard coded, and then I assume the solution would still be to override the template in the theme, with different rules for each field. Personally I think the default behavior is fine.

PR #6301, opened 2016-01-23
Fixes #5593 : Allow disabling HtmlEncode on DynamicForms

### LINKED ISSUE #5593: Dynamic Forms: In form settings Html Encode option is always checked
On the 1.9.x branch Html Encode is always on. When you go to the form settings and uncheck the option, click save, then open the settings again, the option is checked.

TFleury:
If we want to default encode form values, an alternative to this PR would be a checkbox to opt-out HtmlEncoding.
More, HtmlEncoded form values are not correctly diplayed in Form Submissions view.


sebastienros:
We talked about it during the triage, and these kind of options should not exist, as each element should define the behavior. A ScripElement would not encode, a TextElement, would.

TFleury:
I agree for defining this behavior at element level rather than on the form. It allows par field rules.
However, we must be able to choose to encode or not TextElement's values, because there are many cases in which you don't want an encoded string (most cases for me).

PR #6303, opened 2016-01-25
DefaultJsonDeserializer using JSON.NET: Allowing deserializing into any JSON token type

harmony7:
This PR is a very simple change that would allow the use of `DefaultJsonConverter.Deserialize()` to parse a JSON string that may be of types other than a JSON object.

Since the current code uses JObject.Parse(), it can only handle JSON Objects.
http://www.newtonsoft.com/json/help/html/ParseJsonObject.htm

    // Works
    _jsonConverter.Deserialize("{\"key1\": \"value\", \"key2\": 5}");
    // These don't work
    _jsonConverter.Deserialize("[0, \"foo\", 2]");
    _jsonConverter.Deserialize("100");
    _jsonConverter.Deserialize("null");
    _jsonConverter.Deserialize("\"hello, world\"");

My suggestion uses JToken.Parse(), which handles any kind of JSON token.
http://www.newtonsoft.com/json/help/html/ParseJsonAny.htm

    // Works
    _jsonConverter.Deserialize("{\"key1\": \"value\", \"key2\": 5}");
    // Now Works Too
    _jsonConverter.Deserialize("[0, \"foo\", 2]");
    _jsonConverter.Deserialize("100");
    _jsonConverter.Deserialize("null");
    _jsonConverter.Deserialize("\"hello, world\"");

This does not break existing code because JObject is a subclass of JToken.

### DIFF
--- a/src/Orchard/Services/DefaultJsonConverter.cs
+++ b/src/Orchard/Services/DefaultJsonConverter.cs
@@ -15,7 +15,7 @@ namespace Orchard.Services {
         }
 
         public dynamic Deserialize(string json) {
-            dynamic result = JObject.Parse(json);
+            dynamic result = JToken.Parse(json);
             return result;
         }
 


sebastienros:
This was not then initial intent for this interface. Is this your own code that needs to do that? If it's from your code I'd suggest you depend directly on JSon.NET and use it this way.
We could add a non-breaking method like `DeserializeToken` but is it really necessary to abstract it?

harmony7:
You're right about it being my code that needs this, and I have always done what you suggested (making my own module that depends on JSON.NET).

However, I find myself having to do this really often, since so much everywhere uses JSON, so it might make sense as a utility in the framework... and JsonConverter.Deserialize is _so close_ to what I need.

Additionally, this change actually _is_ non-breaking (unless you actually depend on the behavior of throwing when something invalid is passed in), and I figured it wouldn't hurt, so I suggested it.

If you don't think it's necessary then that's ok, I'll continue using my own modules.

sebastienros:
> I'll continue using my own modules

Actually don't ;) I am not saying to use your own module, or to make your own abstraction. I am just saying don't try to abstract it, it's really not necessary, just use Json.net explicitly. For instance this is something we are also doing in Orchard 2, we are not trying to hide JSON.NET anymore, because it is the feature, and it doesn't need to be mocked in unit tests.

harmony7:
Oh I see what you mean.  Just like how I'd use something like System.Net.WebClient directly, I'd go ahead and use JToken.Parse directly.

Can I ask you then, what the purpose was for IJsonConverter originally? you mentioned earlier in this thread about an "initial intent" for this thread.

sebastienros:
It was to abstract the deserialization of json documents, not fragments.

PR #6308, opened 2016-01-27
#6302: Fixes some issues when using Forms API checkboxes

### LINKED ISSUE #6302: Checkboxes issues in elements editors forms in dev branch (+ possible fix)
In the dev branch

1. Open an element editor that have a checkbox option.
2. Check the option (if not already done) and save.
2. Re-open the editor, uncheck the option and save.
3. Re-open the editor, the option is still checked.

In 1.9.x, there is this PR #6301 that solves a same kind of issue #5593 due to an option that is initially checked... But in the dev branch it concerns all checkboxes (defined through the forms api), whether initially checked or not...

The problem is that if a checkbox is not checked, no value is posted on submission. So, here, the beginning of a possible solution in the dev branch by modifying `FormsElementDriver.cs`.

        private void UpdateElementProperty(dynamic formElementShape, ElementEditorContext context) {
            var name = (string)formElementShape.Name;
            if (name != null) {
                var value = context.ValueProvider.GetValue(context.Prefix + name);
                if (value != null) {
                    context.Element.Data[name] = value.AttemptedValue;
                }

                // Added code...
                else if(formElementShape.Metadata.Type == "Checkbox") {
                    if (formElementShape.Value == "true")
                        context.Element.Data[name] = "false";
                    else if (formElementShape.Default != null)
                        context.Element.Data[name] = formElementShape.Default;
                }
            }
        }

Here, if nothing is posted, the mainly used "true" value of the `shape.Value` property is tested, then "false" is assigned... To implement other cases a `shape.Default` property can be simply added through the `DescribeForm()`.

Best

jtkech:
Fixes #6302 

In some scenarios, as soon as a form checkbox is saved as checked, it can't be saved again as unchecked (layout element editors in dev branch). In other scenarios, this happens if the checkbox is initially checked (e.g workflows activity editors).

this is due to the fact that, when a checkbox is unchecked, no value is sent on form submission.

Here, when a checkbox value is "true", it is marked as "boolean" and, if unchecked, an hidden input is added to send a "false" value on form submission.

Best

### DIFF
--- a/src/Orchard.Web/Core/Shapes/Scripts/base.js
+++ b/src/Orchard.Web/Core/Shapes/Scripts/base.js
@@ -252,6 +252,35 @@
 
         });
     });
+    // Boolean checkboxes (data-boolean="true").
+    // Ensures that a value is always sent on submission.
+    // This by adding an hidden input when unchecked.
+    $(function () {
+        $('form').find('input:checkbox')
+            .filter(function () {
+                return $(this).data('boolean');
+            })
+            .change(function () {
+                var _this = $(this);
+                var name = _this.prop('name');
+                var next = _this.next('input[type = hidden]')
+                    .filter(function () {
+                        return $(this).prop('name') == name;
+                    });
+
+                if (_this.prop('checked')) {
+                    if (next.length)
+                        next.remove();
+                }
+                else if (!next.length) {
+                    _this.after($('<input>')
+                        .prop('name', name)
+                        .prop('type', 'hidden')
+                        .val('false'));
+                }
+            })
+            .change();
+    });
 })(jQuery);
 
 ///////////////////////////////////////////////////////////////
--- a/src/Orchard.Web/Modules/Orchard.Forms/Shapes/EditorShapes.cs
+++ b/src/Orchard.Web/Modules/Orchard.Forms/Shapes/EditorShapes.cs
@@ -282,7 +282,12 @@ namespace Orchard.Forms.Shapes {
             }
             if (Value != null) {
                 Value = Value is string ? Value : Display(Value);
-                tag.MergeAttribute("value", Convert.ToString(Value), false);
+                string value = Convert.ToString(Value);
+                tag.MergeAttribute("value", value, false);
+
+                if (Type == "checkbox" && value.ToLower() == "true") {
+                    tag.MergeAttribute("data-boolean", "true");
+                }
             }
             if (Checked.GetValueOrDefault()) {
                 tag.MergeAttribute("checked", "checked");


sebastienros:
What happens if the hidden input is not the next tag? Would the values be separated by comma ?

sebastienros:
What are the scenarios you are mentioning? Shouldn't we fix the missing input instead of trying to be clever?

jtkech:
@sebastienros 

First: it only concerns checboxes define through the `DescribeForm()` of the **Forms API**.

> What happens if the hidden input is not the next tag? Would the values be separated by comma ?

I try to get the next hidden input with the same name if it already exists. This is the one that is added by the same script if the checkbox is initially unchecked, or when an user unchecks it. And here `_this.after()` is used to insert the hidden input just after the related checkbox.

> What are the scenarios you are mentioning?

In the dev branch, in the layout element editors, as soon as a checkbox is saved as checked, you can't save it again as unchecked. In 1.9.x, there is not this **1st issue** for all checkboxes that are initially unchecked. But here the **2nd issue** is that if the checkbox is defined in the `DescribeForm()` as initially checked, then same problem...

In dev branch (not tested in 1.9.x), in workflows editor, i've tried the form of the create user activity. There is not the **1st issue**. But for testing, if i define the checkbox as initially checked, then there is the **2nd issue**.

> Shouldn't we fix the missing input instead of trying to be clever?

I've tried to always add an hidden input for this kind of checkboxes, but, when checked, 2 values are sent and it doesn't work (but i can try it again).

Maybe the best solution is server side, e.g in a driver update editor, it depends on the desired beharvior, but we could use the right default value if nothing is sent, this instead of using a previous posted value (e.g "true").

For the layout element editors i've found a solution server side. But, because the Forms API is already used in different places, this is the only way i've found for a more global solution.


Best

jtkech:
@sebastienros, @sfmskywalker 

Just re-tried to add, in `EditorShapes.cs`, an hidden input for Forms checkboxes. This as seen in the aspnet source code for an html helper. Then, if the checkbox is checked 2 values are posted, and through the value provider we get, as you said, a comma separated string `true,false`.

Then, i assume that, with a regular model binding, mvc only take the 1st value. But in some others scenarios we need to do the same ourselves. This can be easily done for layout elements in `UpdateElementProperty()`. But needs another way for workflows that uses `ToJsonString(formValues)`...

An option would be not to fix the 2nd issue and have a rule to never use a Forms checkbox that is initially checked (only seen for one layout element). And only fix the 1st issue that is only related to layout elements (as seen right now), and that can be easily fixed (without an extra hidden input).

But i thought it would be better to use a more general solution where a "boolean" checkbox value (and only one) is always posted. At least, we need to do something because of the 1st issue.

Best

jersiovic:
What do you think of the solution I propose here? https://github.com/OrchardCMS/Orchard/issues/6527 It is based on the solution of this pull request but avoids using javascript and handles the comma separated true,false values using:
```value.ConvertTo(typeof(bool)).ToString().ToLower()```

jtkech:
@jersiovic just tried your suggestion that adds an hidden field and uses the above code in `FormsElementDriver.UpdateElementProperty()` where model binding is done for element editors.

Indeed it works with elements, but, sorry, this creates a little issue e.g with workflows. If i use a `CreateUserActivity` and i check `Approved`, visually all is ok, but when the worflow executes there is an exception when trying to convert `"true,false"` to a boolean.

So, here also, we would need to change how model binding is done. That's why in this PR i add an hidden field only if needed, but as you i don't like to add jquery code.

With the current code there is no issue with workflows (a checkbox would need to be defined as initially checked). So, if we only focus on elements, maybe we can use #6339 without any hidden input.

This only to share my findings because i like the solution with an hidden field, and if it's not good to add jquery code, then your solution is better. But, you would need to update model binding in different places.

Best

Issue #6333, opened 2016-02-04
Content Culture Selector not changing culture on Non-Default Tenant


jburgard:
The content culture selector is unable to obtain content item from a non-default tenant running under a URL prefix.  This is causing the WorkContext.CurrentCulture to remain the default site culture.

My suggested fix is to account for tenant shell settings URL prefix in order to retrieve the content item.

Repro Steps:

1. Create a new tenant 
2. Enable content localization, culture picker modules
3. Add localization part to Page content type
4. In Tenant Settings, create an additional culture(ex. "es-US")
5. Publish a new Page in default culture
6. Publish variation of previous page in newly created culture
7. Navigate to page in previous step.  The page's common metadata shape is displaying English formatted date

jburgard:
This was fixed in 1.10.  Closing.

PR #6334, opened 2016-02-04
#6333: added tenant url prefix check on content culture selector

jburgard:


jburgard:
Just realized this was fixed on 1.10.  Sorry

PR #6339, opened 2016-02-04
#6302: forms checkboxes issues in layouts elements editors

### LINKED ISSUE #6302: Checkboxes issues in elements editors forms in dev branch (+ possible fix)

jtkech:
Fixes #6302 

1. This only concerns the layouts module in dev, where as soon as a checkbox is saved as checked, it can't be saved again as unchecked (here, whatever the initial state of the checkbox).

2. This doen't solve the same kind of issue that could happen elsewhere, as tried in the worflows module, but here only if a forms checkbox is initially checked (this is not the case right now). For a more general solution, see a possible fix with this PR #6308 that uses some jquery code.

So, right now, these issues are visible only in the layouts module. That's why i propose this PR to fix at least these issues for the layouts module.

Best

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Framework/Drivers/FormsElementDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Framework/Drivers/FormsElementDriver.cs
@@ -78,6 +78,12 @@ namespace Orchard.Layouts.Framework.Drivers {
                 if (value != null) {
                     context.Element.Data[name] = value.AttemptedValue;
                 }
+                else if (formElementShape.Metadata.Type == "Checkbox") {
+                    var shapeValue = formElementShape.Value as string;
+                    if (shapeValue != null && shapeValue.ToLower() == "true") {
+                        context.Element.Data[name] = "false";
+                    }
+                }
             }
         }
 


jtkech:
@sebastienros 

Just re-tried it with the last dev and 1.10, same symptom, as soon as a checkbox is saved as checked, it can't be saved again as unchecked. Applying this PR fix it.

Best

jtkech:
@sebastienros, @sfmskywalker.

Oh, maybe a misunderstood here. When i talk about checkboxes it's those that are used in some layout element editors, so particularly (maybe only) in the `DynamicForms Element Editors`.

As soon as saved as checked, you can't save it again as unchecked. Applying this PR fix it.

Best.

PR #6340, opened 2016-02-04
#6304: Prevents client cache from being disabled when public caching

### LINKED ISSUE #6304: Change: Vary by *
I would like to change how orchard 1.9.1 sets the Vary request header form Vary * to Accept-Encoding.
Tried: in web.config and got response headers of both * and Accept-Encoding.
I serve gziped static files from AWS cloud front and to cache these files for users I need a Accept-Encoding Header
using
This somewhat works to give the header Accept-Encoding: gzip,deflate; 
```<add name="Accept-Encoding" value="gzip,deflate" />```
Still it does not cache these resources on users machines without the proper Vary header.
But when add
```<remove name="Vary" />```
```<add name="Vary" value="Accept-Encoding" />```
With
```<add name="Accept-Encoding" value="gzip,deflate" />```
Then I get a static file error on my sites and a Vary response header of
Vary: *,Accept-Encoding,Accept-Encoding

NO settings in Orchard cache to change Vary to anything it will stay Vary:* no matter what, and you cannot change it or fix it in the web config.

jtkech:
Fixes #6304 (maybe not completely but at least the main drawback)

New hint texts have been provided to me by @DanielStolt.

Best

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -503,13 +503,13 @@ namespace Orchard.OutputCache.Filters {
                     maxAge = TimeSpan.Zero;
                 }
                 response.Cache.SetCacheability(HttpCacheability.Public);
+                response.Cache.SetOmitVaryStar(true);
                 response.Cache.SetMaxAge(maxAge);
             }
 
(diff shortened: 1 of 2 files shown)


DanielStolt:
GTG IMO.

dcinzona:
This line causes a problem with loading locally cached content after authentication.
See issue: https://github.com/OrchardCMS/Orchard/issues/6587

Issue #6320, opened 2016-01-31
Exception while cooking the default recipe from the command line


DanielStolt:
Reproduces on fresh clone of `1.10`. Steps:

1. Start `Orchard.exe`.
2. Execute command `setup /SiteName:"MySite" /DatabaseProvider:SqlCe /AdminUsername:admin /AdminPassword:password /Recipe:"Default"`
3. Check the log files (the exception will be emitted in all of them).

This reproduces both in `Debug` and `Release` configs, and both when running directly from source and when running from a published package.

Here's an excerpt where you can see the exception in context:

```
2016-01-31 12:12:21,387 [1] Orchard.Recipes.Services.RecipeExecutionLogger - (null) - INFO CommandStep: Executing command: layer create Authenticated /LayerRule:"authenticated" /Description:"The widgets in this layer are displayed when the user is authenticated"
2016-01-31 12:12:21,405 [1] Orchard.Recipes.Services.RecipeExecutionLogger - (null) - INFO CommandStep: Executing command: layer create Anonymous /LayerRule:"not authenticated" /Description:"The widgets in this layer are displayed when the user is anonymous"
2016-01-31 12:12:21,452 [1] Orchard.Recipes.Services.RecipeExecutionLogger - (null) - INFO CommandStep: Executing command: layer create Disabled /LayerRule:"false" /Description:"The widgets in this layer are never displayed"
2016-01-31 12:12:21,486 [1] Orchard.Recipes.Services.RecipeExecutionLogger - (null) - INFO CommandStep: Executing command: layer create TheHomepage /LayerRule:"url '~/'" /Description:"The widgets in this layer are displayed on the home page"
2016-01-31 12:12:21,505 [1] Orchard.Recipes.Services.RecipeExecutionLogger - (null) - INFO CommandStep: Executing command: site setting set baseurl
2016-01-31 12:12:21,521 [1] Orchard.Recipes.Services.RecipeExecutionLogger - (null) - INFO CommandStep: Executing command: menu create /MenuName:"Main Menu"
2016-01-31 12:12:21,552 [1] Orchard.Recipes.Services.RecipeExecutionLogger - (null) - INFO CommandStep: Executing command: page create /Slug:"welcome-to-orchard" /Title:"Welcome to Orchard!" /Path:"welcome-to-orchard" /Homepage:true /Publish:true /UseWelcomeText:true
[...]

PR #6341, opened 2016-02-05
#6320: Provides a default baseUrl when not set or empty

jtkech:
Related to #6320 in the context of a setup command line and already fixed but only for the 1.10 release

Other alternatives have been discussed in the related issue, but the way that @sfmskywalker has fixed it for 1.10 seems to me to be the best solution. Here i needed to add `ApplicationVirtualPath` when, in a web context, i clear the baseUrl and use a virtual directory (more details in the related issue).

That's why i propose this PR for dev, just to show you this option.

Best.


src/Orchard/Mvc/MvcModule.cs
```diff
@@ -39,10 +40,13 @@ public class MvcModule : Module {
             // which requires activating the Site content item, which in turn requires a UrlHelper, which in turn requires a RequestContext,
             // thus preventing a StackOverflowException.
 
-            var baseUrl = new Func<string>(() => siteService.GetSiteSettings().BaseUrl);
-            var httpContextBase = new HttpContextPlaceholder(baseUrl);
+            var baseUrl = new Func<string>(() => {
+                var url = siteService.GetSiteSettings().BaseUrl;
+                return !string.IsNullOrWhiteSpace(url) ? url : "http://localhost"
```
DanielStolt: Please use `String.IsNullOrWhiteSpace()` as per our guidelines.

jtkech: Done

jtkech:
Because of a recent merge of 1.10 in dev,

i needed to resync from dev to resolve a conflict. So, now you can see the same kind of fix previously done by sipke in the 1.10 release.

Issue #6321, opened 2016-01-31
Layout snippets feature causes command-line recipe execution to silently terminate


DanielStolt:
This happens when:

- One or more snippets exist in a module which is not a theme
- A setup recipe is executed from the command line
- `Orchard.Layout.Snippets` is enabled by the `<Features>` step of the recipe
- The base URL of the site is set in the recipe *before* the content step
- The recipe contains a `<Content>` step
- The `<Content>` step contains at least one content item with a LayoutPart

Currently repros on tip of `1.10`. Steps:

1. Add a sample snippet to Orchard.Blogs\Views named `SampleSnippet.cshtml`.
2. Add a `<Settings>` element and a `<SiteSettingsPart>` to top of the default recipe.
3. Add a `<Content>` step to recipe.
3. Add the `Orchard.Layouts.Snippets` feature to the `<Features>` element.
4. Add a sample `<Page>` element to the recipe.
5. Start `Orchard.exe` from command line.
6. Execute the command `setup /SiteName:"MySite" /DatabaseProvider:SqlCe /AdminUsername:admin /AdminPassword:password /Recipe:"Default"`.
7. Examine the recipe log and `Settings.txt` file. You will find that the site is still in `Initializing` state and that the recipe had silently terminated without any error, before running to completion.

Here's a full recipe that reproduces the issue:

```
<?xml version="1.0"?>
<Orchard>
    
    <Recipe>
        <Name>Default</Name>
        <Description>The default recipe for an Orchard site that includes pages, blogs, custom content types, comments, tags, widgets and basic navigation.</Description>
        <Author>The Orchard Team</Author>
        <WebSite>http://orchardproject.net</WebSite>
        <Category>Default</Category>
        <Tags></Tags>
        <Version>1.0</Version>
        <IsSetupRecipe>true</IsSetupRecipe>
    </Recipe>

    <Feature enable="Orchard.Blogs,Orchard.Comments,Orchard.Tags,Orchard.Alias,Orchard.Autoroute,
                   TinyMce,Orchard.MediaLibrary,Orchard.ContentPicker,Orchard.PublishLater,
                   Orchard.Resources,Orchard.Widgets,Orchard.ContentTypes,
                   Orchard.Scripting,Orchard.Scripting.Lightweight,PackagingServices,Orchard.Packaging,
                   Orchard.Projections,Orchard.Fields,Orchard.OutputCache,Orchard.Taxonomies,Orchard.Workflows,
[...]

PR #6342, opened 2016-02-05
#6321: Always use CallContext Logical Data in ContextState.cs

jtkech:
Related to #6321 

`ContextSate` uses `CallContext` or `HttpContext.Current` if not null. Now, if no ambient HttpContext, `HttpContext.Current` can be set with a fake one (e.g when view rendering). Then the issue if a state is not saved / retrieved in the same way.

@sfmskywalker already fixed it in 1.10 by checking if it's a fake HttpContext. Another alternative has been discussed and @DanielStolt suggested that i do this PR on dev. The alternative is to always use the `CallContext`.

`CallContext Get/SetData` is only thread local. `HttpContext.Current`, when in a web request, is carried among threads swithching due to `async / await` code. That's why here i always use `CallContext LogicalGet/SetData` that has this feature.

**Warning**

Need to be validated only because data will also be carried when starting explicitly `concurrent child threads`. This could be usefull but data would need to be `thread safe`. One solution is to never write a given data slot in the parent thread. Then, on a first write in a child thread, a non shared object is created.

The only place where i've seen such a case is when all `shells are activated` through a `Parallel.ForEach()`. Here, each child task could use `ContextState` through `ProcessingEngine.AddTask()`. But, all seems to be ok because `AddTask()` is not used in the parent thread.

Best.


DanielStolt:
GTG IMO.

PR #6347, opened 2016-02-05
Fix #6317

### LINKED ISSUE #6317: Orchard.Resources: wrong dependency for jQueryUI_Position resource. Is now jQuery, should be jQueryUI
Stumpled upon a JavaScript error when using the jQueryUI_Position resource:

image

The .ui propery isn't defined as jQueryUI isn't loaded because in Orchard.Resources in jQuery.cs it is stated that the jQueryUI_Position resource is dependent on jQuery but that should be jQuery**UI** as it is a jQueryUI plugin.

````
manifest.DefineScript("jQueryUI_Position").SetUrl("ui/position.min.js", "ui/position.js").SetVersion("1.11.4").SetDependencies("jQuery");
````
Should be(look at the end of the line):
````
manifest.DefineScript("jQueryUI_Position").SetUrl("ui/position.min.js", "ui/position.js").SetVersion("1.11.4").SetDependencies("jQueryUI");
````
When changed, error is gone and positioning works fine.

infofromca:
Fix #6317


infofromca:
this is for orchard dev

PR #6353, opened 2016-02-07
#6357: allow more mime type file to upload 

infofromca:
allow more mime type file to upload


sebastienros:
We decided not to add any extension that we don't server OOB. You will be able to find a related issue with the discussion and the decision. Thanks though.

Issue #6343, opened 2016-02-05
ImageProfileManager null reference with Profile


awyl:
1.10

I have code that call the MediaUrl method with just the profile.  After upgrading to 1.10, there is null reference.

In ImageProfileManager.cs, Line 115, GetImageProfileUrl method:
`    if (customFilters == null || !customFilters.Any()) {`
throws null reference when calling with something like
`    @Display.MediaUrl(Profile: "myProfile", Path:Url.Content("~/Media/Default/Logo.jpg"))`

The issue was caused by when arugment "customFilter" became "params customFilters".  The added check `!customFilters.Any()` didn't account for when null was passed into the params.  So the params array has a null as its item.

If someone can guide me how to make PR (because it is a cool thing to do) =P, I can create one.

Otherwise, the fix is as below.  Not sure if it fix all scenarios though:
`    if ( customFilters == null || !customFilters.Any( c => c != null )) { `

Thanks!

infofromca:
relate to #6231

Issue #6231, opened 2015-12-24
NullReference in ImageProfileManager


CSurieux:
This occurs when you call with a profile

the call where customFilter is null
>public string GetImageProfileUrl(string path, string profileName, FilterRecord customFilter, ContentItem contentItem) {
            return GetImageProfileUrl(path, profileName, contentItem, customFilter);
        }

brings a FilerRcord[1] with a sole and null element in the array, and the test go out of control
>public string GetImageProfileUrl(string path, string profileName, ContentItem contentItem, params FilterRecord[] customFilters) {

There should be some 'filter' for this array of records as
customFilters = (customFilters == null) ? null: customFilters.Where(f => f != null).ToArray();

PR #6364, opened 2016-02-09
Treatment null value for params argument inside ImageProfileManager

neTp9c:
Treatment null value for params argument. (When null value of nullable type T is passed as params argument that it will be converted to array of type T with one null member).

Without it Display.MediaUrl(Profile: "ProfileName", Path: "image-path.jpg") which call GetImageProfileUrl with null customFilter doesn't work.


neTp9c:
fix #6343 and #6231
also related to #6349

PR #6375, opened 2016-02-11
fix #6374 for 1.10

### LINKED ISSUE #6374: jQueryUI_Widgets does not exist in 1.10 now
exception for call 
Script.Require("jQueryUI_Widgets");

infofromca:
fix #6374 for 1.10


sebastienros:
I don't think this is the correct fix. We should declare the widgets resource pointing to jQueryUI this way anyone using it won't be broken.

sebastienros:
@DanielStolt

DanielStolt:
@sebastienros I disagree, this fix is correct. The resource name "jQueryUI_Widgets" no longer exists. The old one (which we kept for compat) is named "jQueryUI_Widget" (singular). I did a few similar fixes like this one yesterday, can't figure out why I missed these two.

sebastienros:
Ok, I get it now. I thought this was about a resource removed in 1.10.

PR #6378, opened 2016-02-11
Fix for checking permissions on upload own media

neTp9c:
User with manage own media permission can not upload media files.


sebastienros:
Are you sure it targets only 1.10 and is not an issue in 1.9.3 ?

neTp9c:
In 1.9.3 there is a lack in same place. Permission check doesn't exists at all there (but its only view action). So this fix should be added to 1.9.3 too. How to deal with this situation? Should I close this pull request and recreate the new one for 1.9.3?

sebastienros:
Yes please.

PR #6389, opened 2016-02-13
always have publish button.fix #6388 for 1.9.x

### LINKED ISSUE #6388: Draftable should only control Save button
if i does not misunderstand, Draftable should only control Save button. and we  should always have publish  button.
i will do a PR for 1.9.3

infofromca:
always have publish button.fix #6388 for 1.9.3


sebastienros:
We have updated the issue with the exact actions to apply. WE don't think your changes are correct, probably only the Views should be changed.

infofromca:
save-draft
this is for draftable

infofromca:
save-nodraftable
this is without draftable

infofromca:
added new commit for 
Action:

    Change "Save" to "Save Draft"
    Show Publish all the time.

sebastienros:
Please target 1.9.x as triaged. thanks.

sebastienros:
Too many issues with this PR, please create a new one based on the comments (including merge issues).

PR #6419, opened 2016-02-18
Fixes #6272 update wording for comparisons

### LINKED ISSUE #6272: Incorrect wording on Edit Query page.
For a custom field of type DateTimeField, when you choose "Is greater than or equal to" on the Edit Filter page, it shows up as "is greater or equal than" on the Edit Query page.

rtpHarry:
Fixes #6272 update wording for comparisons


sebastienros:
Checked on Wikipedia, they agree, fine.

PR #6422, opened 2016-02-19
fixes #5672 add generic default theme.png

### LINKED ISSUE #5672: Update codegen theme `theme.png` to a better default
The default image should be changed to a simple copy of the Orchard logo or something generic.

At the moment it shows the thumbnail from The Theme Machine theme:

https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.CodeGeneration/CodeGenerationTemplates/Theme.png

This is confusing when using the `/BasedOn` parameter.

You could add extra code to pull the `theme.png` from the BasedOn theme but the fallback should still be generic.

rtpHarry:
fixes #5672 add generic default theme.png as using `/basedon` could create confusing themes


sebastienros:
I don't see how the new one is better. The old one is surely not great, but the one you suggest seems worse to me. What about a generic layout image? Other suggestions?

sebastienros:
I thought it was the widget zones png, I was wrong, please target 1.10.x and I will merge it.
Thanks

rtpHarry:
ok I've had a look but I don't see a way to do that, re-opening with a new PR

PR #6427, opened 2016-02-19
Fixes #5907 convert modules admin page to use css flexbox

### LINKED ISSUE #5907: Modules Admin - Display glitches when module has many dependencies
Branch tested: Dev
Browser tested: Chrome

On large screens a module's dependencies may wrap onto a new line and touch the bottom border:

image

On smaller screens the dependencies can disappear from view:

image

(missing the widgets)

Consider implementing an auto-height script such as:

  - https://github.com/liabru/jquery-match-height/

rtpHarry:
Fixes #5907 convert modules admin page to use css flexbox

Summary view now stretches when needed on a per row basis and keeps alignment:

image

Previously "Depends on" would be cut off for long lists like Blogs.

Detail view is converted to column flexbox as well:

image

I artificially inflated the Blog dependencies to confirm that it expands correctly.


sebastienros:
Just a note that this will be replaced by the new admin theme, using the same technique.

rtpHarry:
I will try to keep my emotions under control when that happens.

PR #6437, opened 2016-02-20
Improved snippets harvester.

### LINKED ISSUE #6432: Multiple issues caused by snippet element harvester
The snippet element harvester (part of the `Orchard.Layouts.Snippets` feature) renders all detected snippet razor views on every request, as part of the harvesting process. This happens both on all front-end request and in the admin UI.

This causes a multitude of issues:

- All snippets that call `@Html.TitleForPage()` will cause that title to be added to the composite page title for every page on the site.
- All snippets that include or require any stylesheets or scripts will cause all those resources to be added to all pages on the site.
- When the snippet harvesting happens as part of a recipe execution, the rendering causes an `HttpContext` to be created, suddenly changing the storage of recipe execution state and thereby causing the next step in the recipe to never be executed.
- Stylesheets and scripts referenced by snippets will be rendered in the admin UI while using the layout editor.
- All ad-hoc shapes added to global zones in snippet Razor views will be added for every page on the site.
- Etc etc...

Basically, there is an implicit assumption in MVC in general and Orchard in particular, that Razor views will only be rendered when they are being actually added to the response of the current request. As a consequence, many things commonly done in Razor views have side-effects and affect the state of a request and response beyond only what's being rendered.

The harvester does this rendering to find out what snippet parameters are exposed by each snippet, but this should really only be needed in one situation: in the layout editor in the admin UI when the snippet element is actually placed on the layout.

The behavior of the snippet harvester needs to be fixed such that snippet Razor views are only rendered in that situation, and of course during requests where that snippet is actually displayed.

sfmskywalker:
Fixes #6432

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Providers/SnippetElementHarvester.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Providers/SnippetElementHarvester.cs
@@ -32,12 +32,13 @@ namespace Orchard.Layouts.Providers {
         private readonly Work<IShapeDisplay> _shapeDisplay;
         private readonly Work<ICurrentThemeShapeBindingResolver> _currentThemeShapeBindingResolver;
         private readonly Work<ITokenizer> _tokenizer;
+        private readonly IWorkContextAccessor _wca;
 
         public SnippetElementHarvester(
             IWorkContextAccessor workContextAccessor,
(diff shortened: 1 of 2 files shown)


DanielStolt:
Looks very solid to me. Will do some testing during tomorrow, and if all goes well I'll pull this in.

PR #6449, opened 2016-02-22
[Fixes #6396] - Recent files don't appear on load

### LINKED ISSUE #6396: add media item recent show nothing
add media item recent show nothing after i just upload a image

dcinzona:


sebastienros:
What about 1.9.3 ?

dcinzona:
Well, the recent items folder works in 1.9.2 but not because the bug doesn't exist, but because it is checking to see if the folderName is empty at the API endpoint.  This is the API call that is called on load on 1.9.2 (I have not tested on 1.9.3)
`/Admin/Orchard.MediaLibrary?folderPath=`
That path returns recent items on 1.9.2, but not on 1.10 because the API call on 1.10 is `/Admin/Orchard.MediaLibrary?recent`

Issue #6410, opened 2016-02-17
Extend NotifyType by 'Success'


SzymonSel:
As discussed [here](https://youtu.be/Nf7eMLqFU8w?list=PLuskKJW0FhJfOAN3dL0Y0KBMdG1pKESVn&t=3042)

SzymonSel:
Let me take care of this

PR #6450, opened 2016-02-22
#6410 Added NotifyType.Success 

SzymonSel:
and replaced most usages of `NotifyType.Information` with it. Added appropriate styles in css files.

Some minor fixes - wrong notification types.


sebastienros:
Somehow this is a breaking change as you are using a new css class on the front end. No website will have this style defined if they upgrade, as you changed several front end information to success. It should target dev instead. Wait for more comments.

SzymonSel:
Yup, I'm aware of that, but leaving the message-Information class referring to NotifyType.Success just didn't seem right.

sebastienros:
So as I suggested (indirectly), please move to `dev`

SzymonSel:
Will do

sebastienros:
:watch:

SzymonSel:
Noted!

SzymonSel:
https://github.com/OrchardCMS/Orchard/pull/6634

PR #6456, opened 2016-02-23
fix for #6330

ryanmagoon:
Changed extension manifests to use “orchardproject.net” as the Website

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentPicker/Module.txt
+++ b/src/Orchard.Web/Modules/Orchard.ContentPicker/Module.txt
@@ -1,7 +1,7 @@
 Name: Orchard.ContentPicker
 AntiForgery: enabled
 Author: The Orchard Team
-Website: https://github.com/OrchardCMS/Orchard
+Website: http://www.orchardproject.net
 Version: 1.10
 OrchardVersion: 1.9
 Description: UI for selecting Content Items.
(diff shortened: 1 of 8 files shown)


Jetski5822:
Hey cool - but is there any reason why you are missing the scheme off of the url?

ryanmagoon:
I followed the issue a little too exactly, huh 🤔
-- 
Ryan Magoon

Jetski5822:
haha - i think so!!! Change it to 

https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.CodeGeneration/CodeGenerationTemplates/ThemeManifest.txt#L3

across the board! - but still, good stuff!

ryanmagoon:
Done

sebastienros:
Can you squash your commits? It's simple, checkout the branch of the PR you are working on, reset the branch "softly" to the branch you target (dev), commit with the same message. This way we get a single commit, we don't care about the history in this case. We are not asking that usually but I want to change this habit, using a simple technique. Thanks.

dcinzona:
I wonder if this will work:
(got this from here: https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history)
`git rebase -i HEAD~2`
`s`
Would that work?  I, too, am looking for an easy way to do this.

dcinzona:
Or here is a better explanation: https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

ryanmagoon:
Done

Issue #6446, opened 2016-02-21
MediaLibrary: User sub-library doesn't show recent items


BenedekFarkas:
When a user has Dashboard access with `ManageOwnMedia` permission (not `ManageMediaContent`), their media items are always placed under the `\Users\<username>` folder structure. When such a user goes to `/Admin/Orchard.MediaLibrary`, they should be able to see their recent media items, but the items are not displayed, even after clicking on "Recent".

The reason behind that is the following:
- Users with `ManageMediaContent` permission see all the media items, so when they see the recent items, that means that their root folder is the absolute root (`folderPath` is empty).
- Users with only `ManageOwnMedia` permission see only their own media items, so their root folder is the user media folder (`folderPath` is `Users\<username>`).
- The logic that determines whether recent items need to be displayed compares the current folder to empty string. That is always a match for media admins (since the root folder for them is empty string) when they open up the Media Library and/or click on "Recent", but never matches for users with `ManageOwnMedia` permission, since their root folder is not an empty string.

I've tested the following implementation locally and works exactly as it should (only small additions, no breaking changes):
- IMediaLibraryService extended with the `GetMediaContentItemsRecursive` and `GetMediaContentItemsCountRecursive` methods that do exactly the same as their non-recursive counterparts, but with `recursive` set to `true` when calling `BuildGetMediaContentItemsQuery` (btw that parameter was never used before, because the only time its value is `true`, it's never evaluated due to another condition check above it).
- `RecentMediaItems` in the `AdminController` uses the new recursive methods.
- The current root folder is passed on from the controller to the UI logic to be able correctly determine if the current folder is user's root folder or not (while not changing the behaviour for admins).

@sebastienros I suggest this to be added to 1.9.3/1.10.

dcinzona:
I posted this here: https://github.com/OrchardCMS/Orchard/issues/6396

This is actually caused because hasFolderPath is set to true (since it's being passed in as an empty string in the view).  This line here: https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/Admin/Index.cshtml#L99 should use string.isNullOrWhiteSpace (because on initial load, it's "" by default, not null).  In fact, it will never be null.
https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/AdminController.cs#L44
It defaults to "".

[...]

PR #6461, opened 2016-02-23
Media Library: fixing NRE in FolderController and #6446

BenedekFarkas:
See the description of #6446 for explanation.


BenedekFarkas:
Fix to be pushed directly to 1.10 (instead of 1.9.3 due to interface changes).

PR #6463, opened 2016-02-24
Retrying on tenant startup

sebastienros:
Targeting 1.10 to focus on the changes. Don't merge.

### DIFF
--- a/src/Orchard/Environment/DefaultOrchardHost.cs
+++ b/src/Orchard/Environment/DefaultOrchardHost.cs
@@ -15,6 +15,7 @@ using Orchard.Mvc;
 using Orchard.Mvc.Extensions;
 using Orchard.Utility.Extensions;
 using Orchard.Exceptions;
+using System.Threading;
 
 namespace Orchard.Environment {
     // All the event handlers that DefaultOrchardHost implements have to be declared in OrchardStarter.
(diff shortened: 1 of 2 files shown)


src/Orchard/Environment/DefaultOrchardHost.cs
```diff
@@ -142,16 +144,25 @@ public class DefaultOrchardHost : IOrchardHost, IShellSettingsManagerEventHandle
...
+                    for (var i = 0; i <= Retries; i++) {
+                        bool failed = false;
+                        try {
+                            var context = CreateShellContext(settings);
+                            ActivateShell(context);
+                        }
+                        catch (Exception ex) {
+                            // An exception at this point is always fatal as it literally kills the
+                            // tenant. What is more fatal than something that kills you?
+                            Logger.Error(ex, "A tenant could not be started: " + settings.Name + " Try number: " + i);
```
sebastienros: I don't like the word 'Try', 'Tentative' seems better, any other suggestion?

sebastienros:
The scenario is that at startup there might be some SQL hiccups and the tenant would not start, and you would not know unless you read the error logs frequently. As Azure Websites  restarts the sites regularly, some sites might go down. It happens sometimes on weblogs.asp.net as we have 800 tenants.

dcinzona:
I like this change.  This can happen for us as well.  What is the retry timeout? Exponential?

src/Orchard/Environment/DefaultOrchardHost.cs
dcinzona: Attempt?

sebastienros:
What do people think about retry delay?

PR #6477, opened 2016-02-26
#6470: Dev & 1.10 Dynamic form field binding issue

### LINKED ISSUE #6470: Dev & 1.10: Dynamic form field binding issue
Just re-tried in the last dev branch and 1.10 release. It was working in a previous version but maybe here i'm doing something wrong.

Repro:

- Create a content type with a field of type text.
- Create a dynamic form with the `CreateContent` option and select the above content type.
- Insert a text form field and bind it to the text field of the above content type.
- Display the form on front end, input a text in the form field and submit.
- Edit the created content item in the dashboard, **there is no value**.

Note: Trying this, i could see again the other issue where, for most of the element editors, as soon as a checkbox is saved as checked, it can't be saved again as unchecked. Can be fixed for me with #6339.

Best

jtkech:
Fixes #6470.

Note: still an issue with checkboxes in dynamic forms element editor, see #6339.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.DynamicForms/Services/FormService.cs
+++ b/src/Orchard.Web/Modules/Orchard.DynamicForms/Services/FormService.cs
@@ -263,7 +263,7 @@ namespace Orchard.DynamicForms.Services {
                 ReadElementValues(element, context);
 
                 var value = context.Output[element.Name];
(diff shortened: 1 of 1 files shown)


jtkech:
Just done in 1.10

PR #6478, opened 2016-02-26
Fixes Layout OnIndexing when using a snippet element

jtkech:
Notice that this is already done in `ContentDisplayBase.cs`. But when using a snippet element, the same update needs to be done in `DefaultDisplayManager`.

Because we provide a fake context in the background, sometimes we have to replace `HttpContext != null` with `!HttpContext.IsBackgroundContext()` when it doesn't make sense to do anything.

Here we have to do the opposite to retrieve the `CurrentTheme` needed to find the snippet razor view. So, i've replaced `IsBackgroundContext()` with `!= null`. Notice that the null check may be useless, but i keep it as done in `ContentDisplayBase.cs`.

Best.

### DIFF
--- a/src/Orchard/DisplayManagement/Implementation/DefaultDisplayManager.cs
+++ b/src/Orchard/DisplayManagement/Implementation/DefaultDisplayManager.cs
@@ -11,7 +11,6 @@ using Orchard.DisplayManagement.Shapes;
 using Orchard.Localization;
 using Orchard.Logging;
 using Orchard.Mvc;
-using Orchard.Mvc.Extensions;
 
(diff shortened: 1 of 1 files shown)


sfmskywalker:
I think this looks good. Could you perhaps add a comment to the changed line for future developers to understand what's going on? :)

jtkech:
Ok, i'll do it.

jtkech:
Done.

sebastienros:
Just to punish you @jtkech , please squash the commits into a single one (same PR). If you don't know, ping me.

jtkech:
Done ;)

sebastienros:
1.10.x please

PR #6492, opened 2016-02-26
ContextState Compromise

jtkech:
Ending this discussion https://github.com/OrchardCMS/Orchard/commit/42e71fe9fa2cc5337eb9239d68a08a2b680f3df2#commitcomment-16323433 with @DanielStolt and @sfmskywalker, here a last compromise to try that is, as said by sipke, basically the same as it is now...

- Still use the `HttpContext` during a web request because this is more reliable.
- Use the new `IsBackgroundHttpContext()` extension to check if it's a valid context.
- Nevertheless, replace `CallContext.Get/SetData()` with `CallContext.LogicalGet/SetData()`.
- (The last point is easy to revert if some drawbacks)

Best

### DIFF
--- a/src/Orchard/Environment/State/ContextState.cs
+++ b/src/Orchard/Environment/State/ContextState.cs
@@ -1,12 +1,12 @@
-﻿using System;
+using System;
+using System.Runtime.Remoting;
 using System.Runtime.Remoting.Messaging;
 using System.Web;
-using Orchard.Mvc;
+using Orchard.Mvc.Extensions;
 
 namespace Orchard.Environment.State {
-
     /// <summary>
-    /// Holds some state for the current HttpContext or thread
+    /// Holds some state for the current HttpContext or Logical Context
     /// </summary>
     /// <typeparam name="T">The type of data to store</typeparam>
     public class ContextState<T> where T : class {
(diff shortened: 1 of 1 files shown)


sebastienros:
Unit tests are failing after this PR

sebastienros:
See teamcity.codebetter.com

sfmskywalker:
Checking...

jtkech:
Me too, i'll try unit tests to understand why.

sfmskywalker:
Strange, the failing unit test on TC passes on my local. On TC, the unit test reports a 500 Internal Server Error. Maybe it's temporary?

sfmskywalker:
image

sebastienros:
It it passes locally, we'll wait for the next time it runs to be sure it repros.

jtkech:
`StatusCodeShouldBe404ForUnexistingResources()` also passes on my local environement.

sebastienros:
This time more than 20 tests don't pass

sfmskywalker:
Weird. AppDomainUnloaded exceptions. I'll revert the change set tonight and see if that helps.

jtkech:
@sebastienros, @sfmskywalker just seen your comments.

Because `ObjectHandle` inherit from `MarshalByRefObject`, and you are talking about AppDomain exceptions, not sure at all but i suspect a cross domain lifetime problem. So, maybe worth to add this at the beginning of the `ContextState` class:

    //using System.Runtime.Remoting; // no more used
    ...
    public class ContextState<T> where T : class {

        internal class ObjectHandle : System.Runtime.Remoting.ObjectHandle {
            public ObjectHandle(object o) : base (o) { }
            public override object InitializeLifetimeService() {
                return null;
            }
        }
    ...

Best

jtkech:
Just to try, PR #6498 done.

Sorry again for the downside.

jtkech:
@sebastienros, @sfmskywalker 

My last update solves some exceptions but creates new ones (seen on teamcity that currently runs).

But just seen that i can also run Specs in my local environement (not only unit tests). And i can repro the exact same errors.

So, if possible let me one day more to try to solve this. Note: maybe 2 days because right now i need to leave and i can re-work on this the next night.

Best.

PR #6496, opened 2016-02-27
#6495: host field should be required 

### LINKED ISSUE #6495: multi tenant --one of URL prefix and Host should be required
obviously host field should be required to avoid my mistake at  #6434

infofromca:
fixed #6495


geertdoornbos:
Commented on #6495

dcinzona:
Also commented on #6495

PR #6498, opened 2016-02-27
ContextState: Attempt to fix cross app domain issues on teamcity

jtkech:
Sorry for the downside of #6492.

### DIFF
--- a/src/Orchard/Environment/State/ContextState.cs
+++ b/src/Orchard/Environment/State/ContextState.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Runtime.Remoting;
 using System.Runtime.Remoting.Messaging;
 using System.Web;
 using Orchard.Mvc.Extensions;
@@ -50,6 +49,13 @@ namespace Orchard.Environment.State {
             }
             else {
                 HttpContext.Current.Items[_name] = state;
+            }
+        }
+
+        internal class ObjectHandle : System.Runtime.Remoting.ObjectHandle {
+            public ObjectHandle(object o) : base(o) { }
+            public override object InitializeLifetimeService() {
+                return null;
             }
         }
     }


sfmskywalker:
No worries mate. I'll pull it in since the tests are broken anyway, but could you maybe explain a bit why this may work?

jtkech:
Not sure at all but someone already solved in Orchard the same kind of issue around the cross app domain lifetime. If i remember correctly it was in the context of the command line (so with cross app calls) and for all the objects that inherit from `MarshalByRefObject`. And the `DefaultOrchardHost` uses the `ContextState` (for _tenantsToRestart) during all the app life... But again not sure at all.

sfmskywalker:
Ok. Well, we'll see how it goes.

jtkech:
Ok, thanks to be patient.

I'm not married with Logical CallContext ;)
But if we get rid of it, i would like to understand why...

jtkech:
@sebastienros, @sfmskywalker 

So, because i can now repro exactly the same failing tests on my local, i could work on it and i found how to fix it (see below). But i don't want to disturb the Orchard timeline. Indeed, the easiest way is to revert my 2 last PRs and maybe do a last one to only use the new `IsBackgroundHttpContext()` extension.

Otherwise, if you think it's worth to give it a last try, i can do easily a last PR and here my last findings. But you know maybe other little drawbacks on the dev branch... That said, i could run all unit tests and specflows. Only 4 tests in `Orchard.Profile` failed but it was the same before my updates.

**1st issue**
In a `Cross AppDomain` context during specflows, the first kind of issue was that an object retrieved from the `Logical CallContext` can be a transparent proxy not related to a local instance. The fix was to add this check:

`var data = handle != null && !RemotingServices.IsTransparentProxy(handle) ? handle.Unwrap() : null;`

**2nd issue**
Based on the exception message it was related to an assembly that failed to load. The fix was to replace the following in `Orchard.Resources/web.config`. And i've done the same in `Orchard.Conditions/web.config` because there are the only 2 web.config files where `PublicKeyToken=null` for these asemblies.

                <add assembly="Orchard.Framework, Culture=neutral, PublicKeyToken=null" />
                <add assembly="Orchard.Core, Culture=neutral, PublicKeyToken=null" />

With:

                <add assembly="Orchard.Framework" />
                <add assembly="Orchard.Core" />

So, let me know, i can easily revert the updates and only use the new extension, or try the above one last time. Anyway, i again learned a lot with this.

Best

sfmskywalker:
Well, if this fixes the unit tests and works on both Setup from IIS as well as Orchard.exe, I think we should give it a go. Perhaps you could add some of the green stuff explaining the reasoning (in a nutshell)?

sebastienros:
Apply the fix

jtkech:
Ok, i'll re-fetch the last source and re-run all tests. Then i'll do the PR.

PR #6500, opened 2016-02-28
fixed one of URL prefix and Host should be required

### LINKED ISSUE #6495: multi tenant --one of URL prefix and Host should be required

infofromca:
fixed #6495

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MultiTenancy/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MultiTenancy/Controllers/AdminController.cs
@@ -68,6 +68,10 @@ namespace Orchard.MultiTenancy.Controllers {
                 ModelState.AddModelError("Name", T("Invalid tenant name. Must contain characters only and no spaces.").Text);
             }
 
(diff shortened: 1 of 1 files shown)


geertdoornbos:
Question: If you're enforcing a requirement regarding tenant settings, what would be the safest way (location) to enforce that requirement through the whole system?
What if I would create a tenant using a console or a recipe?

infofromca:
i have not tried console or a recipe , especially the  recipe which i think we should discuss further. but is this another issue?

geertdoornbos:
Other scenario, same issue. Now you know this, how would you solve this issue?

infofromca:
this fix is only for UI situation

infofromca:
and i also got confused about recipe issue:
the purpose to use recipe is export then import.
if you export xml in which there is already URL prefix or Host(because you get it from UI),  is there any problem for importing the xml file?

geertdoornbos:
> the purpose to use recipe is export then import

No, recipes can do much more.

I mentioned recipes as an example for a possible scenario.
You (correctly) added another case in which the requirement should be satisfied. If you're importing from XML then too the same rule should be applied.

There should be one location to enforce the requirements. No duplicate code, right?

infofromca:
but i have not found any problem when i used recipe

geertdoornbos:
Again, I mentioned recipes as a possible scenario that could alter tenant configuration.

infofromca:
make the UI stronger first.
then if you have other scenarios , repo it, and open it and resovle it

PR #6519, opened 2016-03-02
#6518 Highlight all images

### LINKED ISSUE #6518: Media Library: Add button to select all items
There is currently no way to select all images from a folder in the Media Library. If you want them all, you have to individually ctrl+click every image then click the "select" button. I'm proposing adding a button next to the "Select" button called "Hightlight All" which when clicked, loops through all images in the current displayed folder and marks them as selected (green square). 

I'm proposing the phrase "Highlight All" instead of "Select All" because the current "Select" button confirms your choice and closes the Media Library dialog, so this would avoid confusion. This new button would not close the dialog, so it would allow you to navigate to another folder and "Highlight All" from that folder as well, adding to the selection list on the right hand side.

MobesE46:
Fixes #6518 - New button to highlight all images in media library folder.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/Admin/Index.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/Admin/Index.cshtml
@@ -75,6 +75,7 @@
                 <footer>
                     <div id="media-library-main-selection-select">
                         <button class="button-select" data-bind="visible: selection().length > 0">@T("Select")</button>
+                        <button class="button-select-all">@T("Select All")</button>
                         <button class="button-cancel">@T("Cancel")</button>
                     </div>
                 </footer>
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/Admin/Index.cshtml
```diff
@@ -75,6 +75,7 @@
                 <footer>
                     <div id="media-library-main-selection-select">
                         <button class="button-select" data-bind="visible: selection().length > 0">@T("Select")</button>
+                        <button class="button-highlight-all" aria>@T("Highlight All")</button>
```
sebastienros: `Select All` please

sebastienros:
Thanks, as we won't merge this until thursday, would you have some time to try and squash the two commits into one? You don't need to create a new PR, just force push when the squash is done.

MobesE46:
Ok, I think I did the squash and push correctly.  Thanks for pulling this change in!  :)

sebastienros:
Sorry, the 1.9.x branch doesn't exist anymore, can you retarget it to 1.10.x? It will be merged right after it's done.

MobesE46:
Ok, closing PR to retarget 1.10.x

PR #6520, opened 2016-03-02
1.10 Field names duplicated in dropdown list, fixed #6497

### LINKED ISSUE #6497: 2 times TextField in the dropdown when adding a field to a content
Version Dev last today

infofromca:
fixed #6497


sebastienros:
I don't think this is the correct fix. The `ContentDefinitionService.GetFields()` method should be fixed instead not to return duplicates. Because is another module also needs to display the fields, the same bug will be visible. If you change this PR, please squash your commits.

jtkech:
@infofromca as per @sebastienros suggestion you can update `ContentDefinitionService.GetFields()`.

For not to create a custom `IEqualityComparer()`, this can be done with something like this:

            return _contentFieldDrivers.SelectMany(d => d.GetFieldInfo())
                .GroupBy(x => x.FieldTypeName).Select(g => g.First());

Best.

PR #6535, opened 2016-03-07
Fixes 2 little issues in dev around the TextField default values

jtkech:
**Issues in dev** (and 1.10 i think)

1. **When you 1st create** e.g a page that has a text field that has a default value. If you input a non empty string and publish, then the default value is used (not your input). The 2nd time it's ok.

2. Idem but leaving the input empty, then the default value is used, ok, but if it's a token, the token itself is returned, it is not resolved. The 2nd time it's ok.

Note: To allow default values tokenization, a 2nd `TextField` driver has been added that runs after (because of a `Core` dependency in `Orchard.Fields`) the original.

**Fixes**

1. In the original driver i've re-integrated `field.Value = viewModel.Text;`.

2. In the 2nd driver, if the default is not null and the value is equals to the default, then the default is tokenized. if it's not a token, the same value is returned by the tokenizer. Note that if the default is not null, because of 1., here the field value can't be null, but i've kept a null check on it.

There is another issue when you add a field to a content type, `TextField` appears 2 times in the list box. There is an open PR #6520 by @infofromca that will fix it.

Best.

### DIFF
--- a/src/Orchard.Web/Core/Common/Drivers/TextFieldDriver.cs
+++ b/src/Orchard.Web/Core/Common/Drivers/TextFieldDriver.cs
@@ -72,6 +72,7 @@ namespace Orchard.Core.Common.Drivers {
                                         () => shapeHelper.EditorTemplate(TemplateName: "Fields.Common.Text.Edit", Model: viewModel, Prefix: GetPrefix(field, part)));
                 }
 
+                field.Value = viewModel.Text;
                 var settings = field.PartFieldDefinition.Settings.GetModel<TextFieldSettings>();
 
                 if (String.IsNullOrEmpty(field.Value) && !String.IsNullOrEmpty(settings.DefaultValue)) {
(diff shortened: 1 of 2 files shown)


sebastienros:
Can you target it for 1.10? The default value feature is totally unusable right now. Have you tried with the other fields?

jtkech:
@sebastienros as i remember, for comparison i've tried the input field that was working for the default values (Here we are talking about content type fields, not dynamic forms element fields).

I'll try other fields this night, and target 1.10.

Best.

jtkech:
@sebastienros 

Here the other content fields where i've tried default values. I've used tokens for the fields that tokenize their default value.

- **Input field** ok.
- **LinkField** ok.
- **BooleanField** ok. It's a special case where the user always input a value implicitly. So, here the default value is used if the content item is new (no draft and no published version).
- **NumericField** ok. But here the default value is used only if the field is not `Required`. Maybe useful to only check if the value is leaved empty, as for the TextField.
- **DateTimeField** ok and here also, the default value is used only if the field is not `Required`. Maybe because when required we don't want to use the default value in case it's not a valid dateTime. But this can be checked as done for a regular user input.

I will target 1.10 but let me know if i've to update other fields regarding the interaction between th `Required` option and the use of the default value.

Best.

sebastienros:
If you can fix the `Required` issue please do so. But it's time to sleep also.

jtkech:
I agree!

jtkech:
@sebastienros 

Anyway, when `Required` the field can't be left empy because there is client validation (it's not an option as with dyn forms). So, this PR can be left as it is and only needs to target 1.10.

Best

PR #6536, opened 2016-03-07
#6542: Little issue with content picker

jtkech:
Fixes in one place #6542 

When using the new `Add Selected` button that allow to select multiple items in the content picker view, the returned data is not an item but an array of items. Some scripts that consume these data has been updated but not in all places.

Here, because a content menu item only needs one related item, if data is an array we only use the first one.

Note: Same issue to insert an item in a list (Orchard.Lists). Here, we have to iterate through the array to insert multiple items. I need to update `orchard-lists-admin.js` but i need to update my tools to produce the minified version of the script. I'll do it asap.

Best


sebastienros:
Please target 1.10.x

Issue #5089, opened 2015-03-12
Clone content item slug is not made unique


orchardbot:
@dpomt created:
https://orchard.codeplex.com/workitem/21260

Steps to reproduce:

Clone a Content item
--> The Content item has same slug as the the cloned one.

Easiest would be to call AutoRouteService.ProcessPath within DefaultContentManagerClone but not possible since it should not have a dependancy to AutoRoute.

Any ideas how this could be fixed correctly?

Issue #4983, opened 2015-01-21
Cloned items should have their Created date empty


orchardbot:
@sebastienros created:
https://orchard.codeplex.com/workitem/21154

When we create a new item, the creation date is empty and defined on the first publication. It can then be updated as needed. It's very useful for instance when used in conjunction with Publish Later.

When a content item is cloned, it's usefully a Draft version, but the Creation Date is copied too, and it's not possible to delete it. It would be better to make it empty, like a new content item only available as Draft, which a cloned one is.

Issue #4970, opened 2015-01-15
Clone content item drafts: selected taxonomies not cloned


orchardbot:
@dpomt created:
https://orchard.codeplex.com/workitem/21141

If I clone a content item draft (not yet published content item) that has selected taxonomies for some taxonomy fields those taxonomy fields are empty in the clone.
Any idea where in the src code the problem might be?

Thanks

PR #6540, opened 2016-03-07
Content cloning and translation

TFleury:
This PR adds new clone handling to replace export/import logic, and also allow translating a clone of the master content item.

It introduces new `Cloning` and `Cloned` content handlers to specify a clone logic that can be different of import/export ones.

`ContentManager`'s `Duplicate` method don't save a new draft content any more, but create a new content initialized by clone handlers. Admin Content Controller Duplicate action opens a editor to edit the cloned content before saving.

It fixes : #5535, #5089, #4983, #4970, #4538


src/Orchard/ContentManagement/IContentManager.cs
```diff
@@ -42,6 +42,14 @@ public interface IContentManager : IDependency {
         ContentItem Clone(ContentItem contentItem);
 
         /// <summary>
+        /// Makes a clone of the content item with a special context
+        /// </summary>
+        /// <param name="contentItem">The content item to clone</param>
+        /// <param name="cloneContext">The cloning context</param>
+        /// <returns>Clone of the item</returns>
+        ContentItem Clone(ContentItem contentItem, Handlers.CloneContentContext cloneContext);
```
sebastienros: Why is it necessary?

sebastienros:
It needs to target dev, it's a breaking change, big new feature. But it looks great.

src/Orchard/ContentManagement/IContentManager.cs
TFleury: This implementation is used by Localization. It's needed for Translate (Localization), which uses `Clone` with a `TranslateContentContext` (derived from `CloneContentContext`).

Skrypt:
There is something missing. When a content item has it's culture undefined. If we try to clone the content item it throws an error.

src/Orchard/ContentManagement/IContentManager.cs
sebastienros: From what I see and understand it shouldn't be necessary. Clone doesn't have to know about localization. The localization module should just call Clone(item) then change the new content item accordingly.

src/Orchard.Web/Modules/Orchard.Localization/Handlers/TranslateContentContext.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Orchard.ContentManagement;
+using Orchard.ContentManagement.Handlers;
+using Orchard.Localization.Records;
+
+namespace Orchard.Localization.Handlers {
+    public class TranslateContentContext : CloneContentContext {
+        public CultureRecord TagetCulture { get; set; }
```
Skrypt: TargetCulture "typo"

src/Orchard.Web/Modules/Orchard.Localization/Drivers/LocalizationPartDriver.cs
```diff
@@ -145,8 +146,17 @@ public class LocalizationPartDriver : ContentPartDriver<LocalizationPart> {
...
+        protected override void Cloning(LocalizationPart originalPart, LocalizationPart clonePart, CloneContentContext context) {
+            if(context is TranslateContentContext) {
+                var translateContext = context as TranslateContentContext;
+                if(translateContext.TagetCulture != null) {
+                    clonePart.Culture = translateContext.TagetCulture;
+                }
+                clonePart.MasterContentItem = translateContext.ContentItem;
+            }
+            else {
+                clonePart.Culture = originalPart.Culture;
```
Skrypt: if originalPart.Culture != null here

src/Orchard/ContentManagement/IContentManager.cs
Skrypt: He needs to know wich culture can be selected in that new translated "cloned" content item.

TFleury:
As @sebastienros asked, I reopened a new PR based on dev branch. I also removed `TranslateContentContext`.

PR #6545, opened 2016-03-08
#6544: NRE with containable part in Orchard.Lists

### LINKED ISSUE #6544: NRE with containable part in Orchard.Lists
**Repro**

- Create a content type with a containable part.
- This by leaving unchecked `Show Container Picker`.
- Enable the `Orchard.Lists` module.
- Create an item of this type, you'll get an exception each time you will edit it.

Note: if you check `Show Container Picker`, when editing an item you can choose a container. But you still have the same issue when creating an item, and afterwards if you don't select a container.

I have a fix.

Best.

jtkech:
Fixes #6544 

Here, i also fix a little `404 not found` because `nprogress.min.css` doesn't exist.

Best.


sebastienros:
I will wait to merge this one as I'd like it to be on 1.10.x which doesn't exist right now. It will once we ship 1.10. 
If it works in 1.9.x then you can target it instead.

jtkech:
Done

PR #6557, opened 2016-03-09
Output cache route config fixes #6523

### LINKED ISSUE #6523: Outputcache still checks cache even if a route is configured as not to cache
Hi,

Noticed this yesterday,

The outputcachefilter doesn't check the route configs to see if a route is ignored during the `RequestIsCacheable` method (it does check the `IgnoredUrls `field from settings in this method), this means that it will get as far as looking for an item in cache. Once it comes to server the item it does check the configs in the `ResponseIsCacheable` method, realises its disabled in the config (cacheduration set to 0) and never serves the cache item.

This means if your using an external cache provider like redis you can end up doing a request to redis for no reason as far as i can tell?

carlwoodhouse:


src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -314,6 +314,12 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
                 return false;
             }
 
+            // Don't cache ignored routes
```
DanielStolt: Period.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -314,6 +314,12 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
...
+            if (_cacheRouteConfig != null && _cacheRouteConfig.Duration == 0) {
+                Logger.Debug("Request for item '{0}' ignored because route is configured to not be cached.", itemDescriptor);
```
DanielStolt: Would be nice to include the route in the log message here, like so:

    Request for item '{0}' ignored because the route '{1}' is configured to not be cached.

DanielStolt:
Good to go - nice improvement! Thanks @carlwoodhouse

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
carlwoodhouse: Agreed :) updated commit pushed

DanielStolt:
OK, nice - just to confirm @carlwoodhouse before I merge this: did you test it?

sebastienros:
Wait @DanielStolt, we should have it in 1.9.x. Otherwise it will only be shipped with 1.11

DanielStolt:
So what @sebastienros - it's not a bug, only a performance improvement for some scenarios. It should not go into the upcoming releases IMO.

sebastienros:
READ AGAIN SLOWLY

DanielStolt:
Makes sense - @carlwoodhouse sorry, can you retarget to 1.9.x?

carlwoodhouse:
just leaving the office but yeah will do it this evening

Issue #6533, opened 2016-03-06
There was an error installing the requested package.


geertdoornbos:
Probably related to #6485 

Repro:

Fresh 1.10 setup
Install module http://gallery.orchardproject.net/Packages/Orchard.Module.SH.Robots

Error: 
"There was an error installing the requested package." + some text about folder permissions

Quick fix
Click 'download' instead of 'install' to get the module. Then install local module from /Admin/Modules

Cause:
The package cannot be found. PackageInstaller.cs, line 55

```
var package = packageRepository.FindPackage(packageId, packageVersion);
```

Result:

```
            if (package == null) {
                throw new ArgumentException(T("The specified package could not be found, id:{0} version:{1}", packageId, String.IsNullOrEmpty(version) ? T("No version").Text : version).Text);
            }
```

This throws an ArgumentException which will be caught as an Exception, giving the generic error message. Also the message localization is wrong; will never translate correctly.

Fix:

```
            if (package == null)
            {
                var message = String.Format(T("The specified package could not be found, id:{0} version:{1}").Text, packageId, String.IsNullOrEmpty(version) ? T("No version").Text : version);
                throw new OrchardException(new LocalizedString(message));
            }
```

Question remains why the package cannot be found, but can be downloaded from the NuGet feed.

geertdoornbos:
Issue is fixed by in the [Gallery module](https://github.com/OrchardCMS/Orchard.Gallery/issues/19)

Calling client code should still be fixed. PR to which branch? dev, master, 1.9.x, 1.10?

sebastienros:
The issue was on the server code, any client should work fine now. So it should work with any Orchard version.

sebastienros:
Closing the issue for now unless you can repro it since I updated the gallery.

geertdoornbos:
Sebastien, there still is an issue on the client side. The server side failed with package == null. The code looks like it's handling the error, but it doesn't.

BenedekFarkas:
@sebastienros repro: Modify a package install URL under "/Packaging/Gallery/Modules" so that the package ID is invalid. Without change, the exception message will not be displayed as it's caught by `PackageManager.DoInstall()` and wrapped in an `OrchardException`. As suggested by the PR, throwing an `OrchardException` (instead of `ArgumentException`) will surface the original message, so the user will get a better understanding of what's happened.

PR #6562, opened 2016-03-10
#6533 There was an error installing the requested package

geertdoornbos:
Package installer error handling when package not found.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Packaging/Services/PackageInstaller.cs
+++ b/src/Orchard.Web/Modules/Orchard.Packaging/Services/PackageInstaller.cs
@@ -53,8 +53,7 @@ namespace Orchard.Packaging.Services {
             // gets an IPackage instance from the repository
             var packageVersion = String.IsNullOrEmpty(version) ? null : new Version(version);
             var package = packageRepository.FindPackage(packageId, packageVersion);
-            if (package == null)
-            {
+            if (package == null) {
                 var message = String.Format(T("The specified package could not be found, id:{0} version:{1}").Text, packageId, String.IsNullOrEmpty(version) ? T("No version").Text : version);
                 throw new OrchardException(new LocalizedString(message));
             }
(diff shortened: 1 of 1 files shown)


sebastienros:
What are you trying to fix? It's still throwing an exception, just a different one.

sebastienros:
Also formatting is wrong (brace on the same line) and you should not use `String.Format` with `T`.

geertdoornbos:
Currently is wrong on this [line](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Packaging/Services/PackageInstaller.cs#L57): If the package (specified by the given packageid) is not found:
- The T string will never be translated, because the packageid is in the string to be translated. There is no translation for every package signature.
- An ArgumentException is thrown while there is no wrong argument passed. An ArgumentException is thrown when one of the arguments provided to a method is not valid. That's not the case here.
The exception is not handled, resulting in the intended message (the formatted string). Exception is catched [here](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Packaging/Services/PackageManager.cs#L49) resulting in a message that is not of much use 'there was an error installing the requested package' + some permission tips.

With the fix, the string is translated as intended, the exception is catched [here](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Packaging/Services/PackageManager.cs#L46) resulting in a meaningful (as intended) message on the admin panel.

Can you elaborate (on gitter) on the issue with T inside String.Format? The T string calls the .Text property (resulting in the correct string that is used for translation) before it is used in String.Format.

sebastienros:
Do you know how I can simulate the issue, meaning how can I try to download a package which is not found?

BenedekFarkas:
@sebastienros repro: Modify a package install URL under "/Packaging/Gallery/Modules" so that the package ID is invalid. Without change, the exception message will not be displayed as it's caught by `PackageManager.DoInstall()` and wrapped in an `OrchardException`. As suggested by the PR, throwing an `OrchardException` (instead of `ArgumentException`) will surface the original message, so the user will get a better understanding of what's happened.

PR #6566, opened 2016-03-10
Output cache route config fixes #6523

### LINKED ISSUE #6523: Outputcache still checks cache even if a route is configured as not to cache

carlwoodhouse:
retargeting of PR #6557 to 1.9.x

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -71,6 +71,7 @@ namespace Orchard.OutputCache.Filters {
 
         // State.
         private CacheSettings _cacheSettings;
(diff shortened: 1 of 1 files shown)


sebastienros:
@dcinzona has already signed-off the change in prod, and code-reviewed by Daniel.

dcinzona:
Confirmed with breakpoints.  Routes that are not configured are returned without checking the cache.

PR #6569, opened 2016-03-10
#6544: NRE with containable part & Orchard.Lists - Replace #6545 to target 1.9.x

### LINKED ISSUE #6544: NRE with containable part in Orchard.Lists

jtkech:
Fixes #6544
Replace #6545 to target 1.9.x
I also fix a little 404 not found because nprogress.min.css doesn't exist.


sebastienros:
Code looks good. Waiting for someone to try it and confirm it works.

BenedekFarkas:
Tested and works!

Issue #6572, opened 2016-03-11
1.9.3 - Multiple TaxonomyFields with Autocomplete enabled causes terms to be added to empty fields on save


j3ffb:
Repo:

1. Create a content type with 2 taxonomy fields (pointing to the same or different taxonomy's), both with autocomplete enabled.
2. Create a new item of this content type, entering terms in one of the fields and leaving the other one empty.
3. Edit the same item again and just publish.

Expected result: The item is saved with the same terms, one field should be empty.
Actual results: Both fields are populated with the same terms.

j3ffb:
Looked into this and I believe the issue is because when autocomplete is enabled it does not store all terms on the client, they are fetched via the service, for performance. On save all terms for all fields on the item are added to the model, 'TryUpdateModel' does not then set these to `checked = false` as the terms are not in the posted state.

I believe the fix is to have `GetAppliedTerms(part)` also pass the field, so that terms do not bleed across fields.

I will submit a pull request for the fix.

BenedekFarkas:
I followed the steps you described, but couldn't reproduce the issue.

BenedekFarkas:
Update: repro'd on 1.10, PR merged in.

PR #6573, opened 2016-03-11
Issue #6572: Add only existing terms for the field.

j3ffb:
Fix for #6572


sebastienros:
GTG from a code perspective

BenedekFarkas:
Tested the change for both branches.

PR #6610, opened 2016-03-19
Media Library: Add button to select all items, fixes #6518

### LINKED ISSUE #6518: Media Library: Add button to select all items

MobesE46:
Fixes #6518 - New button to select all images in media library folder.

Supersedes PR #6519 to target 1.10.x

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library.js
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Scripts/media-library.js
@@ -509,7 +509,7 @@ $(function () {
         	pickAndClose();
         });
 
-        $("#media-library-main-selection-select > .button-select-all").on('click', function () {
+        $("#media-library-toolbar > .button-select-all").on('click', function () {
             viewModel.selectAll();
         });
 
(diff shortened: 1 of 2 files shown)


sebastienros:
GTG once someone has confirmed it works and doesn't break existing features.

MobesE46:
Here's the new button:
image

After clicking button:
image

sebastienros:
What I see from the code is that `Select` will actually close the window and select items, while `Select All` will just highlight the items. Is that intentional? It would mean to actually select all items you need to click `Select All`, then `Select`. So maybe this button should actually be in the toolbar, so that there is no confusion with the `Select` action of the picker.

MobesE46:
Ok, I moved the button to the Toolbar

image

BenedekFarkas:
Confirmed as working, but I'll push some changes/improvements.

PR #6613, opened 2016-03-20
#6530: [1.10.x] dyn form field default value tokenization

### LINKED ISSUE #6237: token in text field element can not work after db564d1670854459b176459b432833202f98e157 in dev brach
i used both of #{User.Email} and {User.Email},
then publish.
result show this field as {User.Email} in admin and front end (not real email)

### LINKED ISSUE #6530: [1.9? - 1.10] Tokens not working in Dynamic Forms
Create a ContentType with a text field - with either a token as a default value or by setting the property value and binding to a Dynamic Form element within a layout never evaluates, in either case.
e.g. {User.Name} or #{User.Name} results in just the token text itself, not even empty.

My use case is a simple Contact Form binded to a ContactForm Type containing only 4 fields [which now binds correctly thanks to Sipke], one being a Hidden Field with a Value of 'any-token-you-select' results in the aforementioned.

Possibly related:
[#5576](https://github.com/OrchardCMS/Orchard/issues/5576)
[#5407](https://github.com/OrchardCMS/Orchard/issues/5407)
[#6497](https://github.com/OrchardCMS/Orchard/issues/6497) - jtkech's comment about tokenization
[#6237](https://github.com/OrchardCMS/Orchard/issues/6237)

jtkech:
Fixes #6530, fixes #6237 

#6530 is already closed, but for 1.10.x we still need 41e9052 that is only applied to dev.
#6237 is related to dev and still open, but i think it has been fixed after 41e9052.

Here, the copy of 41e9052 but for 1.10.x.

Best

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/EmailFieldElementDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/EmailFieldElementDriver.cs
@@ -77,9 +77,13 @@ namespace Orchard.DynamicForms.Drivers {
         }
 
         protected override void OnDisplaying(EmailField element, ElementDisplayingContext context) {
-            context.ElementShape.ProcessedName = _tokenizer.Replace(element.Name, context.GetTokenData());
-            context.ElementShape.ProcessedLabel = _tokenizer.Replace(element.Label, context.GetTokenData(), new ReplaceOptions { Encoding = ReplaceOptions.NoEncode });
-            context.ElementShape.ProcessedValue = element.RuntimeValue;
+            var tokenData = context.GetTokenData();
+            context.ElementShape.ProcessedName = _tokenizer.Replace(element.Name, tokenData);
+            context.ElementShape.ProcessedLabel = _tokenizer.Replace(element.Label, tokenData, new ReplaceOptions { Encoding = ReplaceOptions.NoEncode });
+
+            // Allow the initial value to be tokenized.
+            // If a value was posted, use that value instead (without tokenizing it).
+            context.ElementShape.ProcessedValue = element.PostedValue != null ? element.PostedValue : _tokenizer.Replace(element.RuntimeValue, tokenData, new ReplaceOptions { Encoding = ReplaceOptions.NoEncode });
         }
     }
 }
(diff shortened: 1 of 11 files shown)


sebastienros:
Please check your commits or squash them, one of the commits is a merge.

jtkech:
@sebastienros 

Yes, i've seen this and i will re-try. But the merge commit was when i've synchronized my fork with the new 1.10.x branche. Then, when i squash a PR, there is always this merge.

Maybe i need to re-start with a new fork. I will see.

Best

jtkech:
@sebastienros 

Squash done

jtkech:
@sebastienros 

Not urgent for me but just for infos if important for those who want to use dyn form tokens.
And because it is ready and an exact copy of what is done on dev.
And it's squashed with the right message ;)

Best

sebastienros:
I trust the changes, I just need someone to confirm it works as expected and doesn't break obvious scenarios. It's for 1.10.x after all.

sebastienros:
Can you confirm the intent of this PR is just to copy what we already have in dev?

jtkech:
Yes

jtkech:
@sebastienros 

Just as a reminder because this one is just a copy of what we already have in dev.
I am never 100% sure of everithing, but i think this one is safe.

Best.

jtkech:
@sebastienros is it worth to solve conflicts here? Or do i have to close this one.

sebastienros:
Sipke and I think that this has already been fixed, or isn't it in 1.10.x ?

jtkech:
It is not in 1.10.x.

If i have to do something here, i would need to resolve the conflict for the hidden field where a null check has been added on `element.RuntimeValue`. So, maybe i have to do the same thing on other fields, or maybe specific to this hidden field ...

And i have to check again if some others updates has been done here on the dev branch.

Best

jtkech:
@sebastienros at least i solved conflicts.

I didn't re-apply the null check of `element.RuntimeValue` in the hidden field because it is now used in `_tokenizer.Replace()` which starts by doing itself this null check.

Best.

jtkech:
And for the hidden field because the value was not tokenized in 1.10.x, the `tokenized` class has been removed. But also in the dev branch (where the value is tokenized) after the last merge of 1.10.x. So, here it is re-introduced and normally it will be in the dev branch after the next merge of 1.10.x.

Issue #6550, opened 2016-03-08
Required fields should not have client side required attribute if a default value is set


sebastienros:
If a default value is provided, and the field is left empty, we can't submit the page as it will be prevented by the browser. We should rely on server side validation in this case instead.

jtkech:
Just seen this issue, i agree.

That's why here #6535 i said `Anyway, when Required the field can't be left empy because there is client validation...`. And i didn't implement default value when `Required`.

So here we also have to update some razor views and revisit some field handlers. I could do a PR for this but not immediately because of other works, but if it can wait.

Best.

jtkech:
@sebastienros 

Just seen the recent merge plan. Maybe these series of PRs need to be re-targeted. If so, let me know in which branch.

I've done them separately because there are many use cases to test. And each field has its own singularities. So, maybe better to validate them step by step and independently.

Best

PR #6616, opened 2016-03-20
#6550: [1.10.x] no required on client side if default value is set (TextField)

jtkech:
Fixes partially #6550.

Re-done to target 1.10.x. Other fields will come, i'll try to group them but i need to test again all options and each field has its special cases.

- If required and a default value is set, then no required on client side, but still with the required class to have the red star, and still required on server side.

- **Special case**: The default value is tokenized in a 2nd driver and the result can be empty. So, the required checking is also done in the 2nd driver. Because, if the default has a value, it has not been done in the 1st driver.

Best


jtkech:
I lost my changes after a bad squash / revert. I'll re-do it asap

jtkech:
@sebastienros Added specflows.

jtkech:
@sebastienros 

Not urgent at all for me, it's just for infos because it's the last one of the serie where i've added specflows.

Note: Other fields are already commited to 1.10.x (and recently merged in dev).

Best.

sebastienros:
I am merging this one, but the goal of squash is to get a single commit, not an extra one named `squash` ;)

jtkech:
Ok, sorry.

PR #6618, opened 2016-03-20
#6550: [1.10.x] no required on client side if default value is set (NumericField)

jtkech:
Fixes partially #6550.

- If required and a default value: no required on client, still with the red star, and still required on server.
- If a default value is used, control and conversion are done by the same code, as for an user input.
- When a default value is a valid number but not validated (limits / scale), it's value is seen on the postback view. This by using in the view a `Value` attribute. Let me know if you think it's ok.
- But, as before, if the default value is not a valid number, there is no direct way to show it on the postback view. So, i've adapted the error message: `{0} or its default value is an invalid number`.

**Note**: The Numeric.Feature spec pass.

Best


jtkech:
@sebastienros 

Specflows updated for the numeric field.
Tomorrow, i'll do it from scratch for the text field.

Best.

PR #6619, opened 2016-03-21
#6550: [1.10.x] no client required if default set (input + link + date fields)

jtkech:
Fixes partially #6550. (todo: enum + boolean fields)

**DateTimeField** (many options to check here)

- There is no client required attribute because of the DateTimeEditor.
- As before, if required and one field is missing, editor data are not parsed.
- But, now, before adding an error, we try to use the default value.

**Special case:** When both Date and Time are required, we consider that both input fields are required. But we use the default value only if all required fields are empty. If not, there is an error message as before, input fields need to be completed or completely cleared to use the default.

Note: The DateTime.Feature spec pass.

**InputField**

- Here, no need to change the InputFieldDriver.
- Just remove client required attribute (not the class) if a default value is set.
- And correct an hint text in the input field settings editor

**LinkField**

- Here also, no need to change the LinkFieldDriver.
- But here, the client required attribute (not the class) was never used.
- So, this attribute is added to the url and link text inputs, unless a default value is set.

Best

### DIFF
--- a/src/Orchard.Specs/DateTime.feature
+++ b/src/Orchard.Specs/DateTime.feature
@@ -89,9 +89,9 @@ Scenario: Creating and using Date fields
     # Required & Date and Time
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
-            | name                                    | value       |
-            | Fields[0].DateTimeFieldSettings.Display | DateAndTime |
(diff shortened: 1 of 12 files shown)


jtkech:
@sebastienros 

Added specflows for input, link and date fields.

PR #6626, opened 2016-03-22
#6550: [1.10.x] no client required if default set (enum + boolean fields)

### LINKED ISSUE #6550: Required fields should not have client side required attribute if a default value is set

jtkech:
Fixes #6550. (with other related PRs)

Note: i didn't have a look on other custom fields as TaxonomyField... Only those in Orchard.Fields.

**EnumerationField**

Unlike other fields, here the default value was not used in the Editor method that updates the model but in the other one, this on first creation to have a predefined value. Let me know if we need this on other fields, or if it's case by case. Right now, i left it.

- Anyway, i also use the default value in the editor method that update the model, as with other fields. Then, when another submission happens with no selected option, we can re-use the default value (even it has been changed in the settings). And this before the required checking on server side.

- When using the default value, we retain only its valid options (according to the options settings).

- I've added the hint message `If no option is selected then the default value will be used.`.

- I've removed the client required attributes when checkboxes are used, because, if required and no default value is set, `you are forced to check all checkboxes on client side`.

Enumeration feature spec pass.

**BooleanField**

- Here, there is no client required attribute. Maybe because, when required, with the controls you can use there is always a valid submitted value. This may change but I let it as before.
- However, the required checking (here if not `Optional`) is still done on server side. So, i've just added the ability to use the default value if not null (here not `Neutral`).
- And i've adapted the related feature specflow not to use the default value while it was expecting a `Mandatory` message. This by setting the default value to `Neutral`.

Best

### DIFF
--- a/src/Orchard.Specs/Boolean.feature
+++ b/src/Orchard.Specs/Boolean.feature
@@ -65,8 +65,9 @@ Scenario: Creating and using Boolean fields
     # The value should be required
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
-            | name                                    | value |
-            | Fields[0].BooleanFieldSettings.Optional | false |
+            | name                                        | value    |
+            | Fields[0].BooleanFieldSettings.Optional     | false    |
+            | Fields[0].BooleanFieldSettings.DefaultValue | Neutral  |
         And I fill in 
             | name                                       | value  |
             | Fields[0].BooleanFieldSettings.NotSetLabel | May be |
--- a/src/Orchard.Specs/Boolean.feature
+++ b/src/Orchard.Specs/Boolean.feature
@@ -53,7 +53,7 @@ Scenario: Creating and using Boolean fields
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Check if the event is active"
     
-    # The default value should be selected
+    # The default value should be used on creation
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value |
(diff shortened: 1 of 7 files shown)


sebastienros:
For all your PRs regarding the required property. I would like someone to try it before merging it, OR you can add a specflow test for that and I won't ask for it.
Because it's 1.10.x it has to work correctly (not that it's not broken already).

The test should be easy to create by copy pasting the existing ones and defining required. You can also check that the HTML doesn't contain `required` when it applies.

jtkech:
@sebastienros Specflows updated for this one.

- I've added the missing tests around the default value.
- Here, the `html required attribute` is only used by the `Enumeration Field` in the `Listbox Mode` where all options can be unselected (useless with dropdown list or radio buttons, and can't be applied when checkboxes are used, because you would force the user to check all options).
- So, in the `Listbox Mode`, i've added the test to check that, when required, the html required attribute is only used if no default value is set.

Best

sebastienros:
That's great! What about the text field, do we need a test for required values? It's also checking on client side and was showing the issue of overwriting a user value.

jtkech:
@sebastienros 

Here, i've updated the specs only for enum and boolean fields. I'm currently working on #6619 to do the same for input, link and date fields. Then, i will do it for the numeric field related to #6618.

Finally, you're right there is the text field related to this PR #6616. I didn't start with this one because it doesn't already have a specflow test. Maybe because it is in the core and was not part of the Orchard.Fields.

For text field, do we need to create a specflow from scratch?

Best.

sebastienros:
For text that'd be awesome. Just because we have default values now and that would prevent any new regression. I will merge the PRs as they get unit tests, thanks.

Issue #4538, opened 2014-05-22
Clone default action should process just as create and care about the Draftable contentitem setting


orchardbot:
@csurieux created:
https://orchard.codeplex.com/workitem/20709

        public ActionResult  Clone(int id, string returnUrl) {
            return Clone(id, returnUrl, contentItem => {
                if (!contentItem.Has<IPublishingControlAspect>() && !contentItem.TypeDefinition.Settings.GetModel<ContentTypeSettings>().Draftable)
                    _contentManager.Publish(contentItem);
            });
        }

public ActionResult Clone(int id, string returnUrl,  Action<ContentItem> conditionallyPublish) {
            var contentItem = _contentManager.GetLatest(id);

            if (contentItem == null)
                return HttpNotFound();

            if (!Services.Authorizer.Authorize(Permissions.EditContent, contentItem, T("Couldn't clone content")))
                return new HttpUnauthorizedResult();
            ContentItem newContentItem = null;
            try {
                newContentItem = Services.ContentManager.Clone(contentItem);
            }
            catch (InvalidOperationException) {
                Services.Notifier.Warning(T("Could not clone the content item."));
                return this.RedirectLocal(returnUrl, () => RedirectToAction("List"));
            }
             conditionallyPublish(newContentItem);
            Services.Notifier.Information(T("Successfully cloned. The clone was saved as a draft."));

            return this.RedirectLocal(returnUrl, () => RedirectToAction("List"));
        }

PR #6630, opened 2016-03-22
Content cloning and translation

TFleury:
Initial PR : #6540
This PR adds new clone handling to replace export/import logic, and also allow translating a clone of the master content item.

It fixes : #5535, #5089, #4983, #4970, #4538, #5031


Skrypt:
Tested and works. I need to review if we could use ILocalizableAspect in some places.

Skrypt:
Going forward, I've set every content types to be listable in Orchard to see if it could uncover missing use cases. 

Here is what can/cannot be cloned : 

- [X] Custom Content Type with all types of fields
- [X] Page
- [X] Layout
- [X] Taxonomies
  - [ ] Taxonomies with it's terms
- [X] Blog
  - [ ] Description, other configurations
- [X] Blog Post
  - [ ] Description, other configurations
- [ ] Comment
  - The clone link is there but it opens a blank form
- [ ] Query
  - _Queries are not listable even if the option is checked_
- [X] Projection
  - [ ] Show on a menu
  - [ ] Show on admin menu
- [ ] Widget
  - HTML, Text ... none of them can be cloned. You can set them as listable but they will only display their titles
- [X] Menu
  - [ ] Menu with it's content
- [X] Media Profiles
  - [ ] Media Profiles with filters
- [ ] Workflows

You get the idea... there's probably more but I think this PR replaces what was already there and it does'nt break anything.

sebastienros:
Merge conflict ... sorry

TFleury:
Rebased

sebastienros:
And a squash would be appreciated too.

@Skrypt how did you tests go with this change?

Skrypt:
Give me some time I'll test this soon after I get my cup of coffee.

src/Orchard.Web/Modules/Orchard.Fields/Drivers/BooleanFieldDriver.cs
```diff
@@ -35,7 +35,7 @@ public class BooleanFieldDriver : ContentFieldDriver<BooleanField> {
 
         protected override DriverResult Editor(ContentPart part, BooleanField field, dynamic shapeHelper) {
             // if the content item is new, assign the default value
-            if(!part.HasDraft() && !part.HasPublished()) {
+            if(!field.Value.HasValue) {
```
jtkech: @TFleury 

Working on field drivers, i noticed that with the boolean field you can use a neutral option in settings, e.g `Maybe`. So, if you select `Maybe`, then `!field.Value.HasValue` will be true without meaning it's a new content item. This because here null is considered as a valid value.

If, for another reason, you still need to change this test, i think we can find the right one.

Best

jtkech:
About my line note on BooleanFieldDriver.cs, without disturbing you, if you can tell me why you have to change this test, i could try to find a compromise.

Best.

src/Orchard.Web/Modules/Orchard.Fields/Drivers/BooleanFieldDriver.cs
TFleury: You are right, need to find another way to initialize whith default value without overriding the cloned one.

Skrypt: This works. Let me know if you like "or not".
```
        protected override DriverResult Editor(ContentPart part, BooleanField field, dynamic shapeHelper) {

            string action = HttpContext.Current.Request.RequestContext.RouteData.Values["action"].ToString().ToLowerInvariant();
            
            // if the content item is new, assign the default value
            if (action == "create") {
                var settings = field.PartFieldDefinition.Settings.GetModel<BooleanFieldSettings>();
                field.Value = settings.DefaultValue;
            } else if (action == "Clone") {
                //check for cloned content item
                if (!field.Value.HasValue) {
                    var settings = field.PartFieldDefinition.Settings.GetModel<BooleanFieldSettings>();
                    field.Value = settings.DefaultValue;
                }
            }

            return ContentShape("Fields_Boolean_Edit", GetDifferentiator(field, part),
                () => shapeHelper.EditorTemplate(TemplateName: TemplateName, Model: field, Prefix: GetPrefix(field, part)));
        }
```

Skrypt:
@sebastienros All the tests I've made have the same results.

src/Orchard.Web/Modules/Orchard.Fields/Drivers/BooleanFieldDriver.cs
jtkech: Didn't seen the details of cloning, but, as i understand, we can't use the previous test because we work on a content item that is being cloning, and therefore has a draft or is published...

So, after a 1st look, and because i think this method is called only during a regular web request, your code looks good for me. Maybe we need a null check on the action string?

Best

Skrypt: The main problem here is that there is no way to know if the Boolean Field has been assigned a null value "explicitly"... since a null value is a null ...

jtkech: You're right, but i was talking about a null check on the `action` string. But, sorry, it was a reflex, it's iseless here ;)

Skrypt:
It is GTG if we fix the small bug with the boolean field.

src/Orchard.Web/Modules/Orchard.Fields/Drivers/BooleanFieldDriver.cs
Skrypt: We need to adapt my code to use IHttpContextAccessor instead of HttpContext wich is pre .NET 3.5

jtkech: You're right, by injecting an `IHttpContextAccessor` and using `_httpContextAccessor.Current()`. Then, one rule is to also use this mvc extension `if (!httpContext.IsBackgroundContext())` when it doesn't make sense to do something, e.g in a background task. But i think that the `Editor()` method will never be executed in the background (unlike the `Display()` method). Anyway, in the background an empty (but not null) route data would be returned.

So, not tested but i propose something like the following:
Note: maybe good to create a route extension `GetActionName()` as already done for `GetAreaName()`.

            ...
            var routeData = _httpContextAccessor.Current().Request.RequestContext.RouteData;

            object action;
            if (routeData.Values.TryGetValue("action", out action)) {
                var actionString = action as string;

                // if the content item is new, assign the default value
                if (!String.IsNullOrEmpty(actionString) && actionString.Equals("create", StringComparison.OrdinalIgnoreCase)) {
                    var settings = field.PartFieldDefinition.Settings.GetModel<BooleanFieldSettings>();
                    field.Value = settings.DefaultValue;
                }
            }
            ...

Skrypt:
I think we should also provide unit tests on this one. At least one for making sure that the identityPart generates a new GUID.

TFleury:
I think there is an issue with all fields (not bound to cloning) :
The default value of the field is set in `Editor` method of the field's driver.
If you create a content item without editing it, or if you add a field to a ContentType with existing content items, they won't take the default value until they are edited.
Default value should be set in a `ContentHandler` on Initializing handler using a `LazyField`. On Loading, the "default value" getter will be replaced with a "stored value" getter (if there is one).

Do you agree with that ?

Skrypt:
I agree. Though, it is not a ship blocker for this PR. If we need to implement such thing, let's propose it in another issue.

TFleury:
But I think the PR won't be accepted with a condition on Action name...

Skrypt:
This is a grey zone. Implement it and let's summon @sebastienros

sebastienros:
I assume either a default value should be implemented for Editor only, or in a Content Handler when an item is created.

sebastienros:
Well, it depends what we think of the default value. Like the first value that is shown to the user, but the user could delete it. Or the value to assign when there is no value, on save.

Skrypt:
Not everyone will want their TextboxField to have forced default values saved when an empty string is submitted by the form. I think that the default values should be only proposed in that scenario. 

Though, if it comes from a background task it should force default values. So IMO it's kind of both.

jtkech:
Indeed, as @sebastienros said there could be different meanings around default values. Initial value on creation, predefined value on editing, default value on submission...

But, in our context, note that currently most of the fields don't use a default value on first editing / creation. As i remember, only the enum and bool fields use it. And the boolean field is a special case where null can be a valid value (if the neutral option is set).

I agree with what @TFleury said about content handler and lazy fields. Or, maybe, by adding a new field driver "event" for this purpose, as done for cloning, through `ContentFieldDriverCoordinator`...

That said, not tested but isn't there any solution by using, in the `BooleanFieldDriver`, the new `Cloning() / Cloned()` "events" to set a private variable that we can check & maybe reset in the 1st `Editor()` method?

Best.

Skrypt:
Here is the result of my search on how Drupal does it.
https://www.drupal.org/project/replicate

jtkech:
@sebastienros @TFleury @Skrypt 

Just tried, great feature! Didn't test everithing because Jasmin did it.

So, if you can check this simple solution in `BooleanFieldDriver.cs` that seems to work for me:

    ...
    private bool _isCloning = false; // new
    ....
        protected override DriverResult Editor(ContentPart part...) {
            ....
            // if (!part.HasDraft() && !part.HasPublished()) { // before
            if (!part.HasDraft() && !part.HasPublished() && !_isCloning) { // and now
            ....
        }
        ....
        protected override void Cloning(ContentPart part, BooleanField originalField....) {
            cloneField.Value = originalField.Value;
            _isCloning = true; // new
        }

Best.

vesko-k:
2 questions
1. When cloning what happens with the owner of the cloned item?
2. When cloning content item, if regeneration of the Identifier of IdentityPart should be done on the source or result item?

Skrypt:
Right now the clone feature opens up a "Create new" form and sets the owner to the current user. The new content item should have then a new IdentityPart since it has the same logic than "Create". So it is done on the created item. Though we would need to assert that if this is done by a background task that it also changes the current owner properly.

TFleury:
@sebastienros which is the best (acceptable) way to solve our problem with default value handling ?
- Condition on action name
- Setting a variable in Driver
- Using ContentHandler and LazyField

Skrypt:
From a discussion with @sebastienros 
> 
This PR depends on the behavior of the "default values". We are not sure about what behavior we want for "default values". So we need to fix the "default values" issue first.
> 

I think Fields should mimic what we can do with Fields in SQL Server. So, my opinion is that default values should only be proposed.

Fields can be attached to a content type even when this content type has content items. If we set a default value to this new field, we don't need to batch update all the content items since the default value can be retrieved by the Field properties. Also, if we export/import those content items, and we decide to ommit this field in the import, I don't think we should then set this field to it's default value.

to continue ...

jtkech:
Not sure that all we can discuss around default values should block this PR. I thought cloning was not directly concerned by this, and, here, we just needed to fix a specific case with the boolean field editor.

About default values we have to separate contexts where different values can be used or not. So, here, for clarity, just more infos on the current implementation related to **default form values for field editing**.

1. `For most of the fields, the default value (if set) is used on submission if the field is left empty`. I think this one is good, consistent with e.g the auto generated autoroute slug if left empty. Note: i think it's good to separate default values for editing and for other contexts.

2. `For some fields, the default value (here the same) is also used on 1st edition as a proposed value`. I think it has been introduced only to adapt some fields to **1.** because some controls always send a value (can't be left empty). So, the need to preset and submit the default if no user action.

- So, when saying `we don't want to have forced default values when an empty string is submitted` and `default values should only be proposed`, we would need to remove the main point **1.** and implement **2.** for all fields. For me it's ok as it is but i've no closed opinion. Maybe we can have **1.** and **2.** with additional options to use them alone or both.

Anyway, regarding default values, the only problem i see here is with the boolean field because null can be a valid value (Neutral), so, when **2.** is used we can't detect if it has no value as in a 1st edition. Maybe we can first find a compromise for this (see above the list suggested by Thierry).

Best.

Skrypt:
Distinction here also that the Autoroute is a Part ... wich is different than a Field.

TFleury:
I've rebased the PR and added a new commit that solves our issue with BooleanField's default value without changing the global behavior.

For consistency, we could move all fields default value setting to ContentHandlers, but it's out of scope and not blocking for the PR.

Let me know if it's GTG so that I squash it all.

sebastienros:
We'll take the default fields PR from JT first, and you don't need to squash github does that for us now ;)

jtkech:
@sebastienros 

Before the default value was mainly used on update editor and with tokens that may be based on an already created item. So, maybe, lazy fields on initializing was not really applicable.

But now we will use, **for all the fields**, the default value as a proposed value, editor only on a new item, and without tokens. So, i think it's the best solution, not only for the boolean but for all the fields.

The only difference is that the default will be used on initializing, even without editing. But, if i recall correctly, SQL also use the default value for an INSERT without edition (at least by setting an option).

So, if you agree, i propose to implement in my PR the lazy fields and handlers for all the fields (in Orchard.Fields). And not to mix things, directly use what's done here by @TFleury for the boolean field.

I'll do my PR asap.

Best.

sebastienros:
@jtkech please ping me on skype if you want to talk about it. I am reading the whole thread again and now I am confused with why Boolean Fields are an issue, with or without default values. Let's assume the default values in the settings are just for the form on the creation. Why should we care for clones, we just copy whatever value is already in the original content item.

Also I thing you should not do the content handler default value feature, let's keep it to the form right now.

sebastienros:
@TFleury JT and I just talked. You will need to remove any default value handling (boolean field handler and lazy field), and just rely on what JT is doing on his default values PR. Meaning that when cloning a content item from the UI, the _empty_ values might get the default from the setting in the editor. We don't want to create complex handler just for this specific case.
What we ask you is to add a warning when the Clone action is executed to diplsay this message: "Some values in this form might have be initialized with their default if the cloned item ones where not provided."

TFleury:
@sebastienros I agree, using `LazyField` and `ContentHandler` is a bit overcomplicated regarding what it's intended to do, and setting the default value in the editor is out of scope of cloning.
But from what I understand from @jtkech's PR, default value will be proposed in the editor when creating (editing a new) content. A cloned content is a new one, so all field values will be overridden with default ones, regardless of if it was defined or not in the original content.
We must leave a door open for keeping cloned values.

jtkech:
@sebastienros @TFleury 

Oh yes you're right, maybe it was too late in the night, indeed we still need to know it is cloning before editing not to overidde all values.

So, right now, just an idea. Because the issue is when editing a cloned item, what do you think about creating a draft version before editing? Not tried, so not sure it could work.

Best.

TFleury:
@jtkech maybe we can use what you proposed 15 days ago
```...
private bool _isCloning = false; // new
....
    protected override DriverResult Editor(ContentPart part...) {
        ....
        if (part.IsNew() && !_isCloning) {
            // Initialize with defaults
        }
        ....
    }
    ....
    protected override void Cloning(ContentPart part, BooleanField originalField....) {
        cloneField.Value = originalField.Value;
        _isCloning = true; // new
    }```

jtkech:
@TFleury 

The only thing is that this variable can be shared by multiple fields using the same driver instance. But i'm not against this.

Just tried adding this before editing a cloned item

    _contentManager.Create(cloneContentItem, VersionOptions.Draft);

Seems to work

I need to leave but i'll work on it this night. I think @sebastienros will give us the best compromise.

Best

sebastienros:
I like the Draft idea, for draftable items, for non-draftable then a new item would be created, just not published, and this way even the empty values would remain empty.
I don't remember the current behavior of the Clone link, this would mean that now it would create a new content item, save it, then go to its editor, with a Success notification mentioning the new item has been cloned.

jtkech:
@sebastienros @TFleury 

So, the following seems to work for me. Only done a few tests with draftable and non draftable types. Maybe need more checking, a notification message...

        public ActionResult Clone(int id) {
            var originalContentItem = _contentManager.GetLatest(id);

            var cloneContentItem = _contentManager.Clone(originalContentItem);

            if (!Services.Authorizer.Authorize(Permissions.EditContent, cloneContentItem, T("Cannot create content")))

            _contentManager.Create(cloneContentItem, VersionOptions.Draft);

            var adminRouteValues = _contentManager.GetItemMetadata(cloneContentItem).AdminRouteValues;
            return RedirectToRoute(adminRouteValues);
        }

Best.

sebastienros:
'ViewPermission' on the original content is required, 'CreatePermission' on the type too.
It needs a notification (but I assume not part of your tests).
Also needs to be a POST.

jtkech:
@sebastienros 

Test done with `[HttpPost]`, seems to work as i've tested. This by re-using, in `Parts.Contents.Clone.SummaryAdmin.cshtml`, the previous `@Url.Action("Clone", "Admin", new {...`. It also seems that we would not need anymore the new `Clone.cshtml`.

I let @TFleury add the right permissions and notifications, maybe do more tests to see if it works in all scenarios, and if he agree with this kind of solution...

Best

TFleury:
Thank you for your tests, I will update the PR this way.

jtkech:
You're welcome, i'll finish my PR asap. Sorry, i think you will need to rebase yours a last time :)

sfmskywalker:
Hey guys, I'm particularly interested in this topic. Please let me know if there's anything I can do to help get this done.

sebastienros:
@sfmskywalker yes, you can test @jtkech 's PRs that are tagged for 1.10.1, then we can merge them, and @TFleury can rebase this branch on top.

TFleury:
Just updated my PR.
It should not conflict with #6720

sebastienros:
Will merge as someone says it works on dev.

Skrypt:
This PR waiting on me to test it ? :wink:

sebastienros:
Let's watch Orchard burning ...

Skrypt:
Oops ! it did not close the thousands issues related to it :smile:

MatteoPiovanelli-Laser:
In light of #7352 I pulled the latest commit related to cloning into a (local) branch off 1.10.x.
I added a few comments in that same commit asking for clarification on some choices in the code.
I hope some one can answer those. In the meantime, I will verify what is broken by the changes here, and a few modifications to the code from the commit to make it so it breaks nothing.

The commit is https://github.com/OrchardCMS/Orchard/commit/87a394854dda9a45b36d757b3d3ce135ac18d24d

/cc @TFleury @Skrypt @sebastienros

PR #6635, opened 2016-03-22
[1.10.x] Include jQuery-UI in the Orchard.Resources csproj and fix jQueryTimeEntry script loading

jbuck79:
jQuery-UI wasn't included in the solution, which prevents it from being copied to the output directory when building.

There was a typo in the resource declaration of the jQueryTimeEntry script which prevented the minified version from loading.


DanielStolt:
@jbuck79 Good change, but can you please retarget to `1.10.x`?

jbuck79:
Oops, I thought I did. I'll try again.

PR #6640, opened 2016-03-23
#6639 : Azure Media Storage don't recognizes its own URLs

TFleury:
Removes port when default in PublicUrl.
`GetStoragePath` recognizes endpoint and public hostnames.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/Media/AzureBlobStorageProvider.cs
+++ b/src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/Media/AzureBlobStorageProvider.cs
@@ -4,6 +4,7 @@ using Orchard.Azure.Services.Environment.Configuration;
 using Orchard.Environment.Configuration;
 using Orchard.Environment.Extensions;
 using Orchard.FileSystems.Media;
+using System;
 
 namespace Orchard.Azure.Services.FileSystems.Media {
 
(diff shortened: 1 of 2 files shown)


sebastienros:
@Piedone don't command on the commit, but on the PR.

src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/Media/AzureBlobStorageProvider.cs
```diff
@@ -64,8 +65,10 @@ public AzureBlobStorageProvider(string storageConnectionString, string container
         /// <returns>The corresponding local path.</returns>
         public string GetStoragePath(string url) {
             EnsureInitialized();
-            if (url.StartsWith(_absoluteRoot)) {
-                return HttpUtility.UrlDecode(url.Substring(Combine(_absoluteRoot, "/").Length));
+            var rootUri = new Uri(_absoluteRoot);
+            var uri = new Uri(url);
+            if((uri.Host == rootUri.Host || (!string.IsNullOrWhiteSpace(_publicHostName) && uri.Host == _publicHostName)) && uri.AbsolutePath.StartsWith(rootUri.AbsolutePath)) {
```
sebastienros: I have learned that if you don't rely on white spaces, you can use `String.IsNullOrEmpty` only as whitespace comparison is expensive.

Issue #6624, opened 2016-03-21
Built-in "Shape" Snippet should only be available when Layout Snippets Feature is enabled


sunkist:
Layout Snippets is a separate feature from Layouts.

However, the Layouts Toolbar shows a "Shape" Snippet when Layouts is enabled and Layout Snippets is disabled. This caused some momentary confusion today, because I was wondering why the Shape Snippet was being found by the Snippet Harvester and not my custom ones. I realized I needed to enable the Layout Snippets Feature, but thought I had because the Shape Snippet was being found.

It would be more consistent if the Shape Snippet was part of the Layout Snippets Feature and not displayed in the toolbar until the feature is enabled.

![shape-snippet](https://cloud.githubusercontent.com/assets/2044049/13936813/c7613a8e-ef96-11e5-85c6-2a72876d2e54.png)

sunkist:
Just a quick note... the idea here is to remove any confusion as to why custom snippets may not be displaying in the toolbar. Seeing the "Shape" Snippet there when Layout Snippets is not enabled adds to that confusion.

I use Snippets a lot and could see other options:

1. A disabled "Shape" Snippet when Layout Snippets is not enabled. This would be a handy reminder that Layout Snippets is not enabled.

2. The Shape Snippet could be enabled like it is now, but maybe a message indicator underneath it that let's the user know Layout Snippets is not enabled and to enable it to get custom snippets to show up in the Snippets Toolbar area.

3. Or, maybe if the Layout Snippets Feature is disabled and custom Snippets are detected in the theme, there could be an indicator in the toolbar that custom snippets have been detected but are not enabled because the feature is not enabled.

Lots of possibilities, but certainly this isn't a high priority item.

sebastienros:
I moved it to another section to avoid confusion.

PR #6645, opened 2016-03-24
Fixes #6644 : Codegen don't creates Themes/Modules in right Solution directory

### LINKED ISSUE #6644: Codegen don't creates Themes/Modules in right Solution directory
Branch master, version 1.10.0
When using codegen with `/IncludeInSolution:true` to create theme/module, it's created at the right place on file system, but it appears at the root of the solution in Visual Studio.

TFleury:
And makes AddToSolution a little more understandable.


sebastienros:
GTG once someone has ensured the command can be called two times in a row and it works as expected.

sebastienros:
Fixes #6624

sebastienros:
Fixes  #6644

PR #6679, opened 2016-03-30
#6659: [1.10.x] [dev] Taxonomy Tags Api issue

### LINKED ISSUE #6659: [dev] Taxonomy Term picker in admin
I'm getting "can't acces the admin" with the Taxonomy Term picker in the admin.
it throws an error in TagsController.cs
`	        if (!_authorizer.Authorize(StandardPermissions.AccessAdminPanel)) {
		        throw new UnauthorizedAccessException("Can't access the admin");
	        }`

jtkech:
Fixes #6659 

**Repro** (both in 1.10.x and dev)

- Use a virtual directory as `OrchardLocal` (or maybe a tenant prefix).
- Add a taxonomy field to a content type, and check the autocomplete option.
- Edit a content item of this type and try to add a term by using the autocomplete field.
- Then you get `can't acces the admin` exceptions.

**Analysis**

- In `admin-taxonomy-tags.js` there is an ajax call that targets `TagsController` (an `ApiController`) by using a relative url (not a full url) with an absolute virtual path `/api/taxonomies/tags`. So, It doesn't work if you use a virtual directory as `OrchardLocal` or, i think, if you use a tenant prefix.

**Fix**

- By initializing a `baseUrl` variable in the razor view and using it as an html data attribute. Then, using this data attribute in the script to construct the full url e.g  `http://localhost:30321/OrchardLoca/api/taxonomies/tags`.

**Note**

In place of using a baseUrl, i could resolve and use directly the full url by using an url helper to target the api controller. But i would need to apply #5823. If applied, using one of the following works:

    // This one because of the fixed priority issue.
    var fullApiUrl = Url.HttpRouteUrl("", new { area = "Orchard.Taxonomies", controller = "Tags" });

    // Or, maybe better, this one because of the new http route default name.
    var fullApiUrl = Url.HttpRouteUrl("Taxonomies.Api", new { controller = "Tags" });

Best.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyField.Autocomplete.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyField.Autocomplete.cshtml
@@ -3,7 +3,7 @@
 @using Orchard.Taxonomies.Models
 @using Orchard.Utility.Extensions;
 @using Orchard.Taxonomies.ViewModels;
-
+@using Orchard.Environment.Configuration
 
 @{
     if (Model.Settings.Autocomplete) {
@@ -18,6 +18,7 @@
     Script.Include("~/Themes/TheAdmin/scripts/admin.js").AtFoot();
     Script.Include("admin-taxonomy-expando.js").AtFoot();
 
+    var baseUrl = Href("~/");
 }
 @functions {
     bool IsTermDisabled(TermPart term) {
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyField.Autocomplete.cshtml
```diff
@@ -18,6 +18,7 @@
     Script.Include("~/Themes/TheAdmin/scripts/admin.js").AtFoot();
     Script.Include("admin-taxonomy-expando.js").AtFoot();
 
+    var baseUrl = Url.Content("~/") + WorkContext.Resolve<ShellSettings>().RequestUrlPrefix;
```
sebastienros: I think you can just do `@Href("~/")`

sebastienros:
I look at the other PR you mention, it looks like the route template for Apis is just missing the prefix, like we do in the default routes for standard controllers.

sebastienros:
This would work but it would be better to fix API routes globally, I will try to understand what is missing.

jtkech:
Indeed with `@Href("~/")`!
And ok, i'll re-try web api but in a tenant context (don't remember if i did it).

jtkech:
@sebastienros 

Indeed, it works with `Href("~/")`, and also in a tenant context. So, i've updated this PR.
And, indeed, it doesn't work in a tenant context with `Url.HttpRouteUrl()`. I'll check this.

So, it's OK for this PR, and i need to update this one #5823 related to web api.

jtkech:
@sebastienros 

**UPDATE:**

So, it works with `Href("~/")` (also in a tenant context). So, it's OK for this PR.

And, if #5823 is applied, it also works, even in a tenant context, with `Url.HttpRouteUrl()`.
In the previous comment i said the opposite, but it was because i didn't refresh my scripts.

So, it also works with web api (if #5823 is applied for the priority issue) and in a tenant context because `ShellRoute.GetVirtualPath()`, that take into account the prefix, is also used for httproute.

Best

sebastienros:
Thanks for all the effort, I think I understand what is happening and I decided to apply the patches directly as they made sense.
I am closing the PRs too. I really appreciate the effort you did put in it, it really helped.

PR #6682, opened 2016-03-30
Orchard.Resources: DateTimeEditor override

### LINKED ISSUE #6674: Orchard.Resources: DateTimeEditor does not render.
I have a painful problem. 
The ```DateTimeEditor.cshtml``` template in ```Orchard.Resources``` module does not override the corresponding template in core common module. So i see simple textbox instead of date picker.

I upgraded from ```1.9.x``` to ```1.10.x```. Orchard.Resources was enabled by default but i disabled Orchard.jQuery manually. 

Any help will be welcomed.

infofromca:
fix #6674


sebastienros:
This patch is already applied on 1.10.x

PR #6685, opened 2016-03-30
Adding missing dependency in Orchard.Resources

### LINKED ISSUE #6674: Orchard.Resources: DateTimeEditor does not render.

armanforghani:
Fixes #6674


sebastienros:
There was a merge issue with another commit I just did so I applied the change directly, thanks.

PR #6689, opened 2016-03-31
Fixes #6642: Media gallery not working with Azure Blob Storage and "M…

### LINKED ISSUE #6642: Media gallery not working with Azure Blob Storage and "ManageOwnMedia" permission
I reproduced this on 1.10.x branch today. Following are the steps to reproduce.

1. Configure Azure Media Storage on site
2. Create a new user with permissions: ManageOwnMedia and AccessAdminPanel

This user will be able to land onto the Media Gallery. This user can also create new Folder BUT cannot navigate to any sub-folder.

Following is the root cause of the issue:
We have following two extension methods in IMediaLibraryService.cs. The first method uses "\\\\" as directory separator and fails to check if the user is in the right directory or not. For root folder this value evaluates correctly since both the values mediaPath and rootPath are same but for sub-folders this gets evaluated wrongly and hence this method returns FALSE.

The second method "GetRootedFolderPath" also used Path.Combine two combine two paths.
Both of these functions should rely on IStorageProvider to combine the paths if required.

We already have a Combine method in IStorageProvider but for the first method we will have to introduce some new method in IStorageProvider to append appropriate path separator. This method will then be implemented in the two storage providers FileSystemStorageProvider as well as AzureFileSystem Provider with appropriate separators.

Please advise if this solution is appropriate.

```
        public static bool CanManageMediaFolder(this IMediaLibraryService service, string folderPath) {
            // The current user can manage a media if he has access to the whole hierarchy
            // or the media is under his personal storage folder.

            var rootMediaFolder = service.GetRootMediaFolder();
            if (rootMediaFolder == null) {
                return true;
            }

            var mediaPath = folderPath + "\\";
            var rootPath = rootMediaFolder.MediaPath + "\\";

            return mediaPath.StartsWith(rootPath, StringComparison.OrdinalIgnoreCase);
        }

        public static string GetRootedFolderPath(this IMediaLibraryService service, string folderPath) {
            var rootMediaFolder = service.GetRootMediaFolder();
            if (rootMediaFolder != null) {
                return Path.Combine(rootMediaFolder.MediaPath, folderPath ?? "");
            }

            return folderPath;
        }
```

hannan-azam:
Fixes #6642


sebastienros:
Is there a way to fix the bug without having to break the current interface? Can't the extension methods use the `Combine` method from the storage provider?

hannan-azam:
@sebastienros I looked into it and it seems that we can use the Combine method to achieve the goal.
I tried three options for the path2 parameter of the Combine method:
1. Empty string ""
2. Some pre-defined string e.g. "root"
3. Blank String with a Space " "

With first two options there were multiple issues and those cannot be used. However the third option seems to be working very nicely. I did a quick test with both storage providers and it seemed to work with both. I will do some more tests on my end and will submit a new PR for this.

```
_storageProvider.Combine(folderPath, " ").Trim();
```

sebastienros:
Are you creating a new PR then?

hannan-azam:
Yes I will create a new PR with minimal changes required to fix this bug.

PR #6690, opened 2016-03-31
Fixes #6686: Prevent double-click on Publish/Save

### LINKED ISSUE #6686: Prevent double-click on Publish/Save
There is already some database constraint preventing duplicate content item version records, but some UI would prevent an exception to be thrown in case it happens.

hannan-azam:
Fixes #6686


sebastienros:
Awesome! Anyone wants to try the feature to ensure it's not breaking any obvious scenario?

donaldboulton:
Combine it with lines 86 - 99

hannan-azam:
@sebastienros , @donaldboulton I have combined it with bulk-actions as well and made it more secure by binding it up with both buttons and inputs of type submit. The code checks if the associated form was previously submitted or not. If already submitted it prevents the action.
I did test with single page edit as well as Content bulk actions.

PR #6700, opened 2016-04-01
Added support for snippet manifests

### LINKED ISSUE #6672: Scripts and styles referenced by layout snippets are emitted in the admin UI
If a layout snippet file references one or more scripts or stylesheets (either via `Require()` or `Include()` those will end up being rendered in the snippet element editor in the admin UI.

Here is an example of when this happens - this particular snippet references stylesheets that are part of the theme and meant for the front-end, which change the styling of fieldsets and their labels:

image

This is what it looks like in the rendered markup:

image

sfmskywalker:
This fixes #6672 

The basic idea is that instead of declaring the fields within the snippet template itself, they are declared in a separate file next to the snippet.

This solves some peculiar side-effects, such as the issue described in #6672, and it is a cleaner implementation.

In addition, having a separate manifest allows us to add more information about a snippet, such as a friendly name, description and toolbox icon for example.

This PR retains full backwards compatibility with the way parameterized snippets currently work. Only when you add a **<MySnippet>.Params.txt** manifest file will the snippet element harvester look into it and use that file instead of invoking the HTML helpers within the snippet itself to discover the snippet fields.

I also added a **Module.md** file as a proposal for allowing module developers to provide documentation alongside their module. We could potentially display these markdown files within the admin somewhere, or even generate online documentation for all modules at some point.


src/Orchard.Web/Web.config
```diff
@@ -225,7 +225,11 @@
             <dependentAssembly>
                 <assemblyIdentity name="Iesi.Collections" publicKeyToken="aa95f207798dfdb4" culture="neutral" />
                 <bindingRedirect oldVersion="0.0.0.0-4.0.0.0" newVersion="4.0.0.0" />
-            </dependentAssembly>            
+            </dependentAssembly>
+            <dependentAssembly>
```
sebastienros: We just removed it, why adding it again?

src/Orchard/Orchard.Framework.csproj
```diff
@@ -419,7 +427,9 @@
     <Compile Include="Services\ClientHostAddressAccessor.cs" />
     <Compile Include="Services\DefaultJsonConverter.cs" />
     <Compile Include="Services\IClientHostAddressAccessor.cs" />
+    <Compile Include="Services\IYamlParser.cs" />
```
sebastienros: We already have a Yaml parser that was made to fix performance issues. Isn't it sufficient to build this feature?

sebastienros: See https://github.com/OrchardCMS/Orchard/blob/issue/6672/parameterized-snippets/src/Orchard.Web/Web.config#L196

src/Orchard.Web/Themes/Themes.csproj
```diff
@@ -123,6 +123,7 @@
     <Content Include="TheAdmin\Views\Pager.cshtml" />
     <Content Include="TheAdmin\Views\OrchardVersion.cshtml" />
     <Content Include="TheThemeMachine\ThemeZonePreview.png" />
+    <Content Include="TheThemeMachine\Views\BadgeOfHonorSnippet.Params.txt" />
```
sebastienros: I assume it was a temporary test that slipped in the commit?

sebastienros:
I don't dislike the idea of __Module.md__ but it looks out of scope for this PR. Doing it only for a project is almost useless as people won't follow continue it. If this is something we want to push it should be at least in the codegen templates. And if there is no actual usage in the admin then it will die by itself. I would favor a new entry in the documentation instead, even a placeholder in the RDT site from @agriffard.

sebastienros:
I like the idea of having a separate file triggering the new behavior, keeping the current one otherwise.

src/Orchard.Web/Web.config
sfmskywalker: Sorry, it got added automatically when adding the YamlDotNet Nuget package. Should I remove it?

src/Orchard/Orchard.Framework.csproj
sfmskywalker: Oh! Where is it? The only "YAML parser" I could find was the one used by the ExtensionHarvester, but it's not general-purpose. Perhaps you are referring to the one used in O2? If so, should we use that one instead?

sfmskywalker: The Web.config was modified by NuGet. Should I remove the JSON binding redirect entry?

src/Orchard.Web/Themes/Themes.csproj
sfmskywalker: Good catch, I'll remove it.

sfmskywalker:
Makes sense regarding Module.md, and it is indeed out of scope. I removed it for now.\, but will update the documentation separately.

DanielStolt:
Agree with `Module.md` being nice but out of scope for this.

Since we intend to do more with the snippet manifest for `1.11` that just parameter declaration (like snippet display name, description and toolbox icon) I think we should avoid naming the manifest something with "Params".

Instead I suggest:

- `SomeSnippet.Info.txt`
- `SomeSnippet.Manifest.txt`
- or even better, simply `SomeSnippet.txt`

Other than that, this is great! I especially love the inclusion of a YAML parser. For `1.11` we should also change the extension harvesters to use this rather than doing their own parsing.

sfmskywalker:
I like Daniel's suggestion of simply naming the manifest `SomeSnippet.txt`.

PR #6701, opened 2016-04-02
ChildAction shoudn't reset _cacheKey for OutputCacheFilter

560889223:
This wasn't raising awareness in Orchard community as the official code base does not use Child Actions (they use shapes).

Previously there is reentrancy-prevention code in OnResultExecuted to filter out Child Actions so that OnResultExecuted only gets executed once per request and only for the parent Action. However, this reentrancy-prevention code is being incorrectly placed in the try-catch block, that the finally-block can get executed more than once if parent Action includes any Child Action. This wasn't an issue until commit 117dd09f (first included in release 1.10) which resets _cacheKey in that finally-block via ReleaseCacheKeyLock. Now for every request that uses Child Actions, the finally block will be executed at least once during Child Action's OnResultExecuted, leaving the _cacheKey as null. Later after all Child Actions finish execution, the parent Action's OnResultExecuted will now see a _cacheKey of null, thus causing NRE in captureStream.Captured event handling code.


DanielStolt:
That is some seriously nice problem analysis and troubleshooting!

I knew @sebastienros shouldn't have messed around with my implementation... ;)

This looks solid to me, but can you retarget on 1.10.x please?

560889223:
Hi @DanielStolt 

Thanks, and thanks for the suggestion. PR #6726 is created. Closing this one.

PR #6708, opened 2016-04-03
Fixes #6642 Media gallery not working with Azure Blob Storage and "Ma…

### LINKED ISSUE #6642: Media gallery not working with Azure Blob Storage and "ManageOwnMedia" permission

hannan-azam:
Fixes #6642

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/IMediaLibraryService.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/IMediaLibraryService.cs
@@ -133,31 +133,19 @@ namespace Orchard.MediaLibrary.Services {
         /// <param name="inputStream">The stream with the file's contents.</param>
         /// <returns>The path to the uploaded file.</returns>
         string UploadMediaFile(string folderPath, string fileName, Stream inputStream);
-    }
 
-    public static class MediaLibrayServiceExtensions {
-        public static bool CanManageMediaFolder(this IMediaLibraryService service, string folderPath) {
-            // The current user can manage a media if he has access to the whole hierarchy
-            // or the media is under his personal storage folder.
+        /// <summary>
+        /// Checks if current user can manage the storage folder
+        /// </summary>
+        /// <param name="folderPath"></param>
+        /// <returns></returns>
+        bool CanManageMediaFolder(string folderPath);
 
-            var rootMediaFolder = service.GetRootMediaFolder();
-            if (rootMediaFolder == null) {
-                return true;
-            }
-
-            var mediaPath = folderPath + "\\";
-            var rootPath = rootMediaFolder.MediaPath + "\\";
-
-            return mediaPath.StartsWith(rootPath, StringComparison.OrdinalIgnoreCase);
-        }
-
-        public static string GetRootedFolderPath(this IMediaLibraryService service, string folderPath) {
-            var rootMediaFolder = service.GetRootMediaFolder();
-            if (rootMediaFolder != null) {
-                return Path.Combine(rootMediaFolder.MediaPath, folderPath ?? "");
-            }
-
-            return folderPath;
-        }
+        /// <summary>
+        /// Combines given folderPath to user's root folder path
+        /// </summary>
+        /// <param name="folderPath"></param>
+        /// <returns>Rooted folder path</returns>
+        string GetRootedFolderPath(string folderPath);
     }
 }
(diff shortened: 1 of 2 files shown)


BenedekFarkas:
This change doesn't fix the issue for me (testing locally with Azure Storage Emulator), but I've found two other issues while testing.
Parameters: A user named "user" has ManageOwnMedia permission and it has media files in the "UserTest" folder (it's path is "Users/user/UserTest"):
- Super user doesn't see the media files in the "UserTest" folder, although they are displayed in "Recent".
- Super user navigates to "/Admin/Orchard.MediaLibrary?folderPath=Users/user/UserTest", but the Media Library navigation doesn't show the "UserTest" folder. If I click on the "user" folder, then it's displayed.

The first one is definitely related to how the file system translates paths: while normal Media Library storage accepts "Users/user/UserTest", Azure storage only works with "Users\\user\\UserTest".

hannan-azam:
@BenedekFarkas I checked the fix using both FileSystemStorage as well as real Azure Storage Account before commiting. Further now i verified it again with Azure Emulator as you described and the user with permission of "ManageOwnMedia" can now create/edit folders and import files within the users storage area. Also the user is able to traverse the folder hierarchy in its own storage area.

I have this user created with name "mediatester" and im able to do all the tests successfully which i wasnt able to do previously. Please see the attached screenshot.

4-7-2016 4-46-57 pm

Kindly can you write down some steps to reproduce or may be attach screenshots on what errors you are seeing there.

sebastienros:
Maybe I am reading the changes wrongly, but isn't this change still breaking as it changes the public interface?

hannan-azam:
@sebastienros there were two extension methods "", "" in class "MediaLibrayServiceExtensions". These extension methods operate on IMediaLibraryService and the new functionality as we discussed requires IStorageProvider. There are two solutions here:
 
1. We implement a new IStorageProvider property in IMediaLibraryService which will then be accessible in the extension methods.
2. We remove these extension methods and implement these two functions as part of IMediaLibraryService interface. IStorageProvider is available as private member and can be sued by the new functions.

I used the second approach and converted the extension methods to member functions of the interface.

Please recommend if there is any better alternative or we should do it someway different.

sebastienros:
So this can't be fixed without a breaking change. In this case I'd rather expose `Combine` at the media service level than expose the `IServiceProvider` itself. It will just call the inner `_serviceProvider` implementation. I assume some other locations might not need `IServiceProvider` anymore after that.

Sorry for the confusion, and thanks for the effort.
Also please go to https://cla2.dotnetfoundation.org to sign the CLA.

BenedekFarkas:
@hannan-azam I'll test it again and come back with the results.

hannan-azam:
@sebastienros i have changed the implementation as you suggested.

PR #6709, opened 2016-04-04
fields localization and display name

### LINKED ISSUE #2337: Localization of DisplayName setting of a ContentPart Custom Field
**Phusion** created:
https://orchard.codeplex.com/workitem/18510

Release: Orchard 1.4

Display names for Content Part custom fields are not localized.

i.e: The DisplayName is stored in the "Orchard_Framework_ContentItemVersionRecord" table.

For more info, please refere to :
http://orchard.codeplex.com/discussions/346682

infofromca:
make localization consistent with input 
make display name consistent 
also fixed #2337

### DIFF
--- a/src/Orchard.Web/Core/Common/Views/EditorTemplates/Fields.Common.Text.Edit.cshtml
+++ b/src/Orchard.Web/Core/Common/Views/EditorTemplates/Fields.Common.Text.Edit.cshtml
@@ -3,7 +3,7 @@
     var isRequired = Model.Settings.Required && String.IsNullOrWhiteSpace(Model.Settings.DefaultValue);
 }
 <fieldset>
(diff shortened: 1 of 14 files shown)


sebastienros:
Can someone point the issue where we explain how translated dynamic content should be handled? The one that describes a feature where we can add new strings and their translations, that would be harvested by the localizer as a fallback context.

sebastienros:
Design proposal: https://github.com/OrchardCMS/Orchard/issues/6750

Skrypt:
As far as I can remember. We had talked about having something close to what I did with the Autoroute culture patterns. Having a control that would add tabs/textbox for each culture installed in the website settings where needed.

ramiassi:
I'm looking ahead until this PR become merged. I think it solves an important issue regarding to the localization.

sebastienros:
As explained even if it works, this is not the final solution so we can't merge it as is. The solution is to implement the suggestion in https://github.com/OrchardCMS/Orchard/issues/6750.
Any one is free to apply this technique in the meantime.

PR #6716, opened 2016-04-06
Issues/6600

jagraz:
Updates to the DefaultLocalizedStringManager.cs file to only harvest dictionaries from active themes.


src/Orchard/Localization/Services/DefaultLocalizedStringManager.cs
```diff
@@ -28,13 +31,15 @@ public class DefaultLocalizedStringManager : ILocalizedStringManager {
             ICacheManager cacheManager,
             ILocalizationStreamParser locationStreamParser,
             ShellSettings shellSettings,
-            ISignals signals) {
+            ISignals signals,
+            ShellDescriptor shellDescriptor) {
```
sebastienros: Are you sure the `ShellDescriptor` only contains the enabled features? There is also `IFeatureManager` which contains a `GetEnabledFeatures`. Also there is a mismatch between enabled features and localization files are these files don't mention features per se, just the module, but I am not sure you can do anything about it.

sebastienros:
If that fixes the issue, then GTG.

jagraz:
Yeah the GetEnabledFeatures won't show themes. Shell Descriptor only shows features and themes that are enabled.

sebastienros:
This is really not intuitive ;)

sebastienros:
Are you sure you don't want it in 1.10.x ?

jagraz:
it's up to you guys. I can put it in 1.10.1. if you feel comfortable with it.

PR #6720, opened 2016-04-07
[1.10.x] Field default value as a proposed value on first edition when creating.

jtkech:
@sebastienros 

Following our discussion just started on the text field, and need little confirmations before continuing.

- So, as a proposed value, editor only and on creating, and without tokens.
- So, no proposed value when editing an existing content item, even if the field has been added to the content type after the creation of this content item.

So, now all fields need to check if the content item is being created. I've seen this in some places:

    if(!part.HasDraft() && !part.HasPublished())

But, regarding on these extensions, i think i can resume this with this new content extension:

        public static bool IsNew(this IContent content) {
            return content.ContentItem.VersionRecord == null;
        }

Let me know if it's ok and if the extension name is ok.

Note: because of this, for the future cloning feature, we will need to also check for all the fields (not only the boolean field) if the content item is being cloning.

Best.


jtkech:
Text and input fields done, todo: others + squash.

sebastienros:
As a reminder you don't need to squash anymore, Github does it for us now.

jtkech:
Revert usage of lazy fields as you asked me.
Text and input fields already done.
Boolean and link fields done.
Todo: Date and enum fields

jtkech:
Enumeration done. Todo: Date field and resolve conflicts.

jtkech:
@sebastienros 

Resolving conflicts, here i reverted d441d83 because `setting.TextPlaceHolder` is defined nowhere.

Todo: Numeric (i forgot this one) and date field.

sebastienros:
Great, thanks, but it's late now ;)

jtkech:
@sebastienros 

Just finished the last date time field but, before committing, i want to review it a little.
Need to leave but this PR will be ready this night.

Best.

jtkech:
@sebastienros 

Now ready but indeed need to be checked because there have been so many changes around the default value behavior ... So, i tested again all the fields (maybe not all possible scenarios), and i ran again each individual field spec.

Best.

jtkech:
Warning: If someone want to check this PR, because it is based on a quite recent 1.10.x, currently there is an unrelated issue with the `publish / save double clicking prevention`. As soon as this is resolved, i'll rebase this PR. But you can revert the related change yourself for your tests: in `TheAdmin/Scripts/admin.js` around line 84, you can remove the related script.

Best.

Issue #6722, opened 2016-04-07
NullReferenceException in SessionLocator.DisposeSession()


560889223:
Hi recently we are seeing NRE in SessionLocator.DisposeSession(). As we use Debug binaries, we managed to pinpoint the exact line that causing the issue:
`_contentManagerSession.Clear();`

I looked at all references to `_contentManagerSession` and it seems to me that the only possible explanation is that `Require()`/`RequireNew()`/`Cancel()`/`Dispose()` was called before `GetSession()` is called. I'm not exactly sure how this (sequence of calls) would happen, but it's happening to us. I'm proposing a simple fix by using the new C# 6.0 syntax:
`_contentManagerSession?.Clear();`

Please let me know your thoughts.

jtkech:
I think it's a good catch.

There are some other issues related to this one. See #6555 where, because i couldn't repro myself, i just also proposed this null check `_session.Transaction != null`. (may also be useful)

But i didn't think about `_contentManagerSession`, maybe because it seemed to me as a normal `IDependency`... This could explain some issues where sessions are not correctly disposed...

But how can `_contentManagerSession` be null? Calling `_contentManagerSession.Clear()` before doesn't explain this. But your debugging proves it.

Right now, just seen that it is not directly injected. It's an auto generated (by autofac) delegate factory func `_contentManagerSessionFactory` that is injected and then used to construct `_contentManagerSession` when creating a new session.

Maybe a thread safety problem when using this factory...

Best.

560889223:
@jtkech 

To help anyone understand how we found this issue, here is the call stack when exception was thrown:
```
2016-04-07 00:00:53,433 [125] Orchard.Tasks.BackgroundService - (null) - ERROR Error while processing background task "Orchard.Indexing.Services.IndexingBackgroundTask" on tenant "Default".
System.NullReferenceException: Object reference not set to an instance of an object.
   at Orchard.Data.TransactionManager.DisposeSession() in C:\projects\orchard\src\Orchard\Data\SessionLocator.cs:line 96
   at Orchard.Data.TransactionManager.RequireNew(IsolationLevel level) in C:\projects\orchard\src\Orchard\Data\SessionLocator.cs:line 68
   at Orchard.Data.TransactionManager.RequireNew() in C:\projects\orchard\src\Orchard\Data\SessionLocator.cs:line 64
[...]

PR #6727, opened 2016-04-07
#6722: Fix NullReferenceException in SessionLocator.DisposeSession()

560889223:
NREs are seen inside DisposeSession() during background task sweeping. By reviewing the code, it seems that this can happen if inside EnsureSession(), _session.BeginTransaction() fails due to DB issues (say DB connection pool maxing out), which leaves _session assigned but not _contentManagerSession.

### DIFF
--- a/src/Orchard/Data/SessionLocator.cs
+++ b/src/Orchard/Data/SessionLocator.cs
@@ -71,7 +71,7 @@ namespace Orchard.Data {
 
         public void Cancel() {
             // IsActive is true if the transaction hasn't been committed or rolled back
-            if (_session?.Transaction?.IsActive == true) {
+            if (_session != null && _session.Transaction != null && _session.Transaction.IsActive) {
                 Logger.Debug("Rolling back transaction");
                 _session.Transaction.Rollback();
                 DisposeSession();
@@ -86,15 +86,15 @@ namespace Orchard.Data {
             if (_session != null) {
                 try {
                     // IsActive is true if the transaction hasn't been committed or rolled back
-                    if (_session.Transaction?.IsActive == true) {
+                    if (_session.Transaction != null && _session.Transaction.IsActive) {
                         Logger.Debug("Committing transaction");
                         _session.Transaction.Commit();
                     }
                 }
                 finally {
-                    // The other changes should have made sure that _session and _contentManagerSession are both assigned or both not assigned.
-                    // However, just to be extreme cautious, we still do a null-checking here using null-condition operator (?.).
-                    _contentManagerSession?.Clear();
+                    if (_contentManagerSession != null) {
+                        _contentManagerSession.Clear();
+                    }
 
                     Logger.Debug("Disposing session");
                     _session.Close();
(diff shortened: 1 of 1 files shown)


560889223:
This is to fix issue #6722

jtkech:
After our discussion, here what i propose. And we can ask @sebastienros to confirm.

So, as you have found, because `_session.BeginTransaction(level)` may fail, we may have a non null `_session` with a null `_contentManagerSession`. So, `_contentManagerSession.Clear()` may fail in `DisposeSession()`. Then this session is not closed, disposed...

- Ok with your fix with a null check on `_contentManagerSession` in `DisposeSession()`.
- Ok for the null checks on `_session.Transaction` as i've seen in some source codes.
- Note: not sure we can use c# 6 syntax for those using previous version of VS.
- Finally, because you have found the cause, i think it's better to let things as they was in `EnsureSession()` without a try / catch. Let nhibernate surface its exceptions (as you do in your catch block) but without disposing the session (that will be disposed because of your fix).

Best

src/Orchard/Data/SessionLocator.cs
```diff
@@ -71,7 +71,7 @@ public class TransactionManager : ITransactionManager, IDisposable {
 
         public void Cancel() {
             // IsActive is true if the transaction hasn't been committed or rolled back
-            if (_session != null && _session.Transaction.IsActive) {
+            if (_session != null && _session.Transaction != null && _session.Transaction.IsActive) {
```
sebastienros: This is not correct. If the goal is to prevent `_session.Transaction` from being called if it's null, then this needs to be split in two `if` statements as the `DisposeSession` would no be called, even if `_session` itself is not null.

jtkech:
@sebastienros About your line note.

> `DisposeSession()` would no be called, even if _session itself is not null.

This is already the case in the current implementation of `Cancel()`, if `_session` is not null but `_session.Transaction.IsActive` is false, then `DisposeSession()` is not called.

So, here, it's just to add a null check on `_session.Transaction` before testing `_session.Transaction.IsActive`.

But, do you mean that in this new use case, `_session` is not null and `_session.Transaction` is null, we have to call `DisposeSession()`?

Best.

sebastienros:
Whatever value is _session.Transaction, is _session is not null it has to be disposed in the case I commented on.

sebastienros:
```csharp
if (_session != null) {
  if ( _session.Transaction != null && _session.Transaction.IsActive) {
    Logger.Debug("Rolling back transaction");
    _session.Transaction.Rollback();
  }

  DisposeSession();
}
```

jtkech:
Oh, yes i see. If not nul, session is always disposed.
 So, here, it was already a problem before. Right?

sebastienros:
Yes, or this would be a half fix. The NRE and the fact the _session is not correctly disposed because of the NRE.

jtkech:
Ok, but just for infos, here the NRE that was first involved is not on `_session.Transaction` in `Cancel()` but on `_contentManagerSession` that was null in `DisposeSession()`. Null check on `_session.Transaction` has been added after for more safety in `DisposeSession()`, and then in `Cancel()`.

sebastienros:
I will apply the change if you agree, and mark Qingyu as the author.

jtkech:
@sebastienros 

Ok, i agree because it fixes the original NRE in `DisposeSession()`.
And ok if you confirm that always disposing the session (if not null) in `Cancel()` (not as before) is good.

Indeed, it also makes for me sense to dispose a session that has no more valid transaction. But i didn't want to create new use cases.

Just seen this PR is already closed :)

Best

jtkech:
@sebastienros I think you're totally right.

I was afraid of the new use case in Cancel(), session not null and transaction not null but not active, where we will dispose the session not as before.

But just seen with the debugger that the transaction is active as soon as we begin the transaction. So, if the transaction is not active, normally it has been committed / rolled back and the session should have already been released.

Thanks.

PR #6729, opened 2016-04-07
Workaround to avoid Transaction issue for PostgreSQL when creating DistributedLocks table

hopla:
When checking if the "Orchard_Framework_DistributedLockRecord" exists, a select on the table is performed. When the table does not exist, this throws an exception.
In PostgreSQL, this sets the transaction in an Error state. Because of this, the transaction accepts no other commands, and the table is never created.

This pull request works around this, by starting a new transaction to create the table  

@OrchardCMS

### DIFF
--- a/src/Orchard/Tasks/Locking/Services/DistributedLockSchemaBuilder.cs
+++ b/src/Orchard/Tasks/Locking/Services/DistributedLockSchemaBuilder.cs
@@ -13,14 +13,6 @@ namespace Orchard.Tasks.Locking.Services {
             _schemaBuilder = schemaBuilder;
         }
 
-        public bool EnsureSchema() {
-            if (SchemaExists())
-                return false;
-
-            CreateSchema();
-            return true;
-        }
-
         public void CreateSchema() {
             _schemaBuilder.CreateTable(TableName, table => table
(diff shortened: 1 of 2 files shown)


src/Orchard/Tasks/Locking/Services/DistributedLockSchemaBuilder.cs
```diff
@@ -36,8 +28,8 @@ public class DistributedLockSchemaBuilder {
 
         public bool SchemaExists() {
             try {
-                var tablePrefix = String.IsNullOrEmpty(_shellSettings.DataTablePrefix) ? "" : _shellSettings.DataTablePrefix + "_";
-                _schemaBuilder.ExecuteSql(String.Format("select * from {0}{1}", tablePrefix, TableName));
+                var tablePrefix = string.IsNullOrEmpty(_shellSettings.DataTablePrefix) ? "" : _shellSettings.DataTablePrefix + "_";
```
sebastienros: Please revert the formatting.

hopla: Formatting reverted.

Issue #6733, opened 2016-04-08
NHibernate IdentifierGenerationException when posting comment to content item of custom content type


560889223:
Repro Steps

1. Fresh 1.10 deployment (I used SQL Server connection, but SQLite should do as well)
2. Enable Audit Trail module
3. Add a custom content type, with `TitlePart`, `CommentsPart` and `AutoroutePart` (so you can navigate to it)
4. Create a new content item of that custom content type
5. Post a comment (watch output in VS if you are debugging the instance, otherwise look for `IdentifierGenerationException` in logs)
  * There will be a first chance exception of `IdentifierGenerationException` (see [sample log file](https://gist.github.com/560889223/724cad40e3723b69901b08632d186b2b "orchard-error-2016.04.07.log")).
6. Try to access Audit Trail dashboard, which will throw an exception telling you the Content Item ID is null.
  * There will be unhandled `FormatException` then rethrown in view (see [sample log file](https://gist.github.com/560889223/724cad40e3723b69901b08632d186b2b "orchard-error-2016.04.07.log")).

This syndrome, is caused by the following sequence of calls:

1. `CommentController.Create(...)` calls to `ContentManager.UpdateEditor(comment, this)`, which in turns calls into `CommentPartDriver.Editor(...)` to validate the information submitted by current request. The new content item of `Comment` has not been **Created** yet. 
2. Since content pipeline has been kicked off by the call to `ContentManager.UpdateEditor(comment, this)`, content pipeline events, such as `handler.Updating` and `handler.Updated` will be called.
3. Audit Trail's `GlobalContentHandler` on the other end will receive such events, and think as a signal of a new version of content item. It will starts collection of information to record an Audit Trail event. This includes:
  1. Export previous version of content item during `handler.Updating` by calling `driver.Export(...)` on applicable content drivers.
  2. Wait for `CommentPartDriver.Editor(...)` to finish updating the content item.
  3. Export current version of content item during `handler.Updated` by calling `driver.Export(...)` on applicable content drivers.

During step 3.iii, export of the updated content item would trigger a DB flush. In our case, current DB session will try to flush the newly **instantiated** `CommentPartRecord`, which only exists in memory and not in database. Note this is before `ContentManager.Create(comment)` is called, so no corresponding `ContentItemRecord` is created yet, not even as scheduled DB changes during the flush. At this time, `CommentPartRecord.ContentItemRecord` will remain null, thus our `IdentifierGenerationException`.

[...]

Issue #6534, opened 2016-03-06
Orchard creates content item even when validation fails.


TwistedWhisper:
I'll confess that the title only speaks the half truth. Now that I have your attention, let's try this out and reproduce the issue. Below is what happening in Orchard when validation fails:

1. Create a content item, let's say create a new **Page** from the dashboard.
2. Leave the form empty and save it.
3. Content item **Page** gets created and about to be inserted into database through a transaction.
4. Validate form. Since the required field **Title** is empty, validation will fail.
5. Transaction is cancelled and rolled back after validation failed.

Why bother creating the content item when the database transaction is gonna get cancelled anyway and waste precious database resources (the Identity counter of the database table `Orchard_Framework_ContentItemRecord` is increased every time a content item is created and never backs down regardless of transaction rollback)?

Why not swap Step `#3` with `#4`? Validate the form first and if it fails, then there's no need to even create the content item thus no transaction rollback necessary.

I'll point you to the Controller's Action in question to save you the trouble, `Orchard.Core.Contents.Controllers.AdminController.CreatePOST(string, string, Action<ContentItem>)`

    _contentManager.Create(contentItem, VersionOptions.Draft); //content item is created first

    var model = _contentManager.UpdateEditor(contentItem, this); //validation happens after content item creation

    if (!ModelState.IsValid) {
        _transactionManager.Cancel(); //database transaction is cancelled when validation fails
        return View(model);
    }

I suggest changing the order to

    var model = _contentManager.UpdateEditor(contentItem, this); //validate the form first

    if (!ModelState.IsValid) {
        _transactionManager.Cancel(); //rollback transaction when validation fails - probably doesn't even require this since database is not even touched yet
        return View(model);
    }

    _contentManager.Create(contentItem, VersionOptions.Draft); //create content item after validation passed

By doing this, we'll be able to tell apart a content item that has been created in the database from one that has not. A new content item (including those that fail validation) has an `Id` of 0. We can use this data in our `ContentPartDriver` to serve different shapes depending on whether if the content item is new or already created.

    protected override DriverResult Editor(MetadataPart part, dynamic shapeHelper) {            
            //only serve a shape if the content item already inserted into the database
            if (part.ContentItem.Id != 0) {
                return ContentShape("Parts_Metadata_Edit", () =>
                    shapeHelper.EditorTemplate(TemplateName: "Parts/Metadata.Edit", Model: _metadataService.BuildEditorViewModel(part), Prefix: Prefix)
[...]

PR #6734, opened 2016-04-08
#6733: Content item of Comment should first be 'Created' before being passed into `UpdateEditor`

qingfx:
Orchard.Blogs.BlogPostAdminController.CreatePOST(int blogId, bool publish) is a good example on how to implement content item creation.

CommentController.Create(...) should ContentManager.Create the CommentPart first, only then use ContentManager.UpdateEditor(...) to update part and validate updated part. After this:
- If everything is correct, all DB changes will be saved upon request finishes. -OR-
- If something goes wrong, say ModelState.IsValid == false, cancel the DB transaction via Services.TransactionManager.Cancel()

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Comments/Controllers/CommentController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Comments/Controllers/CommentController.cs
@@ -27,6 +27,7 @@ namespace Orchard.Comments.Controllers {
                 return this.RedirectLocal(returnUrl, "~/");
 
             var comment = Services.ContentManager.New<CommentPart>("Comment");
+            Services.ContentManager.Create(comment);
             var editorShape = Services.ContentManager.UpdateEditor(comment, this);
 
             if (!ModelState.IsValidField("Comments.Author")) {
@@ -46,8 +47,6 @@ namespace Orchard.Comments.Controllers {
             }
 
             if (ModelState.IsValid) {
-                Services.ContentManager.Create(comment);
-
                 var commentPart = comment.As<CommentPart>();
 
                 // ensure the comments are not closed on the container, as the html could have been tampered manually
@@ -57,13 +56,13 @@ namespace Orchard.Comments.Controllers {
                 }
 
                 var commentsPart = Services.ContentManager.Get(commentPart.CommentedOn).As<CommentsPart>();
-           
+
                 // is it a response to another comment ?
                 if(commentPart.RepliedOn.HasValue && commentsPart != null && commentsPart.ThreadedComments) {
                     var replied = Services.ContentManager.Get(commentPart.RepliedOn.Value);
                     if(replied != null) {
                         var repliedPart = replied.As<CommentPart>();
-                            
+
                         // what is the next position after the anwered comment
                         if(repliedPart != null) {
                             // the next comment is the one right after the RepliedOn one, at the same level
@@ -93,7 +92,7 @@ namespace Orchard.Comments.Controllers {
                             }
                         }
                     }
-                        
+
                 }
                 else {
                     // new comment, last in position


Skrypt:
Could you point on the issues # that it will fix.

jtkech:
With @560889223 and @Skrypt we tried to allow the pattern "create after update" but we had drawbacks with handlers / drivers that create repository record that rely on the content item.

If the driver / handler check if it's a new item (not already created), no exception but the field is not persisted. So we needed to add a 2nd UpdateEditor as suggested in #6534.

But if the driver / handler doesn't check if it's a new item, there are exceptions. So we needed to use again "create before update". It can be fixed at the driver / handler level but we thought that we can't rely on this.

Here, the issue was with the CommentController and an AuditTrail handler. **And i can confirm that this PR fixes this issue**.

Best.

sebastienros:
But calling Create and cancelling the transaction will create transaction logs. Please take a look at the previous change to this class that fixed another issue so that it doesn't conflict with it, or we need to find a compromise.

jtkech:
@sebastienros 

You're right it was since this commit 874bd7d to fixes this work item [19304](https://orchard.codeplex.com/workitem/19304). I'll take a look.

560889223:
@sebastienros 

In that case, work item 19304 sounds not unique to `Comment`; many content items that are recreated before model validation could have similar issues. The problem is that today's `ContentManager.Create` is not being used in the same sense as the user perceived. It's more like creating a placeholder for the content item-to-insert-into-database. To align `ContentManager.Create` with what the user perceives, it seems to me that we have to draw a clear line between validation and creation. The event broadcasting and the work of creating `ContentItemRecord`/`ContentItemVersionRecord` needs to be separated.

I'm suggesting the following:

1. Today's `UpdateEditor` will be renamed to `UpdateEditorImpl` and no longer broadcasting events `Updating`/`Updated`.
2. Today's `Create` will be renamed to `CreateImpl` and no longer broadcasting events `Creating`/`Created`.
3. To respond a user's request to create a content item, one should:
  1. `ContentManager.Validate`, think it as a **dry-run** of `UpdateEditorImpl` that does checks without updating content item. This gives controller/service a chance to fail creation without appending noises to DB transaction logs.
  2. `ContentManager.Create`
    1. Calls `CreateImpl`.
    2. Broadcast `Creating` events with the new `Created` content item. 
    3. Calls `UpdateEditorImpl`.
    4. Broadcast `Created` events with the populated content item.
4.  To respond a user's request to update an existing content item, one should:
  1. `ContentManager.Validate`, think it as a **dry-run** of `UpdateEditorImpl` that does checks without updating content item. This gives controller/service a chance to fail creation without appending noises to DB transaction logs.
  2. `ContentManager.Update`
    1. Broadcast `Updating` events with the existing content item. 
    2. Calls `UpdateEditorImpl`.
    3. Broadcast `Updated` events with the updated content item.

jtkech:
@sebastienros @560889223.

The work item was about using Orchard.Rules (now deprecated) to send an e-mail on comment publication. That's why the create was moved after `UpdateEditor()` and model checking in 874bd7d.

The comments controller uses `Services.ContentManager.Create()` with no `VersionOptions`, so `Published` is used. So, the create action do a create and publish, so the publish event is triggered.

**Compromise 1**

One compromise that doesn't conflict with the work item is to do a create with a `Draft` option.

    Services.ContentManager.Create(comment, VersionOptions.Draft); // no publish event here
    var editorShape = Services.ContentManager.UpdateEditor(comment, this);
    ...
    ...
        if (ModelState.IsValid) {
            Services.ContentManager.Publish(comment.ContentItem); // publish event here

But if you are using, e.g in a workflow, a comment creation event, you may have to swap to a publish one.

**Compromise 2**

As you, i would prefer to use the "create after update" pattern to preserve transaction logs and @560889223 suggestion with a new `ContentManager.Validate` implementation looks very good.

Currently, to use this pattern we have to fix handlers / drivers that do actions relying on a created item but without checking it. Then, we also have to re-call `UpdateEditor()` after the effective create.

So, here we would have to update `GlobalContentHandler`, the one that was failing in this issue. In the `Updating / Updated` events, if it's a new item (not created) then don't record an audit trail event.

For doing this, we can use this new content extension:

    public static bool IsNew(this IContent content) {
        return content.ContentItem.VersionRecord == null;
    }

And in `CommentController` do a 2nd `UpdateEditor()`:

    var comment = Services.ContentManager.New<CommentPart>("Comment");
    // The 1st update simulate the validation step but some drivers / handlers,
    // in their update events need to check if it's a new item before doing some actions
    var editorShape = Services.ContentManager.UpdateEditor(comment, this);
    ...
    ...
        if (ModelState.IsValid) {
            Services.ContentManager.Create(comment, VersionOptions.Draft); // create without publish
            // 2nd update events where actions relying on a created item can be done
            Services.ContentManager.UpdateEditor(comment, this);
            Services.ContentManager.Publish(comment.ContentItem); // publish events only here
            ...

Best.

jtkech:
@sebastienros @560889223 

So, specifically for comments (many items, spams ...), here the best compromise i've found and tested.

In `CommentController.cs`:

            var comment = Services.ContentManager.New<CommentPart>("Comment");
            var editorShape = Services.ContentManager.UpdateEditor(comment, this);
            ...
            ...
            if (ModelState.IsValid) {
                Services.ContentManager.Create(comment, VersionOptions.Draft);
                Services.ContentManager.UpdateEditor(comment, this);
                Services.ContentManager.Publish(comment.ContentItem);
                ...

In `GlobalContentHandler.cs` in both `Updating()` and `Updated` methods:

            var contentItem = context.ContentItem;

            if (contentItem.IsNew())
                return;

At the end of `ContentExtensions.cs`: (will be there with #6720 on default values)

        public static bool IsNew(this IContent content) {
            return content.ContentItem.VersionRecord == null;
        }

Finally, if there are already bad audit trail records created before this fix, to prevent admin audit trail views from failing, in `DiffGramAnalyzer`:

    // Replace this line with the following
    //var index = match == null || isAttributeChange ? default(int?) : Int32.Parse(match) - 1;

    int matchInt;
    var index = match == null || isAttributeChange || !Int32.TryParse(match, out matchInt)
        ? default(int?) : matchInt - 1;


Best.

BenedekFarkas:
[JT's suggested changes](#issuecomment-211128847) were applied on dev with #6891.

PR #6737, opened 2016-04-08
Fixed removal of records in DistributedLockService.

sfmskywalker:
This fixes an issue with creating new distributed lock records while a record with the specified name already exists (but with a different ValidUntilUtc value) by deleting any existing (and expired) records first of the specified name.

### DIFF
--- a/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
+++ b/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
@@ -141,9 +141,19 @@ namespace Orchard.Tasks.Locking.Services {
 
             ExecuteOnSeparateTransaction(repository => {
                 // Try to find a valid lock record in the database.
-                var record = repository.Table.FirstOrDefault(x => x.Name == internalName && (x.ValidUntilUtc == null || x.ValidUntilUtc >= _clock.UtcNow));
+                var records = repository.Table.Where(x => x.Name == internalName).ToList();
+                var record = records.FirstOrDefault(x => x.ValidUntilUtc == null || x.ValidUntilUtc >= _clock.UtcNow);
                 if (record == null) {
-                    // No record existed, so we're good to create a new one.
+                    if (records.Any()) {
+                        // No record matched the criteria, but at least one expired record with the specified name was found.
+                        // Delete the expired records before creating a new one. In theory no more than one record can exist
+                        // due to the unique key constraint on the 'Name' column, it won't hurt to work on a collection.
+                        foreach(var expiredRecord in records) {
+                            repository.Delete(expiredRecord);
+                        }
+                    }
+
+                    // No valid record existed, so we're good to create a new one.
                     Logger.Debug("No valid record was found for lock '{0}'; creating a new record.", internalName);
 
                     repository.Create(new DistributedLockRecord {
--- a/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
+++ b/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
@@ -89,7 +89,7 @@ namespace Orchard.Tasks.Locking.Services {
                 DistributedLock dLock = null;
 
                 // If there's already a distributed lock object in our dictionary, that means
-                // this acquisition is a reentrance. Use the existing lock object from the 
+                // this acquisition is a reentrance. Use the existing lock object from the
                 // dictionary but increment its count.
                 if (_locks.TryGetValue(monitorObj, out dLock)) {
                     Logger.Debug("Current thread is re-entering lock '{0}'; incrementing count.", internalName);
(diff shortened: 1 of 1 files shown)


jtkech:
@sfmskywalker 

I would have to review the details to be sure, but it seems to me that a record could be no more valid because a job is too long. Then, when the job will end and release the lock record, it will fail in `DeleteDistributedLockRecord`.

So, not sure at all, but i was thinking of the idea not to delete the lock record in the release action, and let it be deleted (because it's no more valid) when a new lock record is acquired, as done by your fix.

Best

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -141,9 +141,19 @@ public class DistributedLockService : Component, IDistributedLockService {
 
             ExecuteOnSeparateTransaction(repository => {
                 // Try to find a valid lock record in the database.
-                var record = repository.Table.FirstOrDefault(x => x.Name == internalName && (x.ValidUntilUtc == null || x.ValidUntilUtc >= _clock.UtcNow));
+                var records = repository.Table.Where(x => x.Name == internalName).ToList();
+                var record = records.FirstOrDefault(x => x.ValidUntilUtc == null || x.ValidUntilUtc >= _clock.UtcNow);
                 if (record == null) {
-                    // No record existed, so we're good to create a new one.
+                    if (records.Any()) {
```
DanielStolt: This if statement is redundant.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -141,9 +141,19 @@ public class DistributedLockService : Component, IDistributedLockService {
...
+                        // No record matched the criteria, but at least one expired record with the specified name was found.
+                        // Delete the expired records before creating a new one. In theory no more than one record can exist
+                        // due to the unique key constraint on the 'Name' column, it won't hurt to work on a collection.
+                        foreach(var expiredRecord in records) {
```
DanielStolt: Space after `foreach`.

DanielStolt:
This looks good to me.

@jtkech I don't think your suggestion is good - the `ValidUntilUtc` is intended to be a fail-safe if for some reason a task fails to release its own lock. Normally, most tasks should finish long before the lock record expires. If we do as you suggest, the lock would be held much longer than necessary - the task which originally acquired the lock will in most cases have finished long ago, and now no other task can acquire the lock until the record expires. Not a good situation.

You are correct that there is a theoretical chance that a task runs for so long that its lock record expires and gets deleted by another task, but this is to be expected. If a task runs longer than its lock expiration, then the lock expiration time was much too small to begin with, and the task has been incorrectly designed and **should** experience an exception. In the rare case that tasks have such unpredictable running times that a reasonable expiration cannot be assumed, then they should be designed to "renew" their lock records while running and push the expiration time.

@sfmskywalker's fix is therefore the most correct IMO.

jtkech:
@DanielStolt 

First, i don't remember the details, so here not sure, just to understand.

> the task has been incorrectly designed and should experience an exception.

I agree, the valid time should be well tuned. I said this because in the related issue, a job seemed to be around 5 minutes, and in the current code `JobsQueueProcessor` uses a `maxValidFor of 5 minutes`.

> This if statement is redundant

I don't think so because `records` (all records with `internalName`) is not the same as `record` (1st unexpired record). Maybe the comment should be `No unexpired record existed...`

> If we do as you suggest, the lock would be held much longer than necessary.

I don't think so because it has expired. So, no unexpired record will be found. So with the new code it will be deleted and a new one will be created.

Best.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
Skrypt: It is not. One is record without a "s" and the other is records with a "s"
It validates if "record" has not found any results. Then it looks in the "records" list to delete it.
I suggest renaming the vars to avoid eventual confusion ;)

DanielStolt:
@jtkech 

> > This if statement is redundant

> I don't think so because `records` (all records with `internalName`) is not the same as `record`

That's not why it's redundant. It's redundant because the `records` collection is empty, the loop will simply skip. The `foreach` does the empty checking, so no real reason to do it twice.

> > If we do as you suggest, the lock would be held much longer than necessary.

> I don't think so because it has expired.

I might have misunderstood your suggestion, but as I understood it, you suggest that tasks should not delete their own lock records, but instead the next task should delete the lock record if it finds it expired, no?

The problem with that approach:

1. Task A acquires the lock (creates the lock record).
2. It's hard to predict how long task A is going to take, so to be safe, it sets the expiration to 10 mins (good practice).
3. Task A runs to completion in 2 seconds. It does not delete the lock record (as you suggested)
4. 10 seconds later, the next task B tries to acquire the lock, but finds the lock record which has not yet expired.
5. Task B has to wait/retry for 10 minutes until it can safely delete the existing lock record. In reality, task A only needed it for 2 seconds.

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
DanielStolt: @Skrypt I insist. :) That's not why it's redundant. It's redundant because the `records` collection is empty, the loop will simply skip. The `foreach` does the empty checking, so no real reason to do it twice.

Skrypt: Now you win :)

jtkech:
@DanielStolt 

Indeed for the if statement you're right, sorry i thought you didn't see the difference between record and records. Indeed, the foreach do one of the if statement.

And yes it was my idea but a bad idea, i did too many things in the same time. Indeed, i was only thinking about the case the valid time is too short (as in the related issue with the jobs processor). And i was trying to a find a way not to throw an exception in this case.

But you said it's normal to throw an exception in this case, so all seems to be ok. but what do you think of the current expiration of 5 minutes used by the jobs processor in the source code?

Thanks for your explanation.

DanielStolt:
@jtkech My pleasure. :)

Regarding the jobs processor, I'm not so familiar with it, but it sounds like a generic work execution engine where it is impossible to know how long workloads will take, and as such it should probably be changed to use a sliding expiration approach, i.e. regularly update the lock record and extend the expiration time as long as the task is still running. We should probably open a separate issue for this, rather than discussing it here.

sebastienros:
GTG once feedback has been applied

sebastienros:
:up::date:

PR #6754, opened 2016-04-13
Fine tuning of the Admin Content Types Field and Parts lists. (CSS)

### LINKED ISSUE #6695: Content Definition lists are alternated
At some point a css change has lead to this:

image

Skrypt:
Removed the use of gif to use png files since we needed 4 gifs vs 2 png to have the same result.
The weight of the files is not a concern "under 1ko".
Removed hover that changed the background color for nothing.

closes #6695 

orchard edit content type page


sebastienros:
https://cla2.dotnetfoundation.org

PR #6763, opened 2016-04-13
Update MenuItem.Edit.cshtml

Hazzamanic:
Add returnUrl hidden field so hitting save will redirect you correctly


sebastienros:
Fixes #6764

PR #6766, opened 2016-04-14
[1.10.x] fix double click prevention when client validation

jtkech:
If you click on submit but the form is not submitted because of client validation, then you can't submit anymore.

### DIFF
--- a/src/Orchard.Web/Themes/TheAdmin/Scripts/admin.js
+++ b/src/Orchard.Web/Themes/TheAdmin/Scripts/admin.js
@@ -84,11 +84,14 @@
     //Prevent double-click on buttons of type "submit"
     $("form button[type='submit'], form input[type='submit']").click(function (e) {
         var form = $(this).closest("form")[0];
-        if (typeof(form.formSubmitted) != "undefined") {
+        if (typeof (form.formSubmitted) != "undefined" && form.formSubmitted == true) {
             e.preventDefault();
             return;
         }
         form.formSubmitted = true;
+        setTimeout(function () {
+            form.formSubmitted = false;
+        }, 1000);
     });
 
     // Handle keypress events in bulk action fieldsets that are part of a single form.
--- a/src/Orchard.Web/Themes/TheAdmin/Scripts/admin.js
+++ b/src/Orchard.Web/Themes/TheAdmin/Scripts/admin.js
@@ -84,13 +84,13 @@
     //Prevent double-click on buttons of type "submit"
     $("form button[type='submit'], form input[type='submit']").click(function (e) {
         var form = $(this).closest("form")[0];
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Themes/TheAdmin/Scripts/admin.js
```diff
@@ -84,11 +84,14 @@
     //Prevent double-click on buttons of type "submit"
     $("form button[type='submit'], form input[type='submit']").click(function (e) {
         var form = $(this).closest("form")[0];
-        if (typeof(form.formSubmitted) != "undefined") {
+        if (typeof (form.formSubmitted) != "undefined" && form.formSubmitted == true) {
             e.preventDefault();
             return;
         }
         form.formSubmitted = true;
+        setTimeout(function () {
```
sebastienros: How does it work ?

sebastienros:
Fixes https://github.com/OrchardCMS/Orchard/pull/6766

jtkech:
As i've tested, it releases the formSubmitted lock (implemented here) after 1 second.

So, if you click i.e `Publish` and you are still in the same edit form because of a client validation checking, the you can correct your input and retry a publish action.

sebastienros:
Isn't there a way to not disable the button if the form is not submitted? Not sure if you are an expert at JAvascript, we might have to ask one's opinion ...

jtkech:
I agree

sebastienros:
I am currently working on it getting some tips from gurus.

jtkech:
I am interested to know the final solution. Here i just fixed the current implementation, maybe we have better to use the `submit event` directly.

jtkech:
@sebastienros 

Just tried with the submit event, seems to work by replacing this:

    $("form button[type='submit'], form input[type='submit']").click(function (e) {

with:

    $("form").submit(function (e) {

And now without the `setTimeout()` call.

Best.

jtkech:
Just by curiosity i've done additional tests, then i let you investigate.

So the solution to use the submit event (see below) without `setTimeout()` seems to work. But, if a submit is done i.e through jquery without refreshing the page, then the submit lock will persist. Or without hitting at all the submit event.

    $("form").submit(function (e) {
        if (!!this.submitted) {
            e.preventDefault();
            return;
        }
        this.submitted = true;
    });

So, maybe better to filter by using the click event of the submit button as before. And the only way i've found to fix the issue is by using `setTimeout()`.

    $("form button[type='submit'], form input[type='submit']").click(function (e) {
        var form = $(this).closest("form")[0];
        if (!!form.submitted) {
            e.preventDefault();
            return;
        }
        form.submitted = true;
        setTimeout(function () {
            form.submitted = false;
        }, 1000);
    });

Best.

Issue #6772, opened 2016-04-14
BlockQuote element


sfmskywalker:
In addition to having the **Break** and **Heading h1-6** content elements, I would like to add a **BlockQuote** element, which would map to the `<blockquote>` HTML element.

sfmskywalker:
I created a PR here: #6773

bleroy:
I'm getting a little concerned that we're rebuilding an HTML editor instead of a layout editor. Until now, elements were structural, containers, or higher-level content blocks. For instance, both hr and headings are structural. This one is an HTML element that isn't structural, and it's not a container in the sense that it will not contain any structure richer than text. It looks like it would be marginally more difficult to use a HTML element with a blockquote inside. This looks like an element that should be in a gallery module rather than in the core distro.
An alternative design that would kill more than one bird with a stone would be to make it easy to configure the outer tag (and if there is one) on all elements, like we do for CSS classes.

sfmskywalker:
Do you mean because a Blockquote element is not structural because it is like a container that can contain other HTML elements?

I can see that the line becomes a little bit blurry when dealing with this type of element, because it would arguably make sense as well to be able to add child elements to a Blockquote element (a `<cite>` element for example, but also paragraphs and images) using drag & drop.

I'll give this some thought, and raise it as a topic coming Tuesday.

bleroy:
Pretty much. Where do we draw the line? Are we going to add elements for all HTML tags?

sebastienros:
From the meeting people seem to agree with a generic container tag where you would select the name of the tag, from its editor. Dropdown list with available tags, and textbox to enter a custom one.

jersiovic:
It is important you can set tag when create a new element through UI inheriting from the generic container element. Currently you cannot inherit from containers.

bleroy:
As discussed today at the meeting:

1. CssClass is very analogous to a possible Tag property to all tags. Objections to one almost all work for the other. We might want to think about adding exceptions to CssClass, as a consequence. The absence of an outer tag is one possible exception where the CssClass loses meaning.
2. The text and container elements at least could be enhanced with a new tag property. This would cover the blockquote, cite, and other scenarios.

I'd add that it may even be useful in some cases to suppress the default outer tag entirely, both for text and container.

sebastienros:
Action: 
- New Tag property on Text element to support the `BlockQuote` issue.
- Add Tag property to Canvas element

PR #6773, opened 2016-04-14
Added Blockquote element

sfmskywalker:
For issue #6772

This adds a **Blockquote** element to the toolbox, making it awesome for users to add beautifully styled block quote elements to their page.

image


DanielStolt:
Tested and reviewed, looks and works fine, both code and functionality.

However, I think we should also add a "citation" property to the element, allowing editor control of the rendered `cite` attribute:

``` html
<blockquote cite="http://developer.mozilla.org">
  <p>This is a quotation taken from the Mozilla Developer Center.</p>
</blockquote>
```

sfmskywalker:
Thanks for reviewing! I added the `Citation` property.

sebastienros:
Why not an HTML section instead of content ?

sfmskywalker:
Section, as in a new Category called HTML containing HTML elements? I would not be opposed to that.

PR #6791, opened 2016-04-17
PE tasks list init and start background timer after PE tasks.

### LINKED ISSUE #6742: Start background timer after PE tasks, PE tasks list init and isolation.
**1st concern**

Because PE through ContextState now uses LogicalData (if no ambient HttpContext), data are carried among threads. This could be desired with async / await codes, or not with some concurent codes.

The data sharing begins on the 1st write and can be broken by writing a `new` object. You can do this in a desired parent thread and / or to isolate it from its parent. So, the Processing Engine would need a new `Initialize()` interface method that can be easily done by using `_entries.SetState(new List<Entry>())`.

**The use case**

On startup, each shell starts in a parallel thread where the following is done:

1. Activate the shell.
2. Where the background timer is activated.
3. Execute Processing Engine tasks.

So, if PE tasks are added in 1, because 2 is in the same thread, the PE tasks list will be seen in the shell background tasks. No problem because PE tasks are not evaluated in the background.

But i've seen a background task that uses `AddTask()`, even if useless here. So, if in 3 there is a quite long running job, a background task could add an entry in the PE list while it is evaluted in 3.

**Suggestion**

Create the `PE.Initialize()` method and call it in `BackgroundService` to isolate background tasks. And also on startup in the `Parallel.ForEach()` at the beginning (so for each shell).

**2nd concern and suggestion**

Maybe not good to start executing a background task while a job on startup is still running. Here, my suggestion is to activate the background timer only when the PE tasks ended? (Could solve the above).

Best.

jtkech:
Fixes #6742 

- On startup, for each tenant, explicitly init the PE tasks list in a new (isolated from its parent) logical context, this at the beginning of the desired parent thread.

- On startup, for each tenant, suspend the background timer until the PE tasks list is fully evaluated.

On startup, a shell and its bg timer are activated in the same logical context, so the PE list is shared with bg tasks. So, maybe good to also init (isolate) the PE tasks list at the beginning of the background service. But here it's not needed because bg tasks are suspended while evaluating the PE list.

Best.

### DIFF
--- a/src/Orchard.Tests.Modules/Comments/Services/CommentServiceTests.cs
+++ b/src/Orchard.Tests.Modules/Comments/Services/CommentServiceTests.cs
@@ -203,6 +203,8 @@ namespace Orchard.Tests.Modules.Comments.Services {
 
     public class ProcessingEngineStub : IProcessingEngine {
 
+        public void Initialize() { }
+
         public string AddTask(ShellSettings shellSettings, ShellDescriptor shellDescriptor, string messageName, Dictionary<string, object> parameters) {
             return "";
         }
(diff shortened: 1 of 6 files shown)


sebastienros:
Talked with @jtkech offline about it, and it's legit. GTG from my point of view, waiting for testing.

Issue #6756, opened 2016-04-13
Media Library: Can't migrate to Azure Blob Storage automatically


BenedekFarkas:
Let's consider the following scenario:

- I'm developing an Orchard application locally while using the regular file storage (Orchard.Web\Media) for media files.
- After deploying the application to Azure, I also want to use Blob Storage:
    - Enable the Azure Media Storage Feature.
    - Copy all files from the local file system to the Storage Account.

I would expect the Media Library to continue to work as before, but media files are not loading at all for any folder, except "Recent". That's because Orchard stores the path of the media files based on their location on the file system (at the time they are created): NTFS uses `\` as the directory separator, but AFS uses `/`, so when querying for content items in a specific folder while using AFS won't return any results for media items created on NTFS (and vice versa).

My suggestion is that Media Library should always use one of those (I don't think it matters which one, since both of them are illegal characters for file and folder names).

hannan-azam:
@BenedekFarkas i agree with you, there should be a simple solution to migrate to Azure Blob/File System anytime one decides to go with either storage.

I came across this problem working on a project where the site involved around 6000 media items. We kept working happily with local file system and once we were ready to go live with Azure Blob Storage we got struck with this. Later we solved it by running SQL Queries to convert all those directory separators to "/".

In HTML Content Editors and wherever media urls are used, Orchard uses URL based paths so  "/" separator would be a good candidate for this.

sebastienros:
I agree, the database should have normalized file paths, using '/' might be the simplest.

rtpHarry:
I'm using 1.10.x with Azure Media Storage module enabled and that new bit of code is failing:

![image](https://cloud.githubusercontent.com/assets/1038062/14753954/709f406c-08d0-11e6-93fc-ad942e461e08.png)

![image](https://cloud.githubusercontent.com/assets/1038062/14754061/e7478f12-08d0-11e6-9289-e55a190982a3.png)

It uploaded my kitty test pictures (I can see them in the blob) but now I can't view them.

## Repro
This is an almost out of the box setup, I have cloned 1.10.x at 'b8e138b2d55734c9f6257d08fa944fb48b1676cb' then set up the azure settings and media to be stored in blob. Created a folder called 'cat pictures', dropped three pictures into it and tried to go back to media browser but all pages (folder page and recent page) are showing the display error.

rtpHarry:
Hmm actually it only happens after I have uploaded and clicked "close".

[...]

PR #6792, opened 2016-04-17
6748: Stricter file and folder name validation

### LINKED ISSUE #6748: Stricter file and folder name validation
Windows (NTFS) prohibits the use of the following characters in file and folder names: `"`, `*`, `:`, `<`, `>`, `?`, `|`, `/`, `\`.

Due to the differences in how files are stored in Azure Blob Storage (including the Development Storage of the Storage Emulator) most of these are actually accepted when renaming files in the Media Library when using `Microsoft Azure Media Storage`:
- `"`, `<`, `>`, `|`: Server error when calling the API to query media items ("Illegal characters in path."). For some of the illegal characters, the file name is changed up until the illegal character, but the rest of it is cut off. This error renders the Media Library unusable when opening a folder that contains an affected file.
- `*`, `:`, `?`: The file can't be downloaded to the local file system without changing its name (also true for the previous set of characters).
- `/`: Creates a folder for each slash in the file name (and displayed as such in the Media Library navigation) and the file is at the bottom of the folder structure created by those slashes, but the file is displayed in the correct folder (where it was actually uploaded) in the Media Library (with a file name that has the folder structure defined by the slashes cut off).
- `\`: This is the only illegal character that is handled correctly: A validation error is displayed when saving the file.

**Suggestion**: Trigger a validation error for all illegal characters.

BenedekFarkas:
Fixing that some special characters were accepted in file and folder names.

Reworked the PR to have the following functionality:

- Extended the capability of the `CreateFolder`, `RenameFolder`, `CreateFile` and `RenameFile` functions of `FileSystemStorageProvider` to validate the file/folder name against a larger set of special characters that aren't allowed or considered unsafe.
- The list of characters include those [suggested below](#issuecomment-211580940), except for `~`, which should be allowed.
- This validation now happens on a low level, so removed validation from certain MediaLibrary components and improved the error message handling, so the user is notified of the characters causing trouble.
- Applied the same validation to `AzureFileSystem` too.
- If this PR is accepted, then #7577 can be closed, although I didn't include any client-side validation, IMO we don't win that much with that added complexity (= having to use a regex).


sebastienros:
Can we ask @hannan-azam to test this fix? He's the most knowledgeable about this feature after you ;)

BenedekFarkas:
Sure!

hannan-azam:
Sure, I'll do the testing and will get back with results.

hannan-azam:
@sebastienros @BenedekFarkas During the testing i found another issue with a recent commit 2440e68 (13th April) in media-library.js line 385. Please see the attached screen shot:

1. Script is trying to call a method "contains" which doesn't exist in standard string prototype and causes a javascript error. We should use indexOf method and the return value -1 or positive to write the conditional statement.

2. The Script has a glitch. While using Azure Media Storage, it concludes wrong directory separator for the top level folders. For example in current scenario i had Azure Media Storage turned on. I clicked the top level folder "Users". With the condition `'Users'.contains('/') ? '/' : '\\'` it will always conclude the directory separator to be `\\` instead of `/`.

3. Because of this error, the screen keeps showing the loading icon and stops working.

```
                        var separator = deepestChildPath.contains('/') ? '/' : '\\';
```

4-19-2016 1-48-05 am

BenedekFarkas:
Hm, I tested that one quite extensively though, but you're right about the JS problem anyway. That change btw will be unnecessary once #6756 is fixed - I applied it as a temporary solution, because it was easier to make it work there first (to be able to use the Media Library properly with Azure Storage and test other things).

hannan-azam:
Regarding the actual issue, below are my findings and recommendations:

I read this article https://perishablepress.com/stop-using-unsafe-characters-in-urls/ which references RFC 1738 http://www.ietf.org/rfc/rfc1738.txt with recommendations on not using the characters `blank/empty space and " < > # % { } | \ ^ ~ [ ] `` for the URIs. I have pasted the description from RFC on why these characters are unsafe at the end.

The article also suggests that there are many others characters which should be **URL ENCODED** if they are used in urls. For example character ranges 00-1F hex (0-31 decimal) and 7F, 80-FF hex (128-255 decimal.) Reserved charcaters $ & + , / : ; = ? @. Everything should be URL Encoded.

Following are my recommendations:

1. We are using Path.GetInvalidFileNameChars to get a list of unsafe characters for file system. This is very nice since we make sure that file will have no issues while storing on file system.

2. The HttpUnallowed should include the Unsafe characters mentioned in the RFC. Spaces/Blanks should either be removed or converted to underscores.

3. We should finally URL Encode the filename to make sure that the url will have no issues later on. Azure Blob always does URL Encoding. For example `download §.jpg` will be saved as `download%20%C2%A7.jpg`. Why not make it consistent and we also store the url encoded file name internally as well. The only drawback i see with this approach is that even on File System we will have files with URL Encoded names and not the special chars. But on the other hand using such special chars in file names is a rare scenario and on top of that it will save us from conversions later on.

> Unsafe:
> 
>    Characters can be unsafe for a number of reasons.  The space
>    character is unsafe because significant spaces may disappear and
>    insignificant spaces may be introduced when URLs are transcribed or
>    typeset or subjected to the treatment of word-processing programs.
>    The characters "<" and ">" are unsafe because they are used as the
>    delimiters around URLs in free text; the quote mark (""") is used to
>    delimit URLs in some systems.  The character "#" is unsafe and should
>    always be encoded because it is used in World Wide Web and in other
>    systems to delimit a URL from a fragment/anchor identifier that might
>    follow it.  The character "%" is unsafe because it is used for
>    encodings of other characters.  Other characters are unsafe because
>    gateways and other transport agents are known to sometimes modify
>    such characters. These characters are "{", "}", "|", "\", "^", "~",
>    "[", "]", and "`".
> 
>    All unsafe characters must always be encoded within a URL. For
>    example, the character "#" must be encoded within URLs even in
>    systems that do not normally deal with fragment or anchor
>    identifiers, so that if the URL is copied into another system that
>    does use them, it will not be necessary to change the URL encoding.

sebastienros:
Conflicts. Do we still need it?

BenedekFarkas:
I've been meaning to come back to this. I'll try to check it soon and see what to do.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/MediaLibraryService.cs
```diff
@@ -33,6 +34,8 @@ public class MediaLibraryService : IMediaLibraryService {
             _mediaFactorySelectors = mediaFactorySelectors;
 
             T = NullLocalizer.Instance;
+
+            InvalidCharacters = Path.GetInvalidFileNameChars().Union(HttpUnallowed).OrderBy(c => c).ToArray();
```
sebastienros: Static constructor and field?

src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/MediaLibraryService.cs
```diff
@@ -142,12 +145,12 @@ public class MediaLibraryService : IMediaLibraryService {
 
         public string GetUniqueFilename(string folderPath, string filename) {
 
-            // remove any char which is unallowed in an HTTP request
-            foreach (var unallowedChar in HttpUnallowed) {
-                filename = filename.Replace(unallowedChar.ToString(), "");
+            // Removing characters the are forbidden on the file sytem or unallowed in an HTTP request.
```
agriffard: `the are forbidden on the file sytem` => `that are forbidden on the file system`

BenedekFarkas:
On the latest episode of reviving old PRs, please see the PR description for what changed.
If you can test file system and Azure storage (e.g., with Azurite), that's a bonus (I did, but extra set of eyes help!).

src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/AzureFileSystem.cs
```diff
@@ -260,6 +273,11 @@ public class AzureFileSystem {
             path = ConvertToRelativeUriPath(path);
             newPath = ConvertToRelativeUriPath(newPath);
 
+            if (FileSystemStorageProvider.FileNameContainsInvalidCharacters(Path.GetFileName(newPath), out var invalidCharacters)) {
+                throw new ArgumentException(
+                    string.Format("The new file name contains invalid character(s): {0}", string.Join(", ", invalidCharacters)));
```
sebastienros: If that shows to the user, translate

src/Orchard/FileSystems/Media/FileSystemStorageProvider.cs
```diff
@@ -215,6 +222,13 @@ public class FileSystemStorageProvider : IStorageProvider {
         /// <param name="path">The relative path to the folder to be created.</param>
         /// <exception cref="ArgumentException">If the folder already exists.</exception>
         public void CreateFolder(string path) {
+            // We are dealing with a folder here, but GetFileName returns the last path segment, which in this case is
+            // the folder name.
+            if (FolderNameContainsInvalidCharacters(Path.GetFileName(path), out var invalidCharacters)) {
+                throw new ArgumentException(
+                    T("The directory name contains invalid character(s): {0}", string.Join(", ", invalidCharacters)).ToString());
```
sebastienros: HEre is an example for T

src/Orchard/FileSystems/Media/FileSystemStorageProvider.cs
```diff
@@ -427,6 +458,16 @@ public class FileSystemStorageProvider : IStorageProvider {
             return (di.Attributes & FileAttributes.Hidden) != 0;
         }
 
+        public static bool FolderNameContainsInvalidCharacters(string folderName, out char[] invalidCharacters) {
+            invalidCharacters = folderName.Distinct().Intersect(InvalidFolderNameCharacters).ToArray();
+            return invalidCharacters.Any();
+        }
+
+        public static bool FileNameContainsInvalidCharacters(string folderName, out char[] invalidCharacters) {
+            invalidCharacters = folderName.Distinct().Intersect(InvalidFileNameCharacters).ToArray();
```
sebastienros: string.IndexOfAny()

BenedekFarkas: Addressed in 15dc19ce9c9599496b9502c2666d4817b37c1ff1.

src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/AzureFileSystem.cs
BenedekFarkas: Addressed in 08ae5ddccf2772be2afb2d37fdb7ce6918df0865.
I created a new type of exception to be able to handle such cases correctly and without breaking changes or reaching through abstractions.

MatteoPiovanelli: The strings passed when building those exceptions are still un-translated in AzureFileSystem.cs (while they are localizable in other places). It seems like that message is then never used directly, because in the `catch` statements a different message is "forwarded" as error text: however, I think it would still be better to have these strings localizable, so we can use the exception message directly.

BenedekFarkas: That's what we discussed yesterday with Sebastien and after your comment I dug even deeper into it, but localization doesn't work for `AzureFileSystem`, because it's not a registered type in the DI container, which the `LocalizationModule` relies on to find where translations can be injected. I added an Autofac module to register `AzureFileSystem` and with that, `LocalizationModule` finds it, but translated strings aren't applied still. My guess is that in this case, the Activated handler of this registered component is never called (even if it's registered with `.AutoActivate()`, because it's not injected/used as a dependency (it's a parent class).

I don't think it's worth pursuing this any further.

This is already a big improvement, because:
- This validation is now on the storage provider-level, so it doesn't just apply to Media Library.
- Validation is now applied for the Azure storage provider too with the same logic as the file system.
- Media Library didn't have validation at all possible places in the first place.
- `AzureFileSystem` already had non-localized exception messages, so unless the Azure storage is heavily refactored, we'll just have to live with that.

sebastienros (approved):

PR #6801, opened 2016-04-19
Fixes #5417 External Media (oEmbed) Export/Import Issue

### LINKED ISSUE #5417: External Media (oEmbed) Export/Import Issue
The export and import of OEmbedPart is buggy. Exporting method of this part exports "Source" only. 

```
protected override void Exporting(OEmbedPart part, ContentManagement.Handlers.ExportContentContext context) {
        context.Element(part.PartDefinition.Name).SetAttributeValue("Source", part.Source);
}
```

However the part also contains an indexer which contains very useful information such as the thumbnail, width, height etc which gets lost as it is not exported in the recipe. On a target machine if this recipe is imported the media item is shown with blank thumbnail.

```
public string this[string index] {
    get { return this.As<InfosetPart>().Get("OEmbedPart", index, null); }
    set { this.As<InfosetPart>().Set("OEmbedPart", index, null, value); }
}
```

Any suggestions on how to pull the available Keys and add those to recipe for Export. Googled it and people suggest using reflection to get the Keys from the string indexer.

hannan-azam:
Fixes #5417

### DIFF
--- a/src/Orchard.Web/Core/Contents/Views/Admin/Create.cshtml
+++ b/src/Orchard.Web/Core/Contents/Views/Admin/Create.cshtml
@@ -8,7 +8,7 @@
     Layout.Title = T("New {0}", Html.Raw(typeDisplayName)).Text;
 }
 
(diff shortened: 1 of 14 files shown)


hannan-azam:
@sebastienros Can we also make it part of 1.10.x

infofromca:
@hannan-azam pls also apply to 1.10.x

PR #6809, opened 2016-04-21
add a new method in IDocumentIndex

infofromca:
LuceneDocumentIndex has already implemented this method, but not its interface.
and also it is useful for custom class in custom Module

### DIFF
--- a/src/Orchard/Indexing/IIndexDocument.cs
+++ b/src/Orchard/Indexing/IIndexDocument.cs
@@ -11,7 +11,7 @@ namespace Orchard.Indexing {
         IDocumentIndex Add(string name, int value);
         IDocumentIndex Add(string name, bool value);
         IDocumentIndex Add(string name, double value);
-
+        IDocumentIndex Add(string name, object value);
         /// <summary>
         /// Stores the original value to the index.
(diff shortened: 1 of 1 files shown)


src/Orchard/Indexing/IIndexDocument.cs
```diff
@@ -11,7 +11,7 @@ public interface IDocumentIndex {
         IDocumentIndex Add(string name, int value);
         IDocumentIndex Add(string name, bool value);
         IDocumentIndex Add(string name, double value);
-
+        IDocumentIndex Add(string name, object value);
```
kumards: I kinda feel, the below implementation should be removed from LuceneDocumentIndex.cs. I prefer letting the caller decide how the data type should be handled within the index document. Anyway all it does is stringyfying the object.
```
public IDocumentIndex Add(string name, object value) {
            return Add(name, value.ToString());
        }
```

sebastienros:
I don't like this change for two reasons:
- it's a breaking change
- it doesn't really add any value, because both indexing and consuming the data would need to know how the data is formatted, which means you could just use a `string` directly to format what you want.

What is the reason for the change?

PR #6814, opened 2016-04-22
Add ReSharper support to the Razor view engine

mroskamp:
Previously, ReSharper was unable to resolve views when using Orchard. Someone from the R# team recommended a fix to allow the resolution of views.

Here's the R# recommendation, for reference: https://youtrack.jetbrains.com/issue/RSRP-451935


sebastienros:
Look good to me.
Can you file a bug to link this PR to, so that we can track it has been fixed?

mroskamp:
Sure thing. Here's the issue:  #6815

Issue #6831, opened 2016-04-25
CdnSupportsSsl has no purpose and should be implemented


anoordende:
CdnSupportsSsl has no purpose and should be implemented. The key issue is that some CDN's don't support SSL and in this case the resource should fall back on the local resource, which can naturally be served under the same scheme as the site.

_Note: an example scenario is AzureCDN with a custom domain, even though this is set to change 2016Q2 as per https://feedback.azure.com/forums/169397-cdn/suggestions/1332683-allow-https-for-custom-cdn-domain-names in case you care to vote!)_

**Requirements**

- If the CDN _does_ support SSL, it is greatly encouraged to serve resources from the CDN under SSL, even if the site is not. This also ensures that, when a page such as "Contact Us" is served under SSL, these resources do not have to be reloaded.
- If the CDN does _not_ support SSL but the site is served under SSL, then we should never attempt to load resources from the CDN. The main current browsers will block the resource entirely. The logical fallback is to load the resource locally.

**Specifying Resources**

- When the CDN supports SSL, all that is required is for the URL's passed into SetCdn to start with "https". No need to configure CdnSupportsSsl - its value will be set automatically.
- When the CDN does not support SSL, the full overload of SetCdn should specify "false" for the CdnSupportsSsl parameter. Logically, the URL's passed in start with "http".
- Using protocol relative links is discouraged, if the CDN supports SSL, then always configure the resource's CDN URL's as https.

**Implementation**

1. Pass the Request.IsSecureConnection from CoreShapes as a parameter to the ResourceRequiredContext,
2. If CdnSupportsSsl is not provided, then set it to true if CdnUrl starts with "https",
3. If serving under SSL and CdnSupportsSsl and CDN URL is configured then return CDN URL,
4. If  not serving under SSL and CDN URL is configured then return CDN URL
5. Otherwise return local URL

anoordende:
Pull request #6832

PR #6832, opened 2016-04-25
#6831 Implemented CdnSupportsSsl in Orchard UI Resources

anoordende:
This PR implements **CdnSupportsSsl**  as per the comments on issue #6831.

1. Pass the Request.IsSecureConnection from CoreShapes as a parameter to the ResourceRequiredContext,
2. If CdnSupportsSsl is not provided, then set it to true if CdnUrl starts with "https",
3. If serving under SSL and CdnSupportsSsl and CDN URL is configured then return CDN URL,
4. If not serving under SSL and CDN URL is configured then return CDN URL
5. Otherwise return local URL

Smoketest
(pending)

Build and unit tests pass, including 2 new unit tests added.


sebastienros:
It's a breaking change (minor I agree) and it needs to be evaluated, could you target your PR to dev branch ?

anoordende:
This PR is now superseded by #6848 which targets the dev branch.

Issue #6838, opened 2016-04-26
Workflow Context Tokens Provider


sfmskywalker:
The **Workflows** module currently provides _State tokens_, but there does not seem to be a way to access its _Tokens dictionary_ through a token provider.

Being able to access this token dictionary is useful when for example working with the **SendMail** activity.

Sample use case:

- I have a custom Workflow event (activity) that I trigger using `IWorkFlowManager`, passing in a dictionary (tokens).
- One of these tokens contains a recipient name and email address.
- I want to be able to use these tokens as values for the **To** field in the SendMail activity (see screenshot)

![image](https://cloud.githubusercontent.com/assets/938393/14821230/3fe2f1ba-0bca-11e6-9906-554d7c1e616f.png)

Since we already provide a way to access the State dictionary of the `WorkflowContext`, it seems to make sense to also provide a way to access its Tokens dictionary.

sfmskywalker:
PR: #6839

sebastienros:
Where are the Tokens of the workflow currently used? What's the difference between the State and the Tokens.

sfmskywalker:
Workflow **Tokens** is a dictionary on `WorkflowContext` and are provided to the workflow when triggering a workflow event, and these tokens are used by activities such as `SignalActivity` and `UserTaskActivity`. These tokens are used to bring external information into a workflow instance.

Workflow **State ** is another dictionary on `WorkflowContext` and is used by activities themselves to store state as part of the workflow instance. And example activity using this state dictionary is the `WebRequestActivity` which stores the HTTP response as part of the workflow state.

sebastienros:
Agreed offline with Sipke that this is not necessary, just create custom Token provider for `FormSubmission`.

PR #6839, opened 2016-04-26
Added Workflows.Token token

sfmskywalker:
This PR adds a token called **Workflow.Token:[toke name]**, which enables access to the `WorkflowContext.Tokens` dictionary from workflow activities such as `SendMail`.

This is useful when workflow events are triggered in combination with provided "tokens" (a dictionary of key/valye pairs) and you want to be able to access these values.

Associated issue: #6838


sfmskywalker:
Closing this again as it was decided we don't need this as [mentioned here](https://github.com/OrchardCMS/Orchard/issues/6838#issuecomment-218864776). @li0803 perhaps your use case is different that does warrant a token provider to access workflow tokens, but I think that @sebastienros argued something along the lines of accessing the workflow tokens directly was not a good idea. I forgot the details unfortunately.

PR #6851, opened 2016-04-29
#6850 Added scheme to OutputCacheKey to prevent SSL bypass

### LINKED ISSUE #6850: Outputcache should always vary by scheme to prevent SSL bypass
The Outputcache currently does not differentiate between the scheme (http/https) which will allow a visitor to bypass SSL on pages configured in to be served under https in SecureSocketsLayer module.

**Steps to reproduce issue 2:**

1. Enable OutputCache
2. Enable SecureSocketsLayer
3. Configure a convenient cache time to reproduce the issue, e.g. 600
4. Configure a secure URL, e.g. "~/contact-us"
5. Create a page with the autoroute "contact-us"
6. Important: **LOG OFF**

8. Navigate to the secure URL http://localhost:30321/contact-us
9. You should be redirected to https://localhost:44300/contact-us

11. LOG BACK IN
12. Confirm in Admin/Cache that /contact-us is now cached
13. Important: **LOG OFF**

15. Navigate to the INsecure URL http://localhost:30321/contact-us

**Expected result:** Redirect to Secure URL
**Actual result:** No redirect, serving cached INSecure URL

The reason for this, is that in the first request (step 8) OutputCache does not find the page in its cache therefore the filters are executed. The SSL filter returns a 302 redirecting the browser correctly to the secure URL (step 9), which is subsequently cached and rendered back to the browser.

On the second INSecure request (step 15), the OutputCache now finds the page in its cache, and immediately returns the cached version, however this is the Secure version (from step 9) now being served under http.

There are a few solutions, one of which is putting the SSL filter before the OutputCache, but that just adds more processing to every cached request. By far the simplest solution, possibly also the most hardened, is to add the scheme de facto to the cache key.

Will submit the PR right away.

anoordende:
One-liner to add the scheme (http/https) to the OutputCacheKey. Fixes #6850.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -370,7 +370,7 @@ namespace Orchard.OutputCache.Filters {
             foreach (var p in filterContext.ActionParameters)
                 result.Add("PARAM:" + p.Key, p.Value);
 
-            // Vary by scheme (See #6850)
+            // Vary by scheme.
             result.Add("scheme", filterContext.RequestContext.HttpContext.Request.Url.Scheme);
 
             // Vary by theme.
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -370,6 +370,9 @@ namespace Orchard.OutputCache.Filters {
             foreach (var p in filterContext.ActionParameters)
                 result.Add("PARAM:" + p.Key, p.Value);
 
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -370,6 +370,9 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
             foreach (var p in filterContext.ActionParameters)
                 result.Add("PARAM:" + p.Key, p.Value);
 
+            // Vary by scheme (See #6850)
```
DanielStolt: Remove the issue reference and end the comment in a period (like the one below). I don't think we use the practice of referencing GitHub issues from code (but we could discuss that in a meeting of course).

DanielStolt:
Good to go IMO!

PR #6855, opened 2016-05-02
Fix of removing not current page for breadcrumbs

### LINKED ISSUE #5624: Mistake in constructing the breadcrumbs
There is an option in MenuWidget module: Check to render the current content item as the last element.
If this option is uncheched that Orchard may removes not current content item from breadcrumbs.

It happens because orchard looking for active menu items by the url and considers the menu item as active when the requested page url starts from menu item url. (it's cool, but it does not work correctly with the mentioned option). Then orchard removes last item from breadcrumbs. But last item can be not current (when current page isn't added to a menu).

To fix this mistake:
1.Add a property to Orchard.UI.Navigation.MenuItem class

```
public bool Current { get; set; }
```

2.Change method of Orchard.UI.Navigation.NavigationHelper class

```
private static Stack<MenuItem> SetSelectedPath(IEnumerable<MenuItem> menuItems, HttpRequestBase currentRequest, RouteValueDictionary currentRouteData, bool compareUrls) {
            foreach (MenuItem menuItem in menuItems) {
                Stack<MenuItem> selectedPath = SetSelectedPath(menuItem.Items, currentRequest, currentRouteData, compareUrls);
                if (selectedPath != null) {
                    menuItem.Selected = true;
                    menuItem.Current = false;
                    selectedPath.Push(menuItem);
                    return selectedPath;
                }

                // compare route values (if any) first
                // if URL string comparison is used it means all previous route matches failed, thus no need to do them twice
                bool match = !compareUrls && menuItem.RouteValues != null && RouteMatches(menuItem.RouteValues, currentRouteData);
                bool current = true;

                // if route match failed, try comparing URL strings, if
                if (currentRequest != null && !match && compareUrls) {
                    string appPath = currentRequest.ApplicationPath ?? "/";
                    string requestUrl = currentRequest.Path.StartsWith(appPath) ? currentRequest.Path.Substring(appPath.Length) : currentRequest.Path;

                    string modelUrl = menuItem.Href.Replace("~/", appPath);
                    modelUrl = modelUrl.StartsWith(appPath) ? modelUrl.Substring(appPath.Length) : modelUrl;

                    if (requestUrl.Equals(modelUrl, StringComparison.OrdinalIgnoreCase)) {
                        match = true;
                    }
                    else if ((!string.IsNullOrEmpty(modelUrl) && requestUrl.StartsWith(modelUrl + "/", StringComparison.OrdinalIgnoreCase)))
                    {
                        match = true;
                        current = false;
                    }
                }

                if (match) {
                    menuItem.Selected = true;
                    menuItem.Current = current;

                    selectedPath = new Stack<MenuItem>();
                    selectedPath.Push(menuItem);
                    return selectedPath;
                }
            }

            return null;
        }
```

3.Аdd a check of this property before deleting current page from breadcrumbs (change Display function in Orchard.Core.Navigation.Drivers.MenuWidgetPartDriver class)

```
// inject the current page
                    if (!part.AddCurrentPage && selectedPath != null && result[result.Count - 1].Current)
                    {
                        result.RemoveAt(result.Count - 1);
                    }
```

aipetryakov:
Fix #5624

Last item in breadcrumbs doesn't always represent current page. It's not true for situation when current menu item is found by part of URL (not by full URL).

When add current page option for breadcrumbs is not active. It need to be ensure that breadcrumbs last item represents current page before deleting one.

### DIFF
--- a/src/Orchard.Web/Core/Navigation/Drivers/MenuWidgetPartDriver.cs
+++ b/src/Orchard.Web/Core/Navigation/Drivers/MenuWidgetPartDriver.cs
@@ -100,7 +100,22 @@ namespace Orchard.Core.Navigation.Drivers {
 
                     // inject the current page
                     if (!part.AddCurrentPage && selectedPath != null) {
-                        result.RemoveAt(result.Count - 1);
+                        var lastMenuItem = result[result.Count - 1];
+                        // we need to be ensure that breadcrumbs last item represents current page before deleting one.
+                        if (lastMenuItem.RouteValues != null && NavigationHelper.RouteMatches(lastMenuItem.RouteValues, routeData.Values)) {
+                            result.Remove(lastMenuItem);
+                        }
+                        else {
+                            string appPath = request.ApplicationPath ?? "/";
+                            string requestUrl = request.Path.StartsWith(appPath) ? request.Path.Substring(appPath.Length) : request.Path;
+
+                            string lastMenuItemUrl = lastMenuItem.Href.Replace("~/", appPath);
+                            lastMenuItemUrl = lastMenuItemUrl.StartsWith(appPath) ? lastMenuItemUrl.Substring(appPath.Length) : lastMenuItemUrl;
+
+                            if (requestUrl.Equals(lastMenuItemUrl, StringComparison.OrdinalIgnoreCase)) {
+                                result.Remove(lastMenuItem);
+                            }
+                        }
                     }
 
                     // prevent the home page to be added as the home page and the current page


sebastienros:
Tests should be done also with this Pr : https://github.com/OrchardCMS/Orchard/pull/6899

neTp9c:
I added sign-off-by and also changed the source of commit.

All changes could be wrapper in a function named like IsRepresentCurrentPage and added to NavigationHelper class. But if I do so there will be conflict with #6899. Now it's independent pull requests and they do not affect each other.

BenedekFarkas:
Closing this PR, reason: https://github.com/OrchardCMS/Orchard/issues/5624#issuecomment-524256767

Issue #6860, opened 2016-05-04
Add SASS support to Gulpfile.js


anoordende:
The Gulpfile currently supports LESS, but it would be great to let our SASS users feel some Orchard love too (me for one). This could become increasingly important with the uptake of Bootstrap v4.

Will submit the PR.

anoordende:
Created PR #6861

sebastienros:
Good, I also had to do it on the Orchard2 repository https://github.com/OrchardCMS/Orchard2/blob/master/gulpfile.js

anoordende:
@sebastienros in Orchard2 it seems that the _maxListeners warning is also fixed. Maybe worth adding that to the PR too?

sebastienros:
Feel free. But you can also just keep this PR focused and create a new one in case it blocks it. Up to you.

anoordende:
Agreed, I will leave this PR as is.

PR #6861, opened 2016-05-04
#6860 Added SASS support to Gulpfile.js

anoordende:
This adds SASS support to Gulpfile.js.


thekaveman:
`.sass` is also a valid extension for SASS files (albeit, older and not as popular). From the [`sass` repo](https://github.com/sass/sass)

> The second, older syntax is known as the indented syntax...Files in the indented syntax use the extension `.sass`.

May want to include support for that file extension here as well.

anoordende:
@thekaveman I did consider including .sass but discarded it on the basis of it being older and not as popular ;)

PR #6867, opened 2016-05-06
Fix for profile image creation failing with absolute uri

kumards:
Uri.IsWellFormedUriString fails with absolute file uri validation. So using Uri.TryCreate to validate the uri.


sebastienros:
Looks good.
Can you file a bug to track it's been resolved? Please also explain the context how it happens on the issue.

PR #6870, opened 2016-05-07
#6733: [1.10.x] Fix Comments and Audit Trail issue

### LINKED ISSUE #6733: NHibernate IdentifierGenerationException when posting comment to content item of custom content type

jtkech:
Fixes #6733 

@sebastienros here the best compromise i've found to fix the issue without breaking a first original one.

**Summary**
- If audit trail enabled, on a new comment errors are logged and bad event records are created. The audit trail handler rely on a created item and the comment controller creates it after model validation.

- A previous PR #6734 proposed to do a create before update editor (as in other controllers) but, for comments with many items / spams, it's good to preserve transaction logs when model validation fails.

- Another concern is not to conflict with another issue when sending an e-mail on comment publication, the model need to be validated. That's why the create was originally moved after model checking.

**Best compromise i've found**

- `CommentController.cs`: The create is still after the model validation but with a draft option, so with no publish event, then a 2nd update editor is done before publishing. So, it doesn't conflict with the original issue, and events are triggered in the standard order (good for audit trail).

- `GlobalContentHandler.cs`: In the updating and updated methods, before creating a bad audit trail record (e. version = 0) and logging errors, we check if the item is already created.

- `DiffGramAnalyzer.cs`: Here a null check has been added to prevent admin audit trail views from failing if some bad audit trail records have been created before the fix.

Best.


sebastienros:
This one needs to go in dev as it might impact other module we don't know of. Like Workflows.

PR #6876, opened 2016-05-08
Adding warning notification if Redis configuration string is not set …

Skrypt:
…instead of throwing an exception and making the whole admin/website crash.

orchard manage aliases manage aliases


src/Orchard.Web/Modules/Orchard.Redis/OutputCache/RedisOutputCacheStorageProvider.cs
```diff
@@ -54,17 +54,19 @@ public class RedisOutputCacheStorageProvider : IOutputCacheStorageProvider {
         }
 
         public void Set(string key, CacheItem cacheItem) {
-            if(cacheItem == null) {
-                throw new ArgumentNullException("cacheItem");
-            }
+            if (_connectionMultiplexer != null) {
```
sebastienros:
```
if (_connectionMultiplexer == null) {
 return;
}
```

src/Orchard.Web/Modules/Orchard.Redis/Configuration/RedisNotificationProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public RedisNotificationProvider(ShellSettings shellSettings) {
+            _shellSettings = shellSettings;
+            T = NullLocalizer.Instance;
+        }
+
+        public Localizer T { get; set; }
+
+        public IEnumerable<NotifyEntry> GetNotifications() {
+            var _defaultSettingsKey = "Orchard.Redis.MessageBus";
```
sebastienros: Just call `GetConnectionString()` and check the `null` result.

src/Orchard.Web/Modules/Orchard.Redis/OutputCache/RedisTagCache.cs
```diff
@@ -28,18 +28,25 @@ public class RedisTagCache : ITagCache {
         }
 
         public void Tag(string tag, params string[] keys) {
-            Database.SetAdd(GetLocalizedKey(tag),  Array.ConvertAll(keys, x=> (RedisValue) x));
+            if (_connectionMultiplexer != null) {
```
sebastienros: Same here

src/Orchard.Web/Modules/Orchard.Redis/OutputCache/RedisTagCache.cs
```diff
@@ -28,18 +28,25 @@ public class RedisTagCache : ITagCache {
...
+                Database.SetAdd(GetLocalizedKey(tag), Array.ConvertAll(keys, x => (RedisValue)x));
+            }
         }
 
         public IEnumerable<string> GetTaggedItems(string tag) {
-            var values = Database.SetMembers(GetLocalizedKey(tag));
-            if (values == null || values.Length == 0)
-                return Enumerable.Empty<string>();
-            return Array.ConvertAll(values, x => (string) x);
+            if (_connectionMultiplexer != null) {
```
sebastienros: Same here

src/Orchard.Web/Modules/Orchard.Redis/OutputCache/RedisTagCache.cs
```diff
@@ -28,18 +28,25 @@ public class RedisTagCache : ITagCache {
...
+                if (values == null || values.Length == 0)
+                    return Enumerable.Empty<string>();
+                return Array.ConvertAll(values, x => (string)x);
+            }
+            return new string[0];
         }
 
         public void RemoveTag(string tag) {
-            Database.KeyDelete(GetLocalizedKey(tag));
+            if (_connectionMultiplexer != null) {
```
sebastienros: Same here

Skrypt:
Done

sebastienros:
Can you check if INotificationProvider needs to check if it's in Admin, or if the service which calls it already does that?

Skrypt:
The service already does that. The notification only appears in the admin. Though I added the configuration validation also in the OutputCache folder *.cs files so that it does'nt try to return any cached items.

Skrypt:
Here since there can be multiple connection strings that are defined by the services that calls GetConnectionString(). I added in the Notifier the injection of FeatureManager to validate if each features are enabled. Then I can check if there is a connection string configured per feature. Though, if someone implements a custom feature it will need to implement it's own INotificationProvider.

Another better way of doing this could be to list the enabled features that have a dependency on the Redis feature ... and loop through them to validate if they have a connection string set. Though the connection string would need to have the same name e.g. "Orchard.Redis.OutputCache" than it's feature name. Wich right now is not the case.

sebastienros:
The correct way is to have one notification for each feature, which also has the [ORchardFEature] attribute so that it's enabled with the feature.

sebastienros:
And the other services should check if the connection string is null before doing anything, like you did for the first service.

Skrypt:
Thanks for the precisions. It is now using multiple notifiers. Maybe someone should test it before merging ?

src/Orchard.Web/Modules/Orchard.Redis/Services/RedisCacheNotificationProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+            _redisConnectionProvider = redisConnectionProvider;
+            _featureManager = featureManager;
+            T = NullLocalizer.Instance;
+        }
+
+        public Localizer T { get; set; }
+
+        public IEnumerable<NotifyEntry> GetNotifications() {
+            //verify if module is enabled first
+            if (!_featureManager.GetEnabledFeatures().Where(f => f.Id == "Orchard.Redis.Caching").Any()) {
```
sebastienros: Wrong, you just need [OrchardFeature("Orchard.Redis.Caching")] on the class. Look at the service for the same feature for an example.

src/Orchard.Web/Modules/Orchard.Redis/Services/RedisMessageBusNotificationProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+            _redisConnectionProvider = redisConnectionProvider;
+            _featureManager = featureManager;
+            T = NullLocalizer.Instance;
+        }
+
+        public Localizer T { get; set; }
+
+        public IEnumerable<NotifyEntry> GetNotifications() {
+            //verify if module is enabled first
+            if (!_featureManager.GetEnabledFeatures().Where(f => f.Id == "Orchard.Redis.MessageBus").Any()) {
```
sebastienros: Same thing, [OrchardFeature] attribute instead

src/Orchard.Web/Modules/Orchard.Redis/Services/RedisOutputCacheNotificationProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+            _redisConnectionProvider = redisConnectionProvider;
+            _featureManager = featureManager;
+            T = NullLocalizer.Instance;
+        }
+
+        public Localizer T { get; set; }
+
+        public IEnumerable<NotifyEntry> GetNotifications() {
+            //verify if module is enabled first
+            if (!_featureManager.GetEnabledFeatures().Where(f => f.Id == "Orchard.Redis.OutputCache").Any()) {
```
sebastienros: Idem

src/Orchard.Web/Modules/Orchard.Redis/Services/RedisCacheNotificationProvider.cs
Skrypt: Much cleaner :+1:

sebastienros:
Looks great to me.

PR #6892, opened 2016-05-12
Introduces a contextual HtmlFilter.

### LINKED ISSUE #6193: ContentHandler TokenFilter and Element TokenFilter can't cohabit
**Repro** on dev branch
- In an html widget, use a token that need Content data, e.g #{ Content.Id }.
- So, enable the `Tokens Html Filter` feature (will be used by the body part).
- In an Layout element (text, html...) use such a token, e.g the same as above.
- So, enable the `Element Tokens` feature used by the html, text... elements.

**Result**
- The token in the html widget is not evaluated correctly.
- Note: may depends on the token filters ordering (see below).

**Analyse**

The 1st TokensFilter inherit from ContentHandler, not the 2nd one (IElementFilter), context data are not passed in the same way and can be empty. So, a first running filter can kill the token.

Maybe the 1st filter (that is an html filter) should not be coupled to a content type by exposing a ProcessContent() method with a data context parameter... More details in my last comment here #6155.

**Fix**

To allow such filters to cohabit, in Tokenizer.cs, I've found a fix by testing if the replacement value is null and then do nothing on the token. I will do a PR.

Best

sfmskywalker:
The contextual HtmlFilter can optionally accept context in the form of a dictionary.
Callers can provide any data they have, and the filters can optionally do something with this.
An example is the TokensFilter that uses "Content" if specified.

Fixes #6193


jtkech:
I've tried it and **it works for me**.

So, now only one tokens filter and the one which uses the default html encode option!

But i remember i had drawbacks with filter execution order. E.g when the `TextFieldFilter` (which html encodes) runs after the `TokensFilter` (which already has encoded the token). I couldn't repro in my tests, the tokens filter always run last, but i think it may depend on modules dependencies.

But dealing with the feature priorities, by setting a priority of `-1` to the tokens filter feature, i could simulate the issue where the token is encoded twice. So, here **i propose to add a positive priority** to the `Orchard.Tokens.HtmlFilter` feature (only for this feature) in `Orchard.Tokens/Module.txt`.

I would propose a priority > 5 because many modules depend on `Orchard.Resources` that already have a priority of 5. But because there are not so much filters that html encodes, and currently, as i remember in the source code, only the `TextFieldFilter` in the `Core/Common` module, **maybe a priority of 1** is sufficient.

Sorry for the lenght, best.

Best.

sfmskywalker:
Thanks for trying it out!

Regarding the double encoding issue depending on the filter execution order, I think it makes sense for filters to be configured in order of priority (leveraging the module priority feature). At least I don't really see another clever way of preventing filters from encoding tokens. The Tokens filter really needs to execute first.

jtkech:
@sfmskywalker 

Indeed no perfect solution and maybe not so important. But after many tests some more infos:

- You pointed out the case where the token itself (not its resolved value) has special characters, e.g `#{Content.DisplayText.TrimEnd:é}`. Here, as you said, the token filter would need to run first before being encoded and then not resolved. But i was assuming such a token is not allowed because e.g in an html element, anyway, it will always be first encoded by the tinymce editor.

- So, i was talking about the case where special characters are in the token resolved value. E.g with this page title `Page 1 ééé` and this token `#{Content.DisplayText}`. So, here, the token filter, which encodes its resolved value (default option), **would need to run last**! Or at least not before another filter which may encode the resolved value a 2nd time.

- That's why i proposed to add a positive priority for the token filter (only for this feature in Module.txt), not to run first but to run last or at least to be sure it runs e.g after the `TextFieldFilter `. Here **higher is the priority later is the execution**!

Best

jtkech:
@sfmskywalker just for infos here the best solution i found and tested.

Finally the idea, as you said, is to force the tokens filter to run first (here by giving it a negative priority). Then, if the flavor = `html`, because the content is supposed to be already encoded, the tokens filter needs to encode its resolved value. Otherwise not, it's the responsability of a subsequent filter.

So, in Module.txt:

    Orchard.Tokens.HtmlFilter:
        Name: Tokens Html Filter
        Description: Evaluates tokens in a body.
        Category: Content
        Dependencies: Orchard.Tokens
        Priority: -1

And in TokensFilter.cs:

            return _tokenizer.Replace(text, context.Data,
                String.Equals(context.Flavor, "html", StringComparison.OrdinalIgnoreCase)
                ? new ReplaceOptions { Encoding = ReplaceOptions.HtmlEncode }
                : new ReplaceOptions { Encoding = ReplaceOptions.NoEncode });

Best

sfmskywalker:
@jtkech That is great work, I think this makes a lot of sense. Would it be possible for you to create a PR for this issue branch? Thanks

sebastienros:
Status: waiting for JT's suggestion to get in

jtkech:
@sebastienros, @sfmskywalker 

Done here #6909.

sebastienros:
dev branch

sfmskywalker:
I created a new PR targeting dev.

PR #6899, opened 2016-05-14
When selected path is being looked for menu by comparing string-based url It should select most nested url

neTp9c:
Marking menu item as active is not always right.

It's wrong for situation when first element of second menu level and element of first menu level links to same url:
  * homepage (~/)
  * about us (~/about-us)
    * about as (~/about-us)
    * our news (~/about-us/our-news)
      * news 1 (~/about-us/our-news/news-1)
      * news 2 (~/about-us/our-news/news-2)
  * contacts

News 1 and News 2 are not actually added to menu but have respective url.

When current (routed) page is "news 1" then "about as" page from second menu level is marked active (instead of "our news" page)

This pull request solves that problem.

### DIFF
--- a/src/Orchard/UI/Navigation/NavigationHelper.cs
+++ b/src/Orchard/UI/Navigation/NavigationHelper.cs
@@ -80,20 +80,12 @@ namespace Orchard.UI.Navigation {
         /// <param name="currentRouteData">The current route data.</param>
         /// <returns>A stack with the selection path being the last node the currently selected one.</returns>
         public static Stack<MenuItem> SetSelectedPath(IEnumerable<MenuItem> menuItems, HttpRequestBase currentRequest, RouteValueDictionary currentRouteData) {
-            // doing route data comparison first
-            var path = GetSelectedPath(menuItems, currentRouteData);
+            // doing route data comparison first and if that fails, fallback to string-based URL lookup
+            var path = SetSelectedPath(menuItems, currentRequest, currentRouteData, false) 
+                    ?? SetSelectedPath(menuItems, currentRequest, currentRouteData, true);
 
-            // if route data comparison fails, fallback to string-based URL lookup
-            if (path == null)
-            {
-                var currentUrl = GetUrl(currentRequest.Path, currentRequest.ApplicationPath);
-                path = GetSelectedPath(menuItems, currentUrl, currentRequest.ApplicationPath);
-            }
-
-            if (path != null)
-            {
-                foreach (var menuItem in path)
-                {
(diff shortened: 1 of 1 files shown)


sebastienros:
Please remove the reformat, we can't see what the actual changes are.

neTp9c:
It was not a simple reformat. I split one function into three (it's more like refactoring).
I do it because finding active menu item by url-based string and route values looks like independent methods for me (current method has too many if-else clauses).
Anyway I joined they together just as it was before.

sebastienros:
It needs sign-off and also testing the other part of the code to ensure it's not breaking other scenarios.

neTp9c:
I added sign-off-by.
I don't know what I can do about testing. If you give me direction I will test it.
But I can't imagine any right scenario that could be broken by this code. Picks the menu item with the most nested URL (when using URL comparison) looks best choice for me.

If some one ordered his menu items in a reverse order:
- level-1/level-2
    - level-1
and current page is level-1/level-2/level-3 that in this case only level-1/level-2 item will be marked as active.

sebastienros:
> I can't imagine any right scenario that could be broken by this code

This is one of the funniest comments ever ;) 

> I don't know what I can do about testing

The sign-off label is for someone else than the creator of the PR to test the feature. It's always good to have someone else look at it, because even the most simple changes might break something.

src/Orchard/UI/Navigation/NavigationHelper.cs
```diff
@@ -96,12 +102,17 @@ public static class NavigationHelper {
         /// <param name="compareUrls">Should compare raw string URLs instead of route data.</param>
         /// <returns>A stack with the selection path being the last node the currently selected one.</returns>
         private static Stack<MenuItem> SetSelectedPath(IEnumerable<MenuItem> menuItems, HttpRequestBase currentRequest, RouteValueDictionary currentRouteData, bool compareUrls) {
+            var selectedPathes = new List<Stack<MenuItem>>();
```
rtpHarry: typo, should just be `selectedPaths`

neTp9c:
:)
I fixed the typo.

sfmskywalker:
I'm running into an issue that this PR seems to fix, so I'll take it for a test drive.

sfmskywalker:
I just tested it and it works nicely! The only thing I would change is rename the private `SetSelectedPath` to `GetSelectedPath`, since it's no longer setting `menuItem.Selected = true`; Instead, the public `SetSelectedPath` is doing that now.

GTG

PR #6909, opened 2016-05-19
Tokens filer execution order & html encode

jtkech:
Complement of #6892, related to html filters.


sfmskywalker:
@sebastienros @jtkech It appears this PR was merged into an "issue" branch of Orchard rather than the dev branch. Is that on purpose or by mistake? Looking at the current code in the dev branch, the token filter issue is still there.

jtkech:
Oh yes, don't remember the details and i first thought it is a mistake.

But taking a look in #6938 i can see one of your comments.

>This also includes @jtkech 's work on fixing an encoding issue and execution order of the TokenFilter.

So, maybe intended and we need to finalize and merge #6938. But looking at the code here, i think we can safely use it separately.

PR #6914, opened 2016-05-22
fixes #6913 - extend snippet manifest format

### LINKED ISSUE #6913: Orchard.Layouts - Extend snippet manifest format with category & description
While finishing off implementing the snippet manifest parsing in #6872 I noticed a few other improvements that could be made.

1. Move the default / fallback snippet `ToolboxIcon` value into the `Snippet` class and pull it from there.

1. Add a `Category` field to the yaml manifest to allow the snippets to blend in to whatever category the author chooses.

1. Add a `Description` field to the yaml manifest to customize the existing hard coded description `String.Format("An element that renders the {0} shape.", shapeType)`

As they will extend the manifest further I thought I would get approval before I start. Also, which branch should I target?

rtpHarry:
fixes #6911 - fully parse existing manifest format
fixes #6913 - extend snippet manifest format

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Elements/Snippet.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Elements/Snippet.cs
@@ -1,5 +1,6 @@
 ﻿using Orchard.Environment.Extensions;
 using Orchard.Layouts.Framework.Elements;
+using Orchard.Localization;
 
 namespace Orchard.Layouts.Elements {
     [OrchardFeature("Orchard.Layouts.Snippets")]
@@ -14,6 +15,10 @@ namespace Orchard.Layouts.Elements {
 
         public override bool HasEditor {
             get { return false; }
+        }
+
+        public override string ToolboxIcon {
+            get { return "\uf10c"; }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


thekaveman:
Just wanted to see if this would make it into `1.10.x` as well? The snippet manifest feature was merged there in a5605b7, would be nice to have the expanded functionality.

rtpHarry:
I was told to target dev branch because it makes changes to the manifest file. It's actually a harmless extension to the manifest format and all of the fields parsed are optional. Also it fixes a bug which is present in the system at the moment.

You can drop the two modified files into your local 1.10.x branch and they will work no problems.

I am letting @sfmskywalker direct me with this one as I still don't understand all the nuances between deciding which branch to target but if the whole pr isn't appropriate for the 1.10.x branch I can easily create a second scaled back pr to only implement the bugfix and #6911 fields.

sfmskywalker:
The rule of thumb is that fixes should go into the "maintenance" branch of the latest release (being **1.10.x** currently), while new features go into the **dev** branch.

The snippets manifest file clearly is a new feature, but the reason it was added to the 1.10.x branch is to fix an issue with snippet field discovery in the 1.10 release.

Although this goes against the rule of thumb, it fixes something we made an exception. The compromise was to only add the bare essentials necessary to fix the issue at hand.

Therefore, the rest of the feature should be implemented on the dev branch.

Hope this clarifies things a bit.

sfmskywalker:
After discussing this on today's triage meeting, @sebastienros pointed out to me that the rule of thumb a long time ago, so I am retracting my previous statement.

To summarize, new features can go into the maintenance branch as long as they are not breaking.

So please go ahead and re-target the 1.10.x branch :)

rtpHarry:
closing to retarget 1.10.x

PR #6919, opened 2016-05-25
Fixing that renaming Media Library files to a reserved Windows name should cause an error, fixes #6802

### LINKED ISSUE #6802: ValidatePath problem with files named "con.(something else)".
When I call ValidatePath(@"D:\ConsoleApplication1\ConsoleApplication1\",@"D:\ConsoleApplication1\ConsoleApplication1\bin\Debug\con.img.jpg") I get an ArgumentException("Invalid path")

Problem is I have found .Net method Path.GetFullPath doesn't work properly with filenames equals to  "con" or that start with "con.". This could be because "con" is something of the operating system (not sure)
Here is a sample of the output I get calling Path.GetFullPath:

```
Console.WriteLine(Path.GetFullPath("con")); //Output: \\.\con
            Console.WriteLine(Path.GetFullPath("con.img.jpg")); //Output: \\.\con
            Console.WriteLine(Path.GetFullPath("congress.img.jpg")); //Output: D:\ConsoleApplication1\ConsoleApplication1\bin\Debug\congress.img.jpg

```

So I would need to add a patch here: https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard/Validation/PathValidation.cs to support files with "con.". Any proposal that doesn't compromise security?

jchenga:


src/Orchard.Web/Modules/Orchard.MediaLibrary/MediaFileName/MediaFileNameDriver.cs
```diff
@@ -61,9 +62,12 @@ public class MediaFileNameDriver : ContentPartDriver<MediaPart> {
                                 try {
                                     _mediaLibraryService.RenameFile(part.FolderPath, priorFileName, model.FileName);
                                     part.FileName = model.FileName;
-                                    
+
                                     _notifier.Add(NotifyType.Information, T("File '{0}' was renamed to '{1}'", priorFileName, model.FileName));
                                 }
+                                catch (InvalidWindowsPathException) {
```
sebastienros: You can use `OrchardException` here.

src/Orchard/Validation/PathValidation.cs
```diff
@@ -26,10 +31,22 @@ public static class PathValidation {
...
             }
 
             return mappedPath;
         }
     }
+
+    /// <summary>
+    /// Thrown when an invalid path is encountered on a Windows platform.
+    /// </summary>
+    [Serializable]
```
sebastienros: Not necessary as we can use `OrchardException`

src/Orchard/Validation/PathValidation.cs
```diff
@@ -26,10 +31,22 @@ public static class PathValidation {
             }
 
             if (!valid) {
-                throw new ArgumentException("Invalid path");
+                if (System.Environment.OSVersion.Platform == PlatformID.Win32NT)
```
sebastienros: You don't need to check the environment

src/Orchard/Validation/PathValidation.cs
```diff
@@ -1,11 +1,16 @@
 ﻿using System;
 using System.IO;
+using System.Runtime.Serialization;
```
sebastienros: Remove unnecessary using

jchenga:
@sebastienros 
The environment check would have been necessary if Orchard also ran on Linux because the reported issue is only applicable to Windows platform. Since Orchard doesn't run on Linux platform, the check and the exception are redundant. 

I think I understand what you have in mind now:
1. Throw an OrchardException in ValidatePath
2. Catch the OrchardException in MediaFileNameDriver and display a different error message.

I'll send in another pull request.

sebastienros:
Don't send another PR, just add a new commit to your current branch. Thanks.

jchenga:
To use OrchardException, I need to pass in a localizedString object. How do I do that in PathValidation class?

sebastienros:
Pass a `new LocalizedString()` but use a call to `T()` if you need to display something when you catch it.

jchenga:
OK. done. sorry it took a while.

src/Orchard/Validation/PathValidation.cs
lbcsy: Throwing OrchardException breaks unit testOrchard.Framework.Tests.Orchard.Tests.Storage.FileStystemStorageProviderTests.CreateFileFailsInInvalidPath.

Issue #6924, opened 2016-05-27
Executing recipes shows success even recipe fails


kappas:
Navigate to Admin UI/Modules/Recipes and click "execute" link

If recipe has any errors, there are no error messages showing. However logs show that it failed eg. some invalid step

If same recipe is run via Import, error message is showing.

Seems that when refactoring for recipe execution was made, execute recipe was not modified.

rtpHarry:
Is this the same issue as #4207?

kappas:
I think so

PR #6925, opened 2016-05-27
Show recipe execution errors in Admin UI #6924

kappas:
This fix modifies recipe execution from Admin UI to use RecipeResultAccessor service to show possible errors in recipe execution  #6924

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Modules/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Modules/Controllers/AdminController.cs
@@ -34,6 +34,7 @@ namespace Orchard.Modules.Controllers {
         private readonly IRecipeManager _recipeManager;
         private readonly ShellDescriptor _shellDescriptor;
         private readonly ShellSettings _shellSettings;
+        private readonly IRecipeResultAccessor _recipeResultAccessor;
 
         public AdminController(
             IEnumerable<IExtensionDisplayEventHandler> extensionDisplayEventHandlers,
@@ -44,6 +45,7 @@ namespace Orchard.Modules.Controllers {
             IFeatureManager featureManager,
             IRecipeHarvester recipeHarvester,
             IRecipeManager recipeManager,
+            IRecipeResultAccessor recipeResultAccessor,
             ShellDescriptor shellDescriptor,
             ShellSettings shellSettings,
             IShapeFactory shapeFactory)
@@ -56,6 +58,7 @@ namespace Orchard.Modules.Controllers {
             _featureManager = featureManager;
             _recipeHarvester = recipeHarvester;
             _recipeManager = recipeManager;
+            _recipeResultAccessor = recipeResultAccessor;
             _shellDescriptor = shellDescriptor;
             _shellSettings = shellSettings;
             Shape = shapeFactory;
@@ -112,7 +115,7 @@ namespace Orchard.Modules.Controllers {
             });
         }
 
-        public ActionResult Recipes() {
+        public ActionResult Recipes(string executionId) {
             if (!Services.Authorizer.Authorize(StandardPermissions.SiteOwner, T("Not allowed to manage modules")))
                 return new HttpUnauthorizedResult();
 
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Modules/Controllers/AdminController.cs
```diff
@@ -155,17 +172,19 @@ public class AdminController : Controller {
...
             }
 
-            try {
-                _recipeManager.Execute(recipe);
+            var executionId = _recipeManager.Execute(recipe);
+
+            if (String.IsNullOrEmpty(executionId))
+            {
+                Logger.Error("Error while executing recipe {0} in {1}", moduleId, name);
+                Services.Notifier.Error(T("Error while executing recipe {0} in {1}", moduleId, name));
```
DanielStolt: Trailing period on these two messages would be nice.

DanielStolt:
Nice improvement!

src/Orchard.Web/Modules/Orchard.Modules/Controllers/AdminController.cs
```diff
@@ -155,17 +172,19 @@ public class AdminController : Controller {
...
 
-            try {
-                _recipeManager.Execute(recipe);
+            var executionId = _recipeManager.Execute(recipe);
+
+            if (String.IsNullOrEmpty(executionId))
+            {
+                Logger.Error("Error while executing recipe {0} in {1}.", moduleId, name);
+                Services.Notifier.Error(T("Error while executing recipe {0} in {1}.", moduleId, name));
             }
```
DanielStolt: I think the parameters are reversed on these two (moduleId vs. name)? I see this was the case also in the old code, but now it's also for the notifier, so we might as well fix it.

kappas: good catch!

src/Orchard.Web/Modules/Orchard.Modules/Controllers/AdminController.cs
```diff
@@ -132,6 +135,20 @@ public class AdminController : Controller {
                 .ToList();
             }
 
+            if (!string.IsNullOrEmpty(executionId))
+            {
```
sebastienros: Formatting

src/Orchard.Web/Modules/Orchard.Modules/Controllers/AdminController.cs
```diff
@@ -132,6 +135,20 @@ public class AdminController : Controller {
...
+                var result = _recipeResultAccessor.GetResult(executionId);
+
+                if (result.IsSuccessful)
+                {
+                    Services.Notifier.Information(T("The recipe was executed successfully. Check the logs (recipe execution ID <strong>{0}</strong>) for more information.", result.ExecutionId));
```
sebastienros: Use `.Success`.
Remove the details about this execution ID -> `The recipe was executed successfully`

DanielStolt: @sebastienros I disagree - displaying the execution ID is the only way you can reasonably correlate what you just did to a potentially huge log. Why do you want to remove it?

sebastienros: Because everything went well, I am not suggesting to remove it when there was an issue.

kappas: Would it be useful to redirect to separate result page like on Orchard.ImportExport module when import is executed? That way it would be more convienent to show details of recipe execution results both on success and failure.

sebastienros: That might be the best solution, but we might need to move this screen to the Recipes modules then.

DanielStolt: Ah, I missed that aspect - then I am not opposed to removing it (but nor am I opposed to keeping it). ;)

kappas:
Closing this in favor of #6957

PR #6931, opened 2016-05-30
Fix for getting other content item localzation.

aipetryakov:
LocalizationPartRecord for specific content usually looks like:

| Id | CultureId | MasterContentItemId |
| --- | --- | --- |
| 25 | 2 | 0 |
| 1236 | 3 | 25 |
| 1237 | 4 | 25 |

And when someone try to get another localization item for non-master content item that function returns nothing.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Localization/Services/LocalizationService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Localization/Services/LocalizationService.cs
@@ -29,11 +29,17 @@ namespace Orchard.Localization.Services {
             if (localized == null)
                 return null;
 
+            if (localized.Record.CultureId == cultureRecord.Id) {
+                return localized;
+            }
+
+            int masterContentItemId = localized.HasTranslationGroup ? localized.Record.MasterContentItemId : localized.Id;
+
             // Warning: Returns only the first of same culture localizations.
             return _contentManager
                 .Query<LocalizationPart>(versionOptions, content.ContentItem.ContentType)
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Localization/Services/LocalizationService.cs
```diff
@@ -29,11 +29,17 @@ public class LocalizationService : ILocalizationService {
             if (localized == null)
                 return null;
 
+            if (localized.Record.CultureId == cultureRecord.Id) {
```
sebastienros: Here you can check for the record equality directly like `localized.Record == cultureRecord` but it's not very important, this change is already nice.

sebastienros:
Maybe the only change to do is to actually change the query to

`(l.Id == content.ContentItem.Id || l.MasterContentItemId == content.ContentItem.Id || l.MasterContentItemId == content.Record.MasterContentItemId)`

neTp9c:
I removed the check for equality a seeking and a current content culture. It was added for performance, but i did not do any tests for it.

`(l.Id == content.ContentItem.Id || l.MasterContentItemId == content.ContentItem.Id || l.MasterContentItemId == localized.Record.MasterContentItemId)`
This code will not find master culture by non-master one.

It should be extend to 
`(l.Id == content.ContentItem.Id || l.MasterContentItemId == content.ContentItem.Id || l.MasterContentItemId == localized.Record.MasterContentItemId || l.Id == localized.Record.MasterContentItemId)`

But i think that It looks more complicated.

PR #6935, opened 2016-05-31
Add SiteName to New Account Subject

dalenewman:
When you run more than one Orchard CMS site, and you have to moderate the logins, it's nice to know which site a new account was created on.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Services/MembershipService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Services/MembershipService.cs
@@ -118,7 +118,7 @@ namespace Orchard.Users.Services {
                         template.Metadata.Wrappers.Add("Template_User_Wrapper");
 
                         var parameters = new Dictionary<string, object> {
-                            {"Subject", T("New account").Text},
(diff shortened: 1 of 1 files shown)


infofromca:
I think this fix should also go to 1.10.x

src/Orchard.Web/Modules/Orchard.Users/Services/MembershipService.cs
```diff
@@ -118,7 +118,7 @@ public class MembershipService : IMembershipService {
                         template.Metadata.Wrappers.Add("Template_User_Wrapper");
 
                         var parameters = new Dictionary<string, object> {
-                            {"Subject", T("New account").Text},
+                            {"Subject", T("New account on "  + _orchardServices.WorkContext.CurrentSite.SiteName).Text},
```
sebastienros: The correct way to do it is:
`T("New account on {0}.", _orchardServices.WorkContext.CurrentSite.SiteName).Text`

Otherwise you could not translate it.

sebastienros:
It's changing a translation, not sure it can go in 1.10.x. But it's not breaking anything.
Can you file an issue so we can track the change for the release notes?

dalenewman:
Created issue #6989.  Sorry for the delay.

PR #6938, opened 2016-05-31
#6193: IHtmlFilter and TokenFilter improvements and bugfixes

### LINKED ISSUE #6193: ContentHandler TokenFilter and Element TokenFilter can't cohabit

sfmskywalker:
This changeset unifies the two separate TokenFilter implementations (one for BodyPart, TextField, etc and another one for certain elements such as Html).
It also fixes a bug with the TokenFilter when executing for HtmlWidget, where the wrong content item is being recorded by the handler (the original implementation).
Thirdly, it removes awkward code repetition by moving filter execution into a dedicated HtmlFilterRunner service.

Fixes #6193


sfmskywalker:
This also includes @jtkech 's work on fixing an encoding issue and execution order of the TokenFilter.

src/Orchard.Web/Modules/Orchard.Tags/Feeds/TagFeedQuery.cs
```diff
@@ -1,34 +1,27 @@
...
-using Orchard.Utility.Extensions;
 using Orchard.Core.Feeds;
 using Orchard.Tags.Services;
 using Orchard.Localization;
 using System.Web.Routing;
 using Orchard.Environment.Extensions;
 
-namespace Orchard.Tags.Feeds {
+namespace Orchard.Tags.Feeds
+{
```
sebastienros: What is that ?

src/Orchard/Services/IHtmlFilterRunner.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+using Orchard.ContentManagement;
+
+namespace Orchard.Services {
+    public interface IHtmlFilterRunner : IDependency {
```
sebastienros: I don't like the name `Runner`. We don't have such things right now. We have `Processor`, `Coordinator`, probably other choices but I don't like `runner`.

sebastienros:
Minus the comments it looks good.

jtkech:
Does this one only waiting for a name and a curly brace rule correction?
And maybe better to target 1.10.x?

Best.

BenedekFarkas (approved):

src/Orchard.Web/Modules/Orchard.Tags/Feeds/TagFeedQuery.cs
BenedekFarkas: Fixed in d95f81259e96f5e23c6c263cb49d9913fb45b5cf

BenedekFarkas:
@sebastienros tested, good to go!

Issue #6941, opened 2016-06-01
CdnSupportsSsl + Use CDN doesn't affect protocol relative URLs


thekaveman:
This is somewhat of a follow on to #6831 (and likewise targets `dev`)

We have, for example in [`Orchard.Resources/Bootstrap`](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Resources/Bootstrap.cs#L7) (changed spacing for readability here):

``` csharp
manifest
    .DefineStyle("Bootstrap")
    .SetUrl("bootstrap.min.css", "bootstrap.css")
    .SetVersion("3.3.5")
    .SetCdn(
        "//ajax.aspnetcdn.com/ajax/bootstrap/3.3.5/css/bootstrap.min.css",
        "//ajax.aspnetcdn.com/ajax/bootstrap/3.3.5/css/bootstrap.css",
        true
    );
```

`cdnSupportsSsl` is given `true`, while the URLs are protocol relative (`//domain.com`).

When the **Use CDN** setting is checked, these resources will be served under the protocol relative URL, even though they are available under `https`. As @anoordende mentioned in the linked issue, and pretty widely accepted best-practice:

> If the CDN does support SSL, it is greatly encouraged to serve resources from the CDN under SSL, even if the [Orchard] site is not [served under SSL].

The quick fix is to make all the CDN URLs `https` for CDNs that support it.

A possibly better fix is to always use `https` whenever `cdnSupoprtsSsl` is `true`, regardless of the given URLs.

sebastienros:
Where is the PR?

I assume anywhere we have `//` urls they should be converted explicitly to `https` ?

/cc @bleroy you confirm ?

thekaveman:
I wasn't sure if you wanted the quick fix, or the possibly better fix?

thekaveman:
Submitted a PR, my attempt at the possibly better fix. I targeted `dev` since the work by @anoordende was also merged into `dev` (but not yet `1.10.x`)

sebastienros:
I am not really a fan of trying to alter the url provided by devs, I'd rather educate in using `https` directly in the manifest instead of `//`

BenedekFarkas:
[...]

PR #6944, opened 2016-06-03
Ensure that CDN protocol relative URLs support ssl

thekaveman:
For #6941

### DIFF
--- a/src/Orchard.Tests/UI/Resources/ResourceManagerTests.cs
+++ b/src/Orchard.Tests/UI/Resources/ResourceManagerTests.cs
@@ -106,6 +106,26 @@ namespace Orchard.Tests.UI.Resources {
         }
 
         [Test]
+        public void CdnProtocolRelativePathSupportsSsl() {
+            _testManifest.DefineManifest = m => {
+                m.DefineResource("script", "Script1").SetUrl("script1.js").SetCdn("//cdn/script1.min.js");
+            };
+            _resourceManager.Require("script", "Script1");
+            VerifyPaths("script", new RequireSettings { CdnMode = true }, "https://cdn/script1.min.js", true);
+            VerifyPaths("script", new RequireSettings { CdnMode = true }, "https://cdn/script1.min.js", false);
+        }
+
+        [Test]
+        public void CdnProtocolRelativeDebugPathSupportsSsl() {
+            _testManifest.DefineManifest = m => {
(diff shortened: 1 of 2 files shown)


sebastienros:
Oh, I was actually thinking about changing the hard coded urls instead of doing it automatically ;)

sebastienros:
Not saying it's not better.

thekaveman:
That's what I meant by the "quicker option" :wink:

I'm also thinking of `Script.Include("//cdn.com/asset.js")` type situations; this fix would not cover, correct? Should we be concerned about that scenario?

sebastienros:
The more I think about it the more I think we should just update the urls in the resource files. Because maybe a user really need to use `//`. We should just change the urls we manage.

thekaveman:
Fair enough. I'll close this one and submit a new one with those changes.

While I'm at it though, there is one minor addition we should consider: if the user entered `//`, this should imply `CdnSupportsSsl`, correct? 

In other words, this code that looks for `https` at the beginning:

```csharp
else {
    CdnSupportsSsl = cdnUrl.StartsWith("https", StringComparison.OrdinalIgnoreCase);
}
```

should be expanded to look for `//` **or** `https` at the beginning, right?

Even though we may not output `https`, it is true that if the CDN supports protocol relative, it supports SSL.

anoordende:
Just as a bit of background, the only reason for the assumption `CdnSupportsSsl = cdnUrl.StartsWith("https", StringComparison.OrdinalIgnoreCase);` is backward compatibility, so that (hopefully) the majority of existing setups keep serving under https even with the CdnSupportsSsl flag omitted.

Just a radical thought, what we are heading towards is the convention that if the CDN URL starts with `https` (or `//` if  you like) it supports SSL, otherwise it does not. My gut reaction is that we can deprecate CdnSupportsSsl altogether.

And keep in mind that we are only indicating SSL support, we're not indicating if SSL should actually be used. Control can be given back by an Admin option along the lines of:

- Always use SSL for CDN resources [that support it] even when the site is insecure
- Only use SSL for CDN resources [that support it] when the site is secure
- Never use SSL for CDN resources (this would be a testing thing only I expect)

BenedekFarkas:
@thekaveman and @anoordende thanks for your input!
We've decided deprecate the CdnSupportsSsl parameter in 1.10.x (for 1.10.4) and remove it in dev (for 1.11).

PR #6945, opened 2016-06-03
Little Correction of TinyMce Module.txt

jtkech:
Just because `FeatureDependencies` is not an existing section name, it is never parsed in `Module.txt`.

### DIFF
--- a/src/Orchard.Web/Modules/TinyMce/Module.txt
+++ b/src/Orchard.Web/Modules/TinyMce/Module.txt
@@ -6,5 +6,5 @@ Version: 1.10.1
 OrchardVersion: 1.9
 Description: The TinyMCE module enables rich text contents to be created using a "What You See Is What You Get" user interface.
 FeatureDescription: TinyMCE HTML WYSIWYG editor.
-FeatureDependencies: Orchard.Resources
+Dependencies: Orchard.Resources
 Category: Input Editor


sebastienros:
I am worried there is another issue actually. `Description` and `Dependencies` are supposed to be per feature. We have `FeatureDescription` when we want to use the implicit feature of a module (a module has 1-* features) without having to create a section for the feature. So `FeatureDependencies` would totally make sense. Unless `Dependencies` works this way too already. But if it's not handled then there is no point in changing a wrong property to another wrong one, it's better to fix it or remove it completely.

jtkech:
You're right, i did it too fast. But at least this one is currently useless.
So, maybe the best right now is to remove it.
Let me know.

sebastienros:
Funny as I thought I was commenting on an issue for O2 ;)

jtkech:
Maybe because i've seen and corrected the same in my PR  on O2 :)

jtkech:
@sebastienros 

> We have FeatureDescription when we want to use the implicit feature of a module (a module has 1-* features) without having to create a section for the feature.

You're right, here we need 2 section names, we need the module description and we still need the description of its only one implicit feature.

But we need only one section name for dependencies which is `Dependencies` (maybe `FeatureDependencies` would have been better but it's not the case).

Under the debugger i can see all features and their dependencies. When i use `Dependencies` for `TinyMCE`, i can see the dependency of its only one implicit feature. When i use `FeatureDependencies` it doesn't work.

So, if we really want to make `TinyMce` dependent on `Orchard.Resources` we need to use `Dependencies`. If we want to keep the current behavior we can delete this line that is useless.

Best.

sebastienros:
What should we do with this PR ?

jtkech:
Because TinyMce seems to be dependent on Orchard.Resources (e.g needs jQuery), i think we can merge this PR as it is.

Best

rtpHarry:
Nudge for triage next week. I think this can be merged.

Issue #6937, opened 2016-05-31
Add PasswordChanging event to IUserEventHandler or change PasswordChanged signature


sfmskywalker:
Background for this issue can be found in the discussion here: #6930

In order to be able to implement custom rules such as preventing users from using passwords used in the past, I need to be able to intercept when a password is set for a user using an event handler.

When a password is changed from the `AccountController`in the Users module, the following event is fired when changing a forgotten password:

``` C#
IUserEventHandler.ChangedPassword (IUser user);
```

Ideally, we change this method by replacing the `IUser` argument with a context class containing more details about the event, such as the clear text password being received.

The downside is that we break other modules that implement this interface, but maybe that's better than introducing a new interface.

I propose the following change:

``` C#
IUserEventHandler.ChangedPassword (PasswordChangedContext context);

class PasswordChangedContext {
   IUser AffectedUser;
   string NewPassword;
}
```

Using a complex type argument as the context makes the interface more resilient to future changes.

PR #6955, opened 2016-06-06
Added password to IUserEventHandler.PasswordChanged event

sfmskywalker:
See #6937


DanielStolt:
I see you didn't go with the new complex type approach - how come?

In any case, I thought the name suffix "Context" was inappropriate for the suggested type, because in my mind, that suffix suggests that handlers can interact with the context object to somehow alter the outcome of the event, because that seems to be the convention in Orchard for context objects. "Args" or "Info" would have been better name suffices for this reason IMO.

Anyway, you didn't end up using the object, so... don't know why I'm mentioning it. :)

sfmskywalker:
I am pleased you asked!

In hindsight, other methods on this interface also accept an extra argument called `password`. Introducing a complex type argument for just the one method would then seem arbitrary. So I think it would be better to do it either all the way, or don't do it (and risk needing to break the interface yet again in the future, however unlikely in my mind).

Regarding the suffix "Context" vs "Args" or "Info": I considered the suffix 'Args", but I couldn't find any previous usage, plus it seems to be a convention used with actual .NET events. Then again, they are using "EventArgs" as a suffix. "Info" I hadn't thought of.

But the main question is: should we completely overhaul the interface with complex type arguments, or just a select few members (yet to be determined), or keep my PR as-is?

DanielStolt:
I see - agree with your assessment.

Regarding the main question, I'm not aware of the additional needs beyond this PR, but if your PR as-is solves the issue, then personally I would vote for keeping it.

sebastienros:
Broken PR, I know who did this to you but won't blame them.

Issue #6966, opened 2016-06-09
DateTimeField, proper support for "not required"


ViRuSTriNiTy:
Currently a `DateTimeField` with setting `not required` is saved with value `System.DateTime.Min` which leads to a field that displays the minimum value on the next page refresh.

In case of my setup (german localization) it displays the date as `1.1.0` and time as `0:0`.

I would like to have empty fields when nothing has been selected / entered, so i changed some parts of the Orchard modules.

Can this change possibly be included in the next orchard release or do i have to live with a adjusted version?

ViRuSTriNiTy:
Repro:
1) add a `Date Time Field`  to content type 'A' and make ensure field is not required hence setting `Required` should not be checked
2) create a content from content type 'A', leave the date time field blank and save
-> field will now show date `1.1.0` and time `0:0`

This is caused by line [157](https://github.com/OrchardCMS/Orchard/pull/7248/files?diff=unified#diff-61a3b0f22ff429a68be31f8b51f72cc0L157) in `DateTimeFieldDriver.cs´ where `DateTime.MinValue` is assigned to `field.DateTime` ...

`field.DateTime = value.HasValue ? value.Value : DateTime.MinValue;`

.. which means that the field will hold the min value instead of an empty value.

Therefore, to support an empty value the first change is to change `DateTime` to [DateTime?](https://github.com/OrchardCMS/Orchard/pull/7248/files#diff-8bebe290ece2f50946d98527e4d590c4R9) in `DateTimeField.cs` and then basically change all places where the compiler complains about incompatible types.

The changes are applied in a way that existing code remains untouched and a new code path for "value is null" is inserted via `if` `then` `else` pattern.

mwentz:
@sebastienros can we get the bug fix for this into the next release?

sebastienros:
The change looks fine, but we'd need to target the dev branch as this is a breaking change.

PR #6968, opened 2016-06-09
Orchard.Fields, added proper support for "not required" setting of DateTimeField

ViRuSTriNiTy:
see issue https://github.com/OrchardCMS/Orchard/issues/6966


infofromca:
@ViRuSTriNiTy this fix should go to 1.10.x pls

ViRuSTriNiTy:
@infofromca Is it my task to change to 1.10.x? When this is the case how do i do this?

ViRuSTriNiTy:
Created a new pull request, see #7248

rtpHarry:
For future reference, sometimes you can do this via the GitHub UI now.

Click Edit and the base is selectable:

image

Sometimes it makes a mess though as too much is different between the branches.

The feature was only added in last few months so it probably wasn't available when you asked in June.

PR #6980, opened 2016-06-16
[Fixes #6979] Orchard.Projections - Added operator IsNull and IsNotNull to date time filter handling to be able to filter NULL values of DateTime? columns correctly

### LINKED ISSUE #6979: Projection should support NULL and IS NOT NULL filters for DateTime? columns
Currently there is no way to generate a `MyField IS NULL` or `MyField IS NOT NULL` query condition in Orchard.Projections.

Filter editor should be extended to allow a selection of `not null` and `is not null`.

ViRuSTriNiTy:
see issue #6979

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/DateTimeFilterForm.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/DateTimeFilterForm.cs
@@ -127,7 +127,7 @@ namespace Orchard.Projections.FilterEditors.Forms {
 
             string type = Convert.ToString(formState.ValueType);
 
-            DateTime min, max;
+            DateTime? min, max;
 
             // Are those dates or time spans
             if (type == "0") {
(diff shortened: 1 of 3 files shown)


sfmskywalker:
Instead of adding the `IsNull` and `IsNotNull` operators, would it perhaps be better to simply use the `IsEqual` and `IsNotEqual` operators, and add the ability to specify a NULL value instead?

ViRuSTriNiTy:
@sfmskywalker Yeah that would be another solution, i thought about that too initially, but the implementation would be a bit more complicated as

1) the equal operator implementation uses calculated min / max variables, this is not compatible with NULL values, so there needs to be a new path for the NULL value handling

2) the view to configure the condition needs to be extended with a new input that accepts NULL values instead of the current solution (that is simply hiding the value input controls)

I don't see a benefit of adding additonal complexity by enabling the ability to specify a NULL value here.

Also, imho `IsNull` and `IsNotNull` does communicate clearly what is going on under the hood.

sebastienros:
Can someone try the changes and ensures it doesn't break the current scenarios?

Xceno:
I've tried the changes and it works for existing projections (several years old) and new projections which use the new Null and NotNull filters. The additional null checks also seem to fix some issues (crashing queries that is), I had before when creating projections on DateTime? fields.

ViRuSTriNiTy:
@Xceno Thx, awesome.

ViRuSTriNiTy:
@Xceno I changed the base branch to 1.10.x as requested in https://github.com/OrchardCMS/Orchard/issues/6979#issuecomment-258402945

Xceno:
Does this need a review again? We would still like to have this PR in 1.10.x

Xceno:
Thank you sir! =)

sebastienros:
Bitte schön

PR #6982, opened 2016-06-17
Request/patching framework extensions

arijoon:
Orchard framework extension methods had a couple of issues. I tried to fix them in a modular way but unfortunately there was no way around it aside from patching the actual methods themselves. 

I encountered these issues when I was setting up the Gallery server. The TitleForPage issue arises in the gallery theme. I think a better solution is an implicit conversion from `MvcHtmlString[]` to `String[]`

The second issue in UrlHelpers arises when I was exposing the feed. The problem is if `workContextAccessor` is null for any reason it'll throw up, without even logging the error. 

I must say you guys have a done a fantastic job with Orchard. The modularity is top notch, and the code itself is self documenting. Keep up the good work

### DIFF
--- a/src/Orchard/Mvc/Html/LayoutExtensions.cs
+++ b/src/Orchard/Mvc/Html/LayoutExtensions.cs
@@ -48,6 +48,16 @@ namespace Orchard.Mvc.Html {
             return MvcHtmlString.Create(html.Encode(titleParts[0]));
         }
 
+        public static MvcHtmlString TitleForPage(this HtmlHelper html, params MvcHtmlString[] titleParts) {
+            if (titleParts == null || titleParts.Length < 1)
+                return null;
+
+            string[] parts = titleParts.Select(t => t.ToString()).ToArray();
+            html.AppendTitleParts(parts);
+
+            return MvcHtmlString.Create(html.Encode(titleParts[0]));
+        }
+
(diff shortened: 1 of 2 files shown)


src/Orchard/Mvc/Html/LayoutExtensions.cs
```diff
@@ -48,6 +48,16 @@ public static class LayoutExtensions {
...
         }
 
+        public static MvcHtmlString TitleForPage(this HtmlHelper html, params MvcHtmlString[] titleParts) {
+            if (titleParts == null || titleParts.Length < 1)
+                return null;
+
+            string[] parts = titleParts.Select(t => t.ToString()).ToArray();
+            html.AppendTitleParts(parts);
+
+            return MvcHtmlString.Create(html.Encode(titleParts[0]));
```
sebastienros: An HtmlString is supposed to be encoded already, it should not be encoded again.

sebastienros:
Can you file the related issues so we can understand them and comment?

arijoon:
Two issues here:
 - if `workContextAccessor` is null it'll crash (which as you commented should not ever happen)
 - Somewhere `TitleForPage` extension method is being called with an array of `MvcHtmlString`, no extension method with correct signature or implicit conversion exists to handle it

Issue #6981, opened 2016-06-16
ImageProfile paths need to exclude host to allow 301 redirects on domain moves


anoordende:
SEO and user-friendlyness is probably the main driver for resolving this issue. Currently the function `ImageProfileManager.FormatProfilePath` includes the full URI to calculate the hashes for the processed images, e.g when using Azure storage with a public host: `https://mystore.blob.w.c.net/media/Foo/Bar/image.png`

This means that, when moving domains or switching to e.g. a CDN, it is virtually impossible to create clean 301 redirects because you would have to calculate and redirect each individual hash. Whereas, if the hashes were solely generated on the directory part of the path (ie. discarding protocol, host, app relatives, etc.) it would be as easy as redirecting everything under _Profiles/\* as the hashes will be the same.

Long story short, I propose to _normalize the hashed profile paths just on the path portion so that they become agnostic_.

As per usual, I will create a pull request for this momentarily, probably targeting dev. I think an option to switch this off for existing sites is also in order, to prevent a global recalculation of hashes...

anoordende:
Pull request created: #6984

PR #6984, opened 2016-06-17
#6981 Normalized image profile path hash and added profile purging

anoordende:
This resolves both issue #6981 by normalizing the profile path hash, as well as re-introduces an old CodePlex PR that missed the GitHub boat, namely the ability to purge profiles through the Admin UI.

**Normalizing the image profile path**
Normalizing the image profile path hash "should" be the default for all new sites, as it allows for changing the domain from which images are served, whilst maintaining the already generated profiles. This enables, for example, 301 SEO-friendly redirects when moving from local storage to azure storage or moving to a CDN by simply redirecting one-to-one at the `^_Profiles/.*` level.

**Important for Release Notes**
_Existing sites may opt-out from Normalizing the profile paths, especially those which are not serving from the local Media folder, those which use a container other than "media" in Azure Blob Storage and those sites which use a custom storage provider. This can be achieved by setting the following AppSetting to `false`, which sets the behavior exactly as before:_

`<add key="Orchard.MediaProcessing.NormalizePath" value="false" />`

**Purging Profiles**
As mentioned, this PR also enables 2 purge options form the Admin > Media > Profiles view, forcing profiles to be regenerated and old "orphaned" images will be cleared:
1. Purge all images under an individual profile
2. Purge all obsolete images, including all dynamic profiles (such as the Admin UI thumbs)

This functionality sits quite naturally with the normalization of the hashes, as those who wish to implement the normalization on existing sites immediately have a function to clear the old profiles (using the Purge Obsolete option). Though for SEO purposes you probably want to wait a few months before doing so.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
@@ -15,6 +15,8 @@ using Orchard.MediaProcessing.Media;
 using Orchard.MediaProcessing.Models;
 using Orchard.Tokens;
 using Orchard.Utility.Extensions;
+using Orchard.Environment.Configuration;
+using Orchard.Caching;
 
 namespace Orchard.MediaProcessing.Services {
     public class ImageProfileManager : IImageProfileManager {
@@ -24,6 +26,8 @@ namespace Orchard.MediaProcessing.Services {
         private readonly IImageProcessingManager _processingManager;
         private readonly IOrchardServices _services;
         private readonly ITokenizer _tokenizer;
+        private readonly ISignals _signals;
+        private readonly IAppConfigurationAccessor _appConfigurationAccessor;
 
         public ImageProfileManager(
             IStorageProvider storageProvider,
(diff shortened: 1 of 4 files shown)


src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
```diff
@@ -235,14 +243,66 @@ public class ImageProfileManager : IImageProfileManager {
...
+            //Slice at the protocol, if any. E.g. http:// or https://.
+            var index = path.IndexOf("//", StringComparison.OrdinalIgnoreCase);
+            //Slice at the first directory after the protocol or at 0 if no protocol specified.
+            index = path.IndexOf("/", index < 0 ? 0 : index + 2, StringComparison.OrdinalIgnoreCase);
+            //Return path from the first directory, replacing lcase 'media' (Azure container) with ucase 'Media' (filestorage).
+            return path.Substring(index < 1 ? 0 : index).Replace("media", "Media");
+        }
+
+        private bool ShouldNormalizePath() {
+            var normalizePath = _appConfigurationAccessor.GetConfiguration("Orchard.MediaProcessing.NormalizePath");
```
sebastienros: How often is it called? Shouldn't it be cached?

BenedekFarkas:
@anoordende please enable edits by maintainers.

BenedekFarkas:
Superseded by #8788 due to inactivity.

PR #6990, opened 2016-06-20
#6783: Workaround for PostgreSQL without breaking SQL server.

### LINKED ISSUE #6783: Can't access DistributedLockRecord table on setup + possible fix
In `AutomaticDataMigrations.cs`, before, `_transactionManager.RequireNew()` was called after creating the schema of the `DistributedLockRecord` table. Apparently with `PostgreSQL` we need to do it before, see #6729. But, since this update, i get nhibernate exceptions on a new setup (sqlce).

This because we still need `_transactionManager.RequireNew()` after creating the schema, otherwise, when 1st trying to acquire a lock on setup, we can't access the `DistributedLockRecord` table.

Indeed, when doing this it works

            if (!distributedLockSchemaBuilder.SchemaExists()) {
                // Workaround to avoid some Transaction issue for PostgreSQL.
                _transactionManager.RequireNew(); // line added for PostgreSQL
                distributedLockSchemaBuilder.CreateSchema();
                _transactionManager.RequireNew(); // new line for my tests
            }

But here there are 2 `RequireNew()`. Let me know what we have to do.

Best.

jtkech:
Fixes #6783 

This PR #6729 solved an issue with `PostgreSQL` but introduced a little drawback e.g with `SqlCe`, see in #6783. As said by @sebastienros we would need to revert #6729 ...

Here, i propose the same workaround for `PostgreSQL` without breaking other providers, e.g `SqlServer`.

Best.

### DIFF
--- a/src/Orchard/Data/Migration/AutomaticDataMigrations.cs
+++ b/src/Orchard/Data/Migration/AutomaticDataMigrations.cs
@@ -85,9 +85,16 @@ namespace Orchard.Data.Migration {
             var schemaBuilder = new SchemaBuilder(_dataMigrationInterpreter);
             var distributedLockSchemaBuilder = new DistributedLockSchemaBuilder(_shellSettings, schemaBuilder);
             if (!distributedLockSchemaBuilder.SchemaExists()) {
+
                 // Workaround to avoid some Transaction issue for PostgreSQL.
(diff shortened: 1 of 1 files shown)


sebastienros:
Seems to be the only way to deal with it.

PR #6992, opened 2016-06-22
#6986 Created feature Media Processing Html Filter

### LINKED ISSUE #6986: Feature: Dynamically Media-process images with fixed size in HTML
Images inserted in HTML through, e.g. any content with a Body can be resized through the editor. However, the full size image will still be served whereas it would be more efficient, especially on mobile clients, to serve a pre-sized image.

It is quite straightforward to dynamically cut these images to size using the MediaProcessing module in combination with IHtmlFilter. Of course, it needs to be a feature that can be turned on or off as desired.

The one requirement is that the height/width attributes are provided, which will be when resizing through the TinyMce editor. This also provides an excellent trick when you particularly don't want an image to be resized: omit the height or width.

As I've had this code live for a few years already, the PR is to follow shortly.

anoordende:
Fixes #6986

Enabling this feature registers an implementation of `IHtmlFilter` that works on any part with a flavor of "HTML" that injects and applies the registered IHtmlFilters.

Additional settings appear under in the Admin UI under `Settings > Media` to control to behavior of this feature. It is quite straightforward:

**Density Threshold**
With a threshold of, for example, `@2x`, an image will only be reduced if the natural size of the image is more than 2x the size specified in the attributes.

**Quality**
Only applies to jpeg's, the quality to compress the image at.

**Populate Empty Alt Tags**
If the img has an empty Alt tag, one will be composed based on the image file name

**noresize attribute**
Adding a noresize attribute to any img element will cause the element not to be processed:

`this image will not be processed`

_Note that images already the result of a profile (ie already under /_Profiles/*) will not be processed._

To try out this feature, simply add a suitable image from your media library to e.g. a body part and use the handles in the editor the resize the image. When rendered, you will see that the image comes from a profile and not from the original source. Change the quality or density in the settings and you will see new profiles being generated.

Clean up the dynamic test profiles or force regeneration with the "Purge Obsolete" feature from PR #6984.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaProcessing/Constants.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaProcessing/Constants.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web;
+
+namespace Orchard.MediaProcessing {
+    public static class Features {
+
+        public const string OrchardMediaProcessingHtmlFilter = "Orchard.MediaProcessingHtmlFilter";
+
+    }
+}
--- a/src/Orchard.Web/Modules/Orchard.MediaProcessing/Filters/MediaProcessingHtmlFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaProcessing/Filters/MediaProcessingHtmlFilter.cs
@@ -1,0 +1,166 @@
+﻿using HtmlAgilityPack;
+using Orchard.ContentManagement;
+using Orchard.Environment.Extensions;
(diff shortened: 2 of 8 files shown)


sebastienros:
I would not use HtmlAgilityPack. It seems too heavy for the goal here, probably a compiled Regex or a custom parser would be able to fetch all images. Just thinking about performance here.

sebastienros:
Unless someone proves a regex wouldn't be optimal for this case.

BenedekFarkas:
Note: #6938 has breaking changes.

BenedekFarkas:
Closing due to inactivity, superseded by #8806.

Issue #6993, opened 2016-06-22
Orchard.Blogs permission inconsistency in AdminMenu


thekaveman:
This is on the `1.10.x` branch.

The `Orchard.Blogs` admin menu code applies permission checks inconsistently. I'm not sure if this is a bug or by design, but it seems like a bug (and has a very simple fix).

The issue is in [`BuildMenu`](https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.Blogs/AdminMenu.cs#L28). Obtaining the initial set of blogs to build the menu checks the [`MetaListBlogs`](https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.Blogs/Permissions.cs#L17) permission. Subsequent `menu.Add(...).Permission(...)` check the [`MetaListOwnBlogs`](https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.Blogs/Permissions.cs#L18) permission. 

This means that a user with sufficient permissions to publish, edit, and delete their own posts, but not perform actions against posts of others, will not see a Blog menu. 

The simple fix is to use [`MetaListOwnBlogs`](https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.Blogs/Permissions.cs#L18) in all cases.

thekaveman:
Sorry, pulled the trigger too fast. Looks like this issue is a dupe of #4767. I will close this one.

Issue #4767, opened 2014-09-11
Blogs menu not shown if user doesn't have MetaListBlogs


orchardbot:
@Piedone created:
https://orchard.codeplex.com/workitem/20938

Expected: when users have any permissions to manage at least their own blog posts in any way (edit/delete, publish) they should see the Blogs menu item in the admin.

Actual result: the menu item is only seen if they have a permission that implies MetaListBlogs, e.g. "Edit blog posts for others". This is because Orchard.Blogs/AdminMenu ln 29 checks for this when listing blogs.

Shouldn't we use MetaListOwnBlogs instead?

Issue #3933, opened 2013-09-13
Publishing blog posts requires "Publish or unpublish content" permission, no matter how "Publish or unpublish blog post" permission is set


orchardbot:
__Dunross__ created:
https://orchard.codeplex.com/workitem/20105

I wanted to allow some roles just to edit some specific content types. However, publishing blog posts requires "Publish or unpublish content" permission and only then the "Publish or unpublish blog post" permissions are evaluated.

It is a bug and quite a serious one, because if you want to make it work for users (you need to), you have to give up whole authorization on content-type level.

Found in Orchard version: 1.7

Similar issues:

Media metadata editation requires "Edit content" permission, no matter how "Manage Media" permission is set.
Menu Item editation requires "Edit content" permission, no matter how "Manage main menu" permission is set.

thekaveman:
I believe this is still an issue, and is related to my PR #6994. With that change, given the user has *Publish Own Blog Post* permission, the menu item for **New Post** is available. Additionally, in the Summary Admin screen for blog posts, a **Publish** link becomes available for each owned blog post of the user.

There is an inconsistency though; from the Blog Post edit screen, the user can only **Save**, and not **Publish**.

thekaveman:
An additional twist is that, the `PublishLaterPart` also requires the *Publish Content* permission. I'm having a difficult time with the correct way to override this for Blog Posts (thinking about an `IShapeTableProvider` for the *EditorTemplate* but that seems ugly).

Issue #2575, opened 2012-06-13
Permissions on specific content type


orchardbot:
__gkennedy__ created:
https://orchard.codeplex.com/workitem/18748

When allowing permission to a content type such as a blog post, the relevant links don't show until the permission is granted to the Content feature aswell.

Repro
1) Create new role
2) Allow publish & unpublish own blog post permission on Orchard.Blogs feature
3) Ensure publish & unpublish own blog post permission is not allowed on Content feature
4) Create new user and assign to role
5) Log in as new user and attempt to publish or unpublish a blog post. The view link exists but no publish/unpublish

This also occurs for edit/delete etc.

If the role is granted the equivalent Content feature permission then the user is able to access the links.

PR #6994, opened 2016-06-22
Blogs permission check consistency

thekaveman:
See #3933, #4767, and #6993 for background.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Blogs/AdminMenu.cs
+++ b/src/Orchard.Web/Modules/Orchard.Blogs/AdminMenu.cs
@@ -26,7 +26,7 @@ namespace Orchard.Blogs {
         }
 
         private void BuildMenu(NavigationItemBuilder menu) {
-            var blogs = _blogService.Get().Where(x => _authorizationService.TryCheckAccess(Permissions.MetaListBlogs, _workContextAccessor.GetContext().CurrentUser, x)).ToArray();
+            var blogs = _blogService.Get().Where(x => _authorizationService.TryCheckAccess(Permissions.MetaListOwnBlogs, _workContextAccessor.GetContext().CurrentUser, x)).ToArray();
             var blogCount = blogs.Count();
             var singleBlog = blogCount == 1 ? blogs.ElementAt(0) : null;
 
--- a/src/Orchard.Web/Modules/Orchard.Blogs/Views/Content-BlogPost.SummaryAdmin.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Blogs/Views/Content-BlogPost.SummaryAdmin.cshtml
@@ -1,9 +1,12 @@
-﻿@using Orchard.ContentManagement;
-@using Orchard.Core.Contents
+﻿@using Orchard.Blogs;
+@using Orchard.Blogs.Extensions;
+@using Orchard.Blogs.Models;
+@using Orchard.ContentManagement;
 @using Orchard.Utility.Extensions;
 @{
     Script.Require("ShapesBase");
     ContentItem contentItem = Model.ContentItem;
+    BlogPostPart post = contentItem.As<BlogPostPart>();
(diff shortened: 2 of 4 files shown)


thekaveman:
The most recent commit modifies the Admin so that the **Publish** button and various action links for Blog Posts check the Blog Post permissions and use Blog Post specific action URLs (as opposed to using the standard Contents permissions and actions).

src/Orchard.Web/Modules/Orchard.Blogs/Views/Content-BlogPost.SummaryAdmin.cshtml
```diff
@@ -25,17 +28,17 @@
             @T(" | ")
 
             if (contentItem.HasDraft()) {
-                @Html.Link(T("Publish Draft").Text, Url.Action("Publish", "Admin", new { area = "Contents", id = contentItem.Id, returnUrl = Request.ToUrlString() }), new { itemprop = "UnsafeUrl" })
+                @Html.Link(T("Publish Draft").Text, Url.BlogPostPublish(post), new { itemprop = "UnsafeUrl" })
```
sebastienros: Why this?

thekaveman: @sebastienros for one, the helper already exists in `Orchard.Blogs.Extensions.UrlHelperExtensions`, and I couldn't see it being used anywhere else, so... why not? :smile:  

More importantly though, Publish in BlogPostAdmin controller checks the BlogPostPublish permission, while Publish in the standard Contents AdminController checks the regular PublishContent permission. So for normalizing on the Blog specific permissions, it's important.

sebastienros: I would imagine that when I worked on the permissions for BlogPost I changed it to use the default content controller as much as possible. Did you look at the history of the file? I probably haven't removed the extension methods as it could have broken other modules.

sebastienros: Or is it also part of the fix, and using the Content controller creates the actual issue?
Can I ask you to provide the steps to expose the issues you are fixing?

thekaveman: Correct: using the Contents Admin controller creates the issue, using the BlogPostAdmin controller fixes the issue. This file was added in d31d684 as part of fixing #2575. There have been no other meaningful commits since (that I can tell). 

The permissions checks done in `Content-BlogPost.SummaryAdmin.cshtml` are correctly looking at the Blog Post specific permissions. It is the action links for Publish/Unpublish/Delete that are incorrect, because they forward to Contents Admin Controller actions, and those actions don't check Blog Post specific permissions (but rather, the standard Content permissions).

src/Orchard.Web/Modules/Orchard.Blogs/AdminMenu.cs
```diff
@@ -26,7 +26,7 @@ public class AdminMenu : INavigationProvider {
         }
 
         private void BuildMenu(NavigationItemBuilder menu) {
-            var blogs = _blogService.Get().Where(x => _authorizationService.TryCheckAccess(Permissions.MetaListBlogs, _workContextAccessor.GetContext().CurrentUser, x)).ToArray();
+            var blogs = _blogService.Get().Where(x => _authorizationService.TryCheckAccess(Permissions.MetaListOwnBlogs, _workContextAccessor.GetContext().CurrentUser, x)).ToArray();
```
sebastienros: This change is not necessary, there is an event handler that will do the conversion automatically: `BlogAuthorizationEventHandler`

sebastienros:
The new comment I made can now be worked on by targeting the `blogspermission` branch.

PR #7002, opened 2016-06-28
Update Content-Term.Edit.cshtml

hbulzy:
this file in Module “Taxonomies” 
when Taxonomy tile is "我是中文" the ToSafeName method will remove all chars 
alternates will match this file then bug happens

or just delete this file


sebastienros:
I don't understand how your change is fixing anything.

hbulzy:
@sebastienros  use other language like Chinese create a taxonomy 
when we Add a term we'll get blank page 
like this
image

PR #7035, opened 2016-07-13
Fixed the output caching of non 200  - fixes #7034

### LINKED ISSUE #7034: If an exception is thrown in a view or htmlhelper, the response is still output cached.
It seem's that at the moment if an exception is thrown in a view or html helper, the response is still captured by output caching and stored along with it's status code, meaning subsequent visits to a url throw the same error even if the issue is rectified.

This was spotted using Redis as the cache provider; but i imagine it's the issue whatever the provider.

It seems that the ResponseIsCachable tries to determine the status and not cache if its not 200 or a transformed redirect, BUT at this point the StatusCode is still 200.

The status could be rechecked after the output is captured before storing it, this seems to fix the issue - happy to provide a PR with this fix.

carlwoodhouse:
Fixed the output caching of non 200 responses when exceptions are thrown by views, htmlhelpers, etc.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -227,8 +227,10 @@ namespace Orchard.OutputCache.Filters {
                         // should be resolved from it.
 
                         // Recheck the response status code incase it was modified before the callback.
-                        if (response.StatusCode != 200)
+                        if (response.StatusCode != 200) {
+                            Logger.Debug("Response for item '{0}' will not be cached because status code was set to {1} during rendering.", _cacheKey, response.StatusCode);
                             return;
+                        }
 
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -226,6 +226,10 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
                         // To prevent access to the original lifetime scope a new work context scope should be created here and dependencies
                         // should be resolved from it.
 
+                        // Recheck the response status code incase it was modified before the callback.
+                        if (response.StatusCode != 200)
+                            return;
```
DanielStolt: Let's also add a debug log message here I think. Something like this:
"Response for item '{_cacheKey}' will not be cached because status code was set to {response.StatusCode} during rendering."

carlwoodhouse:
Updated to add the debug message

sebastienros:
GTG

PR #7038, opened 2016-07-19
Changed CultureSelectorResult.Priority type to double

ihrebin:
Orchard culture selectors' priorities are from -5 to -1. This makes impossible to create a culture selector, that returns result with priority between default ones. For example if we want to have default culture per user stored in database, but want to override it with cookie, we'll need to have a selector with priority smaller than `CookieCultureSelector `but greater than all others, for example -2.5.

### DIFF
--- a/src/Orchard/Localization/Services/ICultureSelector.cs
+++ b/src/Orchard/Localization/Services/ICultureSelector.cs
@@ -2,7 +2,7 @@
 
 namespace Orchard.Localization.Services {
     public class CultureSelectorResult {
-        public int Priority { get; set; }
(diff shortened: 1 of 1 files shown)


sebastienros:
The correct solution is to change the boundaries from -5;-1 to something like -100, 100. We use this everywhere and it would be inconsistent to start changing the type.

sebastienros:
Let's use ranges of 20 for the current ones, like -100 for SiteCulture and -80 for BrowserCulture, and so on.

sebastienros:
We agree this is also portentially a breaking change that we'll need to document in the release notes.

PR #7042, opened 2016-07-19
Identity support to menu create command switch

kumards:
Control over menu Identity will help in setting menu identifier in menu items from the recipe data.

### DIFF
--- a/src/Orchard.Web/Core/Navigation/Commands/MenuCommands.cs
+++ b/src/Orchard.Web/Core/Navigation/Commands/MenuCommands.cs
@@ -1,7 +1,8 @@
-﻿using System;
+using System;
 using Orchard.Commands;
 using Orchard.ContentManagement;
 using Orchard.ContentManagement.Aspects;
+using Orchard.Core.Common.Models;
 using Orchard.Core.Navigation.Models;
 using Orchard.Core.Navigation.Services;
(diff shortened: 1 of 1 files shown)


sebastienros:
The code looks good, but I'd like to understand what the scenario is. Could you provide an example?

kumards:
It helps to create menu and menu items interchangeably by using using command element of a recipe like below and after that menu items can be imported through data element of the same recipe. Otherwise both the menu and menu items should be created through command or both the menu and the menu items should be created from the recipe data.

**Note:** 
Current implementation of the menu command uses **menu name** to locate a menu.
Current implementation of menu item import(MenuPartDriver.cs) uses **identifier** to locate a menu.
We do support setting the Identity in the widget command.
```
layer create Default /LayerRule:"true" /Description:"The widgets in this layer are displayed on all pages"

layer create Authenticated /LayerRule:"authenticated" /Description:"The widgets in this layer are displayed when the user is authenticated"

menu create /MenuName:"Main Menu" /Identity:"MainMenu1"

widget create MenuWidget /Title:"Main Menu" /RenderTitle:false /Zone:"Navigation" /Position:"1" /Layer:"Default" /Identity:"MenuWidget1" /MenuName:"Main Menu"

```

PR #7061, opened 2016-07-28
Updating with a precise user message

kumards:
The email is sent only when the user is found in the system, not otherwise. 
When a non existing user tries to recover the password, the current message does not say that the user wouldn't receive the email.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Controllers/AccountController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Controllers/AccountController.cs
@@ -204,7 +204,7 @@ namespace Orchard.Users.Controllers {
 
             _userService.SendLostPasswordEmail(username, nonce => Url.MakeAbsolute(Url.Action("LostPassword", "Account", new { Area = "Orchard.Users", nonce = nonce }), siteUrl));
 
-            _orchardServices.Notifier.Information(T("If you are an existing user, we will immediately send you an email with a link to reset your password."));
+            _orchardServices.Notifier.Information(T("If your username or email is correct, we will immediately send you an email with a link to reset your password."));
             
             return RedirectToAction("LogOn");
         }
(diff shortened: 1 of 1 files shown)


sebastienros:
Can you file an issue so we can talk about it. I am not sure the text adds any value.

kumards:
Adding issue reference #7070 .

src/Orchard.Web/Modules/Orchard.Users/Controllers/AccountController.cs
```diff
@@ -204,7 +204,7 @@ public class AccountController : Controller {
 
             _userService.SendLostPasswordEmail(username, nonce => Url.MakeAbsolute(Url.Action("LostPassword", "Account", new { Area = "Orchard.Users", nonce = nonce }), siteUrl));
 
-            _orchardServices.Notifier.Information(T("Check your e-mail for the confirmation link."));
+            _orchardServices.Notifier.Information(T("If you are an existing user, we will immediately send you an email with a link to reset your password."));
```
adrianbooth: this doesn't really add any value as the email the user submitted isn't provided back to them, so they won't know whether they have made a typo. Something better would be:
```
T("Check your e-mail for the confirmation link. We will be sending the message to: {0}", email)

kumards: I agree with you about adding the given email address in the user message, however it still would not tell the user that recieving email is not guranteed unless his email id matches in the database.
Think of an user who was not sure if he had used his gmail id or microsoft id or yahoo id.

Otherway to work around this issue is, always send out an email to the user. When the id found, send out the recovery steps, else send out that user id not found in the database.

sebastienros: We can't expose the emails, as we could know the username and then get their email address.

sebastienros:
I think what is wrong here is this part: "If you are an existing user". Something more explicit about what we are trying to fix here is : "If your username or email is correct" ...

kumards:
That works too! I just needed some hint to the user that s/he has should know its not always the system will come up with some email.

sebastienros:
Please fix the merge conflict.

BenedekFarkas:
"Merge manually" to 1.10.x (78420a02acf4bfafd35e86258a83fccf8ed312a4).

Issue #7068, opened 2016-08-04
Invalid Last Logged-In Value in Users List


jchenga:
After creating a user account, the Last Logged In field for that account showed "2016 years ago".

sebastienros:
Duplicate of #6440. Already fixed in 1.10.

PR #7069, opened 2016-08-04
#7068 Users List showing invalid Creation and Login Time

jchenga:
Return an empty MvcHtmlString if DateTimeUTC is MinValue. Shows empty text if Creation or Login datetime is null.


sebastienros:
Closing for same reason as issue: #7068

PR #7072, opened 2016-08-07
Fix #6812: Add shape attributes to "CreateElementTagBuilder"

### LINKED ISSUE #6812: Add shape.Attributes to the result from TagBuilderExtensions.GetCommonElementAttributes
I tried to extend the default text field element, to add placeholder attribute to the output html tag, but I'm ended up with some workarounds to add it, but I found that if I set this attribute to shape.Attributes list directly, it will not rendered because this method "TagBuilderExtensions.GetCommonElementAttributes" doesn't check the shape attributes to be included in the output tag, and I add this bug because this is an important feature to be able to extend any element without need to override the element view.
please see the attached file to understand my solution for this issue.
[TagBuilderExtensions.cs.zip](https://github.com/OrchardCMS/Orchard/files/230931/TagBuilderExtensions.cs.zip)

mdameer:
Add shape attributes to "TagBuilderExtensions.CreateElementTagBuilder()", to give developers the ability to add there new attribute from custom layout elements drivers.


sebastienros:
@sfmskywalker ?

sebastienros:
Can you also file an issue for this one?

mdameer:
@sebastienros, this fix for issue #6812.

PR #7074, opened 2016-08-10
Pass the right display type to the widget element.

jtkech:
**In this PR**

- While helping someone which uses a projection element with a custom display type to render each content, we needed to pass the right diplay type to widget elements.

**Another suggestion**

- When in `Design` mode, because the actual `Projection.Design.cshtml` only uses the `ContentItems` collection, to prevent many handler / driver from being executed, we could use this in `ProjectionElementDriver.OnDisplaying()`.

    ```
// after this line
    context.ElementShape.ContentItems = contentItems;

    // Nothing more to do if we are in "Design" mode.
    if (context.DisplayType == "Design") {
        return;
    }
```
But i didn't include this in this PR because maybe we still want to be able to override `Projection.Design.cshtml` if someone want to render more things in design mode ... So let me know.

Best


sebastienros:
What the link between Projections, Widgets, Display Type and "Design" ?

jtkech:
When you edit a layout part (displayType = Design) that has a projection element which queries on contents which have widget elements ...

But regarding the Design mode it's only a suggestion, right now this PR only pass the right displayType to widget elements when rendering in the front end.

PR #7081, opened 2016-08-12
Allow adding HTML classes through ShapeTableBuilder.Describe

judgestone:
Small change to Content.Summary.cshtml to allow module developers to add classes to Content.Summary.cshtml via ShapeTableBuilder.Describe.


sebastienros:
Can you file an issue linking to this PR so we can track it in the release notes?

PR #7100, opened 2016-08-27
Added DesignWidth and DesignHeight to VectorImage element

Yaevh:
In the Layout Designer view, VectorImages are always scaled up to fill available width, which is often undesirable as the image will eventually be scaled down on the front-end by a CSS rule. This causes the image size (and thus layout) to be different between Layout Designer and front-end, and is particularly annoying if the front-end image should be substantially smaller than the container.
By defining DesignWidth and DesignHeight, you can force the image to scale down (or up) in the Layout Designer to better approximate its final front-end dimensions. It has no effect on its front-end dimensions.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Drivers/VectorImageElementDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Drivers/VectorImageElementDriver.cs
@@ -22,7 +22,9 @@ namespace Orchard.Layouts.Drivers {
             var viewModel = new VectorImageEditorViewModel {
                 VectorImageId = element.MediaId.ToString(),
                 Width = element.Width,
-                Height = element.Height
+                Height = element.Height,
+                DesignWidth = element.DesignWidth,
+                DesignHeight = element.DesignHeight
             };
             var editor = context.ShapeFactory.EditorTemplate(TemplateName: "Elements.VectorImage", Model: viewModel);
 
@@ -31,6 +33,8 @@ namespace Orchard.Layouts.Drivers {
                 element.MediaId = ParseVectorImageId(viewModel.VectorImageId);
                 element.Width = viewModel.Width;
                 element.Height = viewModel.Height;
+                element.DesignWidth = viewModel.DesignWidth;
+                element.DesignHeight = viewModel.DesignHeight;
             }
 
             var mediaId = element.MediaId;
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Elements/VectorImage.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Elements/VectorImage.cs
@@ -25,5 +25,17 @@ namespace Orchard.Layouts.Elements {
             get { return this.Retrieve(x => x.Height); }
             set { this.Store(x => x.Height, value); }
         }
+
+        public int? DesignWidth
+        {
+            get { return this.Retrieve(x => x.DesignWidth); }
+            set { this.Store(x => x.DesignWidth, value); }
+        }
+
+        public int? DesignHeight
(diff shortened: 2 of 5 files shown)


sebastienros:
Can you file  bug to track it and also talk about the issue itslef?

I wonder if a better solution wouldn't be to just define a max size for this element in a layout. Seems odd to drop an image then edit its properties just for the design view. Or it would make more sense to have it resizable in the designer with the mouse, and use your new properties to persist the changes. Either way, your change makes sense I assume.

sfmskywalker:
I acknowledge the issue, but I have reservations about having a design time width and height, because a user might expect the manually set size to reflect on the front end as well. Perhaps it makes more sense if there were support for a theme to provide a design-time style sheet that the layout editor would load. This way, the theme developer can apply CSS for vector images on the front end as well as for the layout designer. The name of the style sheet could be based on a convention. The layout editor would try and see if that stuck sheet exists on the "current" theme as stored in SiteSettings.

The advantage is that the user no longer has to manually deal with design time sizing to match the front end appearance. Instead, this happens automatically (provided that the theme developer did a correct job).

sebastienros:
I agree it should be part of the Layout's css that the element does overflow some limits.

To make it extensible, here is a suggestion:
- The element wrapper adds a class with the technical name of the element. 
- The element designer view injects it's own css definitions in the header using the resource manager.
- The element designer can either target its custom classes or the ones the layout is providing for the element.

sebastienros:
I filed this issue to follow up on the matter: https://github.com/OrchardCMS/Orchard/issues/7110

rtpHarry:
I've been thinking about this overnight. 

I think that #7110 is a good idea but doesn't solve the issue raised in this PR elegantly.

When you let a user loose on the layout and they want to upload a vector image and use it the VectorImage element will likely be a different size every time. I don't see how the #7110 idea will help. 

Either you have to have the user remember a bunch of class names which are being supplied by the theme or the user will have to go back and edit the theme each time to add a new rule in to handle it.

I don't think the DesignWidth/DesignHeight solution is quite right either. Instead I think it should be a checkbox the same way widgets can be named but not output the name as the header. 

So:
 - you only set the width and the height once (using the existing properties)
 - that is used in the editor vectorimage.design.cshtml view
 - a checkbox beneath it says if it should be rendered on front end like widget titles:
    image

sfmskywalker:
Perhaps I missed something, but I don't see how the user has to memorize the CSS class names. The idea is that the theme developer would target a class name that either she provides herself when crafting the VectorImage element template, or one that is generated based in the element type name, which is constant. That same theme developer would also provide a style sheet that is loaded by the layout editor UI that matches the rules provided for front-end display as to accommodate the same width and height on either ends in a representative manner.

rtpHarry:
Its because VectorImageElement could represent many different sizes of image and because it's a vector it defaults to 100% of its container. A single rule from a theme developer doesn't make sense.

A vector doesn't have any dimensions the browser can detect so it goes 100% each way. I assume this is why you(?) added the Width and Height properties in the first place (they aren't on the Image element).

The original issue isn't about it overflowing the container, its just that it auto uses the whole width which can really throw the design off.

I think VectorImageElement is a special case?

I guess adding the checkbox solution means that content editors can keep their admin view clean without affecting whatever is in place on the front end. Even with the solution proposed in #7110 it requires either duplicated css rules or a clever css pipeline setup to output a secondary stylesheet based on some core less/sass files. Factor in modules providing custom elements the theme hasn't envisaged and responsive design I can see why a site admin would want to tidy the editor page but not output the dimensions to the front end.

So I guess it comes down to should we add features that are there to help dashboard-only site admins keep things tidy on the editor side?

PR #7137, opened 2016-09-19
Allow custom date format when displaying published state

petedavis:
When overriding the Parts.Common.Metadata.cshtml in a theme, this will allow for changing the date format used to display the date.

### DIFF
--- a/src/Orchard.Web/Core/Common/Shapes.cs
+++ b/src/Orchard.Web/Core/Common/Shapes.cs
@@ -23,12 +23,12 @@ namespace Orchard.Core.Common {
         }
 
         [Shape]
-        public IHtmlString PublishedState(dynamic Display, DateTime createdDateTimeUtc, DateTime? publisheddateTimeUtc) {
+        public IHtmlString PublishedState(dynamic Display, DateTime createdDateTimeUtc, DateTime? publisheddateTimeUtc, LocalizedString customDateFormat) {
             if (!publisheddateTimeUtc.HasValue) {
                 return T("Draft");
             }
 
-            return Display.DateTime(DateTimeUtc: createdDateTimeUtc);
+            return Display.DateTime(DateTimeUtc: createdDateTimeUtc, CustomFormat: customDateFormat);
         }
 
         [Shape]


sebastienros:
Is it breaking anything if the new parameter is not set or null ?

petedavis:
I was a bit concerned with not setting it to default of null, but it looks like extension methods on a dynamic makes almost all parameters optional. I based on the DateTimeShapes.DateTime method. The CustomFormat parameter doesnt have a default value.

With the debugger attached, using the default template
`<div class="published">@Display.PublishedState(createdDateTimeUtc: Model.ContentPart.CreatedUtc, publisheddateTimeUtc: Model.ContentPart.PublishedUtc)</div>`

The `customDateFormat` parameter was null and the output was the format `Sunday, July 31, 2016 9:40:19 AM`

And adding a custom format:
`<div class="published">@Display.PublishedState(createdDateTimeUtc: Model.ContentPart.CreatedUtc, publisheddateTimeUtc: Model.ContentPart.PublishedUtc, customDateFormat: T("MMM dd, yyyy"))</div>`

Yields the format 'Jul 31, 2016'

petedavis:
One change that I did want to fix but was breaking was the casing on the parameter

`publisheddateTimeUtc`

Was driving me bonkers reading that.

rtpHarry:
@petedavis It looks like you might be able to fix that without breaking orchard.

The [DateTimeRelative in DateTimeShapes.cs](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Core/Shapes/DateTimeShapes.cs#L30) has had a similar fix and backwards compatibility maintained:

        [Shape]
        public IHtmlString DateTimeRelative(dynamic Display, DateTime DateTimeUtc, DateTime dateTimeUtc) {
            DateTimeUtc = DateTimeUtc != System.DateTime.MinValue ? DateTimeUtc : dateTimeUtc; // Both capitalizations retained for compatibility.
            var time = _clock.UtcNow - DateTimeUtc;

sebastienros:
Can you create an issue so we can track it when released?

petedavis:
Now worries @sebastienros. This Fixes #7148

PR #7149, opened 2016-09-23
Feature/admin theme Update EditPlacement.cshtml 

ramiassi:
If the script snippet is not located in the footer then jQuery will not be recognized.

### DIFF
--- a/src/Orchard.Web/Themes/TheAdmin/TheAdmin.csproj
+++ b/src/Orchard.Web/Themes/TheAdmin/TheAdmin.csproj
@@ -662,6 +662,9 @@
   <ItemGroup>
     <Content Include="Assets\Less\orchard-general.less" />
   </ItemGroup>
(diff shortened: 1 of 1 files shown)


Skrypt:
That's the kind of thing that will require to be set on the footer if some of your javascript functions are not executed on document.ready(). I prefer having it on top to force me to do so.

sfmskywalker:
Wasn't the issue that the `$` symbol was not available in the first place? If not, then you're right.

Skrypt:
Not supposed to. I think most of main modules are using JQuery in there. Lists whas working and it's using some JQuery plugin. So maybe the problem is that the customizations that @ramiassi did on his own custom modules would need to use document.ready().

sfmskywalker:
That's a good observation. I'll look into this in more detail and see what needs to be done. Thanks for being so vigilant Jasmin, appreciate it.

PR #7160, opened 2016-09-28
Replace Media Item: Replacing an Image/OEmbed with new one

hannan-azam:
Finally Replace Media is here. While developing this following points were kept in mind:

1. **No new media item should be created.** In case of media items with physical files only the physical files should be replaced such as Images and in case of OEmbed the entire Infoset within the OEmbed should be replaced. It also means that the ContentType of the existing item and the ContentType of new media being uploaded should match. In case of a mismatch for example trying to replace and Image with an OEmbed, or a Document with an Audio should display some error.

2. **Try to use out of the box code as much as possible** with minimum changes.

3. **Give the users same UI to Replace as they are used to for Importing new Item.** So for each media item in Media Gallery, once selected a new button "Replace" is available. Once clicked it opens up the same interface of "Import" and users can use File System, Web Search or OEmbed to replace the selected Item. The only catch is, the content type should match. So one can't replace an Image with and OEmbed. All appropriate checks are in place.

Please review this pull request and let me know if any changes are required.


rtpHarry:
@hannan-azam are you finished with your development of this feature or is it ongoing? I'm not trying to rush you but we will be having the weekly triage meeting in about an hour so if it's still being worked Seb needs to know not to merge it yet.

Great job with the work you have done so far by the way! 👍

hannan-azam:
@sebastienros @rtpHarry I'm done with the changes. It can now be integrated.

PR #7210, opened 2016-10-01
[Fixes #7180] Admin Theme - Modules - Module title on same row

### LINKED ISSUE #7180: Admin Theme - Modules - Module title looks like should be on same row
Spacings are there but not floated onto same row:

image

image

rtpHarry:
[Fixes #7180] Admin Theme - Modules - Module title on same row


rtpHarry:
I'm using rebracer, I thought that magically took care of these issues. Will investigate.

PR #7248, opened 2016-10-05
[Fixes #6966] Orchard.Fields - Added proper support for "not required" setting of DateTimeField

### LINKED ISSUE #6966: DateTimeField, proper support for "not required"

ViRuSTriNiTy:
See issue #6966


sebastienros:
This should target the dev branch

PR #7249, opened 2016-10-05
[Fixes #6967] Orchard.Projections - Fixed handling of nullable System.DateTime values

### LINKED ISSUE #6967: Projection does not handle nullable DateTimeFields correctly
Currently when no value is selected for a DateTimeField then `GetFilterPredicate()` determines `min` and `max` incorrectly as it expects a string formatted as `yyyy-MM-dd HH:mm:ss`.

The following change is needed to pass `min` and `max` as empty string (to avoid exception in `IHQLCriterion.FormatValue()`) or `0` when `asTicks` is `true`:

ViRuSTriNiTy:
See issue #6967


sebastienros:
Waiting for a repro of the issue. See comments in issue.

ViRuSTriNiTy:
No need for this pull request anymore, it's already part of PR #6980 which has been merged already.

PR #7275, opened 2016-10-09
[Fixes #7155] Orchard.Framework - GroupInfo generates routes with spaces

### LINKED ISSUE #7155: Orchard.Framework - Using GroupInfo() class generates urls with spaces in it
The [GroupInfo](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard/ContentManagement/GroupInfo.cs) class is used to represent a settings part in the menu.

If the settings part has a space in the name then the url does too:

    context.Metadata.EditorGroupInfo.Add(new GroupInfo(T("Audit Trail")));

Results in `/Admin/Settings/Audit Trail/` which works but isn't correct.

From my simple testing [.HtmlClassify()](https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard/Utility/Extensions/StringExtensions.cs#L58) seems to create a better url. 

Is there a more appropriate helper method tucked away somewhere in Orchard? If not I will create a PR.

rtpHarry:
[Fixes #7155] Orchard.Framework - GroupInfo generates routes with spaces

### DIFF
--- a/src/Orchard/ContentManagement/GroupInfo.cs
+++ b/src/Orchard/ContentManagement/GroupInfo.cs
@@ -1,11 +1,12 @@
 ﻿using Orchard.Localization;
+using Orchard.Utility.Extensions;
 
 namespace Orchard.ContentManagement {
     public class GroupInfo {
         private string _position = "5";
 
         public GroupInfo(LocalizedString name) {
-            Id = name.TextHint;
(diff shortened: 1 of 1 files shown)


rtpHarry:
I went with `.ToSafeName()` in the end which created nicer urls than my initial proposal.

PR #7295, opened 2016-10-11
Fixes Intenet Explorer 10 + flex display issues

jonvee:


sfmskywalker:
Can you please provide the issue number that describes the issue? Thanks

jonvee:
Sorry, only the second PR I've ever done, first on Orchard ;)
Hopefully the latest commit was OK

src/Orchard.Web/Orchard.Web.csproj
```diff
@@ -236,6 +236,9 @@
   <ItemGroup>
     <Folder Include="App_Data\" />
   </ItemGroup>
+  <ItemGroup>
+    <WCFMetadata Include="Service References\" />
+  </ItemGroup>
   <Target Name="MvcBuildViews" AfterTargets="AfterBuild" Condition="'$(MvcBuildViews)'=='true'">
```
Skrypt: I don't think this is needed. Revert that change for the PR.

sfmskywalker:
No worries, we're glad to have you on board :)

sfmskywalker:
Please also resolve the merge conflicts. You can do so by merging in the *feature/admin-theme* branch into your own branch, then push the new commit.

jonvee:
Pretty much making a mess over here with git... I'll have to look at it tonight. Noob issues ;P

sfmskywalker:
I hear ya. No rush.

jonvee:
What can I do to prevent these issues (unwanted file changes, rebracer settings etc in the future?

sebastienros:
@jonvee You don't have to commit all the files that have been changed, for instance with a tool like Git Extensions you can select precisely which files should be part of the commit, or which lines. You can also restore files selectively if you want to cancel some changes.

rtpHarry:
@jonvee the way to prevent the future mistakes is to make all of the mistakes several times until they stick in your head :) 

In your case that rebracer.xml is just because the plugin has updated its default config and we haven't reflected this in the repo yet (I've just opened a PR for that).

The other issue is because `TheAdmin.css` is automatically generated from all of the other `less` files. While there are a lot of people working on the new admin theme at the moment it means that when one person's PR gets accepted with the latest generated admin css it breaks everyone elses PRs because the amends in your version of the file would wipe out the just-accepted versions updates.

So you didn't actually do anything wrong in either case.

jonvee:
Thanks for all the feedback guys!!! Had a lot of interruptions trying to sort out git, but I assume since it was merged I finally got it right..? ;)

sebastienros:
yep, the files looked good

PR #7302, opened 2016-10-12
[Fixes #6494] SSL missing settings banner showing on frontend

### LINKED ISSUE #6494: SSL module is showing admin notification to all users on all pages
When Socure Sockets Layer module is enabled and not configured it will notify with message "You need to configure the SSL settings." ... This is clearly intended for administrators but is displayed to all users (including anonymous) on frontend.

```
public void OnActionExecuted(ActionExecutedContext filterContext) {
    var settings = _sslService.GetSettings();

    if (!settings.Enabled) {
        _orchardServices.Notifier.Warning(T("You need to configure the SSL settings."));
    }
}
```
I don't think this message is THAT important to have it displayed on all pages even for administrators, should probably appear only on administration for admins similarly to how "The SMTP settings need to be configured." message from **Orchard.Email** works.

rtpHarry:
[Fixes #6494] SSL missing settings banner showing on frontend

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Filters/SecureSocketsLayersFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Filters/SecureSocketsLayersFilter.cs
@@ -3,7 +3,6 @@ using System.Web.Mvc;
 using Orchard.Localization;
 using Orchard.Mvc.Filters;
 using Orchard.SecureSocketsLayer.Services;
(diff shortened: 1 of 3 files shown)


rtpHarry:
The issue was tagged as 1.10.x so I targeted that but then I got naughty and improved the error message so this probably wants to be put to dev branch now.

sfmskywalker:
Perhaps it's not a big deal that the error message was changed, and we can still target 1.10.x? @sebastienros

PR #7305, opened 2016-10-12
Fixed an issue with layout element lookup

### LINKED ISSUE #7304: GetElementDescriptorByTypeName should not be case sensitive
The issue becomes apparent when you create an [element snippet](http://www.ideliverable.com/blog/parameterized-snippets) and you want to turn that snippet into a widget (leveraging the [Element Wrapper part](http://www.ideliverable.com/blog/orchard-layouts)).

When you configure the `ElementWrapperPart` with the type name of the snippet element, you currently need to be aware that this type is based on the shape type name of the Razor view that represents the snippet. The Razor shape template harvester, however, generates a shape type name based on the Razor view file name, using all lower case.

For example, if I create a Razorfile called `"MyLogoSnippet.cshtml"`, the shape type name will become `"mylogosnippet"`. This name is used in the `SnippetElementHarvester` as the _element's type name_.

If I now configure the `ElementWrapperPart` with an element type name of MyLogoSnippet, the part will not be able to find any elements by that name, because `GetElementDescriptorByTypeName` is case-sensitive:

```C#
public ElementDescriptor GetElementDescriptorByTypeName(DescribeElementsContext context, string typeName) {
    var elements = DescribeElements(context);
    var element = elements.SingleOrDefault(x => x.TypeName == typeName);

    return element;
}
```

To fix this situation, a case-insensitive predicate should be used:

```C#
var element = elements.SingleOrDefault(x => String.Equals(x.TypeName, typeName, StringComparison.OrdinalIgnoreCase));
```

That way, you don't have to be aware of using the all lower case "mylogosnippet" value for the element type name on the ElementWrapperPart.

This seems to be a non-breaking change, so I would propose to fix this as part of 1.10.x.

sfmskywalker:
Fixes #7304

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Services/ElementManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Services/ElementManager.cs
@@ -79,7 +79,7 @@ namespace Orchard.Layouts.Services {
 
         public ElementDescriptor GetElementDescriptorByTypeName(DescribeElementsContext context, string typeName) {
             var elements = DescribeElements(context);
-            var element = elements.SingleOrDefault(x => x.TypeName == typeName);
+            var element = elements.SingleOrDefault(x => String.Equals(x.TypeName, typeName, StringComparison.OrdinalIgnoreCase));
 
             return element;
         }


src/Orchard.Web/Modules/Orchard.Layouts/Services/ElementManager.cs
```diff
@@ -79,7 +79,7 @@ orderby elementDescriptor.DisplayText.Text
 
         public ElementDescriptor GetElementDescriptorByTypeName(DescribeElementsContext context, string typeName) {
             var elements = DescribeElements(context);
-            var element = elements.SingleOrDefault(x => x.TypeName == typeName);
+            var element = elements.SingleOrDefault(x => String.Equals(x.TypeName, typeName, StringComparison.OrdinalIgnoreCase));
```
sfmskywalker: True, a dictionary would be much faster, provided that we cache it (otherwise there would be no gain, since it would to iterate over all elements to build the dictionary in the first place). We could use the type name (lower case) as the key.

I still don't buy the other argument against `SingleOrDefault`, because if you _know_ there will only ever be one match at the most, then it is a great help that it will throw an exception in case the set contains more than one result, as this would indicate a bug. Anything else would cover up the bug.

sfmskywalker: Well, in this particular case, imagine a layout containing an element provided by Feature A. Now, Feature A is disabled, so when the layout is now loaded, the element pointing to the element descriptor provided by Layout A can no longer be found. I think this is a valid case for returning `null` rather than throwing an exception. hence, `SingleOrDefault` instead of `Single`.

But perhaps you meant something else.

PR #7306, opened 2016-10-13
Fixed an issue with incorrect log message at Indexing module

### LINKED ISSUE #7225: Incorrect log message in Indexing module.
It's not that important, but I just want to note that a log message is not correct. :)

Take a look at [Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs#L298](https://github.com/OrchardCMS/Orchard/blob/a6f46d17cdfb46cd5573a3e4e9aebfd6a5d042a9/src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs#L298).

```c#
_indexProvider.Delete(indexName, deleteFromIndex);
Logger.Information("Added content items to index: {0}", addToIndex.Count);
```

viniciusdfx:
Fixes #7225

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
+++ b/src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
@@ -295,7 +295,7 @@ namespace Orchard.Indexing.Services {
             try {
                 if (deleteFromIndex.Count > 0) {
                     _indexProvider.Delete(indexName, deleteFromIndex);
-                    Logger.Information("Added content items to index: {0}", addToIndex.Count);
+					Logger.Information("Deleted content items from index: {0}",  deleteFromIndex.Count);
                 }
             }
             catch (Exception ex) {
--- a/src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
+++ b/src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
@@ -295,7 +295,7 @@ namespace Orchard.Indexing.Services {
             try {
                 if (deleteFromIndex.Count > 0) {
                     _indexProvider.Delete(indexName, deleteFromIndex);
-					Logger.Information("Deleted content items from index: {0}",  deleteFromIndex.Count);
+			    Logger.Information("Deleted content items from index: {0}",  deleteFromIndex.Count);
                 }
             }
             catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
```diff
@@ -295,7 +295,7 @@ public class IndexingTaskExecutor : IIndexingTaskExecutor, IIndexStatisticsProvi
             try {
                 if (deleteFromIndex.Count > 0) {
                     _indexProvider.Delete(indexName, deleteFromIndex);
-                    Logger.Information("Added content items to index: {0}", addToIndex.Count);
+					Logger.Information("Deleted content items from index: {0}",  deleteFromIndex.Count);
```
sfmskywalker: It seems that you inserted a tab or two. Please use spaces for indentation instead. Thanks

viniciusdfx: Now it's correct? Sorry for the mistake

rtpHarry: No it still has mostly tabs in it. Orchard follows some specific code formatting conventions such as spaces instead of tabs and putting the { on the same line, javascript style, instead of on a new one.

Luckily you can automate this with a free plugin for Visual Studio called [rebracer](https://visualstudiogallery.msdn.microsoft.com/410e9b9f-65f3-4495-b68e-15567e543c58). 

Once installed as you make new edits it will automatically follow these rules but to fix this file you will need to:

  1. install rebracer
  1. open IndexingTaskExecutor.cs
  1. press <kbd>CTRL-K</kbd> + <kbd>CTRL-D</kbd> to apply source formatting

rtpHarry:
The only last thing was that you haven't followed the correct base branch. When an item is triaged (reviewed each thursday in a skype video meeting) we look at the issue and if it is a valid problem then a Milestone is applied.

When you start your fix you should base your forked branch on the milestone branch. So if you had checked #7225 you would have seen the milestone say 1.10.x which means you should have targeted the 1.10.x branch.

As it's a minor edit I have swapped the target branch for the issue over to dev so you don't need to make more changes on your side. I'll leave it for @sfmskywalker to give it a final review and confirm this is acceptable.

viniciusdfx:
Thank you for the tips, @rtpHarry 

I'll keep in mind those things

PR #7324, opened 2016-10-15
[Fixes #7251, Fixes #1659] Orchard.Setup - Blog recipe doesn't create widgets correctly

### LINKED ISSUE #1659: The command "widget create" doesn't contain which blog to show
__Yinhaixia__ created:
https://orchard.codeplex.com/workitem/17829

Log on the Orchard site as admin, and create some blogs
Run bin>orchard.exe, then run "widget create RecentBlogPosts /title:1234 /zone:Header /position:1 /layer:Default"
The command treats the latest blog as default, so if I want to use the older one, I cannot find a place to choose the blog by command

### LINKED ISSUE #7251: Orchard.Setup - Blog recipe doesn't create widgets correctly
I thought #7169 and friends would be simple issues but they have caught me out.

My [blog recipe](https://github.com/OrchardCMS/Orchard/pull/7166) doesn't actually work as intended. I checked the admin after it had run and the UI showed the widgets (Archives, Recent) with the correct blog selected but this was just because that was the only value in the list. The actual `BlogId` on the part is `0` which means when posts are added it doesn't list them.

I can't see any way to pass a blog id in through the `widget create RecentBlogPosts` command.

I'm thinking instead I should catch the blog id being 0 and attempt to set a sensible default in `RecentBlogPostsPartHandler.cs` for example:

    using Orchard.Blogs.Models;
    using Orchard.Blogs.Services;
    using Orchard.ContentManagement.Handlers;
    using Orchard.Data;

    namespace Orchard.Blogs.Handlers {
        public class RecentBlogPostsPartHandler : ContentHandler {
            public RecentBlogPostsPartHandler(
                IRepository<RecentBlogPostsPartRecord> repository,
                BlogService blogService) {

                Filters.Add(StorageFilter.For(repository));
            
                OnCreated<RecentBlogPostsPart>((context, part) => {
                    // If the part.BlogId = 0 then assume its being created via a command (ui shouldn't allow it)
                    // the widget create command can't specify a blog id so we attempt to infer it here:                
                        // get list of blogs ordered by last modified date
                        // set Part.BlogId to the last modified blog
                });
            }
        }
    }

This should mean that in a recipe as long as the recent/archive parts are created after the blog they want to be assigned to they will be attached to that blog.

//cc @sebastienros before I code it, is this a good approach?

rtpHarry:
[Fixes #7251, Fixes #1659] Orchard.Setup - Blog recipe doesn't create widgets correctly

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Blogs/Commands/BlogWidgetCommands.cs
+++ b/src/Orchard.Web/Modules/Orchard.Blogs/Commands/BlogWidgetCommands.cs
@@ -1,0 +1,165 @@
+﻿using System;
+using System.Linq;
+using Orchard.Blogs.Models;
+using Orchard.Blogs.Services;
+using Orchard.Commands;
+using Orchard.ContentManagement;
+using Orchard.ContentManagement.Aspects;
+using Orchard.Core.Common.Models;
+using Orchard.Security;
+using Orchard.Settings;
+using Orchard.Widgets.Models;
+using Orchard.Widgets.Services;
+
+namespace Orchard.Blogs.Commands {
+    public class BlogWidgetCommands : DefaultOrchardCommandHandler {
+        private readonly IWidgetsService _widgetsService;
+        private readonly IBlogService _blogService;
+        private readonly ISiteService _siteService;
+        private readonly IMembershipService _membershipService;
+        private readonly IContentManager _contentManager;
+
+        private BlogPart blog;
+
+        public BlogWidgetCommands(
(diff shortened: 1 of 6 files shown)


rtpHarry:
This fixes the issue but not in the cleanest way I would have liked.

As was targeted to 1.10.2 I couldn't change any interfaces. I will provided a refactored version shortly for dev branch. It will update the `IWidgetService` and move the duplicated code which creates a standard widget. Then all three commands can share the same boilerplate.

PR #7325, opened 2016-10-15
[Fixes #7071] Orchard.Users - Users last login time is not correct

### LINKED ISSUE #7071: Users last login time is not correct
This is in version 1.10.1.0, so if it got fixed please disregard. I did a search and didn't see anything about it. I have attached a screen cap of the problem but basically it say my last login time is 2016 years ago. I must be really old.
issue

rtpHarry:
[Fixes #7071] Orchard.Users - Users last login time is not correct


rtpHarry:
Backporting @dcinzona's PR #6441 to close this blocking issue for 1.10.2

PR #7333, opened 2016-10-19
- added EventHandler in order to participate in cachekey generation

HermesSbicego-Laser:
introduce partial support to #5790

Issue #5790, opened 2015-09-11
Add events to the Output Cache filter in order to allow modules to extend it


paynecrl97:
3rd party (or even 1st party) modules are currently very restricted in the way that they can interact with the Output Cache functionality.

I propose adding an event handler interface to allow developers to hook into specific events offered by the Output Cache filter.

This could look something along the lines of this:

    public interface IOutputCacheEventHandler : IEventHandler
    {
        void Caching(string cacheKey, CacheItem cacheItem);
        void Cached(string cacheKey, CacheItem cacheItem);
        void Retrieving(string cacheKey);
        void Retrieved(string cacheKey, CacheItem cacheItem);
    }

Example use cases would be:

* Altering markup immediately before the item is cached (for example to remove the Glimpse console)
* Adding additional logging or diagnostics

DanielStolt:
:+1:

sebastienros:
I also prefer this suggestion to the old PR. One comment would be to have a CachedItemContext and CachingItemContext instead of two parameters, and also add a way to Cancel in the Caching one.

DanielStolt:
@sebastienros A question on that. When we introduce the `IOutputCacheControlProvider` interface it will be encouraged to implement that interface to "cancel" caching of a request. If we also add a cancellation mechanism to the `Caching` event here, there will be two ways to prevent caching. Do you see an issue with this, or do you think it's fine?

sebastienros:
Well I think these event provide enough extensibility and the other interface is not necessary,

DanielStolt:
Well... the plan is for the other interface to have much more functionality for cache control such as:

- Control whether request can be served from cache
- Control response cacheability
- Participate in cache key generation
- Control duration
- Control gracetime
- Control cache storage locations

All these things will be necessary for the next step which is to make output cache configuration much more composable and collaborative between whatever parts and filters participate in a given request.

If we add these things to this event handler approach, then I agree, the cache control interface is of course unnecessary. But I think that would not be a very good design, because it would become tricky to determine which events should be used for which cache control operations, and also if the same "context" object is passed around through all event handlers it becomes trickier to implement things like collaboratively agreeing on the max duration allowed by all parts involved in the rendering...

Thoughts?

sebastienros:
All your points make even more sense for me with event handlers, as many parts can affect the process. As long as the context classes support these extensibility points. One handler could invalidate another's changes, or append something in the cache key, add more time to a gracetime, ...

[...]

PR #7334, opened 2016-10-19
Feature/5790 participate in cache key generation

HermesSbicego-Laser:
- typo fix #7333

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -41,7 +41,7 @@ namespace Orchard.OutputCache.Filters {
         private readonly ICacheService _cacheService;
         private readonly ISignals _signals;
         private readonly ShellSettings _shellSettings;
-        private readonly ICachingEventHandler _chachingEvents;
+        private readonly ICachingEventHandler _cachingEvents;
         private bool _isDisposed = false;
 
         public ILogger Logger { get; set; }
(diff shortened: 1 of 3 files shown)


rtpHarry:
Thanks for your contributions!

It looks like you have opened a second PR for the same thing. When you make a commit to the same branch that you used in the first pull request it automatically updates that pull request.

If you look back at your first PR then you will see the typo commits are in there already so you can close this second PR unless I'm missing some other changes (I've just had a quick glance).

src/Orchard.Web/Modules/Orchard.OutputCache/ICachingEventHandler.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Orchard.Events;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Web;
+
+namespace Orchard.OutputCache {
+    public interface ICachingEventHandler : IEventHandler {
+        StringBuilder ParticipateInCacheKey(StringBuilder key);
```
sebastienros: It should be something like `KeyGenerated`. 
It should not return anything.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -610,6 +613,14 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
                 }
             }
 
+            //make CacheKey morphable by external modules
+            try {
```
sebastienros: This is unnecessary, just let it throw if there is an issue, remove the try/catch.

sebastienros:
This is a very limited feature, but we can always add more events later.

HermesSbicego-Laser:
I will apply changes tomorrow.

HermesSbicego-Laser:
Can partially umplement what described in Add events to the Output Cache filter in order to allow modules to extend it #5790

Issue #4718, opened 2014-08-18
Jobs queue service causes exception


orchardbot:
**kobowi** created:
https://orchard.codeplex.com/workitem/20889

I'm building some functionality to do background processing (couple of web service requests, update db), using the Jobs Queue module. Jobs Queue exposes an interface called IJobsQueueService:
public interface IJobsQueueService : IEventHandler {
    QueuedJobRecord Enqueue(string message, object parameters, int priority);
}

However when calling this from my code I get an exception like this:
2014-08-18 16:59:05,280 [17] Orchard.Exceptions.DefaultExceptionPolicy - Default - An unexpected exception was caught
System.InvalidOperationException: Sequence contains no elements
 at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source)
 at Orchard.Events.Extensions.GetGenericMethod(Type t, String name, Type[] genericArgTypes, Type[] argTypes, Type returnType) in e:\Code\EA\Sites\experimental\src\Orchard\Events\EventsInterceptor.cs:line 48
 at Orchard.Events.EventsInterceptor.<Adjust>b__7(Type type) in e:\Code\EA\Sites\experimental\src\Orchard\Events\EventsInterceptor.cs:line 41
 at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
 at Orchard.Events.EventsInterceptor.Adjust(IEnumerable results, Type returnType) in e:\Code\EA\Sites\experimental\src\Orchard\Events\EventsInterceptor.cs:line 41
 at Orchard.Events.EventsInterceptor.Intercept(IInvocation invocation) in e:\Code\EA\Sites\experimental\src\Orchard\Events\EventsInterceptor.cs:line 28
 at Castle.DynamicProxy.AbstractInvocation.Proceed()
 at Castle.Proxies.IJobsQueueServiceProxy.Enqueue(String message, Object parameters, Int32 priority)
 at EA.Development.Controllers.TestController.Queue(String name)
 at lambda_method(Closure , ControllerBase , Object[] )
 at System.Web.Mvc.ReflectedActionDescriptor.Execute(ControllerContext controllerContext, IDictionary`2 parameters)
 at System.Web.Mvc.ControllerActionInvoker.InvokeActionMethod(ControllerContext controllerContext, ActionDescriptor actionDescriptor, IDictionary`2 parameters)
[...]

PR #7335, opened 2016-10-20
Correctly implementing IJobsQueueService

kumards:
Fix for #4718. IJobsQueueService supposed to be just a service than an event handler.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.JobsQueue/Services/IJobsQueueService.cs
+++ b/src/Orchard.Web/Modules/Orchard.JobsQueue/Services/IJobsQueueService.cs
@@ -1,8 +1,7 @@
-﻿using Orchard.Events;
 using Orchard.JobsQueue.Models;
 
 namespace Orchard.JobsQueue.Services {
-    public interface IJobsQueueService : IEventHandler {
+    public interface IJobsQueueService : IDependency {
         QueuedJobRecord Enqueue(string message, object parameters, int priority);
     }
 }


sebastienros:
Why? Isn't the solution to fix the bug in the even bus?

kumards:
Yeah, it fix solved my problem. I believe by mistake IJobsQueueService inherited IEventHandler instead of IDependency.

sebastienros:
No, it's by design, on purpose that it's an `IEventHandler`, but there is a bug in event handlers, which needs to be fixed, c.f. linked issues. I am closing this PR, please create one that fixes the event handler, it should be explained in this one: https://github.com/OrchardCMS/Orchard/issues/4584

kumards:
Does that mean, there can be multiple implementations of `IJobsQueueService `?

I guess if `IJobsQueueService  `is going to inherit from `IEventHandler`, then it has to be injected like collection `IEnumerable<IJobsQueueService> jobsQueueServices`, not as a single instance.

sfmskywalker:
Although technically speaking one could of course create multiple implementations, you don't inject an enumerable, but a single instance. This baffled me too at first. Check this out for more details (inclyuding the comments): http://www.ideliverable.com/blog/ieventhandler

kumards:
I am just curious, why is it intentionally inheriting from `IEventHandler` when multiple implementations of it is not expected?

sfmskywalker:
The reason is that this enables other modules to use the `IJobsQueueService` without having to actually add a project reference to *Orchard.JobsQueue*. This is actually a very nice pattern to implement modules with truly optional dependencies, where the dependent module would still build even if there was no *Orchard.JobsQueue* available. The only sad part is the poor choice of name, because to your point, `IEventHandler` would indicate a handler of an event or events. I wouldn't be surprised if this will be renamed in O2, or if at least an additional interface with a more appropriate name will be provided that leverages the same event bus in the same way.

PR #7337, opened 2016-10-21
Fix for Jobs queue service throwing exception

kumards:
EventsInterceptor fails when it tries to locate a generic method on an implementation with a return type. A null check has been added to avoid the exception.

Related issues: #4718, #4584 

Thanks @sebastienros, @sfmskywalker and @CSurieux for the fix.


sebastienros:
LGTM 
Can you rebase your change on top of 1.10.x as triaged?

kumards:
Created a new PR #7360 on 1.10.x.

PR #7361, opened 2016-10-28
Changed Audit Trail trimming task to delete all records in one batch

gcsuk:
I've changed the Audit Trail trimming task to delete all records in one batch, rather than iterating through the table records. This will be a far more efficient method of clearing up the audit trail.

I brought in ITransactionManager instead of changing IRepository to support multiple deletions, as this would have less of an impact on the codebase.


sfmskywalker:
Makes a lot of sense to me.

Issue #7362, opened 2016-10-30
Problem when reading SiteSettings from the Routes class


ThaerAlAjlouni:
I am trying to read **SiteSettings** in the **Routes** class, because I need to optionally add a route depending on some setting values, but it will throw an error each time I try to do that because the **WorkContext** instance will be _null_. I knew that a similar issue ( #6991 ) has been resolved and I think the same fix can be applied here.

As you can see below (**DefaultOrchardShell** class) the OwinMiddlewares and Orchard's  Shell events are wrapped with a **WorkContextScope** while Routes are not:
![defaultorchardshell](https://cloud.githubusercontent.com/assets/22820843/19833814/2066b0ba-9e1e-11e6-90e1-eceb2e28452b.png)

sebastienros:
I am afraid that in this case the _routeProviders and _httpRouteProviders instances should be resolved from the same work context.

ThaerAlAjlouni:
So the fix will be extending the **scope** of the created work context to everything inside the **Activate** method, right?

PR #7363, opened 2016-10-30
Fix  the problem preventing reading SiteSettings from the Routes class

ThaerAlAjlouni:
This is a proposed fix to the issue #7362


sebastienros:
Have you tried to run the Specflow tests, just to be sure we are not breaking anything? Or have you deployed it to test?
Can you try with a middelware from a module, that would use services too?

sebastienros:
Note to myself, it's on dev so we would have time to find any issue.

ThaerAlAjlouni:
No I haven't run Specflow tests, but I have deployed it and tested it in a middleware which I created in a custom module.

ThaerAlAjlouni:
I am sorry, that happened by mistake

Issue #7372, opened 2016-11-02
Active theme language file overwritten by other themes


adrianbooth:
when you have multiple themes installed with the same msgid but different msgstr there is no guarantee that the theme currently in use is going to take issue - PR to follow

adrianbooth:
I propose we treat the active theme differently to inactive themes like this: https://github.com/OrchardCMS/Orchard/pull/7373

sebastienros:
Could you provide some repro steps? I don't understand how you would get multiple msgstr with different values. Are you using .po files in each of the themes? And you changed one of them with a different value?

adrianbooth:
if we were to have a view that had a greeting like this:

`T("Welcome to our site")`

then we had two themes, possibly an easter and a christmas theme, each with a po file which attempt to replace the message
easter 
`msgid "Welcome to our site"
msgstr "Happy Easter from fakeSite"'

christmas
'msgid "Welcome to our site"
msgstr "Merry Christmas from fakesite'

the theme files would be gven equal priority and enabling the easter theme may give a Christmas message or vice versa

PR #7373, opened 2016-11-02
prioritise current theme over inactive themes to resolve issue 7372

adrianbooth:


src/Orchard/Localization/Services/DefaultLocalizedStringManager.cs
```diff
@@ -118,70 +121,49 @@ public class DefaultLocalizedStringManager : ILocalizedStringManager {
         private IDictionary<string, string> LoadTranslationsForCulture(string culture, AcquireContext<string> context) {
             IDictionary<string, string> translations = new Dictionary<string, string>();
             string corePath = string.Format(CoreLocalizationFilePathFormat, culture);
-            string text = _webSiteFolder.ReadFile(corePath);
```
adrianbooth: extracted this block (and others below) as it seemed a lot of repetition

src/Orchard/Localization/Services/DefaultLocalizedStringManager.cs
```diff
@@ -118,70 +121,49 @@ public class DefaultLocalizedStringManager : ILocalizedStringManager {
...
-                    }
+                    LoadTranslationsFromLanguageFilePath(modulePath, translations, context, true);
                 }
             }
 
-            foreach (var theme in _extensionManager.AvailableExtensions()) {
-                if (DefaultExtensionTypes.IsTheme(theme.ExtensionType) && _shellDescriptor.Features.Any(x => x.Name == theme.Id))
-                {
-
+            var currentTheme = _workContextAccessor.GetContext().CurrentTheme;
```
adrianbooth: this is the only functional change, splitting out the currently active theme and the inactive ones so inactive themes can be loaded then any overrides in the current theme will not be lost

src/Orchard/Localization/Services/DefaultLocalizedStringManager.cs
```diff
@@ -118,70 +121,49 @@ public class DefaultLocalizedStringManager : ILocalizedStringManager {
...
-                        if (!DisableMonitoring) {
-                            Logger.Debug("Monitoring virtual path \"{0}\"", themePath);
-                            context.Monitor(_webSiteFolder.WhenPathChanges(themePath));
-                        }
-                    }
+                    LoadTranslationsFromLanguageFilePath(themePath, translations, context, true);
                 }
             }
 
+            string currentThemePath = string.Format(ThemesLocalizationFilePathFormat, currentTheme.VirtualPath, culture);
```
adrianbooth: this would previously be handled the same as an inactive theme above and possibly not take priority

adrianbooth:
this is in relation to Issue:7372 https://github.com/OrchardCMS/Orchard/issues/7372

sebastienros:
I am afraid that the `DefaultLocalizedStringManager` class is actually called only once at startup, and then the `CurrentTheme` call would just fix the issue for the one that is the first to be called.

adrianbooth:
Ah of course, I suppose this is a reflection of how we (ab?)use Orchard. We have many themes available to us in our environments but they aren't a number of themes for one site, they are completely seperate sites and we would never switch between them.

While I still think this change is required, to make the fix work more generally we need to listen for the "SiteCurrentTheme" signal being triggered in addition to "culturesChanged" as an indicator of when the language file cache should be refreshed.

src/Orchard/Localization/Services/DefaultLocalizedStringManager.cs
```diff
@@ -118,70 +122,53 @@ public class DefaultLocalizedStringManager : ILocalizedStringManager {
...
+                    LoadTranslationsFromLanguageFilePath(modulePath, translations, context, true);
                 }
             }
 
-            foreach (var theme in _extensionManager.AvailableExtensions()) {
-                if (DefaultExtensionTypes.IsTheme(theme.ExtensionType) && _shellDescriptor.Features.Any(x => x.Name == theme.Id))
-                {
+            var sitePart = _workContextAccessor.GetContext().CurrentSite as IContent;
+            var themePart = sitePart.ContentItem.Parts.FirstOrDefault(e => e.PartDefinition.Name == "ThemeSiteSettingsPart");
+            var currentThemeName = themePart?.Retrieve<string>("CurrentThemeName");
```
sebastienros: Haven't looked in depth at the PR but there shouldn't be any reason to look into the settings directly, it should use the Theme selectors instead, with the ThemeManager.

PR #7390, opened 2016-11-09
Fixes #7374.

### LINKED ISSUE #7374: Field Link.Edit.cshtml problem with migration
Upon writing a custom module that uses the following Field migration -
```
`.WithField("LinkButton", cfg => cfg.OfType("LinkField")
                    .WithDisplayName("Button URl Link"))
```
The Link.Edit template renders the error -
Compiler Error Message: CS0103: The name 'isRequired' does not exist in the current context

If I include the following line of code from commit be323ab, the code runs fine? 

`var isRequired = settings.Required && String.IsNullOrWhiteSpace(settings.DefaultValue);`

benschi11:
With the merge from f51c8a72789dc546fec08c3058f372edec9e4a78 a few things were getting lost.
Merged the defaultValue feature with the current code.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Fields/Drivers/BooleanFieldDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Fields/Drivers/BooleanFieldDriver.cs
@@ -49,7 +49,12 @@ namespace Orchard.Fields.Drivers {
                 var settings = field.PartFieldDefinition.Settings.GetModel<BooleanFieldSettings>();
 
                 if (!settings.Optional && !field.Value.HasValue) {
-                    updater.AddModelError(field.Name, T("The field {0} is mandatory.", T(field.DisplayName)));
+                    if (settings.DefaultValue.HasValue) {
+                        field.Value = settings.DefaultValue;
+                    }
+                    else {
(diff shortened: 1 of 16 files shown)


sebastienros:
I think we removed some things intentionally. @jtkech you agree?

jtkech:
Seems to be mainly around default values (and their tokenizations). As i remember, they was first used to provide default values when nothing is submitted. But after a long discussion they are now used as proposed values when 1st editing on item creation, and then without tokenization.

But i'm open on everything.

Best.

BenedekFarkas:
So what should we with this PR @sebastienros @jtkech?

jtkech:
If we still agree that it is better to use default values as proposed values when 1st editing on item creation, personally i think that you can close this PR, but i'm not against any other suggestion.

BenedekFarkas:
Thanks @jtkech, I think I remember this discussion now.

So the idea is that PR would change the behavior from the current one, so instead of applying the default value to the editor when creating a content item, it would always (every save) apply when the value is empty. I think I'd prefer to keep the current implementation or apply this PR with the option to disable enforcing it. But if that's too much, the logic in this PR can actually be implemented in a 3rd-party module, so we don't need to merge, but it still can be used.

I'll close it for now then, but @benschi11 please comment if you'd like to discuss further options.

PR #7399, opened 2016-11-11
[Fixes #7378] fixed tenent name validation regex, not to allow numbers.

### LINKED ISSUE #7378: Error when starting a new tenant with a name starting with the '3' character.
`Orchard.Environment.DefaultOrchardHost - (null) - A tenant could not be started: 3aaaa Attempt number: 0 [(null)]
System.Xml.XmlException: Name cannot begin with the '3' character, hexadecimal value 0x33. Line 4, position 4.`

Should a fix, disallow tenant name begging with numbers or should we try to make this work?

SzymonSel:


sebastienros:
Why the `+`, it would mean it has to contain 2 chars at least. A `*` assume you can have zero or more, so there could be a single char if you want. Remember the `[a-zA-Z]` imposes at least one char.

SzymonSel:
My bad, you're right.

PR #7400, opened 2016-11-11
Cloning and Translation fixes

TFleury:
Small corrections based on @MatteoPiovanelli-Laser comments on 87a394854dda9a45b36d757b3d3ce135ac18d24d


MatteoPiovanelli-Laser:
I have one more comment on the Translate method in the AdminController of the Localization module:

The redirect you do at the end should probably be to the EditorRouteValues of the metadata:
I think those two match for some simple contentItem cases, but not for more advanced ones (e.g. blogposts seem to be ok; blogs are not).

Issue #7404, opened 2016-11-15
Users' own Media folder PathValidation


LorenzoFrediani-Laser:
A user with ManageOwnMedia permission can upload files to a folder that can't be accessed by other users which is Media/{Tenant name}/Users/{User.Name}.
User.Name can accept any char es:"*?", when user with these chars on username try adding media file to ContentItem  an error of ValidationPath occour  and folder can't be created.

Other problem, i think that to expose username can create Security Risk.

Other problem, changing username -> folder can't be accessed by user

I suggest a solution for the 3 problems 
change Media/{Tenant name}/Users/{User.Name}  to Media/{Tenant name}/CustomMedia/{User.Id}

sebastienros:
- Normalizing the username would introduce a security issue as there would be collisions
- Using the id is a breaking change

The solution is not to normalize but to encode the username so that forbidden chars are replaced by supported ones.

Example: 
username: lorenzo*?
would become: lorenzo_2A_3F

It's using ASCII encoding, for any char that is not alphanumeric.

Also, if we want to be able to hide the username from urls, we need to create an extensibility extension. Let's create an `IMediaUsernameProvider` that would convert a username to the folder name. The default implementation is the one described above. Any other module can provide alternative implementations.

sebastienros:
One example could be to attache an IdentityPart to the User type, and use it as the folder name.

PR #7405, opened 2016-11-15
- Implements what described into #7404

LorenzoFrediani-Laser:
- Manage Own Media permission, created Users/UserName folder that sounds as having some problems. Now creates CustomMedia/User.Id folder in rder to prevent:
 1. Cannot access Media folder when User Name changes
 2. Media Folder name broken with special chars
 3. Potential security issue exposing username within public URL


sebastienros:
@MatteoPiovanelli-Laser  and @HermesSbicego-Laser will explain why

PR #7406, opened 2016-11-15
Order contained content items in ascending order

ramiassi:
As widgets and all other elements are ordered by position in ascending order. Why the contained items are ordered in descending order?


sebastienros:
Can you provide a screenshot of the end result? Or some repro showing the issue?

ramiassi:
Right now what is happening is like this:
a) If in admin -> widgets: 
widget1.Position = 10;
widget2.Position = 20;
widget3.Position = 30;
widget4.Position = 40;
They are displayed in this order:
1. widget1
2. widget2
3. widget3
4. widget4
b) after attaching the containable part to a content type and allowing the user to set position:
if the user set the positions like this:
containedItem1.Position = 10;
containedItem2.Position = 20;
conatinedItem3.Position = 30;
containedItem4.Position = 40;
then they are displayed in this order in the container’s item detailed view page:
1. containedItem4
2. containedItem3
3. containedItem2
4. containedItem1

Issue #6840, opened 2016-04-26
Publish later tasks on older content item versions still execute/are not removed when a newer version is published


andrewcartwright1:
When using the audit trail part on a content item, if a content item is saved by pressing publish later, but then published and a newer version is also published (i.e version 2) when the publish later time is hit, the older version of the content item will be published.

Replication Steps: As an admin
- Enable the audit trail module and attach the part to a content item
- Create a content item and click ​_save_​
- With the same content item set a publish later date and click ​_publish later_​
- With the same content item click ​_publish now_​
- With the same content item click ​_save_​
- With the same content item click ​_publish now_​

Wait until the publish later time has passed and then check the audit trial and the front end site

Example:

![zappshot - 2016-04-26_04-31-21-pm](https://cloud.githubusercontent.com/assets/10059155/14824330/7fdc95f2-0bcd-11e6-9652-884e98c3ecdf.png)

sebastienros:
Does it happen also when you are not using the audit trail part? Trying to understand if the issue is related to it or not.

andrewcartwright1:
Yes, It also looks to be an issue without the part, audit trail below is from Orchard.AuditTrail

![zappshot - 2016-04-29_08-36-47-am](https://cloud.githubusercontent.com/assets/10059155/14910477/990bbf7a-0de5-11e6-9c2c-dd2af5260592.png)

sebastienros:
Thanks, it will help us to replicate and fix the issue.

gcsuk:
Is anyone working to fix this? If not, I will look into it

@sebastienros @sfmskywalker

sfmskywalker:
All yours. Thanks!

gcsuk:
@sfmskywalker PR open 😄

PR #7407, opened 2016-11-15
RemovePublishLaterTaskIfNewerVersionIsPublished

hacko-bede:
Added a functionality that will check when a new version is published, if the previews version has a publish later option. If so it will remove it, in order not to override already newly published content.

Link to issue:
https://github.com/OrchardCMS/Orchard/issues/6840


gcsuk:
:+1:

Issue #7394, opened 2016-11-11
Issues in implementation of ILocalizationService


MatteoPiovanelli-Laser:
related to #7386 and #7352.
As discussed with @sebastienros, I am opening an issue about the methods from ILocalizationService.

Method: `GetLocalizedContentItem(IContent content, string culture, VersionOptions versionOptions)`
(`versionOptions` is optional because there is a version of the method without that parameter)
**Expected behaviour:** Call a "LocalizationSet" the set of all contents that are localizations one of the other. I would expect this method to return the item from `content`'s LocalizationSet whose Culture property matches the one given in the string `culture`.
**Actual behaviour:** That only happens if `content` is the MasterContentItem for its set. If that's not the case the method returns:
 - `null` if `content`'s Culture does not match `culture`.
 - `content` otherwise.
This is all down to the query used in this method:
```
return _contentManager
    .Query<LocalizationPart>(versionOptions, content.ContentItem.ContentType)
    .Where<LocalizationPartRecord>(l =>
        (l.Id == content.ContentItem.Id || l.MasterContentItemId == content.ContentItem.Id) // <-- I THINK THE ISSUE IS HERE
        && l.CultureId == cultureRecord.Id)
    .Slice(1)
    .FirstOrDefault();
```
As things are now (with #7386), given `var localized = content.As<LocalizationPart>();`, I would change the line I marked above to something like:
```
(l.Id == content.ContentItem.Id || l.MasterContentItemId == content.ContentItem.Id || (localized.MasterContentItem != null && l.MasterContentItemId == localized.MasterCotnentItem.Id))
```
I think that change would allow to at least look into the set of items that share `content`'s MasterContentItem.

Method: `GetContentCulture(IContent content)`
**Actual behaviour:** This method returns `content`'s culture if `content` has a `LocalizationPart`. It returns the Site's culture otherwise (which is debatable).

Method: `SetContentCulture(IContent content, string culture)`
**Actual behaviour:** This method is fine. If `content` has a `LocalizationPart`, it will wet its culture.

Method: `GetLocalizations(IContent content, VersionOptions versionOptions)`
(`versionOptions` is optional because there is a version of the method without that parameter)
**Expected behaviour:** I would expect this method to return what we called "LocalizationSet" above.
**Actual behaviour:** This mostly works as expected. However, because of #7386, LocalizationSets are fragmented whenever some localizations are not created directly starting from a common MasterContentItem.

Issue #7386, opened 2016-11-08
Localizations do not have same master


MatteoPiovanelli-Laser:
I noticed this behaviour, that in light of the recent discussions on localization #7352 we may want to change. I am writing it here in case we actually want to change this behaviour.
This is the scenario:
 - In a tenant enable 3+ cultures. In my example I have en-US, it-IT, fr-FR, de-DE.
 - Pick an en-US content item (I use the "Welcome to Orchard!" page) that I'll call enItem. Call its ID enID.
 - From enItem, create a new translation to the it-IT culture. Call the content item itItem, and its ID itID. In the localization records, you will see that the MasterContentItemId for itItem is enID.
 - From itItem, create a new translation to the fr-FR culture. Call the content item frItem, and its ID frID. In the localization records, you will see that the MasterContentItemID for frItem is itID.

The behaviour I would have expected was to have the MasterContentItemID for frItem to be enID, since that is the master for itItem.
As things stand, I could then create an en-US localization of frItem, because the localization part ignores the fact that itItem is already the translation of enItem.
On top of that, If I were to delete itItem I would lose all relations between enItem and frItem (I realize that the record would still be in the LocaliationPartRecord table).

I would suggest that the MasterContentItem is the same for all ContentItems in a translation set.

sebastienros:
This totally looks wrong. I think it would be logical to always use the same master id, but maybe it's by design,.
Let's ask @Jetski5822 what he thinks.

Issue #7348, opened 2016-10-24
BUG: content item localization sometimes causes culture duplicates


MatteoPiovanelli-Laser:
I have not been able to replicate this behaviour wih 100% repeatability, so I am not too sure of the actual steps.

After enabling the Localization feature, I tried creating translations of a content item.
In the "new" content, I get to choose the culture, and of course the one I am translating from is not available. SO I select a culture and hit "Publish Now".
Sometimes, and again, I am not sure under what conditions, the translation is marked as having the same culture as the original (see picture).

![image](https://cloud.githubusercontent.com/assets/19283945/19652614/bcd0200a-9a11-11e6-9bb1-7962efc1b925.png)

Again, I have not been able to replicate this behaviour reliably, so I could not isolate its cause yet.

sebastienros:
We'll need either a repro or someone that can corroborate.

sebastienros:
We'll close it in two weeks if there is no update.

MatteoPiovanelli-Laser:
I was checking things again about this. I found the reason why this happens.

There are two places where I am clicking the "New Translation" link:

1. In the SummaryAdmin view (Views/Parts/Localization.ContentTranslations.SummaryAdmin.cshtml)
2. In the Editor view (Views/EditorTemplates/Parts/Localization.ContentTranslations.Edit.cshtml)

In the former, the `id` query string parameter to the `Translate` action is set to the Id of the ContentItem whose SummaryAdmin this is.
In the latter, that same parameter is computed as
```
var contentItemId = Model.MasterContentItem != null ? Model.MasterContentItem.Id : Model.ContentItem.Id;
```

This means we have now repro steps to have the "wrong" behaviour from this issue:

- Enable the Localization module.
- Start from an instance with at least 3 enabled cultures. For example, consider `en-US`, `fr-FR`, `it-IT`.
- Have a ContentType that has a LocalizationPage. For example, the ContentType `Page`.
- Create a `Page` and assign `en-US` to its culture. Call this `Page` EnPage.
- Create a new Translation of EnPage and assign `fr-FR` to its culture. Call this `Page` FrPage.
- From the list of content, start a new Translation clicking on the "New Translation" link from FrPage's SummaryAdmin view. Assign 'it-IT' to its culture. Call this `Page` ItPage.
[...]

PR #7416, opened 2016-11-18
Implemented LocalizationSetId and relate logic.

MatteoPiovanelli-Laser:
Functionally, this replaces the MasterContentItem. However, the old service has not been touched to allow retrocompatibility.

This does not solve #7394 because the "old" LocalizationService was not changed. Rather, a new service was created that lives beside it, so that anyone using the original buggy ones and making them work for their needs will see no change.

This is a solution to #7386, because the new service maintains the relationship inside the LocalizationSet.

This may be a to #7348: in testing this PR on my machine I have never seen the issue I saw there in #7348.

This time I have also been a good boy and squashed all my commits.

The migration may take a while on systems with a lot of items. I have been thinking about ways to make it faster, but I'd first like to see if it's needed, because they are not very easy to implement and test.


MatteoPiovanelli-Laser:
@sebastienros Is it okay if I go ahead and close this? With the changes we have done to the Localization module this has become obsolete.

Issue #7419, opened 2016-11-21
OpenId Module


ThaerAlAjlouni:
For your consediration, the **Orchard.OpenId** is a new module that enables your site to authenticate users using the **_OpenId connect protocol_**. Out of the box, this module provides five Open Id providers:

- Facebook
- Google
- Twitter
- Azure Active Directory 
- Active Directory Federation Services

Each of these providers is implemented as a separate **feature** so it can be _enabled/disabled_ individually.   
Each provider's **settings** can be easily managed via the admin's dashboard under the **Open Id** menu item under the Site settings menu. 

> The login of usual orchard users will not be affected, but the **LogOn** page shape will be overridden to show the enabled OpenId providers.

The module was built in a way to easily add custom Open Id providers by simply implementing **IOpenIdProvider**, and then it will automatically be discovered and registered at runtime.

Piedone:
Wow, very nice! Why not demo on one of Tuesday's meetings?

Piedone:
It was demoed but I wasn't there? :-)

sebastienros:
Two weeks ago, awesome!

Piedone:
I'm glad my suggestion was so good that @ThaerAlAjlouni has taken it two weeks in advance!

armanforghani:
Wow, great work.

ThaerAlAjlouni:
Merged to the dev branch #7420

armanforghani:
When will this be available in 1.10.x or master branch?

PR #7420, opened 2016-11-21
Feature/openid

ThaerAlAjlouni:
This PR adds OpenId to Orchard. See #7419


src/Orchard.Web/Modules/Orchard.OpenId/Views/EditorTemplates/Parts.AzureActiveDirectorySettings.cshtml
```diff
@@ -0,0 +1,63 @@
...
+             Example: service1=https://yoursite.onmicrosoft.com/some-guid-for-service1")
+    </span>
+</fieldset>
+<fieldset>
+    @Html.LabelFor(m => m.AppKey, T("App Key"))
+    @Html.TextBoxFor(m => m.AppKey, new { @class = "text large" })
+</fieldset>
+<fieldset>
+    @Html.CheckBoxFor(m => m.BearerAuthEnabled)
+    @Html.LabelFor(m => m.BearerAuthEnabled, T("Enable Bearer Token Authentication").ToString(), new { @class = "forcheckbox" })
```
sebastienros: Don't use `LabelFor` and `T.ToString` at the same time or there will be a double encoding. Use a `<label>` tag instead. Everywhere in this PR.

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        public void Load() {
+            _sessionLock.EnterReadLock();
+            try
+            {
+                if (InMemoryTokenCache.ContainsKey(_cacheId))
+                    Deserialize(InMemoryTokenCache.Where(x => x.Key == _cacheId).First().Value);
+            }
+            catch (Exception) { /* TODO: Log the error */ }
```
sebastienros: Please implement the TODOs

src/Orchard.Web/Modules/Orchard.OpenId/Views/EditorTemplates/Parts.AzureActiveDirectorySettings.cshtml
ThaerAlAjlouni: Done

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
ThaerAlAjlouni: Done

src/Orchard.Web/Modules/Orchard.OpenId/Constants.cs
```diff
@@ -0,0 +1,36 @@
...
+        public const string AuthenticationErrorUrl = "~/Authentication/Error";
+        public const string LogonCallbackUrl = "~/Users/Account/LogonCallback";
+        public const string OpenIdOwinMiddlewarePriority = "10";
+        public const string LocalIssuer = "LOCAL AUTHORITY";
+        public const string FormsIssuer = "Forms";
+
+        public const string AzureActiveDirectoryObjectIdentifierKey = "http://schemas.microsoft.com/identity/claims/objectidentifier";
+
+        // Certificate Subject Key Identifier
+        public const string VeriSignClass3SecureServerCA_G2 = "A5EF0B11CEC04103A34A659048B21CE0572D7D47";
```
sebastienros: What the recommendation with updating these strings? When, how?

sebastienros: I see it seems to be only for Twitter. Could we make these site settings so it can be updated without having to redeploy, or get a new Orchard deployment? It could be an advanced site setting along with the current twitter settings, hidden in a collapsible section as it's just noise. And an admin banner that would be triggered from a date that is also in these settings, based on the first certificate to expire because at some point it will stop working and people will never find out why.

sebastienros: Remember to cache the values once taken from the settings.

sebastienros: Is there the same issue for Azure AD, I can remember they regularly recycle some keys or thumbprints.

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public void Load() {
+            _sessionLock.EnterReadLock();
+            try
+            {
+                if (InMemoryTokenCache.ContainsKey(_cacheId))
+                    Deserialize(InMemoryTokenCache.Where(x => x.Key == _cacheId).First().Value);
+            }
+            catch (Exception) {
+                _sessionLock.ExitReadLock();
```
DanielStolt: Why not move this to finally block?

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
```diff
@@ -0,0 +1,82 @@
...
+                HasStateChanged = false;
+
+                if (InMemoryTokenCache.ContainsKey(_cacheId))
+                    InMemoryTokenCache.Remove(_cacheId);
+
+                InMemoryTokenCache.Add(_cacheId, Serialize());
+            }
+            catch (Exception)
+            {
+                _sessionLock.ExitReadLock();
```
DanielStolt: Should probably be `ExitWriteLock()`. (And why not move to finally block?) ;)

src/Orchard.Web/Modules/Orchard.OpenId/Module.txt
```diff
@@ -0,0 +1,38 @@
...
+Description: Enables Orchard to authenticate users using OpenId
+Category: OpenId Providers
+Features:
+    Orchard.OpenId:
+        Description: Enables Orchard to authenticate users using OpenId
+		Category: Authentication
+		Name: OpenId
+    Orchard.OpenId.Facebook:
+        Description: Enables Orchard to authenticate users using their Facebook Accounts
+		Category: OpenId Providers
```
DanielStolt: I think the category of the providers should be "Authentication" just like the root feature.

src/Orchard.Web/Modules/Orchard.OpenId/Constants.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace Orchard.OpenId {
+    public class Constants {
```
DanielStolt: I think the structure would be nicer if these constants were moved into feature-specific constant files. Since it's a possible extensibility point to write more providers.

DanielStolt:
Truly great contribution! I wrote some code-level comments on your PR.

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
sebastienros: Why not a concurrent dictionary, I think it has the primitives you need (`TryGetOrAdd`)

src/Orchard.Web/Modules/Orchard.OpenId/Constants.cs
ThaerAlAjlouni: I couldn't find any formal facts on when those values will expire, but to be in the safe side they are now configurable via admin dashboard as you suggested

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
ThaerAlAjlouni: No need for locks anymore, the code has been updated to use a Concurrent Dictionary.

src/Orchard.Web/Modules/Orchard.OpenId/Module.txt
ThaerAlAjlouni: Done :)

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
```diff
@@ -22,51 +21,30 @@ public class InMemoryCache : TokenCache {
             Load();
         }
 
-        public static Dictionary<string, byte[]> InMemoryTokenCache {
+        public static ConcurrentDictionary<string, byte[]> InMemoryTokenCache {
```
sebastienros: Does it have to be public? If not then remove it and put the initialization in the ctor

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
```diff
@@ -1,13 +1,12 @@
...
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using Microsoft.IdentityModel.Clients.ActiveDirectory;
 
 namespace Orchard.OpenId {
     public class InMemoryCache : TokenCache {
-        private static Dictionary<string, byte[]> _inMemoryTokenCache;
-        private static ReaderWriterLockSlim _sessionLock = new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
+        private static ConcurrentDictionary<string, byte[]> _inMemoryTokenCache;
```
sebastienros: Does it have to be static? Or should it be different per tenant?

ThaerAlAjlouni: The keys of the dictionary are unique user ids (the id of the user in the open id provider system, not Orchard's id), so even across multiple tenants the dictionary will be safe.

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
ThaerAlAjlouni: The InMemoryCache class has been updated according to your comment :)

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
sebastienros: Can you be more specific about the keys? Are they the actual user id or Guids?

ThaerAlAjlouni: Guids generated and assigned to the user by the identity provider. This id will be sent back by the provider after authenticating the user inside the JWT.

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
+
+namespace Orchard.OpenId {
+    public class InMemoryCache : TokenCache {
```
sebastienros: I looked at the documentation for the token cache. And I think it would be much better if you actually had it registered as a singleton service such that each tenant would have its own list of tokens (not the case right now), and that this list would then be cleared when a tenant is restarted (not the case right now). Also, you would then not need to instantiate a new `InMemoryCache` instance for each request. Also, it needs to be a service such that when we use multiple nodes, we can use a token cache based on Redis for instance. And this service should be tied to the Azure feature.

src/Orchard.Web/Modules/Orchard.OpenId/Constants.cs
ThaerAlAjlouni: Done :)

src/Orchard.Web/Modules/Orchard.OpenId/InMemoryCache.cs
ThaerAlAjlouni: @sebastienros, I am having trouble understanding exactly how to convert this class to a Singleton service, and I couldn't find the documentation of this class, can you please provide me a link.

sebastienros: Just have this class implement `ISingletonDependency` and then resolve it instead of instantiating it.

ThaerAlAjlouni: Done

src/Orchard.Web/Modules/Orchard.OpenId/Services/AzureActiveDirectory/InMemoryCache.cs
```diff
@@ -1,39 +1,51 @@
...
-            _cacheId = String.Concat(_userObjectId, CacheIdSuffix);
-
-            if (_inMemoryTokenCache == null)
-                _inMemoryTokenCache = new ConcurrentDictionary<string, byte[]>();
+namespace Orchard.OpenId.Services.AzureActiveDirectory {
+    [OrchardFeature("Orchard.OpenId.AzureActiveDirectory")]
+    public class InMemoryCache : TokenCache, ISingletonDependency {
+        public InMemoryCache() {
+            // On tenant start clear cache.
+            Clear();
```
sebastienros: Not really necessary as the instance is new on tenant restart. Or there is an issue with the design.

src/Orchard.Web/Orchard.Web.csproj
```diff
@@ -14,7 +14,7 @@
     <AssemblyName>Orchard.Web</AssemblyName>
```
sebastienros: Revert the changes in this file

sebastienros:
Just a few little changes more.

src/Orchard.Web/Orchard.Web.csproj
ThaerAlAjlouni: Done

src/Orchard.Web/Modules/Orchard.OpenId/Services/AzureActiveDirectory/InMemoryCache.cs
ThaerAlAjlouni: This was an extra step, it is redundant so I will remove it

ThaerAlAjlouni:
Is there anything else to be amended?

sebastienros:
I think it's good, should be merged on thursday

Issue #7432, opened 2016-11-22
Problem in Queries with filter on DateTime field


GiuseppeMusso-Laser:
Hi all.
I'm experiencing a problema with queries when I have a filter on DateTime fields.
I'm working on the back-end site and I created a new Query.
I added a filter on field "Creation date" of "Common Part Record" and my condition is "less than 2016-11-20", for instance. 

Problem: when I execute this query, I face a problem with date-time format in SQL query.
This seems to be caused by a format string used in Orchard.ContentManagement.HqlRestrictions.FormatValue(object, bool) (in file IHqlCriterion.cs).
This method convert a DateTime value into a String using format "yyyy-MM-dd HH:mm:ss".
I found out changing this format string to "yyyy-MM-ddTHH:mm:ss" (ISO 8601) solves this issue.
I'm using SQL Server Express.
I'm wondering if this solution works well for other databases too and if there is a better way to fix this problem.

ViRuSTriNiTy:
Related issue #6973

PR #7433, opened 2016-11-22
Fix for Query Filter on DateTime fields

GiuseppeMusso-Laser:
DateTime value is now parsed to String in ISO 8601 format.
This PR should solve issue #7432.


ViRuSTriNiTy:
This issue is already fixed, see pull #7059

sebastienros:
This one targets 1.10.x though

Issue #860, opened 2010-12-10
Using PublishLater and then editing the item again and picking 'Save' does not cancel the schedule


orchardbot:
@infinitiesloop created:
https://orchard.codeplex.com/workitem/17028

The item is still scheduled after 'save'. The updated content is published at the original publish date. To edit a scheduled item and keep it scheduled you should have to save it with the publish later button, again.

sebastienros:
Suggestion is to create a content handler that will detect the scheduled date has been removed, then delete the scheduled task. It should be on Versionning as it has to work with Save and Publish.

sebastienros:
The service will probably need to be changed.

sebastienros:
https://github.com/OrchardCMS/Orchard/issues/7056

Issue #7056, opened 2016-07-25
Rename "Publish Later" in new blog post page to "Publish Later At"


VahidNaderi:
I suggest replacing the text and order of elements like this
![orchardissue](https://cloud.githubusercontent.com/assets/3517756/17103031/7ddb8c94-5291-11e6-99ca-13f7a6a99f82.jpg)

It makes more sense and prevents user's first failed try frustration.

sfmskywalker:
I agree, much clearer.

VahidNaderi:
@sfmskywalker do you want me to make a pull request for it?

sfmskywalker:
I do, but we should wait until we triaged this suggestion with the rest. We do triage meetings every Thursday, hopefully we will get to this one. Thanks

sebastienros:
I have another suggestion, to replace the Publish Later button by a checkbox, which would show the DateTime editor. And then use the standard Publish button.

Skrypt:
The "publish later" button should open a modal in which we could then pick a date. That modal would have a cancel and "publish later" button ... That way we prevent users from not picking up a date before pushing the "publish later" button.

VahidNaderi:
IMO a modal is a little too much for this task, and what Wordpress folks did is nice but does not match current Orchard's overall design.

Maybe a check box as suggested by @sebastienros would be a good option too, something like this:
![orchardissue2](https://cloud.githubusercontent.com/assets/3517756/17239198/8b73d62a-5577-11e6-8024-7f7ac27c1fe9.jpg)

sfmskywalker:
I agree. I like that the Publish Later checkbox and date time controls are now vertically stacked, looks cleaner, and that there is no longer two separate "Publish" buttons, making it less confusing.

Skrypt:
let's "at least" for the new admin theme use a better looking datetime picker like this one : https://tarruda.github.io/bootstrap-datetimepicker/

sebastienros:
@Skrypt priorities ...

Skrypt:
I agree on that. Though it's not bad to keep those links somewhere. We never know !

sebastienros:
Good point from @HermesSbicego-Laser, it's ambiguous what Save and Publish will do when the checkbox is checked. Will Publish publish now AND later, or just later? Then if just later, what is the difference with Save?

PR #7438, opened 2016-11-24
Remove Publish Later task

GiuseppeMusso-Laser:
Publish Later module: added button "Remove Publish Later" to cancel publish later task.
This PR should partially solve issue #860.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.PublishLater/Drivers/PublishLaterPartDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.PublishLater/Drivers/PublishLaterPartDriver.cs
@@ -11,6 +11,7 @@ using Orchard.PublishLater.Models;
 using Orchard.PublishLater.Services;
 using Orchard.PublishLater.ViewModels;
 using Orchard.Services;
+using Orchard.Tasks.Scheduling;
 
 namespace Orchard.PublishLater.Drivers {
     public class PublishLaterPartDriver : ContentPartDriver<PublishLaterPart> {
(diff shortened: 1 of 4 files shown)


sebastienros:
There is a breaking change here.

Also, can anyone find the issue where we talked about changing the UI for this feature?

HermesSbicego-Laser:
@sebastienros, is it this one #7056? if yes, someone is working on that? Our report was intended to have a way to cancel the publish later Task (there is no way to do that by now), so it's a functional proposition not only a UI refactoring. 

Also, i don't really like what is proposed there, because it's difficult to understand what the `Publish` button should do when the "Publish later" is checked: a `Publish later` or a `Publish now`? And also in order to cancel the `Publish later Task` what the user should do? It should to remove the flag and press `Save`? 

Thanks

src/Orchard.Web/Modules/Orchard.PublishLater/Services/IPublishLaterService.cs
```diff
@@ -6,5 +6,6 @@ namespace Orchard.PublishLater.Services {
     public interface IPublishLaterService : IDependency {
         DateTime? GetScheduledPublishUtc(PublishLaterPart publishLaterPart);
         void Publish(ContentItem contentItem, DateTime scheduledPublishUtc);
+        void RemovePublishLaterTasks(ContentItem contentItem);
```
sebastienros: Remove it.

src/Orchard.Web/Modules/Orchard.PublishLater/Services/PublishLaterService.cs
```diff
@@ -31,5 +31,9 @@ public class PublishLaterService : IPublishLaterService {
             IScheduledTask task = _publishingTaskManager.GetPublishTask(publishLaterPart.ContentItem);
             return (task == null ? null : task.ScheduledUtc);
         }
+
+        public void RemovePublishLaterTasks(ContentItem contentItem) {
```
sebastienros: Delete this

src/Orchard.Web/Modules/Orchard.PublishLater/Drivers/PublishLaterPartDriver.cs
```diff
@@ -105,6 +105,9 @@ public class PublishLaterPartDriver : ContentPartDriver<PublishLaterPart> {
                 }
             }
 
+            if (httpContext.Request.Form["submit.Save"] == "submit.RemovePublishLaterTasks") {
+                _publishLaterService.RemovePublishLaterTasks(model.ContentItem);
```
sebastienros: Use the `IPublishTaskManager` service directly

sebastienros:
Check if the fields are empty when versioning, if so, remove any task. We don't add a button right now, we'll find a better UI later.

Issue #7441, opened 2016-11-25
Orchard export steps no longer triggered by file export /Steps flag


bedegaming-aleksej:
During the recent refactor of export and import steps, `IEventHandler` interface has been removed from export steps. However, console export command still calls `CustomStepsStep`, which relies on other steps implementing that interface to listen to messages published on the event bus. This means that, for example, it is currently not possible for user to execute any custom steps through command line except those that still implement the obsolete interfaces (which there are none left in the Orchard solution). The flexibility provided with flags supported by the export file command is currently very limited and, among other things, does not allow exporting user roles, workflows, aliases, etc.

**Acceptance criteria**

- All export steps provided by Orchard are supported by the `/Steps` flag of the `file export` command.

PR #7442, opened 2016-11-25
Adapt export steps flag to work with refactored steps

bedegaming-aleksej:
This PR resolves issue raised in [#7441](https://github.com/OrchardCMS/Orchard/issues/7441)


src/Orchard/Recipes/Services/IRecipeBuilderStepResolver.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
```
bedegaming-aleksej: Added RecipeBuilderStepResolver and RecipeExecutionStepResolver for convenience.

Issue #7443, opened 2016-11-25
Visibility of Jobs Queue menu item


GiuseppeMusso-Laser:
Hi all.
I need to grant access to the back-end to some users (contributors) but I don't want to show them "Jobs Queue" menu item.
I can add SiteOwner permission to this menu item to solve this issue as they did in Workflows menu item.
Is this a good solution?

PR #7444, opened 2016-11-25
Jobs Queue menu item visibility

GiuseppeMusso-Laser:
Jobs Queue menu item is now visible to site owners only.


GiuseppeMusso-Laser:
This PR solves issue #7443.

Issue #7430, opened 2016-11-22
Autoroute is regenerated wrongly after editing a localized content


MarcoViglione-Laser:
I added a LocalizationPart to a content type for which I modified the default autoroute pattern. I didn't check the option to use different patterns for different cultures because I wanted to use the same pattern for all cultures. When creating the content item the pattern was generated as I expected. Then I edited the content, I deleted the autoroute to force its regeneration, I saved and the result was different: the modified pattern was applied only to the contents with a culture corresponding to the site culture, but for the other cultures Orchard applied the {Content.Slug} pattern.

**Steps to reproduce the issue:**

1. Create a content type with an Autoroute Part and a Localization Part.
2. In the Autoroute Part settings check "Allow custom patterns", uncheck everything else and edit the default pattern into something different.
3. Create a content in a language different from the site language. The autoroute should be generated with the modified pattern.
4. Delete the autoroute from the content and save. The autoroute will be generated with the {Content.Slug} pattern.

As an alternative to steps 1 and 2 you can attach a Localization Part to a content type already defined like that, i.e. a Taxonomy Term or a Blog Post.

Looking at the `AutorouteService`, in the `GetDefaultPattern` method, this is intentional. Why such an approach was chosen? I think that a culture neutral pattern should be applied to all cultures, not only to the site culture.

Skrypt:
Will look at it. Need to put back my head into it first.

Skrypt:
So the problem here I think is that I hardcoded the {Content.Slug} pattern and we just need to make this using the default dynamic slug.

MarcoViglione-Laser:
I did some tests and I found some other problems regarding the autoroute localization support:

1. The concept of culture neutral pattern is not present. Even if you don't check the "Use culture pattern(s)" option, the only editable pattern is saved with the current culture of the site.
2. When saving a content definition that contains an AutoroutePart it automatically generates a pattern (in the default {Content.Slug} format) for each language defined in the website settings that has no pattern associated. Those patterns are used by the GetDefaultPattern method even if the "Use culture pattern(s)" option is unchecked.
3. In the same scenario, if I modify the only editable pattern, only the pattern associated to the site culture is updated, while the other cultures still keep the {Content.Slug} pattern.
4. Some modules (i.e. Taxonomy) currently create in the Migration some patterns with null culture, but when the content definition is saved again they are migrated to the site culture.

[...]

PR #7449, opened 2016-11-29
Added support for culture neutral patterns

MarcoViglione-Laser:
Fixed support for localized and culture neutral patterns (issue #7430)

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
@@ -1,10 +1,23 @@
-﻿using Orchard.ContentManagement.MetaData;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Orchard.Autoroute.Models;
+using Orchard.Autoroute.Settings;
+using Orchard.ContentManagement;
+using Orchard.ContentManagement.MetaData;
 using Orchard.Core.Contents.Extensions;
 using Orchard.Data.Migration;
+using Orchard.Localization.Services;
 
-namespace Orchard.Autoroute
-{
-	public class Migrations : DataMigrationImpl {
+namespace Orchard.Autoroute {
+    public class Migrations : DataMigrationImpl {
+        private readonly IContentManager _contentManager;
+        private readonly ICultureManager _cultureManager;
+
+        public Migrations(IContentManager contentManager, ICultureManager cultureManager) {
+            _contentManager = contentManager;
+            _cultureManager = cultureManager;
+        }
 
         public int Create() {
             SchemaBuilder.CreateTable("AutoroutePartRecord",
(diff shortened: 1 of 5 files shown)


src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs
```diff
@@ -135,32 +135,44 @@ public class AutorouteService : Component, IAutorouteService {
...
+                    }
+                    else {
+                        return new RoutePattern { Name = "Title", Description = "my-title", Pattern = "{Content.Slug}", Culture = culture };
+                    }
                 }
-            }
 
-            // Return a default pattern if set.
-            var patternCultureSearch = settings.Patterns.Any(x => String.Equals(x.Culture, culture, StringComparison.OrdinalIgnoreCase)) ? culture : null;
+                // If patterns for the specified culture exist I search one of them, else I search a culture neutral pattern.
```
sebastienros: Make your comments impersonal: no "I", "He", "She" ...

src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
```diff
@@ -49,5 +62,46 @@ public class Migrations : DataMigrationImpl {
 
             return 4;
         }
+
+        public int UpdateFrom4() {
+            // Adding some culture neutral patterns if they don't exist
+            var types = _contentManager.GetContentTypeDefinitions().Where(t => t.Parts.Any(p => p.PartDefinition.Name.Equals(typeof(AutoroutePart).Name)));
```
sebastienros: You can do the test in the loop, as you are also looking for the part later.

src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
```diff
@@ -49,5 +62,46 @@ public class Migrations : DataMigrationImpl {
...
+            foreach (var type in types) {
+                var typeDefinition = ContentDefinitionManager.GetTypeDefinition((type.Name));
+                if (typeDefinition != null) {
+                    var settings = typeDefinition.Parts.First(x => x.PartDefinition.Name == "AutoroutePart").Settings.GetModel<AutorouteSettings>();
+                    if (!settings.Patterns.Any(x => String.IsNullOrWhiteSpace(x.Culture))) {
+                        string siteCulture = _cultureManager.GetSiteCulture();
+                        string newPatterns = "";
```
sebastienros: Make `newPatterns` a `List<string>` then at the end you can just join all of them using `String.Join(", ")`. This way you don't have to deal with concatenating `,` and then having to remove the last one.

src/Orchard.Web/Modules/Orchard.Autoroute/Migrations.cs
HermesSbicego-Laser: It seems to us that filtering types having AutoroutePart before looping is more efficient, or are we misunderstanding your comment?

sebastienros: var typeDefinitions = ContentDefinitionManager.ListTypeDefinition( ... where has autoroute part);

sebastienros: I am sure you can even get the ContentTypePartDefinition with AutoroutePart for all type

Skrypt:
looks like good work will test it

Skrypt:
I don't know why but I can't see the culture neutral tab. You guys took another direction?

MarcoViglione-Laser:
The culture neutral tab is called "All cultures", you can find it in the content definition.

cultureneutraltab

Skrypt:
Sorry I presumed it was on the dev branch but it's not been merged yet from 1.10.x branch. Tested it on 1.10.x branch and I see it now.

PR #7454, opened 2016-12-01
Issue/7404 users own media folder i media username provider

LorenzoFrediani-Laser:
Implemented what described on issue 7404
Manage Own Media permission, created Users/UserName folder that sounds as having some problems. Now creates Users/UserNameEncoded folder in order to prevent:

Media Folder name broken with special chars

Added interface for change the default behavior

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Providers/IMediaFolderProvider.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Providers/IMediaFolderProvider.cs
@@ -1,7 +1,7 @@
-﻿using Orchard.ContentManagement;
+﻿using Orchard.Security;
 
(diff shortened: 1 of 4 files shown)


src/Orchard.Web/Modules/Orchard.MediaLibrary/Providers/IMediaFolderProvider.cs
```diff
@@ -0,0 +1,7 @@
+﻿using Orchard.Security;
+
+namespace Orchard.MediaLibrary.Providers {
+    public interface IMediaFolderProvider : IDependency {
```
sebastienros: IMediaUsernameProvider

sebastienros: IUserMediaFolderProvider

PR #7456, opened 2016-12-02
[Fixes #7413] problem with permissions for taxonomies and terms

### LINKED ISSUE #7413: Problem with permissions for taxonomies
I want to create a user able to manage only taxonomies and no other content, but if I disable content permissions and enable taxonomies permissions he's not able to manage them.

This is how I was able to find the issue:
1. Create a new user
2. Assign a single role to a new user (i.e. Editor)
3. In the settings for that role, uncheck everything under "Content Feature" and check Manage Taxonomies under "Orchard.Taxonomies Feature"
4. Log as the new user
5. Open the Edit page of a Taxonomy or Term

**Result**
cannotedit

Maybe this behavior happens because taxonomies and terms use the standard content controller instead of a dedicated controller like Blogs?

HermesSbicego-Laser:
Fixes #7413

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Views/Admin/Index.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Views/Admin/Index.cshtml
@@ -39,12 +39,12 @@
                 <tr class="@(taxonomyEntry.IsInternal ? "internal" : null)">
                     <td>
                         <input type="hidden" value="@Model.Taxonomies[taxonomyIndex].Id" name="@Html.NameOf(m => m.Taxonomies[ti].Id)" />
-                        @if (!taxonomyEntry.IsInternal || Authorizer.Authorize(Orchard.Security.StandardPermissions.SiteOwner) || Authorizer.Authorize(Permissions.ManageTaxonomies)) {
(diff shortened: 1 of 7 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Views/Admin/Index.cshtml
```diff
@@ -36,12 +39,12 @@
                 <tr class="@(taxonomyEntry.IsInternal ? "internal" : null)">
                     <td>
                         <input type="hidden" value="@Model.Taxonomies[taxonomyIndex].Id" name="@Html.NameOf(m => m.Taxonomies[ti].Id)" />
-                        @if (!taxonomyEntry.IsInternal || Authorizer.Authorize(Orchard.Security.StandardPermissions.SiteOwner)) {
+                        @if (!taxonomyEntry.IsInternal || Authorizer.Authorize(Orchard.Security.StandardPermissions.SiteOwner) || Authorizer.Authorize(Permissions.ManageTaxonomies)) {
```
sebastienros: SiteOwner check is redundant as it will have ManageTaxonomies

sebastienros: And don't forget all the other occurences of the same test in this class

Issue #6890, opened 2016-05-12
Admin menu not showing "Content" when only a Custom Content Type permission is allowed


natebunton:
I created a custom content type called "Partners", then I created a new role that only has permissions to publish, edit and delete.  When a user logs in with the new role, a link shows up in the Admin menu to create a new "Partner", but the "Content" link in the admin menu does not show up.  However, the user can directly access the Content List in the admin.

http://localhost:30321/OrchardLocal/Admin/Contents/List
![2016-05-12_14-23-00](https://cloud.githubusercontent.com/assets/7373116/15227719/369f172e-184e-11e6-9601-b634cde3eb20.jpg)
![2016-05-12_14-22-12](https://cloud.githubusercontent.com/assets/7373116/15227756/5f7597cc-184e-11e6-8784-71ae6259f575.jpg)

PR #7464, opened 2016-12-07
Admin Menu: fix for "Content" item visibility.

GiuseppeMusso-Laser:
Edit permission added in "Content" menu item for each securable content type.
If user has Edit Permission on a custom content type, "Content" is visible in admin menu.
See issue #6890.


sebastienros:
The solution is to actually check the `Listable` content types, and not the `Securable` ones as this link will display only the `Listable` ones.
Then for each content type, you create a new content item, and check `EditPermission` on each of them, then stop at the first success.

GiuseppeMusso-Laser:
Thank you, Sebastien. I updated my PR according to your suggestions.

PR #7465, opened 2016-12-07
Fix for mp4 files not streaming from Azure blob

kumards:
Fix for #7472.

By default Azure blob storage uses service version "2009-09-19". mp4 files served with this version have to be fully downloaded to play.
More info can be found here http://blog.thoughtstuff.co.uk/2014/01/streaming-mp4-video-files-in-azure-storage-containers-blob-storage/.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/AzureFileSystem.cs
+++ b/src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/AzureFileSystem.cs
@@ -75,7 +75,14 @@ namespace Orchard.Azure.Services.FileSystems {
             // Get and create the container if it does not exist
             // The container is named with DNS naming restrictions (i.e. all lower case)
             _container = _blobClient.GetContainerReference(ContainerName);
-
+ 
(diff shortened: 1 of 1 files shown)


sebastienros:
Can you file an issue to track the fix?
Also, doesn't this code need to run only once? In this case maybe a migration would be more appropriate.
Can it break existing features or sites or containers?

kumards:
Version setting should happen before the container gets created, so that the container gets its default service version, especially when anonymous requests are made to the files inside the container.

I guess its safe enough since its backward compatible. 
Refer https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/versioning-for-the-azure-storage-services

> The current version of the Azure storage services is 2015-12-11, and using that version is recommended where possible. 

> **Requests Via Anonymous Access**
> 
> If a request to the Blob service does not specify the x-ms-version header, and the default version for the service has not been set using Set Blob Service Properties, then the earliest version of the Blob service is used to process the request. However, if the container was made public with a Set Container ACL operation performed using version 2009-09-19 or newer, then the request is processed using version 2009-09-19.

PR #7466, opened 2016-12-07
[Fixes #7425] Session Locator | Isolation Levels not respected in Sites.config

### LINKED ISSUE #7425: Session Locator | Isolation Levels not respected in Sites.config
Isolation level changes for default session locator sessions and migration sessions are not respected in the current version of sites.config. 

The reason Autofac can no longer set these is that these properties moved to datamigration manager and transaction manager respectively in Orchard 1.8.

This change updates the commented configuration to respect where these settings actually live. This may be related to other issues (e.g. 5095).

joshnogales:


sebastienros:
WE'll merge it tomorrow, and also copy this change to 1.10.x

joshnogales:
@sebastienros Part of the reason for the mixup that caused this issue, was that TransactionManager was refactored out of SessionLocator code.  When this was done, the implementation of TransactionManager was left in SessionLocator.cs even though there is TransactionManager.cs file that holds ITransactionManager and some helper.

If there is interest, I have already cleaned TransactionManger back into TransactionManger.cs and can submit a pull request of it.

PR #7473, opened 2016-12-12
Fixes usage of undeclared variables in Link.Edit.cshtml.

jtkech:
As requested by @DanielStolt 

Best.


sebastienros:
Was it the only known merge issue?

jtkech:
Sorry, i don't know.

Just done this one because @DanielStolt had this issue and asked me to fix it.

But i've quickly checked field views, seems that only this one was using undeclared variables.

agriffard:
See #7374 .
I think the pb is in dev, master and 1.10.x

jtkech:
Oh yes and the workaround was to rather declare variables and init them by checking default values.

    var isRequired = settings.Required && String.IsNullOrWhiteSpace(settings.DefaultValue);
    var isTextRequired = settings.LinkTextMode == LinkTextMode.Required && String.IsNullOrWhiteSpace(settings.TextDefaultValue);

Not sure at 100% but because now default values are proposed values when 1st creating items. I think that now we don't need to check default values here.

Before, when a default value was provided, if the user didn't input a value on a required field, it was allowed because the default value was used on server side. But now, default values are used as proposed values on 1st creation, not as default value when the field is posted with an empty value. But just as i remember.

PR #7485, opened 2016-12-22
On Import, the modified url only gets updated in AutoroutePart and not as new Alias (Friendly Url)

### LINKED ISSUE #6316: On Import, the modified url only gets updated in AutoroutePart and not as new Alias (Friendly Url)
I reproduced this bug on Orchard 1.9.3 branch.

Following are the steps to reproduce this issue. On a fresh installation, enable ImportExport:
1. Import the attached file [Import1.txt](https://github.com/OrchardCMS/Orchard/files/109642/Import1.txt). It will create an item of type Page with friendly url "/test-page"
2. Import the attached file [Import2.txt](https://github.com/OrchardCMS/Orchard/files/109641/Import2.txt). Observe that on edit mode, Autoroute Part says "modified/test-page" whereas the friendly url still is "/test-page". i.e. the new Autoroute url won't be registered as the new Alias (friendly url).

The main issue is, when IContentManager.Import method tries to Publish the item at following lines of code:

```
            if (status == null || status.Value == Published) {
                Publish(item);
            }
```

The Publish method checks if the Record is already Published or not. If it is already published it simply ignores and returns without calling the handlers.

```
            if (contentItem.VersionRecord.Published) {
                return;
            }
```

This works very nicely on the first Import process where the record was created anew. It went ahead and called all those handlers. For example Autoroute OnPublished which registers the friendly url:

```
            OnPublished<AutoroutePart>((ctx, part) => PublishAlias(part));
```

If user manually Publishes the item then this error goes away since in that case a new draft is created which gets published and in turn calls all the handlers.

This should work with Import as well. If user changes the friendly url, at the end of Import process it should register the new friendly url automatically.

hannan-azam:
Fixes #6316

@sebastienros i have used a property to get the aliasMaps. In that property im using lock to make sure no other call is updating it at the same moment. If the count is returned as 0 we call IAliasHolderUpdater.Refresh to refresh the map. Please let me know if this approach is not correct and we need to make any adjustments.

The other shape file had an extra column in the Admin/IndexManaged which has now been removed.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Alias/Implementation/Holder/AliasHolder.cs
+++ b/src/Orchard.Web/Modules/Orchard.Alias/Implementation/Holder/AliasHolder.cs
@@ -3,14 +3,27 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Collections.Concurrent;
 using Orchard.Alias.Implementation.Map;
+using Orchard.Alias.Implementation.Updater;
 
 namespace Orchard.Alias.Implementation.Holder {
     public class AliasHolder : IAliasHolder {
-        public AliasHolder() {
+        private readonly Lazy<IAliasHolderUpdater> _aliasHolderUpdater;
+        private readonly ConcurrentDictionary<string, AliasMap> _aliasMaps;
+
+        public AliasHolder(Lazy<IAliasHolderUpdater> aliasHolderUpdater) {
+            _aliasHolderUpdater = aliasHolderUpdater;
             _aliasMaps = new ConcurrentDictionary<string, AliasMap>(StringComparer.OrdinalIgnoreCase);
         }
 
-        private readonly ConcurrentDictionary<string, AliasMap> _aliasMaps;
+        private ConcurrentDictionary<string, AliasMap> aliasMaps {
+            get {
+                lock (_aliasMaps) {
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Alias/Implementation/Holder/AliasHolder.cs
```diff
@@ -3,14 +3,27 @@
...
+        private readonly Lazy<IAliasHolderUpdater> _aliasHolderUpdater;
+        private readonly ConcurrentDictionary<string, AliasMap> _aliasMaps;
+
+        public AliasHolder(Lazy<IAliasHolderUpdater> aliasHolderUpdater) {
+            _aliasHolderUpdater = aliasHolderUpdater;
             _aliasMaps = new ConcurrentDictionary<string, AliasMap>(StringComparer.OrdinalIgnoreCase);
         }
 
-        private readonly ConcurrentDictionary<string, AliasMap> _aliasMaps;
+        private ConcurrentDictionary<string, AliasMap> aliasMaps {
```
sebastienros: Convert to a method `GetOrRefreshAliasMaps()`

hannan-azam:
@sebastienros required changes are done. Please let me know if more changes are required.

Issue #5419, opened 2015-06-24
Add Glimpse support to Orchard out of the box


paynecrl97:
Following on from the Tuesday night demonstration of the Glimpse plugin that I wrote for Orchard, it was suggested that the functionality should be pulled into the out of the box download of Orchard.

Glimpse out of the box can give us information about filter and controller execution, and I have extended this to provide information about layers, widgets, the Content Manager and other Orchard specific features.

As this would be a major new feature, I'd like to suggest we kick off a conversation about the intended feature set, as well as talk about some potential design decisions.

For example:
- Do we build this functionality into the default implementation of services? Or do we decorate existing services?
- Do we create a generic interface for timing and performance monitoring? Or do we tie into Glimpse?
- Do we allow third party module developers to plug into Glimpse by consuming the Glimpse interfaces? Or do we provide an abstraction?
- What would be the required feature set for the first iteration?
- Do we move shape tracing into Glimpse?
- Is the Glimpse license compatible with Orchard's?

DanielStolt:
This would be an awesome addition, so glad you're getting the conversation started.

I think this should definitely be implemented using decorators wherever possible, as we need to be able to enable and disable this feature - it should not be baked into any existing services.

I personally do not believe in creating an abstraction over Glimpse. I think it would be very difficult to design an abstraction that would both allow us to harness all the power of Glimpse **and** be generic enough to substitute Glimpse for something else. Difficult to achieve and for very little benefit IMHO. I would rather just clearly define this feature as a "Glimpse" feature, rather than some generic abstract timing and performance monitoring feature.

As for feature set... what have you got? :)

Sounds interesting to move shape tracing into it - do you think that would be feasible? Don't know much about Glimpse, but during your demo it seemed to be that it was more geared towards timing and performance, than DOM inspection and visualization of the rendered result?

Glimpse is Apache 2.0 licensed it seems - I don't know enough to tell whether it's permissive enough for Orchard's license to mirror it.

sebastienros:
License is fine.
Yes, deprecate Shape Tracing in favor of a Glimpse plugin.
Don't hide Glimpse.

Very useful plugin is the SQL one, as in Mini Profiler.

I would say that the minimum level of feature to ship is to have the times taken by the methods and shapes render.

paynecrl97:
I mainly have timings. There is also a log4net plugin that we have successfully installed.

[...]

PR #7502, opened 2017-01-04
Glimpse

paynecrl97:
Issue number #5419, previous PR #6243.

This PR adds the first set of Glimpse functionality to Orchard.

On top of Glimpse's default MVC and .Net instrumentation, the following Orchard tabs are also available:

- Authorizer
- Cache Service
- Content Manager
- Content Parts
- Layers
- Shapes
- SQL (See issue below)
- Widgets

Each tab exists in its own feature, meaning the instrumentation can be toggled on or off as required.

The following extension points exist:

- Ability for other modules to add data to Glimpse
- Ability for other modules to consume the messages sent to Glimpse (example use cases- build a live stats dashboard, push the data into a db to be aggregated, push data to a 3rd party service eg New Relic etc...)

*There are many more areas of Orchard that could be instrumented, but this is just a start.*

There is an issue with the SQL tab where the number of open connections is being over-reported. Someone with more NHibernate ability than me will have to investigate the cause.

There are a number of concessions, but I feel that the value of Glimpse is too great to not make these concessions:

- All Glimpse config is contained within the web.config as opposed to being configurable in the back office
- Glimpse's DI framework will not currently allow for Orchard services to be injected into it
- Any module that adds a tab to the Glimpse console needs to be referenced by the `Orchard.Web` project

All of these issues could probably be corrected over time.

This is also the first usage of `IDecorator` (other than tests) in the Orchard source.


sebastienros:
Demo on Tuesday?

It doesn't affect anything outside Orchard.Glimpse which is good. Should we also make ShapeTracing obsolete BTW ?

paynecrl97:
Will demo on Tuesday.

As it stands, the current shapetracing feature is more complete than what Glimpse offers. No reason not to port it over to Glimpse at some point however.

Jetski5822:
Hey Chris,

Have you looked much at Glimpse on Dotnet Core?

N

paynecrl97:
@Jetski5822 

I have not (yet).

I spoke with Anthony from the Glimpse project a while back and he walked me through what they were wanting to do with Glimpse v2 which I believe is Dotnet Core (and I see you have 2 commits in this project), but that's about it.

I am planning on porting this over to Orchard v2 at some point once this PR is merged and I get some free time.

paynecrl97:
@sebastienros Release transform to disable Glimpse added, and merge conflicts resolved

Issue #7410, opened 2016-11-16
Authorize attribute does not honor LastLogoutUtc security validation


kumards:
When a user logs from two different browsers, the first logout action invalidates all other ongoing sessions based on `IMembershipValidationService.CanAuthenticateWithCookie`. I assume it is a security measure.

The problem I see here is, `IOrchardServices.WorkContext.CurrentUser` returns null as intended but the `Authorize` attribute decorated over controllers and actions does not validate `IMembershipValidationService.CanAuthenticateWithCookie`. So the web requests go past the MVC/WebAPI authorization but fails at `IAuthorizer.Authorize`.

sebastienros:
I assume it fails _correctly_, right? It just doesn't fail soon enough?

kumards:
Yeah, it has to fail soon. 
I guess the request should be challenged in the authorization filter so that the request does not reach the controller or its action.

sebastienros:
@MatteoPiovanelli-Laser did you recent PR fix this issue?

MatteoPiovanelli:
I don't think so, no

PR #7504, opened 2017-01-06
Invalidating a cookie based session when user has logged out from another browser.

kumards:
Fix for #7410.


src/Orchard.Web/Modules/Orchard.Users/Services/SecurityFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+            Logger = NullLogger.Instance;
+        }
+
+        public ILogger Logger { get; set; }
+
+        public void OnAuthorization(AuthorizationContext filterContext) {
+
+            var accessToAuthorizedUserOnly = filterContext.ActionDescriptor.GetCustomAttributes(typeof(AuthorizeAttribute), true).Any() || filterContext.ActionDescriptor.ControllerDescriptor.ControllerType.GetCustomAttributes(typeof(AuthorizeAttribute), true).Any();
+
+            ////When user has logged out from a different browser, we have to invalidate all other browser sessions too.
```
sebastienros: Fix the comment style

sebastienros:
Please fix the merge conflict

BenedekFarkas:
@sebastienros the issue is marked for 1.10.x, but the PR targets dev, which one should we merge this into?

BenedekFarkas:
@sebastienros I resolved the merge conflicts and tested the functionality. The CanAuthenticateWithCookie method will actually tell you if the User has logged out, because it's comparing LastLogoutUtc to LastLoginUtc, so it's actually working as expected and correctly logs the user out of the current browser if they logged out in another already (login is still working correctly). It looks g2g for me!

sebastienros:
Where and when is the `LastLogoutUtc` stored? Does this filter imply a sql query for each http request?

src/Orchard.Web/Modules/Orchard.Users/Services/SecurityFilter.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public void OnAuthorization(AuthorizationContext filterContext) {
+            var accessToAuthorizedUserOnly =
+                filterContext.ActionDescriptor.GetCustomAttributes(typeof(AuthorizeAttribute), true).Any() ||
+                filterContext.ActionDescriptor.ControllerDescriptor.ControllerType.GetCustomAttributes(typeof(AuthorizeAttribute), true).Any();
+
+            // When user has logged out from a different browser, we have to invalidate all other browser sessions too.
+            // _membershipValidationService.CanAuthenticateWithCookie returns false if the user has logged out from a different browser.
+            if (accessToAuthorizedUserOnly && filterContext.RequestContext.HttpContext.Request.IsAuthenticated &&
+                !_membershipValidationService.CanAuthenticateWithCookie(_authenticationService.GetAuthenticatedUser())) {
```
MatteoPiovanelli, line 28: _membershipValidationService.CanAuthenticateWithCookie() is already called in _authenticationService.GetAuthenticatedUser() so we probably don't need it here as well

Issue #5408, opened 2015-06-22
Dynamic module compilation and C# 6 features


Xceno:
Hi all,

this ain't a high priority thing but I thought I mention it, so it's on the radar.

I wanted to try out some of the new C# 6 features which isn't a problem if you compile your module and then run Orchard.

However, if Orchard is running and the dynamic compilation kicks in, an exception get's thrown when it hits code that utilizes those features (like the new null conditional operator)

``` C#
mailListPart.Record.QueryPartRecord = memberSourceQueryPart?.Record;
```

Here's the exception:

```
System.Web.HttpCompileException (0x80004005): c:\Work\Stash\Orchard\src\Orchard.Web\Modules\MyModule\Handlers\MailListPartHandler.cs(85): error CS1525: Invalid expression '.'.
   at System.Web.Compilation.BuildManager.PostProcessFoundBuildResult(BuildResult result, Boolean keyFromVPP, VirtualPath virtualPath)
   at System.Web.Compilation.BuildManager.GetBuildResultFromCacheInternal(String cacheKey, Boolean keyFromVPP, VirtualPath virtualPath, Int64 hashCode, Boolean ensureIsUpToDate)
   at System.Web.Compilation.BuildManager.GetVPathBuildResultFromCacheInternal(VirtualPath virtualPath, Boolean ensureIsUpToDate)
   at System.Web.Compilation.BuildManager.GetVPathBuildResultInternal(VirtualPath virtualPath, Boolean noBuild, Boolean allowCrossApp, Boolean allowBuildInPrecompile, Boolean throwIfNotFound, Boolean ensureIsUpToDate)
   at System.Web.Compilation.BuildManager.GetVPathBuildResultWithNoAssert(HttpContext context, VirtualPath virtualPath, Boolean noBuild, Boolean allowCrossApp, Boolean allowBuildInPrecompile, Boolean throwIfNotFound, Boolean ensureIsUpToDate)
   at System.Web.Compilation.BuildManager.GetVPathBuildResult(HttpContext context, VirtualPath virtualPath, Boolean noBuild, Boolean allowCrossApp, Boolean allowBuildInPrecompile, Boolean ensureIsUpToDate)
   at System.Web.Compilation.BuildManager.GetCompiledAssembly(String virtualPath)
[...]

PR #7508, opened 2017-01-10
Fixing that Dynamic Compilation broke on C# 6 language features (#5408)

BenedekFarkas:


BenedekFarkas:
I re-initialized the fork since then, so I'll probably close this PR and create another one, but let's keep it here until then.

Issue #7498, opened 2017-01-03
Need a "Do Not Vary by Query String Parameters" cache setting


joshberry:
It would be very useful to have a negative counterpart to the VaryByQueryStringParameters property for cache settings. In many cases, it's cumbersome (and error prone) to list all the query string parameter used on the site when you only want to keep a few of parameters from varying. The example we run into all the time is the Google AdWords gclid parameter. We never want to vary the cache on this since it's only used client-side by analytics.js.

I propose that we add a new field to the cache settings where you can list the query string parameters that you don't want to vary the cache on. It would look something like this...

![image](https://cloud.githubusercontent.com/assets/1396816/21615435/2377fad0-d1a3-11e6-9c63-dc12fec97f43.png)

We are already using this modification on several sites. If folks are interested, I'd be happy to submit it as a PR.

DanielStolt:
I agree that this feature is needed, but the proposed UI is a bit confusing IMO as you would effectively always leave one of the fields empty (any other combination would be redundant).

It also doesn't address the scenario "don't vary by any query string parameters at all" which is also missing today. And btw the current UI where an empty field means unfiltered, and when you enter in a value it suddenly flips the semantic to additive, is also a bit illogical.

Instead I propose something like the following:

![image](https://cloud.githubusercontent.com/assets/3889090/21752263/73b82b60-d5d5-11e6-85b8-975986344299.png)

joshberry:
I like your approach much better. I'll update and submit a PR.

DanielStolt:
Look forward to reviewing it! Thanks @joshberry!

joshberry:
Check it out: #7509. Let me know if you have any suggestions for improvement.

DanielStolt:
Not much, wrote a couple of minor comments on the PR.

PR #7509, opened 2017-01-10
Output cache vary by query string modes

joshberry:
Introducing inclusive and exclusive modes for the output cache module's vary by query string parameters functionality. Details of the change are described in issue #7498. 

Here's how the change looks on the cache settings page:
image

I also added a migration to check if an existing installation is already using the vary by query string parameters and adjust the mode setting to ensure it doesn't experience a change in behavior after upgrading. New and existing installations that don't use the feature will be defaulted to exclusive mode.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Controllers/AdminController.cs
@@ -87,6 +87,7 @@ namespace Orchard.OutputCache.Controllers {
                 DefaultCacheDuration = settings.DefaultCacheDuration,
                 DefaultCacheGraceTime = settings.DefaultCacheGraceTime,
                 DefaultMaxAge = settings.DefaultMaxAge,
+                VaryByQueryStringIsExclusive = settings.VaryByQueryStringIsExclusive,
                 VaryByQueryStringParameters = settings.VaryByQueryStringParameters,
                 VaryByRequestHeaders = settings.VaryByRequestHeaders,
                 VaryByRequestCookies = settings.VaryByRequestCookies,
@@ -115,6 +116,7 @@ namespace Orchard.OutputCache.Controllers {
                 settings.DefaultCacheDuration = model.DefaultCacheDuration;
                 settings.DefaultCacheGraceTime = model.DefaultCacheGraceTime;
                 settings.DefaultMaxAge = model.DefaultMaxAge;
+                settings.VaryByQueryStringIsExclusive = model.VaryByQueryStringIsExclusive;
                 settings.VaryByQueryStringParameters = model.VaryByQueryStringParameters;
                 settings.VaryByRequestHeaders = model.VaryByRequestHeaders;
                 settings.VaryByRequestCookies = model.VaryByRequestCookies;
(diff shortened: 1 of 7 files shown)


src/Orchard.Web/Modules/Orchard.OutputCache/Controllers/AdminController.cs
```diff
@@ -87,6 +87,7 @@ public class AdminController : Controller {
                 DefaultCacheDuration = settings.DefaultCacheDuration,
                 DefaultCacheGraceTime = settings.DefaultCacheGraceTime,
                 DefaultMaxAge = settings.DefaultMaxAge,
+                VaryByQueryStringExclusive = settings.VaryByQueryStringExclusive,
```
DanielStolt: The term should be renamed `VaryByQueryStringIsExclusive`.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -388,8 +388,17 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
             // Vary by configured query string parameters.
             var queryString = filterContext.RequestContext.HttpContext.Request.QueryString;
             foreach (var key in queryString.AllKeys) {
-                if (key == null || (CacheSettings.VaryByQueryStringParameters != null && !CacheSettings.VaryByQueryStringParameters.Contains(key)))
+                if (key == null)
                     continue;
+
+                // In exclusive mode, don't vary if the key matches
+                if (CacheSettings.VaryByQueryStringExclusive && (CacheSettings.VaryByQueryStringParameters != null && CacheSettings.VaryByQueryStringParameters.Contains(key)))
```
DanielStolt: Might want to do a `String.IsNullOrWhitespace()` rather than a simple null check on the settings property.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -388,8 +388,17 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
...
+                    continue;
+
+                // In inclusive mode, don't vary if the key doesn't match
```
DanielStolt: Might want to do a `String.IsNullOrWhitespace()` rather than a simple null check on the settings property.

src/Orchard.Web/Modules/Orchard.OutputCache/Migrations.cs
```diff
@@ -56,5 +65,17 @@ public class Migrations : DataMigrationImpl {
 
             return 7;
         }
+
+        public int UpdateFrom7() {
+            var cacheSettings = _orchardServices.WorkContext.CurrentSite.As<CacheSettingsPart>();
+            if (cacheSettings.VaryByQueryStringParameters != null) {
```
DanielStolt: Might want to do a `String.IsNullOrWhitespace()` rather than a simple null check on the settings property.

DanielStolt:
Overall, very nice contribution!

joshberry:
Thanks for the feedback! I've updated the PR. 

The VaryByQueryStringParameters property in the filter is an IEnumerable, not a string. Whitespace is already handled.

joshberry:
Is there something I need to do for the CLA?

sebastienros:
You need to accept the CLA. You should have received an email to do so, but otherwise you can go directly to this url: https://cla2.dotnetfoundation.org

sebastienros:
Please comment on this PR once you have signed so that it pops up in the PRs list for triage.

joshberry:
Just signed it

PR #7513, opened 2017-01-12
[fixes #7348]Changed the SummaryAdmin view to have the same "New Translation" link as the Editor view

### LINKED ISSUE #7348: BUG: content item localization sometimes causes culture duplicates

MatteoPiovanelli-Laser:
This change should prevent the creation of culture duplicates that was caused by having a different "New 
Translation" link between the two views.

fixes #7348

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Localization/Drivers/LocalizationPartDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Localization/Drivers/LocalizationPartDriver.cs
@@ -20,6 +20,21 @@ namespace Orchard.Localization.Drivers {
         }
 
         protected override DriverResult Display(LocalizationPart part, string displayType, dynamic shapeHelper) {
+
+            return Combined(
+                ContentShape("Parts_Localization_ContentTranslations",
+                             () => shapeHelper.Parts_Localization_ContentTranslations(Id: part.ContentItem.Id, MasterId: ActualMasterId(part), Culture: GetCulture(part), Localizations: GetDisplayLocalizations(part, VersionOptions.Published))),
+                ContentShape("Parts_Localization_ContentTranslations_Summary",
+                             () => shapeHelper.Parts_Localization_ContentTranslations_Summary(Id: part.ContentItem.Id, MasterId: ActualMasterId(part), Culture: GetCulture(part), Localizations: GetDisplayLocalizations(part, VersionOptions.Published))),
+                ContentShape("Parts_Localization_ContentTranslations_SummaryAdmin", () => {
+                    var siteCultures = _cultureManager.ListCultures();
+
+                    return shapeHelper.Parts_Localization_ContentTranslations_SummaryAdmin(Id: part.ContentItem.Id, MasterId: ActualMasterId(part), Culture: GetCulture(part), Localizations: GetDisplayLocalizations(part, VersionOptions.Latest), SiteCultures: siteCultures);
+                })
+                );
+        }
+
+        private int ActualMasterId(LocalizationPart part) {
             var masterId = part.HasTranslationGroup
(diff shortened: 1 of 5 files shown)


src/Orchard.Web/Modules/Orchard.Localization/Drivers/LocalizationPartDriver.cs
```diff
@@ -48,7 +51,7 @@ public class LocalizationPartDriver : ContentPartDriver<LocalizationPart> {
                 SelectedCulture = GetCulture(part),
                 MissingCultures = missingCultures,
                 ContentItem = part,
-                MasterContentItem = part.HasTranslationGroup ? part.MasterContentItem : null,
+                MasterContentItem = _contentManager.Get(part.Record.MasterContentItemId, VersionOptions.Latest),//part.HasTranslationGroup ? part.MasterContentItem : null,
```
sebastienros: Remove the comment

src/Orchard.Web/Modules/Orchard.Localization/Views/EditorTemplates/Parts/Localization.ContentTranslations.Edit.cshtml
```diff
@@ -51,7 +52,8 @@
                 }
 
                 if (Model.MissingCultures.Any()) {
-                    var contentItemId = Model.MasterContentItem != null ? Model.MasterContentItem.Id : Model.ContentItem.Id;
+                    var masterContentItem = WorkContext.Resolve<IContentManager>().Get(((dynamic)Model.ContentItem).Record.MasterContentItemId, VersionOptions.Latest);
```
sebastienros: I think you don't need that

MatteoPiovanelli-Laser:
I made the changes as per @sebastienros comments.

I also made a couple more changes to the driver because I found that if I deleted a MasterContentItem and then attempted to create a New Translation from a content, the list of cultures to choose from would not be correct.

I am reluctant to change the Setter used for the MasterContentItem (see in LocalizationPartHandler) because I am not 100% what that would impact.

src/Orchard.Web/Modules/Orchard.Localization/Drivers/LocalizationPartDriver.cs
```diff
@@ -23,7 +23,10 @@ public class LocalizationPartDriver : ContentPartDriver<LocalizationPart> {
             var masterId = part.HasTranslationGroup
                                ? part.Record.MasterContentItemId
                                : part.Id;
-
+            if (_contentManager.Get(masterId, VersionOptions.Latest) == null) {
```
sebastienros: Yes

MatteoPiovanelli-Laser:
I made that change we discussed in the comments to the earlier version.
I also fixed the line where the MasterContentItem is assigned in the TranslatePOST method of the AdminController.

MatteoPiovanelli-Laser:
The way things work in this PR, I use the fact that part.MasterContentItem is not null when the MasterContentItem has been deleted, so at this point I am happy with the way the Setter for it works.

Down the line, it may make sense to consider how to rewrite that so there's less hoops to jump through to make things happen. I would expect no major rework for it: it may be enough to add a check in the LocalizationPart's MasterContentItem's getter.

Issue #7517, opened 2017-01-15
The settings of Orchard.OpenId features are inaccessible


ThaerAlAjlouni:
After spinning a fresh vanilla site using the **Default** recipe, and enabling one of the **OpenId** features successfully, the settings page of that feature will not be accessible and will throw the following error:

> An exception of type 'System.Web.HttpCompileException' occurred in System.Web.dll but was not handled in user code

And will show the following screen:
![error in the view](https://cloud.githubusercontent.com/assets/22820843/21962239/96f757b4-db29-11e6-9abc-4d07bb52ed3f.png)

ThaerAlAjlouni:
Fixed with PR #7518

PR #7518, opened 2017-01-15
Fix the error that was preventing the load of the settings' views of the OpenId features

ThaerAlAjlouni:
This is a proposed fix of issue #7517


DanielStolt:
@ThaerAlAjlouni I also found another minor issue you may want to address in this PR - Orchard.OpenId references different versions of the Newtonsoft.Json, Iesi.Collections and NHibernate packages than the rest of Orchard core modules, which leads to NuGet consolidation warnings.

sebastienros:
Good to merge now?

sebastienros:
Merging as the CI is broken currently because of this issue

PR #7525, opened 2017-01-21
change form action url to Orchard.OpenId inside openid module

infofromca:


sebastienros:
@ThaerAlAjlouni do you agree with this change?

sebastienros:
@ThaerAlAjlouni ?

ThaerAlAjlouni:
I think it is not needed, since the first will be overridden by the second when the **OpenId** module is enabled and configured correctly.

sebastienros:
If I understand correctly the route pointing to OpenId is supposed to intercept anyone request to /LogOn to be handled by the open id module instead on the users module. In this case then I agree this PR is wrong.

@infofromca or do you think we need the two urls to be able to work at the same time?

infofromca:
suppose in future, we change mind that Orchard.OpenId will depends on another Module called Orchard.BetterLogin , in this case, we only need to change one place which is in router.

But original case, should change two places and easy to forget this place because vs build can not tell us.
@sebastienros @ThaerAlAjlouni

PR #7526, opened 2017-01-21
add Dependencies to openid module

infofromca:


src/Orchard.Web/Modules/Orchard.OpenId/Module.txt
```diff
@@ -5,6 +5,7 @@ Website: http://orchardproject.net
 Version: 1.0
 OrchardVersion: 1.10.1
 Description: Enables Orchard to authenticate users using OpenId
+Dependencies: Orchard.Users
```
sebastienros: this one can go directly to `Orchard.OpenId` section, as it will only be applied to it.

PR #7528, opened 2017-01-22
Fix ChangeExpiredPassword routing in Orchard.OpenId

infofromca:
It fixes this error:
The resource cannot be found.

Description: HTTP 404. The resource you are looking for (or one of its dependencies) could have been removed, had its name changed, or is temporarily unavailable.  Please review the following URL and make sure that it is spelled correctly. 

Requested URL: /OrchardLocal/Users/Account/ChangeExpiredPassword


sebastienros:
Looks good. @ThaerAlAjlouni do you agree with this change?

src/Orchard.Web/Modules/Orchard.OpenId/Controllers/AccountController.cs
```diff
@@ -68,11 +69,12 @@ public class AccountController : Controller {
             }
 
             var membershipSettings = _membershipService.GetSettings();
+
             if (user != null &&
                 membershipSettings.EnableCustomPasswordPolicy &&
                 membershipSettings.EnablePasswordExpiration &&
                 _membershipService.PasswordIsExpired(user, membershipSettings.PasswordExpirationTimeInDays)) {
-                return RedirectToAction("ChangeExpiredPassword", new { username = user.UserName });
+                return RedirectToAction("ChangeExpiredPassword", "Account", new { Area = "Orchard.Users", username = user.UserName });
```
sebastienros: In the same PR you change this to the Users module, but you also create routes for the OpenId module, on the same Controller/Action names, is that on purpose?

src/Orchard.Web/Modules/Orchard.OpenId/Routes/OpenId.cs
```diff
@@ -44,6 +44,57 @@ public class OpenIdRoutes : IRouteProvider {
                 new RouteDescriptor {
                     Priority = 10,
                     Route = new Route(
+                        "Users/Account/LogOn",
```
sebastienros: Is the goal to replace the one from Users ?

ThaerAlAjlouni:
Yes I agree, this was not considered at first.

src/Orchard.Web/Modules/Orchard.OpenId/Routes/OpenId.cs
infofromca: yes

src/Orchard.Web/Modules/Orchard.OpenId/Controllers/AccountController.cs
infofromca: this will redirect action to ChangeExpiredPassword in Orchard.Users because there is no ChangeExpiredPassword in opened module.

The purpose of the  routes will redirect other action.(and also consistent to original code)

For this specific ChangeExpiredPassword action, considerate both of the above,  it will redirect the ChangeExpiredPassword in Orchard.Users module with my test.

ThaerAlAjlouni: But if the user is logged in with one of the OpenId providers, we can't change his/her password, because we don't own it!
So I think we need to implement custom logic to detect whether the user is logged in using a local account and then redirect to the change password action.

src/Orchard.Web/Modules/Orchard.OpenId/Routes/OpenId.cs
ThaerAlAjlouni: No need to replace the routes, once the user sets the settings correctly of one of the enabled OpenId features the app will register the routes automatically after the first restart.
This was intended, no need to override the routes always, only override when the replacement is valid.
I think a more proper solution will be notifying the user that an app restart is required as @sebastienros earlier suggested.

BenedekFarkas:
@sebastienros what's the next step for this PR?

BenedekFarkas:
@sebastienros should we close this?

sebastienros:
@BenedekFarkas I let you decide

BenedekFarkas:
I'll close this for now due to unaddressed feedback + needs comprehensive testing. Please open a new PR if the change requests are addressed and provide testing steps to make sure that there are no breaking changes.

PR #7533, opened 2017-01-24
adding publishlater/archive later to audit log

jimasp:
adding publish later/archive later to audit log.

the part value needs to be updated so that the export/diffgram picks up the change,

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
@@ -239,6 +239,19 @@ namespace Orchard.Projections {
                 DisplayName = T("Body Part Text").Text,
                 Description = T("The text from the Body part").Text
             });
+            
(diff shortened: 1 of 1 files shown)


sebastienros:
Do you mean that `ScheduledArchiveUtc` is never used by the drivers and should have been?

sebastienros:
It seems to be a lazy loaded field with a custom loader to get the value from the database on demand (a proxy). Maybe the solution would be to add another property and make you code use this new one instead. With some comment explaining that this is just for audit trail reasons.

Or if you test thoroughly your change, I assume this should work too.

jimasp:
thanks, i'm happy with the change as is. it'll get thoroughly tested before it goes live. 
(we've had users complaining of publish later apparently not working in production but we can't recreate the problem. so i wanted to audit that property, as currently we can't prove whether or not/when the date/time fields were filled in by the user, or e.g. they filled it in but hit save instead of publish later).

Issue #7546, opened 2017-01-29
Openid: disable, then enable google feature produce error


infofromca:
![googledisable-enable](https://cloud.githubusercontent.com/assets/297675/22401123/7913bedc-e596-11e6-99fd-41462aa7e983.png)

infofromca:
i found it
the seeting of LogonCallbackUrl should be  /Users/Account/LogonCallback

sebastienros:
Is it documented somewhere?

infofromca:
yes, on the setting page, the hint said using
~/Users/Account/LogonCallback

ThaerAlAjlouni:
The **`Microsoft.Owin.PathString`** `struct` is not accepting the "~" character at the beginning of the string passed to its `constructor` which is causing the error. We need to remove it from the default values (in the ` Constants.General` class) which will reflect elsewhere in the solution.

ThaerAlAjlouni:
@infofromca @sebastienros 
Shouldn't this be closed as the fix was merged?

PR #7556, opened 2017-02-03
Fix the error thrown on enabling the "OpenId.Google" feature

ThaerAlAjlouni:
Remove the **tilde ~** from the `LogonCallbackUrl` relative path to prevent errors after enabling the `OpenId.Google` feature.

This will fix issue #7546


infofromca:
I think this just resovled the half.
because the user can still input ~/Users/Account/LogonCallback to the LogonCallbackUrl textbox.
My idea is to disable user input in the textbox of LogonCallbackUrl

ThaerAlAjlouni:
I think adding validation will be better than not allowing the user to edit the callback URL.

src/Orchard.Web/Modules/Orchard.OpenId/Constants/General.cs
```diff
@@ -1,7 +1,7 @@
 ﻿namespace Orchard.OpenId.Constants {
     public class General {
         public const string AuthenticationErrorUrl = "~/Authentication/Error";
-        public const string LogonCallbackUrl = "~/Users/Account/LogonCallback";
+        public const string LogonCallbackUrl = "/Users/Account/LogonCallback";
```
sebastienros: Have you tried on multi-tenancy with a route prefix in this case?

sebastienros:
@infofromca is the PR ok with you now?

src/Orchard.Web/Modules/Orchard.OpenId/Constants/General.cs
ThaerAlAjlouni: No to be honest, I haven't test it on multi-tenancy. I will test it and update the PR if required.

ThaerAlAjlouni: @sebastienros, PR updated after modifying code to support multi-tenancy

infofromca:
I guess the purpose of GetCallbackPath can be found somewhere in the old code.
but it is difficult to find.
long time ago, I suggested that we moved common use code to  a central module ( or Utility project or folder) , if it is not easy for orchard 1 , we should consider it in orchard 2

sebastienros:
I think the `IsValid` property should be removed from the class, or at least converted to a method so that is doesn't get serialized in the db.

sebastienros:
@infofromca this is handled automatically in ASP.NET Core (hence in Orchard Core) because they introduced a `PathBase` in the request object. And everything is local to this. So we assign the tenant prefix now.

src/Orchard.Web/Modules/Orchard.OpenId/OwinMiddlewares/Google.cs
```diff
@@ -27,7 +29,7 @@ public class Google : IOwinMiddlewareProvider {
             var authenticationOptions = new GoogleOAuth2AuthenticationOptions {
                 ClientId = settings.ClientId,
                 ClientSecret = settings.ClientSecret,
-                CallbackPath = new PathString(settings.CallbackPath)
+                CallbackPath = new PathString(GetCallbackPath(workContext, settings))
```
sebastienros: Isn't it fine if the users adds the prefix by themself? We can just document that the url they enter needs to contain the tenant prefix if there is one. I would prefer this option if it works.

sebastienros:
@infofromca what do you think about my comments, does it solve your concerns?

infofromca:
the code is ok 
because we do not have PathBase right now.
ask user to input tenant prefix to too technical, and not every editor is developer.
IsValid seems not used in db

src/Orchard.Web/Modules/Orchard.OpenId/OwinMiddlewares/Google.cs
ThaerAlAjlouni: Why to add an extra step of responsibilty on the user?
I prefer less intervention from the user in such scenarios.

src/Orchard.Web/Modules/Orchard.OpenId/Views/EditorTemplates/Parts.AzureActiveDirectorySettings.cshtml
```diff
@@ -2,31 +2,31 @@
 
 <h2>Azure Active Directory Settings</h2>
 <fieldset>
-    @Html.LabelFor(m => m.Tenant, T("Tenant"))
+    <label for="AzureActiveDirectorySettings_Tenant">@T("Tenant")</label>
```
sebastienros: `@Html.IdFor()` everywhere please

sebastienros: :up::date: ?

ThaerAlAjlouni: I will try to do it the upcoming Friday.

sebastienros:
As a reminder, anyone who has write access to the repository is allowed to help some PRs when they are blocked by some small patches. I'll try to remind it during tomorrow's meeting.

src/Orchard.Web/Modules/Orchard.OpenId/Views/EditorTemplates/Parts.AzureActiveDirectorySettings.cshtml
ThaerAlAjlouni: @sebastienros Done

Issue #7530, opened 2017-01-22
methods in AccountController of in openid module should consider returnUrl


infofromca:
methods in AccountController of in openid module should consider returnUrl like orchard.Users

Devqon:
It does..: https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Users/Controllers/AccountController.cs#L74

ThaerAlAjlouni:
@sebastienros Yes I think it does, it will be nice to have this, but the way **OpenId** works by default will prevent that because it relies on redirecting back to a specific action (_LogOnCallback_ in our case).

I think this is doable by saving the requested URL (in memory or session or another mechanism) before redirecting to the Id provider and then handling it again on successful login.

What do you think?

sebastienros:
Using Session directly is not recommended. Try to use TempData whic is used for return url too.

PR #7558, opened 2017-02-05
Consider "returnUrl" in the `OpenId` actions

ThaerAlAjlouni:
Consider **`returnUrl`** in the **OpenId** actions. Fixes issue #7530

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OpenId/Controllers/AccountController.cs
+++ b/src/Orchard.Web/Modules/Orchard.OpenId/Controllers/AccountController.cs
@@ -18,6 +18,7 @@ using Orchard.Users.Events;
 namespace Orchard.OpenId.Controllers
 {
     [Themed]
+    [AlwaysAccessible]
     [OrchardFeature("Orchard.OpenId")]
     public class AccountController : Controller {
         private readonly IEnumerable<IOpenIdProvider> _openIdProviders;
@@ -47,16 +48,17 @@ namespace Orchard.OpenId.Controllers
         public Localizer T { get; set; }
 
         [HttpGet]
-        public ActionResult LogOn() {
+        public ActionResult LogOn(string returnUrl) {
             if (Request.IsAuthenticated) {
                 return Redirect(Url.Content("~/"));
             }
+
+            ViewData["ReturnUrl"] = returnUrl;
 
             return View(_openIdProviders);
         }
 
         [HttpPost]
-        [AlwaysAccessible]
         [ValidateInput(false)]
         [SuppressMessage("Microsoft.Design", "CA1054:UriParametersShouldNotBeStrings", Justification = "Needs to take same parameter type as Controller.Redirect()")]
         public ActionResult LogOn(string userNameOrEmail, string password, string returnUrl, bool rememberMe = false) {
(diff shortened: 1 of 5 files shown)


src/Orchard.Web/Modules/Orchard.OpenId/Controllers/AccountController.cs
```diff
@@ -81,16 +83,20 @@ public class AccountController : Controller {
...
+        [SuppressMessage("Microsoft.Design", "CA1054:UriParametersShouldNotBeStrings", Justification = "Needs to take same parameter type as Controller.Redirect()")]
+        public void Challenge(string openIdProvider, string returnUrl) {
             _userEventHandler.LoggingIn(openIdProvider, String.Empty);
 
             if (String.IsNullOrWhiteSpace(openIdProvider))
                 openIdProvider = OpenIdConnectAuthenticationDefaults.AuthenticationType;
 
             if (Request.IsAuthenticated) {
-                Redirect(Url.Content("~/"));
+                Redirect(returnUrl);
```
sebastienros: Don't use `returnUrl` anywhere in a redirect without checking the destination is allowed, like `RedirectLocal` does or if it needs to go to a provider's domain, then check it too.

This comment applies to every change in this PR.

ThaerAlAjlouni: The code is updated to use the `RedierctLocal` extension method instead.

src/Orchard.Web/Modules/Orchard.OpenId/Views/EditorTemplates/Parts.AzureActiveDirectorySettings.cshtml
```diff
@@ -59,5 +59,8 @@
     @Html.LabelFor(m => m.GraphApiUrl, T("Graph API URL"))
     @Html.TextBoxFor(m => m.GraphApiUrl, new { @class = "text large" })
     <span class="hint">@T("Typically https://graph.windows.net")</span>
+
+    @Html.LabelFor(m => m.GraphApiKey, T("Graph API Key"))
```
sebastienros: I will merge the PR, but can you create a new one that removes the usage of LabelFor and instead use a `<label>` tag? Otherwise there can be issues with double-encoding (try to use an accent for instance, or `<` in the text)

src/Orchard.Web/Modules/Orchard.OpenId/Models/AzureActiveDirectorySettingsPart.cs
```diff
@@ -71,6 +71,12 @@ public class AzureActiveDirectorySettingsPart : ContentPart {
             set { this.Store(x => x.AppKey, value); }
         }
 
+        public string GraphApiKey
+        {
```
sebastienros: Format

src/Orchard.Web/Modules/Orchard.OpenId/Views/EditorTemplates/Parts.AzureActiveDirectorySettings.cshtml
ThaerAlAjlouni: Sure :)

src/Orchard.Web/Modules/Orchard.OpenId/Models/AzureActiveDirectorySettingsPart.cs
ThaerAlAjlouni: OK :)

PR #7562, opened 2017-02-09
TextFieldDriver doesn't export its tag if content is empty

### LINKED ISSUE #7493: TextFieldDriver doesn't export its tag if content is empty
I reproduced this bug on 1.10.x
TextFieldDriver doesn't export its tag in recipe if content is empty. There is a check in Exporting function which exports the tag only if the value is not null or empty.

```
        protected override void Exporting(ContentPart part, TextField field, ExportContentContext context) {
            if (!String.IsNullOrEmpty(field.Value))
                context.Element(field.FieldDefinition.Name + "." + field.Name).SetAttributeValue("Text", field.Value);
        }
```

Because of this, when importing recipe on target application it skips the TextFieldDriver and doesn't replace the content with empty value.

That condition should be removed just like we have in BodyPartDriver

hannan-azam:
Fixes #7493

@sebastienros  we have reverted the code changes made by @DanielStolt in the commit 1edf94756c671b9c7040403154e5d99753498318 that were related to recipe cleanup.
However, since the default value issue for fields has not yet been sorted out, the Exporting method of all the drivers is still not creating the attributes in recipes as the function SetAttributeValue doesn't create attribute if the value is null and for all the drivers we have null value as default.


DanielStolt:
@hannan-azam I just merged your other PR #7510. Why have you created this one too? They seem to contain the exact same commits?

Closing this one as all these commits are already merged. Feel free to reopen if I have misunderstood something.

Issue #7554, opened 2017-02-01
Projection widget fails when on SQL CE and the pager is displayed


paynecrl97:
More specifically, the `DefaultHqlQuery.Count()` method fails because the SQL generated is not compatible with SQL CE.

Example query generated to get count of content items:
```
select count(distinct contentite0_.Id) as col_0_0_ 
from Orchard_Framework_ContentItemVersionRecord contentite0_ 
  inner join Orchard_Framework_ContentItemRecord contentite1_ 
    on contentite0_.ContentItemRecord_id=contentite1_.Id 
  inner join Orchard_Framework_ContentTypeRecord contenttyp2_ 
    on contentite1_.ContentType_id=contenttyp2_.Id 
where (contenttyp2_.Name in ('BlogPost')) 
  and contentite0_.Published=1
```

SQL CE does not support `count(distinct )`, and so this query fails, which results in an exception being thrown, and if the query is being executed within the context of the `ProjectionPartDriver`, the projection shape will not be displayed.

paynecrl97:
I should add that this is against the latest dev.

sebastienros:
Look at the file history, we did a change in the past weeks that must have done that. It was encapsulating the query instead. Safe is to revert, unless you find a way that is best for the two issues.

sebastienros:
@jersiovic this is about the change you made on HQL I think

sebastienros:
@paynecrl97 Try to revert this to see if it fixed your issue: https://github.com/OrchardCMS/Orchard/commit/020cb8b8bf402864a8974ef2770cde9c5d712a2e

paynecrl97:
@sebastienros it does. https://github.com/OrchardCMS/Orchard/pull/7563

PR #7563, opened 2017-02-09
Revert "Improved performance of Count Hql queries"

paynecrl97:
Reverts OrchardCMS/Orchard#7501 (see #7554).

`count(distinct x)` is not supported in SqlCE


sebastienros:
Fixes #7554

Issue #3954, opened 2013-09-20
Style.Require strips query string


orchardbot:
**woodvorg** created:
https://orchard.codeplex.com/workitem/20126

Running Orchard 1.6.
manifest.DefineStyle("Site").SetUrl("Site.css?d200913");

Style.Require("Site").AtHead();

The querystring "?d200913" is stripped at the output... but only if the file Site.css exists. If the file doesn't exist it doesn't strip the output.

jbuck79:
This appears to have been fixed by #7564 but it hasn't been merged into master yet.

sebastienros:
Fixed in dev

madmox:
Tried migrating our solution to Orchard 1.10.3 to check if this issue was fixed. Looks like it's not!

This keeps the querystring:
```
Script.Include(Url.Content("foo.js?foo=bar")).AtHead()
// Generates: <script src="/Modules/MyModule/scripts/foo.js?foo=bar" type="text/javascript"></script>
```

When the file `foo.css` exists, this doesn't:
```
Style.Include(Url.Content("foo.css?foo=bar")).AtHead()
// Generates: <link href="/Modules/MyModule/Styles/foo.css" rel="stylesheet" type="text/css" />
```

madmox:
Looks like the whole PR #7564 is not included in Orchard 1.10.3 (at least not in [the revision](https://github.com/OrchardCMS/Orchard/commit/a53d30dd6a94d80aa02ae98e617d288ffdc23f64) associated to the tag).

PR #7564, opened 2017-02-10
Implements resource file hash-based cache busting

DaRosenberg:
This PR adds automatic file hash-based cache busting to the Orchard resource manager.

The resource manager now uses a new abstraction `IResourceFileHashProvider` to add a file content hash to the query string of URLs for any scripts and stylesheets sourced from the local file system. The default implementation `ResourceFileHashProvider` provides a simple but efficient (and thread-safe) implementation, which uses stream-based MD5 hashing, and caches generated hashes based on the last modified time on the file on disk.

Resource file hashing can be disabled/enabled through a new site setting, exposed in the settings UI. It will be enabled by default for new sites, but **not** for existing sites.

### DIFF
--- a/src/Orchard.Tests/Stubs/StubWorkContextAccessor.cs
+++ b/src/Orchard.Tests/Stubs/StubWorkContextAccessor.cs
@@ -66,7 +66,7 @@ namespace Orchard.Tests.Stubs {
                     set { throw new NotImplementedException(); }
                 }
 
-				public string SiteCalendar {
+                public string SiteCalendar {
                     get { throw new NotImplementedException(); }
                     set { throw new NotImplementedException(); }
                 }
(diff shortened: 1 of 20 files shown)


jbuck79:
Is there a reason why this never made it into the 1.10.2 release? This code fixes #3954 and it would be really nice to have it. 

I copied over just this code from StylesheetBindingStrategy (lines 92-100):
```
var shape = ((dynamic)displayContext.Value);
      var output = displayContext.ViewContext.Writer;
      ResourceDefinition resource = shape.Resource;
      string url = shape.Url;
      string condition = shape.Condition;
      Dictionary<string, string> attributes = shape.TagAttributes;
      ResourceManager.WriteResource(output, resource, url ?? hit.fileVirtualPath, condition, attributes);
      return null;
```
And it works. But my team tries not to modify the Orchard source code whenever possible to make merging in new releases easier.

Would the fix for #3954 be more likely to make it into an official release if I submitted a pull request with just that bit of code?

Issue #7565, opened 2017-02-10
validation of Comment submission for SiteName field is not represented for anonymous users.


siyamandayubi:

siyamandayubi:
This issue is addressed in PR #7567

PR #7566, opened 2017-02-10
validation of Comment submission for SiteName field is not represented for anonymous users. #7565

siyamandayubi:
validation of Comment submission for SiteName field is not represented for anonymous users. #7565


siyamandayubi:
contains old commits

Issue #7570, opened 2017-02-13
Default value for LastPasswordChangeUtc in 10.2 raises exception in MySql databases


siyamandayubi:
Changing the deafult value from min to 1990/1/1 causes an exception in MySql database

Issue #4780, opened 2014-09-17
AoutroutePartRecord Index under MySQL Issue


orchardbot:
**veskok** created:
https://orchard.codeplex.com/workitem/20951

Unfortunately MySQL does not support text columns in indexes longer than 767 bytes.
 In Orchard.Autoroute migrations we have 
  SchemaBuilder.AlterTable("AutoroutePartRecord", table => table
                .CreateIndex("IDX_AutoroutePartRecord_DisplayAlias", "DisplayAlias"));

, but DisplayAlias is 2048 bytes long, far from the constraint of 767 bytes.

A workaround of this is to change like:
  SchemaBuilder.AlterTable("AutoroutePartRecord", table => table
                .CreateIndex("IDX_AutoroutePartRecord_DisplayAlias", "DisplayAlias(767)"));

, which probably will work only on MySQL.

The same issue we have in Orchard.MediaLibrary
 SchemaBuilder.AlterTable("MediaPartRecord", t => t
                .CreateIndex("IDX_MediaPartRecord_FolderPath", "FolderPath"));

, where FolderPath is far from 767 bytes long, so the workaround is
 SchemaBuilder.AlterTable("MediaPartRecord", t => t
                .CreateIndex("IDX_MediaPartRecord_FolderPath", "FolderPath(767)"));

Actually we can add the following in migrations
   private readonly ISessionFactoryHolder _sessionFactoryHolder;
        private readonly Dialect _dialect;

```
    public Migrations(ISessionFactoryHolder sessionFactoryHolder)
    {
        this._sessionFactoryHolder = sessionFactoryHolder;
        var configuration = _sessionFactoryHolder.GetConfiguration();
        this._dialect = Dialect.GetDialect(configuration.Properties);
    }
```

and to check if dialect is MySQL 
 var dialect = this._dialect as MySQLDialect;
            if (dialect != null)
            {
            SchemaBuilder.AlterTable("AutoroutePartRecord", table => table
                .CreateIndex("IDX_AutoroutePartRecord_DisplayAlias", "DisplayAlias(767)"));
            }
            else
            {
            SchemaBuilder.AlterTable("AutoroutePartRecord", table => table
                .CreateIndex("IDX_AutoroutePartRecord_DisplayAlias", "DisplayAlias"));
            }

luciofassio:
I'm trying:
Orchard 1.9.2
MySql 5.6

I get:
[...]

Issue #4779, opened 2014-09-17
Some column types in Orchard tables must be changed in MySQL, because default types offers too small text size.


orchardbot:
**veskok** created:
https://orchard.codeplex.com/workitem/20950

Some column types in Orchard tables must be changed, because defaul types offers too small text size.

TINYBLOB, TINYTEXT       L + 1 bytes, where L < 2^8    (255 Bytes)
BLOB, TEXT               L + 2 bytes, where L < 2^16   (64 Kilobytes)
MEDIUMBLOB, MEDIUMTEXT   L + 3 bytes, where L < 2^24   (16 Megabytes)
LONGBLOB, LONGTEXT       L + 4 bytes, where L < 2^32   (4 Gigabytes)

Table: Orchard_Framework_ContentItemVersionRecord
Issue: Change type of the "Data" column from "TEXT" to "MEDIUMTEXT"

Table: Common_BodyPartRecord
Issue: Change type of the "Text" column from "TEXT" to "MEDIUMTEXT"

Table: Orchard_Framework_ContentItemRecord
Issue: Change type of the "Data" column from "TEXT" to "MEDIUMTEXT"

PR #7571, opened 2017-02-13
Fix issue #7570, #5366, #4780 and #4779 

siyamandayubi:
Orchard 10.1 cannot be installed by default on a MySql database due to having indexes on too long nvarchar columns.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Migrations.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Migrations.cs
@@ -7,7 +7,7 @@ namespace Orchard.Users {
     public class UsersDataMigration : DataMigrationImpl {
 
         public int Create() {
-            SchemaBuilder.CreateTable("UserPartRecord", 
+            SchemaBuilder.CreateTable("UserPartRecord",
                 table => table
                     .ContentPartRecord()
                     .Column<string>("UserName")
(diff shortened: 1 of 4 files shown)


src/Orchard.Web/Modules/Orchard.Users/Migrations.cs
```diff
@@ -23,7 +23,7 @@ public class UsersDataMigration : DataMigrationImpl {
                     .Column<DateTime>("CreatedUtc")
                     .Column<DateTime>("LastLoginUtc")
                     .Column<DateTime>("LastLogoutUtc")
-                    .Column<DateTime>("LastPasswordChangeUtc", c => c.WithDefault(new DateTime(1990, 1, 1)))
+                    .Column<DateTime>("LastPasswordChangeUtc", c => c.WithDefault(new DateTime(1990, 1, 1).ToString("yyyy-MM-dd")))
```
sebastienros: Move to `DefaultDataMigrationInterpreter` with 

```
return "'" + ((DateTime)value).ToString("yyyy-MM-dd HH:mm:ss") + "'";
```

src/Orchard/Data/Migration/Interpreters/MySqlCommandInterpreter.cs
```diff
@@ -82,5 +84,31 @@ public class MySqlCommandInterpreter : ICommandInterpreter<AlterColumnCommand> {
                 return tableName;
             return _shellSettings.DataTablePrefix + "_" + tableName;
         }
+
+        public string[] CreateStatements(AddIndexCommand command) {
```
sebastienros: Put a static prefix length value. 
Also reference the actual bugs  #4780 and #4779? and #5366

PR #7577, opened 2017-02-15
7572: FileSystemStorageProvider shouldn't allow creating folders with "&" in their name

### LINKED ISSUE #7572: FileSystemStorageProvider shouldn't allow creating folders with "&" in their name
The "&" in the name of a MediaLibrary folder creates an error when the Media Url is requested.
repro steps
- Create a folder POI&Events in Media Library
- Upload an image
- request the image URL
you will obtain an error

2 possible solutions:
1. prevent folder creation when the "&" is present
2. encode the "&" as %26 (as done for spaces " " > %20)

MatteoPiovanelli:
Changes made to prevent ampersand in folder names, especially for Media.
Fixes #7572 

To avoid the issues with the '&' character in folder names I made changes on different levels:

- Input validation in the forms where the user inserts the folder name, both for new folders and while renaming them.
- Validation in MediaLibraryService, in the CreateFolder and RenameFolder methods, in case we get around the html validation.
- Validation in FileSystemStorageProvider, because the same path issues will present themselves also for files that we do not access through the MediaLibrary Module.

I tested these correctly prevented me to create/edit folder with an '&' character in the name.


MatteoPiovanelli-Laser:
The list of invalid characters is obtained at runtime and is based on web.config.
Additionally, this also handles the fact that the '/' and the '\' characters would cause folders to move, or creation of subfolders. Those two characters are added to those obtained from web.config, in case the defaults have been changed.

src/Orchard.Web/Modules/Orchard.MediaLibrary/ViewModels/MediaManagerFolderEditViewModel.cs
```diff
@@ -1,11 +1,21 @@
...
+using System.Web.Configuration;
 
 namespace Orchard.MediaLibrary.ViewModels {
     public class MediaManagerFolderEditViewModel {
         [Required]
         public string Name { get; set; }
         public string FolderPath { get; set; }
+        public string InvalidCharactersPattern {
+            get {
+                string invalidChars = ((SystemWebSectionGroup)WebConfigurationManager.OpenWebConfiguration(null).GetSectionGroup("system.web")).HttpRuntime.RequestPathInvalidCharacters;
```
sebastienros: This code is replicated 3 times.

I'd suggest that you just create a static Regex that checks for any known chars, without having to read the web.config. It would be perfect but it adds unnecessary complexity in this case.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/Folder/Create.cshtml
```diff
@@ -4,29 +4,31 @@
...
+</p>
+</div>
 
-@using (Html.BeginFormAntiForgeryPost()) { 
+@using (Html.BeginFormAntiForgeryPost()) {
     @Html.ValidationSummary()
     <fieldset>
         <label for="Name">@T("Folder Name")</label>
-        <input id="Name" class="text medium" name="Name" type="text" autofocus="autofocus"/>
+        <input id="Name" class="text medium" name="Name" type="text" autofocus="autofocus" pattern="@Model.InvalidCharactersPattern" title="@T("The name contains invalid characters.")" />
```
sebastienros: If you can't reuse the static Regex, then just ignore this attribute.

MatteoPiovanelli-Laser:
@sebastienros I added the static regex in MediaFolder and used it in the viewmodels and the services. The view is unchanged, as it reads the pattern from the viewmodel

src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/Folder/Create.cshtml
sebastienros: Also, if you just want to encode the pattern, don't expose a property to do that, encode based on the context, here from the View directly, you can use JavascriptEncode.

src/Orchard.Web/Modules/Orchard.MediaLibrary/ViewModels/MediaManagerFolderCreateViewModel.cs
```diff
@@ -1,10 +1,17 @@
...
+using System.Collections.Generic;
+using System.Web.Configuration;
 using Orchard.MediaLibrary.Models;
 
 namespace Orchard.MediaLibrary.ViewModels {
     public class MediaManagerFolderCreateViewModel {
         public string Name { get; set; }
         public string FolderPath { get; set; }
         public IEnumerable<IMediaFolder> Hierarchy { get; set; }
+        public string InvalidCharactersPattern {
```
sebastienros: Please remove

BenedekFarkas:
@MatteoPiovanelli-Laser can you please take care of the review feedback? I'll review/test after so we can include this in 1.10.4/1.11.
Related: #6748 / #6792

MatteoPiovanelli:
I think I updated this PR correctly. I also added the characters referenced in RFC 1378 as discussed in #6792.

BenedekFarkas:
@MatteoPiovanelli-Laser can you please update the PR so I can push a few changes to this branch (Allow edit from maintainers) or I alternatively I can push to my PR (#6792) some of your changes including the following adjustments/remarks:

1. In the template changes, the title values of the fields contain the validation error messages, but that can be misleading if the input is empty or has valid value. We can change it to describe what it's for and list the characters that are not allowed. Adding client-side validation is nice btw, I only added server-side validation.
2. In your PR, the path validation is applied to CreateFolder and RenameFolder. I think it should applied to file operations too (upload, rename, move, copy).
3. In your PR, the path validation is applied to not just to the MediaLibraryService, but also to the FileSystemStorageProvider. That's a good idea, because you can use file storage (in this case file system, but Azure Blob Storage too) without Media Library. I think we should have some common logic in the Framework (StorageHelper or something like that) to pass the strings into for validation. This should include everything from the 0-31 control characters + RFC 1378 (so this would cover `HttpRuntimeSection.RequestPathInvalidCharacters` and RFC 3986, which is for URIs). I would apply all this to MediaLibraryService, FileSystemStorageProvider and AzureBlobStorageProvider.

@sebastienros do you agree?

MatteoPiovanelli:
The PR is set to allow edit from maintainers already.
1. Yes, I did not notice that I set that value for the title. It should be changed.
2. Don't files already have something? There is a GetUniqueFilename method, but I don't remember where/how it's used.
3. I agree there should be a "centralized" place where invalid characters are controlled.

BenedekFarkas:
Thanks Matteo for your work on this and sorry that it took this long to finally get this improved validation done, #6792 was inspired by this PR too.

PR #7581, opened 2017-02-19
allow the anonymouse user to access Logon of openid

infofromca:


src/Orchard.Web/Modules/Orchard.OpenId/Controllers/AccountController.cs
```diff
@@ -46,6 +46,7 @@ public class AccountController : Controller {
         public ILogger Logger { get; set; }
         public Localizer T { get; set; }
 
+        [AlwaysAccessible]
```
ThaerAlAjlouni: I think that this is not required, the whole controller is always accessible by default.

infofromca: To consistent to Orchard.User module, and also to defend some condition changes in future,
[AlwaysAccessible] can still work inside whole orchard system

Issue #7590, opened 2017-03-01
Missing index on ContentTagRecord table


GiuseppeMusso-Laser:
Hi all.
I'm experiencing a performance problem saving a content item.
I analized the queries in my log and I found a query filtering on field TagsPartRecord_Id in table Orchard_Tags_ContentTagRecord.
I noticed this query takes about 3 sec. when I run it on SQL Management Studio.
So I propose to create an index on this field.

PR #7591, opened 2017-03-01
Tags: added index on ContentTagRecord table.

GiuseppeMusso-Laser:
Added index on field TagsPartRecord_Id in table ContentTagRecord for performance enhancement.
See issue #7590.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Tags/Migrations.cs
+++ b/src/Orchard.Web/Modules/Orchard.Tags/Migrations.cs
@@ -36,6 +36,13 @@ namespace Orchard.Tags {
 
             return 2;
         }
+
+        public int UpdateFrom2() {
+            SchemaBuilder.AlterTable("ContentTagRecord", table => table
+                .CreateIndex("IDX_TagsPartRecord_Id", "TagsPartRecord_Id")
+            );
+            return 3;
+        }
     }
 
     [OrchardFeature("Orchard.Tags.TagCloud")]


sebastienros:
Is the actual query on this field only or the two fields of this table. Can you point to the line of code that does the query?
If it's using the two of them (as I would expect) then we need to include both fields in the index. We might also need two indexes btw.

GiuseppeMusso-Laser:
I enabled SQL logging in log4net.config file:

	
	<logger name="NHibernate.SQL">
	  <priority value="DEBUG" />
	</logger>

Then I edited a ContentItem containing tags and I got the following SQL query in my log:

SELECT tags0_.TagsPartRecord_id as TagsPart3_2_, tags0_.Id as Id2_, tags0_.Id as Id561_1_, 
    tags0_.TagRecord_id as TagRecord2_561_1_, tags0_.TagsPartRecord_id as TagsPart3_561_1_, 
    tagrecord1_.Id as Id562_0_, tagrecord1_.TagName as TagName562_0_ 
FROM Orchard_Tags_ContentTagRecord tags0_ 
    left outer join Orchard_Tags_TagRecord tagrecord1_ on tags0_.TagRecord_id=tagrecord1_.Id 
WHERE tags0_.TagsPartRecord_id=@p0;
@p0 = 1837 [Type: Int32 (0)]

Here the only where condition is on field TagsPartRecord_id.

Issue #7593, opened 2017-03-01
Missing index on TermContentItem table


GiuseppeMusso-Laser:
Hi all.
I'm working on some performance problems and I noticed some query filtering on field TermsPartRecord_id in table Orchard_Taxonomies_TermContentItem. This field has no index on it.
So I propose to create one.

PR #7594, opened 2017-03-01
Taxonomies: added index on TermContentItem table.

GiuseppeMusso-Laser:
Added index on field TermsPartRecord_id in table TermContentItem for performance enhancement.
See issue #7593.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Migrations.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Migrations.cs
@@ -94,5 +94,11 @@ namespace Orchard.Taxonomies {
             );
             return 6;
         }
+        public int UpdateFrom6() {
+            SchemaBuilder.AlterTable("TermContentItem", table => table
+                .CreateIndex("IDX_TermsPartRecord_id_Field", "TermsPartRecord_id", "Field")
+            );
+            return 7;
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


sebastienros:
Same question, what is the actual query that is used, so we can include all the necessary fields in the index.
Or did you get the index from SQL Server Advisor?

GiuseppeMusso-Laser:
I got the queries from log4net.
In this case I found two different kind of queries involving TermsPartRecord_id:

SELECT terms0_.TermsPartRecord_id as TermsPar4_1_, terms0_.Id as Id1_, terms0_.Id as Id394_0_, terms0_.Field as Field394_0_, terms0_.TermRecord_id as TermRecord3_394_0_, terms0_.TermsPartRecord_id as TermsPar4_394_0_ 
FROM Orchard_Taxonomies_TermContentItem terms0_ 
WHERE terms0_.TermsPartRecord_id=@p0;
@p0 = 2201 [Type: Int32 (0)]

select termconten0_.Id as Id394_, termconten0_.Field as Field394_, termconten0_.TermRecord_id as TermRecord3_394_, termconten0_.TermsPartRecord_id as TermsPar4_394_ 
from Orchard_Taxonomies_TermContentItem termconten0_ 
where termconten0_.TermsPartRecord_id=@p0 
and termconten0_.Field=@p1;
@p0 = 2201 [Type: Int32 (0)], 
@p1 = 'Categoria' [Type: String (4000)]

The first one has one where condition while the second one has an additional condition on "Field".
Is it better to create two indexes, the first one on TermsPartRecord_id and the second one on (TermsPartRecord_id, Field)?

sebastienros:
Yes, please try to add the two of them and see if it works.

PR #7618, opened 2017-03-21
Orchard.Azure.MediaServices module updates

mconverti:
- Upgraded Azure Media Services legacy components (like the encoder and presets)
- Included Azure Media Player v2.0.0.9 for the video playback experience
- Added support to auto-configure the Orchard.Azure.MediaServices module (Media Services and Storage credentials) based on application settings
- Added new 'Video Portal' recipe (based on 'Default' recipe) to enable Orchard.Azure.MediaServices module
- Improved Setup experience to avoid asking for connection string if it's already available in the application settings


src/Orchard.Web/Modules/Orchard.Setup/Services/WebconfigOverrides.cs
```diff
@@ -0,0 +1,39 @@
...
+        public string DataConnectionString { get; set; }
+        public string DefaultRecipe { get; set; }
+
+        public static WebconfigOverrides Load()
+        {
+            var result = new WebconfigOverrides();
+
+            if (ConfigurationManager.ConnectionStrings[ConnectionStringDefaultConnectionString] != null)
+            {
+                result.DataConnectionString = ConfigurationManager.ConnectionStrings[ConnectionStringDefaultConnectionString].ConnectionString;
```
sebastienros: You need to prefix with the Tenant's name like `tenantName + ":" + ConnectionStringDefaultConnectionString` and fallback to one without the tenant name. Because otherwise this would override the value for subsequent tenants.

sebastienros (commented):
Is it possible to have the AMS recipe be available only in the setup if we build the image for Azure?
This would be a new Target in the `Orchard.proj` file that would copy the file from the module to the publication folder after build and before packaging.

src/Orchard.Web/Modules/Orchard.Setup/Services/WebconfigOverrides.cs
sebastienros: Maybe you can use our `IPlatformConfigurationAccessor` that will do it for you.

src/Orchard.Web/Modules/Orchard.Setup/Services/WebconfigOverrides.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public string DataProvider { get; set; }
+        public string DataConnectionString { get; set; }
+        public string DefaultRecipe { get; set; }
+
+        public static WebconfigOverrides Load()
+        {
+            var result = new WebconfigOverrides();
+
+            if (ConfigurationManager.ConnectionStrings[ConnectionStringDefaultConnectionString] != null)
```
mconverti: Yes, when running in an Azure App Service environment, the configuration settings specified in the Portal (through ARM) are available as environment variables at runtime. For each app setting, two environment variables are created; one with the name specified by the app setting entry, and another with a prefix of APPSETTING_. Both contain the same value.

For .NET apps in particular, these settings are also injected into the .NET configuration AppSettings/ConnectionString at runtime, overriding existing settings. That's why we can transparently use the ConfigurationManager class to read the settings.

src/Orchard.Web/Modules/Orchard.Setup/Services/WebconfigOverrides.cs
mconverti: @sebastienros how can we get the tenant name? Should we inject any dependency to be able to get this value?

The problem with the tenantName prefix approach is that it will always fallback to the setting without the tenant name, because the current ARM template is only configuring the "defaultConnection" connection string.

Perhaps a better approach would be to skip the default connection string configuration is tenantName != "Default". I think this should be also performed for the Media Services module default configuration as well.

@sebastienros We will wait for your input on this feedback. Thanks!

mconverti:
@sebastienros regarding the AMS recipe, we are not familiar with the Orchard CMS build process and we currently ran out of cycles to apply this improvements. Perhaps you can handle this change with your team.

Thanks!

src/Orchard.Web/Modules/Orchard.Azure.MediaServices/Views/Parts/CloudVideo.cshtml
```diff
@@ -1,9 +1,21 @@
 ﻿@using Orchard.ContentManagement;
 
+@{
+    string alternateText = string.Empty;
+    try
```
mconverti: This is not a nullability issue. The original code was failing when embedding a cloud media video item in a page because of a binder issue with the As<> extension method.

This is a quick workaround for the existing error in the code. I think that doing a IContent cast before calling the As<> extension method will fix it; I will push the changes later.

BenedekFarkas:
We can close this (and other Azure Media Services) PR and issue due to #8759.

PR #7621, opened 2017-03-23
Redo "Improved performance of Count Hql queries"

jersiovic:
This PR redos the performance improvement reverted on this PR https://github.com/OrchardCMS/Orchard/pull/7563
It is an alternative way of improving performance of Count Hql query compatible with SQLCE.
I have had to use a nested sql to avoid SQLCE limitation. Problem is those are not supported by NHibernate HQL so I translated main HQLquery  to SQL and then used it as nested sql of the general sql doing a count. Problem with this query is cannot be cached by NHibernate second level cache, so I removed SetCacheable line. Now performance is better than with reverted code, so I would prefer to use this version than previous one despite now the query is not cached by NHibernate. I submit this PR just in case it is interesting for you merging it. @sebastienros @paynecrl97


sebastienros:
@jersiovic have you tried to repro this issue with and without this PR ? https://github.com/OrchardCMS/Orchard/issues/7554

jersiovic:
@sebastienros Yes, in fact I repro the issue with SQLCE using my original code for improving performance then I modified code till I found this way of improving performance but avoiding the problem with SQLCE..

src/Orchard/ContentManagement/DefaultHqlQuery.cs
```diff
@@ -207,12 +207,18 @@ public class DefaultHqlQuery : IHqlQuery {
...
+            var sql = ToSql(true);
+            sql = "SELECT count(*) as totalCount from (" + sql + ") t";
+            return Convert.ToInt32(_session.CreateSQLQuery(sql)
+                    .AddScalar("totalCount", NHibernateUtil.Int32)
+                    .UniqueResult());
+        }
+
+        public string ToSql(bool count) {
+            var sessionImp = (ISessionImplementor)_session;
+            var translatorFactory = new ASTQueryTranslatorFactory();
```
sebastienros: To you need to have a new factory on every call:
- can it be per DefaultHqlQuery instance
- can it be static

src/Orchard/ContentManagement/DefaultHqlQuery.cs
```diff
@@ -207,12 +207,18 @@ public class DefaultHqlQuery : IHqlQuery {
...
         public int Count() {
             ApplyHqlVersionOptionsRestrictions(_versionOptions);
-            var hql = ToHql(true);
-            hql = "select count(Id) from Orchard.ContentManagement.Records.ContentItemVersionRecord where Id in ( " + hql + " )";
-            return Convert.ToInt32(_session.CreateQuery(hql)
-                           .SetCacheable(true)
-                           .UniqueResult())
-                ;
+            var sql = ToSql(true);
+            sql = "SELECT count(*) as totalCount from (" + sql + ") t";
```
sebastienros: Have you tried `count(*) with MySql or Postgres?

jersiovic: I've just checked now this code works properly with MySql and Postgres executing a projection query with pagination.

src/Orchard/ContentManagement/DefaultHqlQuery.cs
sebastienros: Ping

jersiovic: ah ok I didn't remember about that

jersiovic: done

src/Orchard/ContentManagement/DefaultHqlQuery.cs
```diff
@@ -5,6 +5,8 @@
 using System.Linq;
 using System.Text;
 using NHibernate;
+using NHibernate.Engine;
+using NHibernate.Hql.Ast.ANTLR;
```
jersiovic: I forgot to add those changes originally copy pasting changes through GitHub web UI for this PR. For last commit I used a regular Git client so, on compiling I detected this PR was broken because of that.

src/Orchard/ContentManagement/DefaultHqlQuery.cs
```diff
@@ -216,8 +220,7 @@ public class DefaultHqlQuery : IHqlQuery {
 
         public string ToSql(bool count) {
             var sessionImp = (ISessionImplementor)_session;
-            var translatorFactory = new ASTQueryTranslatorFactory();
-            var translators = translatorFactory.CreateQueryTranslators(ToHql(count), null, false, sessionImp.EnabledFilters, sessionImp.Factory);
+            var translators = TranslatorFactory.CreateQueryTranslators(ToHql(count), null, false, sessionImp.EnabledFilters, sessionImp.Factory);
```
sebastienros: I checked here: https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Hql/Ast/ANTLR/ASTQueryTranslatorFactory.cs#L27

And it seems you can use the static method directly, no need to instantiate anything.

jersiovic: Well, the method you pointed is static but also is private so I cannot use it.

Issue #7630, opened 2017-03-29
Localization Extension: js bug on Content Picker Field


GiuseppeMusso-Laser:
Hi all.
I noticed a js bug in Content Picker Field when you enable feature Orchard.ContentPicker.LocalizationExtensions.
When a content picker field is hidden by a Placement.info file (in a custom module or an alternate, for instance), you get a javascript error because a test on existance of the content picker field in the web page DOM is missing.
I'll post a solution soon.

PR #7631, opened 2017-03-29
Fix on Content Picker Field rendering

GiuseppeMusso-Laser:
Fix on Content Picker Field rendering when feature Orchard.ContentPicker.LocalizationExtensions is enabled, as described in issue #7630.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentPicker/Views/EditorTemplates/Fields/ContentPickerLocalization.Edit.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.ContentPicker/Views/EditorTemplates/Fields/ContentPickerLocalization.Edit.cshtml
@@ -17,13 +17,17 @@
 
     //We will use a script to find the fieldset for the field we are currently processing.
     //The fieldset contains a span of class "hint". We will add tryTranslateMsg to it.
+    //We will check if the field exists in the page DOM, just in case a Placement.info suppresses it.
     string dataPartName = HttpUtility.JavaScriptStringEncode(Model.Part.PartDefinition.Name);
     string dataFieldName = HttpUtility.JavaScriptStringEncode(Model.Field.PartFieldDefinition.Name);
 }
 @using (Script.Foot()) {
     <script type="text/javascript">
         $(function () {
-            $("fieldset[data-part-name='@dataPartName'][data-field-name='@dataFieldName']").find("span.hint")[0].innerText += "@tryTranslateMsg";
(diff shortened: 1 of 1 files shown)


sebastienros:
The code you are modifying has an issue. The localized translations should not be passed directly in JavaScript form because you will have two issues:
- The localized text will be double-encoded ( `.Text` is called then `@`)
- If the localization contains a `"` then it will break the javascript

The solution to read localized values in JS from a view is to put the value in a tag like this:
`<input type="hidden" id="deleteMessage" value="@T["Are you sure you want to delete this folder?"]" />`

then read it using `document.getElementById('deleteMessage')`

can you please change it on this view?

GiuseppeMusso-Laser:
I modified the code in order to avoid double-encoding and I put the value in a hidden field as you suggested.
Then I realized it is all fine till the translation does not contain a quotation mark (eg.: "Selected \"items\" with ...").
In this case, the quotation mark breaks the value.
To avoid this problem I got the translated string using T("...").Text.Replace("\"", "&quot;") and put it in html tag (hidden field value) using Html.Raw().
Is there a better way to deal with this problem?
I'm also aware quotation mark is the problem because hidden field value is enclosed in quotation marks.
I could have a similar problem with single quotes if my value was enclosed in single quotes.

sebastienros:
Not right.

if you just insert the translation like this:

`<input type="hidden" value="@T["BLAH"]" />` then the quote will be encoded in HTML.

Don't call `.Text` and you won't have any issue.

GiuseppeMusso-Laser:
Sorry, I think I'm misunderstanding something.
If I write in cshtml view:

`<input type="hidden" id="tryTranslateMsg" value="@T("One \"two\" three.")" />`

I get in the page source in my browser:

`<input type="hidden" id="tryTranslateMsg" value="One "two" three." />`


so js code:

`hint.innerText = $("#tryTranslateMsg").val();`

sets hint value to "One " because double quotes breaks hidden field value.

It looks like @T("...") does not HTML-encode double quotes.
Could you please help me to understand where I'm wrong?

sebastienros:
I will investigate, looks like an issue in how the text is encoded.

sebastienros:
Found the solution, you need to use `.Text` as you were doing, but in the `input` tag.

```
<input type="hidden" value="@T("\"Foo\"").Text"/>
```

GiuseppeMusso-Laser:
Great, now it works fine for me.
Thank you, Sébastien!

PR #7635, opened 2017-03-30
add comments to audit module

infofromca:


src/Orchard.Web/Modules/Orchard.AuditTrail/Models/AuditTrailEventRecord.cs
```diff
@@ -2,21 +2,64 @@
...
 
 namespace Orchard.AuditTrail.Models {
+    /// <summary>
+    /// Audit Trail Event Record in the database.
+    /// </summary>
     public class AuditTrailEventRecord {
         public virtual int Id { get; set; }
+
+        /// <summary>
+        /// The happend time of the event.
```
DanielStolt: Bad English, rephrase as:
"The time when the event occurred."

src/Orchard.Web/Modules/Orchard.AuditTrail/Models/AuditTrailEventRecord.cs
```diff
@@ -2,21 +2,64 @@
...
+        /// </summary>
         public virtual DateTime CreatedUtc { get; set; }
+
+        /// <summary>
+        /// The user name of the user who made the event..
```
DanielStolt: Better phrased as:
"The user name of the user whose change caused the event to occur."
And remove double periods at the end.

src/Orchard.Web/Modules/Orchard.AuditTrail/Models/AuditTrailEventRecord.cs
```diff
@@ -2,21 +2,64 @@
...
 
+        /// <summary>
+        /// The comment of the event.
+        /// </summary>
+
         [StringLengthMax]
         public virtual string Comment { get; set; }
+
+        /// <summary>
+        /// The ip address of the user who made the event.
```
DanielStolt: Rephrase as:
"The IP adress of the user who caused the event to occur."

src/Orchard.Web/Modules/Orchard.AuditTrail/Models/AuditTrailEventRecordResult.cs
```diff
@@ -1,6 +1,16 @@
 namespace Orchard.AuditTrail.Models {
```
DanielStolt: Apply trailing periods consistently.

src/Orchard.Web/Modules/Orchard.AuditTrail/Services/IEventDataSerializer.cs
```diff
@@ -1,8 +1,22 @@
 using System.Collections.Generic;
 
 namespace Orchard.AuditTrail.Services {
+    /// <summary>
+    /// Serialize or deserialize event data.
```
DanielStolt: Dependencies should be described with more context IMO. Perhaps better phrased as something like:
"A service responsible for serializing and deserializing audit trail event data."

DanielStolt:
I like the initiative of providing XML comments!

I've written a few comments - if you could go over the PR and make it a bit more consistent with regards to punctuation etc, that would be truly great!

src/Orchard.Web/Modules/Orchard.AuditTrail/Models/AuditTrailEventRecordResult.cs
infofromca: applied

DanielStolt:
Nice work, thanks.

PR #7642, opened 2017-04-04
Allow Feature Prefix override in TableDbName

acherney:
Schema builder can be used with tables from more than one feature.  In this case, it assumes the "current" feature prefix for all tables, which will be incorrect for some tables.  Allow a manual override of the feature prefix in these cases.  

For example:
public int UpdateFrom18()
        {
            string productPartRecordTableName = SchemaBuilder.TableDbName(nameof(ProductPartRecord));
            string contentItemVersionRecordTableName = SchemaBuilder.TableDbName(nameof(ContentItemVersionRecord), "Orchard_Framework");
...
}

### DIFF
--- a/src/Orchard/Data/Migration/Schema/SchemaBuilder.cs
+++ b/src/Orchard/Data/Migration/Schema/SchemaBuilder.cs
@@ -33,8 +33,8 @@ namespace Orchard.Data.Migration.Schema {
         /// <summary>
         /// Translate Table name into database table name - including prefixes.
         /// </summary>
-        public virtual string TableDbName(string srcTable) {
-            return _interpreter.PrefixTableName(String.Concat(FormatPrefix(FeaturePrefix), srcTable));
+        public virtual string TableDbName(string srcTable, string FeaturePrefixOverride = null) {
+            return _interpreter.PrefixTableName(String.Concat(FormatPrefix(FeaturePrefixOverride??FeaturePrefix), srcTable));
         }
 
         /// <summary>
--- a/src/Orchard/Data/Migration/Schema/SchemaBuilder.cs
+++ b/src/Orchard/Data/Migration/Schema/SchemaBuilder.cs
@@ -33,8 +33,8 @@ namespace Orchard.Data.Migration.Schema {
         /// <summary>
         /// Translate Table name into database table name - including prefixes.
         /// </summary>
-        public virtual string TableDbName(string srcTable, string FeaturePrefixOverride = null) {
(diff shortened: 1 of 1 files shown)


sebastienros:
Why don't you create a DataMigartion class or each of the features? You just need to add the [OrchardFeature] attribute to it.
You don't want to create the tables or change them if the feature is not enabled, so your solution could break it.

acherney:
I do have a DataMigration class for my feature, which is where **UpdateFrom18** is located.  Within my Migration I need to know the  _actual_ table name for the ContentItemVersionRecord, which is usually just "Orchard_Framework_ContentItemVersionRecord" but **can** vary in a multi-tenant environment.  Because of this, I was using the SchemaBuilder.TableDbName method for the ContentItemVersionRecord, but all that method does is prefixing MY feature name to the front of the Orchard table name, which seems like erroneous behavior.  I think that the method "TableDbName" is a misnomer.  All it does is prepend the tenant prefix and the feature prefix for the feature that your migration is in - and if the table you need access to is in a framework Orchard table (like ContentItemVersionRecord), the method simply returns the wrong result.  

Maybe we just add comments to the TableDbName method that make it clear it will only function correctly for tables belonging to the feature you are migrating.  To me it is not obvious.

BenedekFarkas:
If I understand correctly, your use-case is that you have multiple features in a module (or separate modules, doesn't matter), let's say `A` and `B` and both have migrations that modify a specific table. You can use `IUpgradeService.GetPrefixedTableName(string tableName)`, but you have to define the table name that includes the feature prefix.
At the same, I'd advise against that if you can avoid, here's why:

1. If A and B have no dependency between them, then you don't have control over which will be executed first. It is possible that in your specific case with that specific DB the migrations will be executed, because you run the migrations as you write them, but if you run a new setup, they can easily be tangled up.
2. If they have dependency between them, then you know which feature's migration will run first, but a new setup can still go wrong if the update steps (across the different features) have a chronological order to them, for example:
2.1. `A` Create method creates `MyTable`.
2.2. `B` Create methods adds `MyColumn` to `MyTable` (you won't be able to do this now without using `GetPrefixedTableName` if there's a tenant prefix).
2.3. `A` UpdateFrom1() adds index to `MyTable.MyColumn`.

So it's a bit dangerous to modify a table outside of the feature that created it.

BenedekFarkas:
@sebastienros what do you think? We can merge this and doesn't break anything, just opens up some possibilities.

PR #7646, opened 2017-04-04
7638: C# 5 compatibility for Orchard.Glimpse

### LINKED ISSUE #7638: Dev: Remove C# 6 code for Dynamic Compilation to work again
Although Dynamic Compilation has other problems on dev.
#7637 should be fixed first and the rest of the code should be modified after 1.10.x/1.10.2 has been merged into dev.
After that, fixing #7640 will make it easier to discover all the changes necessary for this, but it will break the build.

LombiqTechnologies:
Adding LangVersion 5 to Orchard.Glimpse.csproj and removing C# 6 code from the Glimpse module, fixes #7638


BenedekFarkas:
Someone who worked on this module should review it.

BenedekFarkas:
1.11 will ship with C# 7.3 compatibility, so this downgrade is not needed anymore.

PR #7652, opened 2017-04-07
[Fixes #7599] Orchard.Resources - removed file jquery.calendars.min.js from Assets

### LINKED ISSUE #7599: Orchard.Resources - jquery.calendars.min.js in Assets folder instead of being an Gulp artefact
Hi there,

currently i'm reworking some modules to use the [client-side asset pipeline](http://docs.orchardproject.net/en/latest/Documentation/Processing-client-side-assets).

I came across the Orchard.Resources module whereas the `Assets` folder contains a `Js\Calendars\jquery.calendars.min.js`.
Isn't this file supposed to be a Gulp artefact?

With the current configuration a file `Scripts\Calendars\jquery.calendars.min.min.js` is generated.

I think this is a misconfiguration. Should i provide a PR?

ViRuSTriNiTy:
see issue #7599


sebastienros:
Because it's not a bug per se I am waiting for 1.10.2 to be released before pulling it in.

PR #7655, opened 2017-04-10
[Fixes #5345] Orchard.ContentManagement, StorageFilterBase.cs, fixed sanity check in Versioning() and Versioned() causing possible exceptions

### LINKED ISSUE #5345: Exception in StorageFilterBase.cs
Hi there,

i get an exception log in my Orchard 1.9 logs when e.g. editing & saving a user and tracked it down to StorageFilterBase.cs:

```
src\Orchard\ContentManagement\Handlers\StorageFilterBase.cs

namespace Orchard.ContentManagement.Handlers {
    public abstract class StorageFilterBase<TPart> : IContentStorageFilter where TPart : class, IContent {

      ...

        void IContentStorageFilter.Versioning(VersionContentContext context) {
            if (context.ExistingContentItem.Is<TPart>() || context.BuildingContentItem.Is<TPart>())
                Versioning(context, context.ExistingContentItem.As<TPart>(), context.BuildingContentItem.As<TPart>());
        }

        void IContentStorageFilter.Versioned(VersionContentContext context) {
            if (context.ExistingContentItem.Is<TPart>() || context.BuildingContentItem.Is<TPart>())
                Versioned(context, context.ExistingContentItem.As<TPart>(), context.BuildingContentItem.As<TPart>());
        }
```

Here you can see that the **.Is<>() comparisions** are linked via a **logical or**. The expression is imho wrong as it calls **Versioning()** and **Versioned()** when TPart is not welded to _existing content item_ or the _building content item_ resulting in an **null reference exception** in

```
src\Orchard\ContentManagement\Handlers\StorageFilter.cs

namespace Orchard.ContentManagement.Handlers {

  ...

    public class StorageFilter<TRecord> ... {

      ...

        protected override void Versioning(VersionContentContext context, ContentPart<TRecord> existing, ContentPart<TRecord> building) {
            building.Record = existing.Record;
        }
```

due to acessing _building_ or _existing_ without sanity check.

Other handlers overriding the **Versioning()** and **Versioned()** method might be affected too.

I temporarily changed the code in my Orchard sources to 

```
.Is<TPart>() && ... .Is<TPart>
```

It would be nice to have this issue fixed in future releases.

So lonG

ViRuSTriNiTy:
see issue #5345


sebastienros:
Following up on the issue, I need a repro to get to the state that a single item is null. I am worried that the code was done this way intentionally and this code would break another scenario.

BenedekFarkas:
See issue #5345 for discussion.

PR #7656, opened 2017-04-10
modified interface IThemeSelector

zhhbo:
like this   https://github.com/OrchardCMS/Orchard2/pull/680


BenedekFarkas:
@sebastienros the PR with the same concept for Orchard Core seems to be rejected. What do you think about this? I'm not sure if we need this here, but the code actually looks like it can be factored out to an extension, so it's not useless.

sebastienros:
I think we should close this PR, I don't understand what it's doing.

BenedekFarkas:
@zhhbo I'll close the PR then for housekeeping reasons, but we'll be happy to pick up the conversation if you can provide more information on what you want to achieve with these changes.

PR #7665, opened 2017-04-14
Use /lib/nuget/nuget.exe for Package Restore

iamdjones:
If nuget isn't installed, or not on the PATH, deploy will fail with "nuget not recognized..."

### DIFF
--- a/deploy.cmd
+++ b/deploy.cmd
@@ -26,6 +26,10 @@ IF NOT DEFINED DEPLOYMENT_SOURCE (
   SET DEPLOYMENT_SOURCE=%~dp0%.
 )
 
+IF NOT DEFINED NUGET_PATH (
+  SET NUGET_PATH=%~dp0%lib\nuget\nuget.exe
+)
+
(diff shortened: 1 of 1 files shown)


sebastienros:
This is supposed to be a stable environment, so when you say "If nuget isn't installed, or not on the PATH" do you mean that is actually always fails?

Should we do that in 1.10.x instead ?

iamdjones:
No, it doesn't always fail, only when the machine doesn't have nuget installed. This deploy script uses the machines nuget installation, where as you expect it to use the nuget binary found under /lib.

I tried switching to 1.10.x but i might need to toss the PR and start from scratch.

sebastienros:
> only when the machine doesn't have nuget installed

But this script is to be executed on Azure Kudu machines ... what am I missing?

iamdjones:
Ahh, I see now. I was using deploy.cmd to package the application locally. I hadn't realized it was meant for a certain environment.

sebastienros:
For local usage, do `build Precompiled"

PR #7670, opened 2017-04-21
[Fixes #6323] Rename media folder fails with Azure Media Storage

### LINKED ISSUE #6323: Media FolderNames updated incorrectly when Edited
**To replicate:**
1. Enable Azure Media and configure for use.
2. In Media Library, create a folder structure with nested folder.
3. Select the most nested folder
4. Import an image into the most nested folder. Close.
5. Click Edit and rename the most nested folder to something else
6. Save
7. Click the most nested folder to display the images

==> No images appear. The image you uploaded should be displayed and it is not. 

**The problem:**

Orchard.MediaLibrary.Services.RenameFolder() uses '\' as a separator in the new FolderName. But folders are created with whatever separator the StorageProvider uses, and BuildGetMediaContentItemsQuery() doesn't normalize folder names, so it doesn't find MediaParts that have been renamed with mismatched separators.

**Fix:**

Change the test for Folder Names in BuildGetMediaContentItemsQuery() as follows:

```
      // There is a bug that allows FolderNames to be stored with either '\' or '/' separators
      // so test for both variations
        if (!String.IsNullOrEmpty(folderPath)) {
            folderPath = folderPath.Replace(@"\","/");
            var folderPathAlt = folderPath.Replace("/", @"\" );
            if (recursive) {
                query = query.Join<MediaPartRecord>().Where(m => m.FolderPath.StartsWith(folderPath) || m.FolderPath.StartsWith(folderPathAlt));
            }
            else {
                query = query.Join<MediaPartRecord>().Where(m => m.FolderPath == folderPath || m.FolderPath == folderPathAlt);
            }
        }
```

TFleury:


src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/MediaLibraryService.cs
```diff
@@ -328,15 +328,16 @@ public class MediaLibraryService : IMediaLibraryService {
             Argument.ThrowIfNullOrEmpty(newFolderName, "newFolderName");
 
             try {
-                var segments = folderPath.Split(new char[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar }, StringSplitOptions.RemoveEmptyEntries).ToArray();
-                var newFolderPath = String.Join(Path.DirectorySeparatorChar.ToString(), segments.Take(segments.Length - 1).Union(new[] { newFolderName }));
+                var parentIndex = folderPath.LastIndexOfAny(new char[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar });
```
sebastienros, line 331: Is this change only to make the code simpler?

src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/MediaLibraryService.cs
```diff
@@ -328,15 +328,16 @@ public class MediaLibraryService : IMediaLibraryService {
...
+                var parentPath = parentIndex > 0 ? folderPath.Substring(0, parentIndex) : String.Empty;
+                var newFolderPath = _storageProvider.Combine(parentPath, newFolderName);
 
                 var mediaParts = BuildGetMediaContentItemsQuery(_orchardServices.ContentManager, folderPath, true).List();
                 foreach (var mediaPart in mediaParts) {
                     mediaPart.FolderPath = newFolderPath + mediaPart.FolderPath.Substring(folderPath.Length);
                 }
 
-                _storageProvider.RenameFolder(folderPath, _storageProvider.Combine(Path.GetDirectoryName(folderPath), newFolderName));
+                _storageProvider.RenameFolder(folderPath, _storageProvider.Combine(parentPath, newFolderName));
```
sebastienros: Shouldn't it use `newFolderPath` ?

src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/AzureFileSystem.cs
```diff
@@ -239,7 +239,8 @@ public class AzureFileSystem {
                 }
 
                 if (blob is CloudBlobDirectory) {
-                    string foldername = blob.Uri.Segments.Last();
+                    var blobDir = (CloudBlobDirectory)blob;
+                    string foldername = blobDir.Prefix.Substring(blobDir.Parent.Prefix.Length);
```
sebastienros, line 243: Can `Parent` be null or what happens on the root folder?

src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/MediaLibraryService.cs
TFleury: No, String.Join with Path.DirectorySeparatorChar replaces "/" with "\\" in folder path. It is the primary cause of the issue.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/MediaLibraryService.cs
TFleury: That's right.

src/Orchard.Web/Modules/Orchard.Azure/Services/FileSystems/AzureFileSystem.cs
TFleury: There is always at least tenant's folder, or "Default" folder without multitenancy.

PR #7685, opened 2017-05-05
Log error when Recipient emailMessage.Recipients is null

jardg:
When emailMessage.Recipients is null throws so the error is being logged by Orchard.Exceptions instead by the logger in this condition.
Repro steps:
Create a workflow that sends emails when a custom type has been created. 
Attach an email activity when this type has been created and dont fill it.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
+++ b/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
@@ -57,7 +57,7 @@ namespace Orchard.Email.Services {
                 Attachments = (IEnumerable<string>)(parameters.ContainsKey("Attachments") ? parameters["Attachments"] : new List<string>())
             };
 
-            if (emailMessage.Recipients.Length == 0) {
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
```diff
@@ -57,7 +57,7 @@ public class SmtpMessageChannel : Component, ISmtpChannel, IDisposable {
                 Attachments = (IEnumerable<string>)(parameters.ContainsKey("Attachments") ? parameters["Attachments"] : new List<string>())
             };
 
-            if (emailMessage.Recipients.Length == 0) {
+            if (string.IsNullOrWhiteSpace(emailMessage.Recipients) || emailMessage.Recipients.Length == 0) {
```
sebastienros: You can remove the other condition which has no effect with your change.

jardg: Done

PR #7687, opened 2017-05-08
Update CreateDescribeContext to query latest content

### LINKED ISSUE #7686: Element.OnDisplaying context requires the content item to be publish in order for context.Content to be populated
I'm running into an issue when creating a new element, where I want to reference the owning content item in my element display, both on the front-end and in the admin design. I'm finding that my the `ElementDisplayingContext.Content` is populated with the correct `IContent` when I'm viewing my published content on the website. However, prior to publishing the content for the first time, the `ElementDisplayingContext.Content` is set to a new `IContent`, not my current content item, so my design template isn't populating correctly.

To describe the issue a little better, here are some sample reproduction steps to illustrate the issue:

1. Create a new Page content item
2. Set a breakpoint in `HeadingElementDriver.OnDisplaying()`
3. Add a Heading element to the Layout, set the heading as 'h1' and save the editor
4. The breakpoint gets hit and `context.Content` is some new content (Id = 0). _This is expected, since our page hasn't been created yet._
5. Continue through the breakpoint
6. Set the title and save the page as a draft
7. The breakpoint gets hit again.  This time, the `context.Content` is correctly set.
8. Add a new heading element to the layout, set as 'h2' and save the editor.
9. The breakpoint gets hit yet again. I'm expecting my `context.Context` to be correctly set to the same content ID as in step 7, but that's not the case. Instead my `context.Content` is some new content yet again (Id=0).
10. Now, publish the content item and continue through the breakpoints.
11. Add yet another heading element, set as 'h3' and save the editor.
12. The breakpoint is hit and this time the ID is correctly set to the ID of the content that's being edited.

You'll notice that step 7, when OnDisplaying is hit after a save, has the content populated correctly. I suspect this is due to the event stemming from the default admin edit action being called (`Orchard.Core.Contents.Controllers.Edit`), which queries the content manager with `VersionOptions.Latest`.

On the other hand, when building the content display from `ElementController.Edit()` or `ElementController.Update()`, the `ElementController.CreateDescribeContext()` method is called, which queries the content manager without specifying any `VersionOptions`, so the `DefaultContentManager` defaults to `VersionOptions.Published`.

I think the solution is to change `ElementController.CreateDescribeContext()` to query the content manager with `VersionOptions.Latest`, matching the query in `Orchard.Core.Contents.Controllers.Edit`.

I'll submit a pull request shortly.

mroskamp:
Fixes #7686 , where the ElementDriver.OnDisplaying ElementDisplayingContext.Content not being populated properly when the content item is saved, but not yet published. Instead of query the content only for the published version, use the latest version instead.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Layouts/Controllers/ElementController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Layouts/Controllers/ElementController.cs
@@ -151,7 +151,7 @@ namespace Orchard.Layouts.Controllers {
             if (contentId == null && contentType == null)
                 return DescribeElementsContext.Empty;
 
-            var part = contentId != null && contentId != 0 ? _contentManager.Get<ILayoutAspect>(contentId.Value)
+            var part = contentId != null && contentId != 0 ? _contentManager.Get<ILayoutAspect>(contentId.Value, VersionOptions.Latest)
                 ?? _contentManager.New<ILayoutAspect>(contentType)
                 : _contentManager.New<ILayoutAspect>(contentType);
 


sebastienros:
@sfmskywalker I won't merge this PR unless you think the fix is correct. I can't assess the implications of this change. Or it will go in dev.

BenedekFarkas:
@sfmskywalker can you take a look at this, please? I think I agree with the change, but we need more information on the potential side-effects.

sfmskywalker:
Totally missed this. On its own it look fine, but I have to remind myself what all of this does in the first place 🤭 
Will check it out and let you know.

sfmskywalker:
Yeah I think this change makes sense too. The content item being retrieved represents the content item containing the LayoutPart, and it's used during edit/update of this page. It's not used for displaying it anywhere on the front-end, so it seems safe in any case.

BenedekFarkas:
Thanks Sipke!

sfmskywalker (approved):

PR #7691, opened 2017-05-10
issue-7690

JohnStov:
Don't allow Redirect to forcibly terminate page processing
See https://msdn.microsoft.com/en-us/library/t9dwyts4(v=vs.110).aspx and https://msdn.microsoft.com/en-us/library/a8wa7sdt(v=vs.110).aspx
Redirect(url) forces a thread abort which forcibly ends page processing.
This in turn causes NHibernate to rollback any pending transactions.
Redirect(url, false) allows page processing to finish before returning the redirect response


sebastienros:
Closing as you said it was not solving the issue.

Issue #7055, opened 2016-07-25
Terms output in a strange order when paging in the Terms admin interface


pilotbean:
Terms output in a strange order when paging in the Terms admin interface.

I believe this is because sort is applied after paging is applied.

https://github.com/OrchardCMS/Orchard/commit/356e8230ab2c4704ab300fb6915606137873bf22

pilotbean:
Replacing line 45 in TermAdminController.cs with the following code fixes the issue:

```
            var terms = TermPart.Sort(_taxonomyService.GetTermsQuery(taxonomyId).ForPart<TermPart>().List());

            if (pager.PageSize > 0)
            {
                terms = terms.Skip(pager.GetStartIndex()).Take(pager.PageSize);
            }
```

This applies the paging after the sort.

sebastienros:
Your change is just adding another issue, as it would load all the terms in memory before sorting and paging it.
However we agree there is an issue here, first the paging is not based on an order, so loading different pages doesn't mean we'll get the correct page, and also different hierarchies can be mixed in the same page.

The solution is not simple, but there is one. We need to change the FullPath property to have its components padded. Instead of have "1/2/3" we'd have "00000001/00000002/00000003" this way it is lexicographically orderable in database.

1- Change the code to store it this way.
2- Add an index to this field as we'll sort it in database, and also because we filter by prefix to load children of a specific term. Add the sort order in the paged query of the terms.
3- Make TermsComparer obsolete so that we can force users to use the database sorting.
4- Create an upgrade step in the Update module to update all existing terms.

jersiovic:
Wouldn't it be better to add a new int property to taxonomy termpart called "Level".
Its value will be the number of '/' in FullPath. In migrations we should set it for existing terms.
Then the order will be obtained ordering by Level and Weight.

This is how we have solved this issue recently. Furthermore it helped us to add methods that get child elements of a node limiting the number of levels returned in an efficient way.
Don't you think it is more elegant and efficient than using padding?

jersiovic:
Problem of that solution is ignores the weight of the terms. Other more complex solution would be to have a FullWeigthPath Property that stores weights of parents: instead of its id: "2/5/2".

The extra complexity would be caused because after update the weight of a term we will need to visit all their children updating the FullWeigthPath to update the related value.

sebastienros:
[...]

PR #7709, opened 2017-05-23
#7055 changes taxonomy weight

jardg:
This purposed changes add a new property with the full weight of a term(FullWeight) which is made using his weight and id and his parent's weight and id. This way we can lexicographically sort the terms via this property using simple sql and avoid algorithmic logic to sort them. 

The terms update his FullWeight property whenever they are updated in the controller or the driver(including his children terms).Also they can be updated to use this new functionality importing all of them in the Import method using an 'old' recipe or migrating them in the Orchard.Upgrade module if they exist already.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TermAdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TermAdminController.cs
@@ -49,7 +49,7 @@ namespace Orchard.Taxonomies.Controllers {
             var allTerms = _taxonomyService.GetTermsQuery(taxonomyId).OrderBy(x => x.FullWeight);
             var termsPage = pager.PageSize > 0 ? allTerms.Slice(pager.GetStartIndex(), pager.PageSize) : allTerms.Slice(0, 0);
 
-            var pagerShape = Shape.Pager(pager).TotalItemCount(termsPage.Count());
+            var pagerShape = Shape.Pager(pager).TotalItemCount(allTerms.Count());
 
             var entries = termsPage
                     .Select(term => term.CreateTermEntry())
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TermAdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TermAdminController.cs
@@ -46,11 +46,10 @@ namespace Orchard.Taxonomies.Controllers {
 
             var taxonomy = _taxonomyService.GetTaxonomy(taxonomyId);
 
-            var allTerms = TermPart.Sort(_taxonomyService.GetTermsQuery(taxonomyId).List());
+            var allTerms = _taxonomyService.GetTermsQuery(taxonomyId).OrderBy(x => x.FullWeight);
+            var termsPage = pager.PageSize > 0 ? allTerms.Slice(pager.GetStartIndex(), pager.PageSize) : allTerms.Slice(0, 0);
 
-            var termsPage = pager.PageSize > 0 ? allTerms.Skip(pager.GetStartIndex()).Take(pager.PageSize) : allTerms;
-
-            var pagerShape = Shape.Pager(pager).TotalItemCount(allTerms.Count());
+            var pagerShape = Shape.Pager(pager).TotalItemCount(termsPage.Count());
 
             var entries = termsPage
                     .Select(term => term.CreateTermEntry())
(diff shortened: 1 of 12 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TermPartDriver.cs
```diff
@@ -93,7 +94,22 @@ public class TermPartDriver : ContentPartDriver<TermPart> {
         }
 
         protected override DriverResult Editor(TermPart termPart, IUpdateModel updater, dynamic shapeHelper) {
-            //Moved to TermPartHandler to work also with Taxonomy Localization feature
+            updater.TryUpdateModel(termPart, Prefix, null, null);
```
jardg, line 97: The driver wasnt updating the term on dev branch

src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/TermAdminController.cs
```diff
@@ -90,7 +89,9 @@ public class TermAdminController : Controller, IUpdateModel {
 
                     foreach (var entry in checkedEntries) {
                         var term = _taxonomyService.GetTerm(entry.Id);
-                        _taxonomyService.DeleteTerm(term);
+                        if (term != null) {
+                            _taxonomyService.DeleteTerm(term);
```
jardg, line 93: When I deleted all the terms on the admin panel via UI got an exception because some of them were null at this point since I deleted the parent first so it was already removed and term was null.

sebastienros:
And there is an update task, that's a great effort, thank you.

Can you paste some example values of FullWeight to understand simply how it works?

jardg:
This is an example of what you get ordering by FullWeight in sql. 
Edit: @sebastienros  . Have you seen the examples?

orderedterms

src/Orchard.Web/Modules/Orchard.Taxonomies/Models/TermPart.cs
```diff
@@ -51,6 +51,11 @@ public class TermPart : ContentPart<TermPartRecord> {
             set { Store(x => x.Weight, value); }
         }
 
+        public string FullWeight {
```
sebastienros, line 54: Because the value is just for querying and  we don't need it when loading a content item, shouldn't we ignore this property in the Infoset? This will prevent to have to keep them both in sync.

sebastienros:
I totally forgot how weights affect the terms order in the hierarchy, can you refresh my mind?

src/Orchard.Web/Modules/Orchard.Taxonomies/Models/TermPart.cs
jardg: Changed. The way to do this is just setting and storing the property on the Record so Retrieve and Store doesnt save the property in the infoset, am I right?

jardg:
Weights are just used to order terms having same level(path) in the sort method. Using fullweight its the same if they have the same path they are ordered from higher weight to lower.

sebastienros:
@jardg Can you sign the CLA ?

@jersiovic @urbanit would you mind testing the change? I assume it will make taxonomies faster.

jersiovic:
@sebastienros I've been designing this feature with @jardg we work side by side on the same room at XKProject/Pccom. On next week it is expected we will test it on preproduction server.

jersiovic:
@sebastienros I've been designing this feature with @jardg we work side by side on the same room at XKProject/Pccom. On next week it is expected we will test it on preproduction server.

urbanit:
@sebastienros I will try it during weekend hopefully...

sebastienros:
Just waiting for @urbanit then we can merge

urbanit:
Status update: I am testing it now in a multilanguage enviroment with 3 levels and some ci related...

urbanit:
Hi, test was completed.. As said, in a multilingual enviroment. Works nice, except from a bug that probably is not due to this PR but on localization [Quick Bug Description: cannot translate a 2nd level (child) term].
Reordering works smoothly.

Performance wise I cannot be precise, @jardg  & @jersiovic may tested the changes more thoroughly.

jersiovic:
Improvements in family weight basically allow you to use pagination when you render families cause till now it was not a good option because to get the absolute order needed for pagination you need to load all the terms in the taxonomy for returning  each page, so it had no sense cause that had poor performance than not to paginate. Now with this feature you can paginate the rendering of a taxonomy tree using pagination with better performance cause you only load what you are going to render.
When you don't paginate a minor improvement that we have not meassured how much it improves is that you don't need to order the terms in memory after loading them cause actually they come ordered from db.

urbanit:
Seems logical. Thank you for clarification @jersiovic .

Issue #6978, opened 2016-06-16
Allow extending content queries


ViRuSTriNiTy:
I want to extend the content queries executed by `DefaultContentManager` and `DefaultContentQuery` with a left join of my ValidityPart to return only "valid" content items.

In case of  `DefaultContentManager` i can inherit from it and override `Get(int aId, Orchard.ContentManagement.VersionOptions aOptions, Orchard.ContentManagement.QueryHints aHints)`, this works without a hassle.

But in the case of `DefaultContentQuery` i cannot inherit from it because the methods of interest `Slice()` and `Count()` are private as almost everything in this class. Last resort would be to implement IContentQuery myself but this is overkill and results in basically copying `DefaultContentQuery.cs`.

Therefore i added a protected `BeforeExecuteQuery`in `DefaultContentQuery.cs` to be able to extend the query before it is executed:

```
namespace Orchard.ContentManagement {
    public class DefaultContentQuery : IContentQuery {
        ...

        protected virtual void BeforeExecuteQuery(ICriteria contentItemVersionCriteria) {
        }

        private IEnumerable<ContentItem> Slice(int skip, int count) {
            ...

            BeforeExecuteQuery(criteria);

            return criteria....;
        }

        int Count() {
            ...

            BeforeExecuteQuery(criteria);

            return criteria...;
        }
    }
}
```

My `ContentQuery` class looks like this:

```
[Orchard.Environment.Extensions.OrchardSuppressDependency("Orchard.ContentManagement.DefaultContentQuery")]
public class ContentQuery : Orchard.ContentManagement.DefaultContentQuery
{ 
    protected override void BeforeExecuteQuery(NHibernate.ICriteria aContentItemVersionCriteria)
    {
        base.BeforeExecuteQuery(aContentItemVersionCriteria);

        // here comes my code to extend the query by adding additional criteria to aContentItemVersionCriteria
        }
    }
}
```

Do you consider this a valid approach and is it possible to include this change in a future Orchard release?

I also thought about using a handler pattern here as it is done almost everywhere in Orchard but this would required the need to resolve all available handlers each time a query is executed which might affect performance negatively.

HermesSbicego-Laser:
Same needs here, @sebastienros  could you suggest a good approach and from where to begin?

Xceno:
[...]

PR #7711, opened 2017-05-24
Issue/6978 extending content queries

GiuseppeMusso-Laser:
A possible solution to issue #6978 without suppressing dependencies.


sebastienros:
You need to rebase your changes on top of dev because we are seeing commits from the past in this one.

PR #7716, opened 2017-05-26
Fix for displaying aside zones

hmG3:
When using additional aside first or second zone there is an bug in default theme when displaying content zone after adding support for rtl layout.
image


sebastienros:
Should we apply it in 1.10.x too?

hmG3:
Yes

Issue #7713, opened 2017-05-25
Feature request: Support static Razor compilation


jimasp:
Is it possible to perform view compilation in Orchard?

Using the off-the-shelf latest dev branch, and setting `MvcBuildViews` to `true` results in all kinds of reference issues.

As an alternative, I tried doing this:

1. `build Precompiled` - this works fine, no problem
2. `%windir%\Microsoft.NET\Framework64\v4.0.30319\aspnet_compiler -f -v / -p c:\_src\orchard\build\Precompiled c:\temp\aspnetcompilerout`

But (2) fails, with the following error:

`c:\Windows\Microsoft.NET\Framework64\v4.0.30319\Temporary ASP.NET Files\root\fd596974\9555b8ff\App_Web_4psv3d0u.0.cs(34): error CS0246: The type or namespace name 'Lucene' could not be found (are you missing a using directive or an assembly reference?)`

Am I pointing the tool at the wrong folder? Is it possible to create a folder that would work from the targets?
Or is it just not possible to compile views in orchard without major divergence from the code base?

This guy seemed to manage it, and I think? i'm doing the same steps, although I don't have a "Web" folder, so his post is most likely out of date:
https://blogs.msdn.microsoft.com/shacorn/2014/11/13/asp-net-precompilation-speeding-up-orchard-cold-start/

(The reason i am doing this is not for performance, but because there has been the occasional view compilation problem, which has not shown up until well into the testing phase).

sebastienros:
The web folder that is mentioned is the one where the application is deployed. It's up to you to name it `web` too or replace by the name you are using.

jimasp:
yeah, i just misread that.

back to the question:
Is there any way to compile views? (without messing with lots of references). 

mvcbuildviews doesn't work due to reference issues.
aspnet_compiler doesn't work with the build\Precompiled output directory 

The aspnet_compiler failure seems really odd to me, as given that this is the folder that gets deployed, surely it must implicitly be compileable by the aspnet compiler on the web server, or the website wouldn't work?

jimasp:
On further investigation, there are 2 problems:
A. a couple of minor view/ reference issues (**fixed in attached PR**)
B. the requirement that the aspnet compiler requires local copies of referenced dlls.

Publishing from Visual Studio works, but we cannot do this on a build machine.
Attempting to publish from the command line with the same publish properties file seems to fail:

e.g.
```
[...]

PR #7719, opened 2017-05-28
Fixing references / view file due to view compilation errors

jimasp:
see: https://github.com/OrchardCMS/Orchard/issues/7713

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Dashboards/Web.config
+++ b/src/Orchard.Web/Modules/Orchard.Dashboards/Web.config
@@ -7,7 +7,7 @@
         </sectionGroup>
     </configSections>
     <system.web.webPages.razor>
-        <host factoryType="System.Web.Mvc.MvcWebRazorHostFactory, System.Web.Mvc, Version=5.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
+        <host factoryType="System.Web.Mvc.MvcWebRazorHostFactory, System.Web.Mvc, Version=5.2.3, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
         <pages pageBaseType="Orchard.Mvc.ViewEngines.Razor.WebViewPage">
             <namespaces>
                 <add namespace="System.Web.Mvc" />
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.Dashboards/Web.config
```diff
@@ -7,7 +7,7 @@
         </sectionGroup>
     </configSections>
     <system.web.webPages.razor>
-        <host factoryType="System.Web.Mvc.MvcWebRazorHostFactory, System.Web.Mvc, Version=5.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
+        <host factoryType="System.Web.Mvc.MvcWebRazorHostFactory, System.Web.Mvc, Version=5.2.3, Culture=neutral, PublicKeyToken=31bf3856ad364e35" />
```
sebastienros, line 10: I assume the specflow web.config file also needs to change

src/Orchard.Web/Modules/Orchard.Templates/Views/DefinitionTemplates/Parts.Shape.cshtml
```diff
@@ -14,13 +14,6 @@
 
 <fieldset>
     <div>
-        @Html.LabelFor(m => m.Name, T("Name"))
-        @Html.TextBoxFor(m => m.Name, new { @class = "text medium" })
-        <span class="hint">@T("The name of this template, which can be used to reference from code (as when using shape type names) and tokens.")</span>
-    </div>
-</fieldset>
-<fieldset>
-    <div>
```
BenedekFarkas, line 23: I don't think removing this is correct, although I also see build errors with view compilation from this piece of code - I'll look into it.

src/Orchard.Web/Modules/Orchard.Dashboards/Web.config
BenedekFarkas: @sebastienros I think that was also discovered when there was a problem with running the site with IIS Expess, but I don't remember now where we got with that.

src/Orchard.Web/Modules/Orchard.Templates/Views/DefinitionTemplates/Parts.Shape.cshtml
BenedekFarkas: Oh, actually this needs to be removed as Template has TitlePart!

sebastienros:
Should we merge this one or wait for more changes from you @BenedekFarkas ?

BenedekFarkas:
I think we can merge this, as it definitely brings us closer to the goal, but let's not close the issue yet.
I'll open another PR with my changes after this one is merged, then we can discuss how to proceed (judging from what @jimasp commented on the issue, he's testing from VS, but I think validating with MSBuild is also necessary, see my comments on the issue about that).

Issue #7724, opened 2017-06-01
TaxonomyExtensionsService does not check for duplicates in RegenerateAutoroute.


MarcoViglione-Laser:
This is a problem related to the TaxonomyExtensionService that's currently in the dev branch.
The RegenerateAutoroute method does not call the ProcessPath method of the AutorouteService, which means that potentially it can create a permalink conflict.
I'm working on a fix right now.

PR #7725, opened 2017-06-01
Added check for duplicate path

MarcoViglione-Laser:


MarcoViglione-Laser:
This solves issue #7724

Issue #7729, opened 2017-06-16
Text tokens enhancements


HermesSbicego-Laser:
2 possible enhancements there:
1. Not all "Text" Tokens are "chainable" so some cannot be chained eachother:
e.g. Limit: and Format: canot be chained
2. We should insert a TrimStart Token

We are working on a PR just now.

PR #7730, opened 2017-06-16
[#7729] Text tokens enhancements

HermesSbicego-Laser:
- Adds the "TrimStart:" text token
- Adds the ability to chain text to "Limit:", "TrimEnd:", "Format:" and "TrimStart:" tokens


sebastienros:
No tests were altered, just new ones, sounds good.

And looking at this code I am happy we have chosen to use a template library in OC instead.

HermesSbicego-Laser:
Yes, chaining tokens with paramters are not so easy to write, neither intuitive. Wich template library has been adopted in OC? Could be adopted in O1 too?

PR #7732, opened 2017-06-16
#7713: View compilation with MSBuild

BenedekFarkas:
A fix for #7713 that includes the same changes as #7719 by @jimasp plus a few more stuff:

- Modified the `<Private>` property (aka `Copy Local`) for the references to `Orchard.Framework` and `Orchard.Core` in each project file to have the value `$(MvcBuildViews)`. This ensures that those `DLL`s are present in the extension's `bin` folder during view compilation (but they won't be copied if they aren't needed), otherwise view compilation fails.
- Simplified the way the path for view compilation is defined in each project file.
- Added references to `Orchard.Framework` and `Orchard.Core` in some extension's `Web.config`, where necessary for view compilation (also fixes `Razor IntelliSense` for those). Also removed some references where they are needed (which can be determined by not having a corresponding project reference).
- Removed the `MvcBuildViews` build target from `Orchard.Web.csproj` as it's not needed (since `Orchard.Web` doesn't have views on its own) and it causes an error during view compilation.
- Added the `BuildViews` build target to `Orchard.proj`. This is just for self-checking locally and for a CI build, not for deployment as it does not have an actual affect on the output.

### DIFF
--- a/Orchard.proj
+++ b/Orchard.proj
@@ -129,11 +129,12 @@
     <MSBuild
       Projects="$(Solution)"
       Targets="Build"
-      Properties="Configuration=$(Configuration);OutputPath=$(CompileFolder)" />
+      Properties="Configuration=$(Configuration);OutputPath=$(CompileFolder);MvcBuildViews=false" />
     <!-- Compile to "regular" output folder for devs using VS locally -->
     <MSBuild
       Projects="$(Solution)"
-      Targets="Build"/>
+      Targets="Build"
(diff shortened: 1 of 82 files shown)


sebastienros:
Looks good and might not break anything, at least view compilation didn't work before so it won't break this.

Have you changed the codegen templates?
Also maybe just use the existing Precompiled target?

BenedekFarkas:
Code generation templates: Good idea, I'll add that!
I replied on the issue about the Precompiled target.

sebastienros:
Maybe you can also update the documentation. In the deployment page?

Issue #7588, opened 2017-02-27
Consistency displaying login errors


jardg:
This was issued in #1397 too.

I have to add to the issue refered that the same problem arises when users must verify their email address before being able to login. If this option is enabled and the user doesn't validate his e-mail because he didn't notice or the e-mail wasn't sent for any reason He may think that he can't login because of wrong username/email or password. This happens because the ValidateUser method in MembershipService returns null in the three cases where the user can't login.
```
           if ( user == null || ValidatePassword(user.As<UserPart>(), password) == false )
                return null;

            if ( user.EmailStatus != UserStatus.Approved )
                return null;

            if ( user.RegistrationStatus != UserStatus.Approved )
                return null;
```

Should this return an object with IUser and a error property or getting the error in the calling method somehow so the methods using this service could handle the validation with more accuracy?. I would be able to contribute myself with a pull request solving this issue if you let me know which could be the best way to solve this problem.

sebastienros:
Ok. Like return some kind of `IdentityResult` that would describe the current static.

PR #7734, opened 2017-06-19
Create IUSerIdentityResult to show custom login errors to user

jardg:
This way when a login isn't allowed We don't always display the same error for a invalid login.
The issue related is #7588


src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/XmlRpcHandler.cs
```diff
@@ -60,7 +60,7 @@ public class XmlRpcHandler : IXmlRpcHandler {
             UrlHelper url) {
 
             var user = _membershipService.ValidateUser(userName, password);
```
sebastienros: Whenever you call ValidateUser, use a new variable, for instance validationResult, then keep the `user` variable to be `validationResult.User`. This will prevent all the `user.User` changes.

sebastienros:
I have just one concern with this change, it's that it allows for username checks. We can guess if an email is registered or not with this kind of information.
If there is already another endpoint that would allow that then we can ignore this concern, but I can't think of any.
I will also ask what the security engineers if it's ok. And look at what is done in ASP.NET Identity default templates.

jardg:
This wont allow for username checks. Whether user is null or the combination user/password is invalid always returns The username or e-mail or password provided is incorrect. It was just  I made a change returning an incorrect message from "The username or e-mail or password provided is incorrect."  to "Invalid password".
Also where can I look to understand what is a ASP.NET Identity template?Thanks

src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/XmlRpcHandler.cs
jardg: Changed

sebastienros:
With "The username or e-mail or password provided is incorrect." you don't know if the user exists. With your message you know it exists.

Create a new ASP.NET MVC application (dotnet core or not) then try to log with a non-existent user or with a wrong password (existing user). Compare the messages.

jardg:
I explained bad myself. In the first pull the message told you that the password was incorrect as you say. But I updated the pull request to correct this. With the changes the message "The username or e-mail or password provided is incorrect." is always displayed wether the user doesn't exist or the password is incorrect. So its safe in this concern. The messages included on the pull "You must verify your email" or "You must be approved before being able to login" are only displayed if the password/user validation done before is correct but the username didn't verify the mail or the user isn't approved yet. Otherwise it will show "The username or e-mail or password provided is incorrect.". I have tested these cases to work as expected.

sebastienros:
Two requests:
- Translate the message from the service, not from the caller. You can return a LocalizedString.
- Return the error message as an `out` parameter in `ValidateUser`, this will make less changes in the code as you don't use it all the time.

jardg:
Changed to return List of LocalizedString as it can output one or more errors. Changed to output these messages.

Issue #7747, opened 2017-06-25
Captcha always submitted twice, causing validation error


sfmskywalker:
**TLDR**
When you try to submit a comment with the **ReCaptchaPart** attached, the ReCaptcha driver's `UpdateEditor` method is invoked twice. This causes the captcha to be submitted twice to Google. Causes problems.

**Repro Steps**

1. Setup a new Orchard site (develop branch)
2. Enable the Anti-Spam feature
3. Attach the ReCaptchaPart to the Comment content type
4. Configure the ReCaptcha settings with a valid public and private key (generate them [here](http://www.google.com/recaptcha/admin))
5. Create a dummy Blog and Blog Post so you can test submitting a comment
6. Submit a comment
7. Observe the validation error:

![image](https://user-images.githubusercontent.com/938393/27517005-88b4a0de-59c4-11e7-85bc-31566cdab595.png)

**The Issue**
The core issue is that the captcha is submitted twice. The first time the response is succesful, but the second time, Google says (rightfully so) that the captcha was already submitted. This causes a validation error.
The reason the `UpdateEditor` method is invoked twice is because the `CommentController` invokes `UpdateEditor` on the Comment content item twice.

**Possible Remedies**
There are a number of ways to handle this.

1. The root cause is the fact that the `CommentController` invokes `UpdateEditor` on the Comment twice. There's probably a good reason for this, so we may not want to change this behavior. At the same time, the need for having to call `UpdateEditor` twice might be a symptom of an underlying problem with the way `ContentManager.Create` works on a content item on which `UpdateEditor` was already called.
2. If we can't update the `CommentController` to only invoke `UpdateEditor` once, we can update the `ReCaptchaPartDriver` to "remember" that it already submitted the captcha.
3. Alternatively, we can simply check the return code from Google, and ignore the "duplicate" error code by not adding a validation error.

PR: #7748

sfmskywalker:
I'm gonna try and see why the UpdateEditor method is called twice from the CommentController first. Ideally we can just remove the second call, or come up with a better fix than having to call UpdateEditor after ContentManager.Create.

sfmskywalker:
Everything seems to work just fine if I remove the second call to UpdateEditor, so I'm going to submit a PR. If anyone remembers why we do need to call UpdateEditor twice, I'll send another PR with a comment explaining the second call, in addition to proposed solution 2 or 3.

sfmskywalker:
PR: #7748

[...]

PR #7748, opened 2017-06-25
Issues/7747/recaptcha

sfmskywalker:
Fixes issue #7747 where the UpdateEditor of all content parts on the newed-up Comment content item is invoked twice, causing an issue with the ReCaptchaPartDriver.

Also corrected a small typo.

See issue #7747.


sfmskywalker:
I updated the PR by moving the call to `ContentManager.Create` up so that it executes before the call to `ContentManager.UpdateEditor` to prevent the issue with Audit Trail as described in #6733.

sebastienros:
As commented on the issue, we moved Create down on purpose, otherwise it would create transaction log, even on spam.

BenedekFarkas:
Closing this as the origin issue is already closed and I can't repro the problem on either of `1.10.x` and `dev`.

PR #7752, opened 2017-07-04
Adds support to LocalizationStreamParser for po entrys spanning multiple lines

andrewcartwright1:
After spending some time making our gulp localization task work I came across an issue were some PO entries were half complete and noticed that they had been shifted into a new line by the task - after reviewing it, it wraps at 78 characters which seemed like an issue, but after some research, string wrapping is perfectly fine: http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html

The summary of my change is instead of reading line by line and determining what we are working with later on... do the following.

- Work out what scope we are looking at
- Check the next line, if its not another scope, append it to the current line and loop through till we find another scope (hence building the key, string, comment or scope)

Afterwards, check what scope we started with and do the required action.

### DIFF
--- a/src/Orchard/Localization/Services/LocalizationStreamParser.cs
+++ b/src/Orchard/Localization/Services/LocalizationStreamParser.cs
@@ -1,8 +1,7 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
-using Orchard.Logging;
 
 namespace Orchard.Localization.Services {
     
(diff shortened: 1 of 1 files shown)


gcsuk:
:+1:

src/Orchard/Localization/Services/LocalizationStreamParser.cs
```diff
@@ -103,34 +137,17 @@ public class LocalizationStreamParser : ILocalizationStreamParser {
             return sb == null ? str : sb.ToString();
         }
 
-        private string TrimQuote(string str) {
+        private static string TrimQuote(string str) {
             if (str.StartsWith("\"") && str.EndsWith("\"")) {
-                if (str.Length == 1) {
-                    // Handle corner case - string containing single quote
-                    Logger.Warning("Invalid localization string detected: " + str);
```
andrewcartwright1: Need to readd this.

PR #7761, opened 2017-07-09
Update WebRequest help with correct way to format JSON (fixes #7760)

### LINKED ISSUE #7760: The help text for executing a WebRequest activity with JSON is incorrect
The help text is: 

> For JSon, enter a valid JSon string

But due to #7759 this will be executed as a token and remove the valid json. The code handles JSON by replacing {} with double brackets so `{ test: 'test' }` should be inputted as `(( test: 'test' ))`

Hazzamanic:


src/Orchard.Web/Modules/Orchard.Workflows/Forms/WebRequestForm.cs
```diff
@@ -39,7 +39,7 @@ public class WebRequestForm : IFormProvider {
                         _FormValues: New.Textarea(
                             Id: "FormValues", Name: "FormValues",
                             Title: T("Form Values"),
-                            Description: T("For KeyValue, enter one line per key=value pair to submit when using the POST verb. For JSon, enter a valid JSon string"),
+                            Description: T("For KeyValue, enter one line per key=value pair to submit when using the POST verb. For JSON, enter a string where the curly braces are replaced by double brackets e.g. { test: 'test' } becomes (( test: 'test' ))."),
```
sebastienros, line 42: 'brackets' ? `(( ... ))`

sebastienros:
I fixed what I think was a mistake. Is ok with you now?

Hazzamanic:
I don't think so, it should be double brackets like this (( ... ))

The line in the web request activity is:
var json = formValues.Replace("((", "{").Replace("))", "}");

sebastienros:
And I wrote this code ;) sorry, I will fix and merge. But they are not called brackets, but parens, or parenthesis.

Hazzamanic:
Ah okay, sorry for the confusion. In British English brackets generally refers to (), quite rare to hear the term parentheses.

Issue #7767, opened 2017-07-11
Media FolderName Validation


LorenzoFrediani-Laser:
Media FolderName does not have validation to be a valid url.

User can create media foldername with characters valid for systemfolder but not valid for pageurl

Steps to reproduce :
1- Create Media Foldername "a&b" 
2- add image
-The direct url of the Image is not valid

PR #7768, opened 2017-07-11
Add Validation FolderName to generate valid url link

LorenzoFrediani-Laser:
- Used same regex used in AutorouteService.IsPathValid
this solve issue #7767

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/FolderController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/FolderController.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Web.Mvc;
 using Orchard.Localization;
 using Orchard.Logging;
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/FolderController.cs
```diff
@@ -62,17 +63,21 @@ IMediaLibraryService mediaManagerService
             }
 
             try {
-                _mediaLibraryService.CreateFolder(viewModel.FolderPath, viewModel.Name);
-                Services.Notifier.Information(T("Media folder created"));
+                bool valid = String.IsNullOrWhiteSpace(viewModel.Name) || Regex.IsMatch(viewModel.Name, @"^[^:?#\[\]@!$&'()*+,.;=\s\""\<\>\\\|%]+$");
+                if (!valid) {
+                    throw new ArgumentException(T("Folder contains invalid characters").ToString());
```
sebastienros, line 68: Won't it be double-encoded? Can you try with some text that contains HMTL elements?

LorenzoFrediani-Laser: I try insert Html elements but i'm blocked by form validation that deny this behavior,.
I have looked foldername in updatemodel and he is not encoded.

sebastienros: I don't think you need to translate a message in an exception, unless this exception is explicitly reused from the controller to display error messages.

PR #7771, opened 2017-07-11
ContentPart names should be sanitized upon creation. Fixes #7770.

### LINKED ISSUE #7770: ContentType and ContentPart names should be sanitized upon creation
It's currently possible to create ContentParts with trailing whitespace via the Dashboard.
Such ContentParts cannot be edited again, following the link will result in a 404 (e.g. _/Admin/ContentTypes/EditPart/PartWithTrailingSpace%20_)

The `Orchard.ContentType` AdminController already [sanitizes part-name](https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.ContentTypes/Controllers/AdminController.cs#L385) (*suggestion.ToSafeName()*). The same needs to be done in [*CreatePartPOST*](https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.ContentTypes/Controllers/AdminController.cs#L389).

Xceno:
Added the **ToSafeName()** sanitation already used in *CreatePart*, to *CreatePartPOST*.
This ensures that only valid technical names can be used.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentTypes/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.ContentTypes/Controllers/AdminController.cs
@@ -390,6 +390,8 @@ namespace Orchard.ContentTypes.Controllers {
             if (!Services.Authorizer.Authorize(Permissions.EditContentTypes, T("Not allowed to create a content part.")))
                 return new HttpUnauthorizedResult();
 
+            viewModel.Name = viewModel.Name.ToSafeName(); // The name must be sanitized first.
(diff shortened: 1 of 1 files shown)


sebastienros:
Shouldn't it be a validation message instead, if the ToSafeName result is different?

BenedekFarkas:
I'll decline this PR because I found that creating and editing both Parts and Types lack some validation against special characters and the change in this PR needs more work (I'll commit the fixes soon).

PR #7786, opened 2017-07-24
Exclude file from gulp

chanondw:
Files listed in Assets.json can be included as wildcard or per file, but there is no way to include all except some files. If you put `!` in front of the file name in Assets.json, it will not exclude that file from compiling and copying to destination.


sebastienros:
Looks good, @DanielStolt ?

DaRosenberg:
Looks good to me too!

Issue #7789, opened 2017-07-26
ShapeMenuItem lacks IdentityPart


thekaveman:
And thus cannot be imported/exported properly. Fix incoming.

PR #7790, opened 2017-07-26
Adding Identity to ShapeMenuItem

thekaveman:
This is a fix for #7789

### DIFF
--- a/src/Orchard.Web/Core/Navigation/Migrations.cs
+++ b/src/Orchard.Web/Core/Navigation/Migrations.cs
@@ -69,6 +69,7 @@ namespace Orchard.Core.Navigation {
                     .WithPart("ShapeMenuItemPart")
                     .WithPart("MenuPart")
                     .WithPart("CommonPart")
+                    .WithIdentity()
                     .DisplayedAs("Shape Link")
                     .WithSetting("Description", "Injects menu items from a Shape")
                     .WithSetting("Stereotype", "MenuItem")
@@ -153,6 +154,7 @@ namespace Orchard.Core.Navigation {
                     .WithPart("ShapeMenuItemPart")
                     .WithPart("MenuPart")
                     .WithPart("CommonPart")
+                    .WithIdentity()
                     .DisplayedAs("Shape Link")
                     .WithSetting("Description", "Injects menu items from a Shape")
                     .WithSetting("Stereotype", "MenuItem")
@@ -178,6 +180,14 @@ namespace Orchard.Core.Navigation {
             );
 
             return 6;
+        }
+
+        public int UpdateFrom6() {
+            ContentDefinitionManager.AlterTypeDefinition("ShapeMenuItem", cfg => cfg
+                .WithIdentity()
+            );
+
+            return 7;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


thekaveman:
I see that this was already added in the `dev` branch from #7040 (there wasn't an issue so I missed it before submitting mine), but I feel like it should be in the `1.10.x` branch, no?

src/Orchard.Web/Core/Navigation/Migrations.cs
```diff
@@ -177,5 +181,13 @@ public class Migrations : DataMigrationImpl {
 
             return 6;
```
sebastienros: `return 7;`

thekaveman: Isn't that the `UpdateFrom5` method? 

`return 7;` is in `UpdateFrom6` below.

sebastienros: Right, the diff confused me. But then move all new WithIdentity you added to the `UpdateFrom6` as new installs will always go there.

src/Orchard.Web/Core/Navigation/Migrations.cs
```diff
@@ -30,6 +30,7 @@ public class Migrations : DataMigrationImpl {
             ContentDefinitionManager.AlterTypeDefinition("Menu", cfg => cfg
                 .WithPart("CommonPart", p => p.WithSetting("OwnerEditorSettings.ShowOwnerEditor", "false"))
                 .WithPart("TitlePart")
+                .WithIdentity()
```
sebastienros: This should only be in `UpdateFrom6`

src/Orchard.Web/Core/Navigation/Migrations.cs
```diff
@@ -68,6 +69,7 @@ public class Migrations : DataMigrationImpl {
                     .WithPart("ShapeMenuItemPart")
                     .WithPart("MenuPart")
                     .WithPart("CommonPart")
+                    .WithIdentity()
```
sebastienros: This should only be in `UpdateFrom6`

src/Orchard.Web/Core/Navigation/Migrations.cs
```diff
@@ -101,6 +103,7 @@ public class Migrations : DataMigrationImpl {
             ContentDefinitionManager.AlterTypeDefinition("Menu", cfg => cfg
                 .WithPart("CommonPart")
                 .WithPart("TitlePart")
+                .WithIdentity()
```
sebastienros: This should only be in `UpdateFrom6`

src/Orchard.Web/Core/Navigation/Migrations.cs
```diff
@@ -151,6 +154,7 @@ public class Migrations : DataMigrationImpl {
                     .WithPart("ShapeMenuItemPart")
                     .WithPart("MenuPart")
                     .WithPart("CommonPart")
+                    .WithIdentity()
```
sebastienros: This should only be in `UpdateFrom6`

thekaveman:
Thanks for the feedback. Since `UpdateFrom5` already adds identity to `Menu`, I've just left that one out completely.

Issue #7794, opened 2017-07-30
Some shape's placement cannot be overridden in the Edit Placement feature


Hazzamanic:
It is caused by shape type keys in the shape table descriptors have differing cases to the name specified by Content Shape. For example the owner edit shape is stored as a placement setting on the content type as: Parts_Common_Owner_Edit but the corresponding shape in the descriptor dictionary is all lowercase: parts_common_owner_edit. So when it tries to find the shape type key to register a placement override it cant. This comes from Orchard.ContentTypes.Services.ContentTypePlacementStrategy

sebastienros:
Should we do a case insensitive search instead?

Hazzamanic:
I guess so, but being a dictionary I think we'd need to change the dictionary to have a case insensitive key comparer?

Hazzamanic:
Fixed in https://github.com/OrchardCMS/Orchard/pull/7904

PR #7795, opened 2017-07-30
Updates the placement editor screen to support moving shapes into tabs

### LINKED ISSUE #7791: You can't organise editors into tabs using Edit Placement
From version 1.9 (or 10?) you can move editors into tabs using #tab_name in your module's Placement file. However, you can't control this from the admin area using Edit Placement. That screen should be revamped to allow tabs to be created and editors moved between them

Hazzamanic:
fixes #7791


sebastienros:
Looks interesting, would you mind recording a little screencast showing how it looks like, or demoing it during a Tuesday meeting?

Hazzamanic:
Yeah, or I could have at least added a little more information to the pull request about what I did. My bad. 

https://youtu.be/xNiSS6kI9bs

What you can do:

- Add new tabs
- Move content between tabs
- Content tab is fixed at position one because the tab renderer always puts the Content tab at the start
- Delete tabs - deleting a tab will move all editors into the bottom of content
- Content tab cannot be deleted
- By default things will just appear in Content - ie. when you restore, so hopefully there is no clever logic that could lead to editors ever getting "lost"
- Tabs can be reordered 
- Empty tabs will disappear on Save
- Editors are now hidden by default, with a toggle show/hide editor shape button by each editor shape title (I was having a hard time dragging things around with the editor shapes around)

The eagle eyed of you may have noticed that when deleting a tab and the editors being moved into the bottom of content, after saving they are no longer at the bottom. This is due to #7794. This also means some editors (owner and created date amongst them) cannot be moved into tabs or reordered. But I'm sure Sebastien will fix it <3 

I also have a lot of tabs open I know. And my YouTube name is very old and from a time when I probably should have known better but sadly I just didn't. Please don't mock me.

jean:
@Hazzamanic can the _Content_ tab be renamed?

Hazzamanic:
@jean I'm afraid not. The _Content_ tab isn't actually a tab, it is just where anything not in a tab will be put. So Orchard forces its existence at the start of the tab list. This is why the tab management UI doesn't allow you to move the Content tab around. Tabs are a bit of a hack into the existing shape rendering to be honest. 

There is a simple fix for this. We just need to skip the rendering of the Content tab if it is empty. Then you can just move all the shapes out of the Content tab into your own, so you wouldn't actually rename the content tab, just empty it and never render it. If you are running this code now and want this update now, change the Orchard.Core.Shapes.CoreShapes.HarvestAndSortTabs method to this.

```
public static IEnumerable<string> HarvestAndSortTabs(IEnumerable<dynamic> shapes) {
            var orderedShapes = Order(shapes).ToArray();
            var tabs = new List<string>();
            var contentTabRequired = false;

            foreach (var shape in orderedShapes) {
                var tab = (string)shape.Metadata.Tab;

                if (String.IsNullOrEmpty(tab)) {
                    contentTabRequired = true;
                    continue;
                }

                if(!tabs.Contains(tab))
                    tabs.Add(tab);
            }

            // If we have any tabs, make sure we have at least the Content tab and that it is the first one,
            // since that's where we will put anything else not part of a tab.
            if (contentTabRequired && tabs.Any()) {
                tabs.Remove("Content");
                tabs.Insert(0, "Content");
            }

            return tabs;
        }
```

A caveat is that another bug (https://github.com/OrchardCMS/Orchard/issues/7794) breaks this because you cant move some shapes out of the Content tab

PR #7820, opened 2017-08-16
Add placeholder editor form and attribute for applicable fields

mdameer:
issue #6336


src/Orchard.Web/Modules/Orchard.DynamicForms/Elements/PlaceholderableFormElement.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Orchard.Layouts.Helpers;
+
+namespace Orchard.DynamicForms.Elements {
+    public abstract class PlaceholderableFormElement : LabeledFormElement {
```
sfmskywalker: Although it's creative, I'm not a fan of the word "placeholderable" :)
Also, why not add `Placeholder` to the `LabeledFormElement` directly, instead of subclassing?

PS. Great to see you back! :)

src/Orchard.Web/Modules/Orchard.DynamicForms/Forms/PlaceholderForm.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Orchard.Forms.Services;
+using DescribeContext = Orchard.Forms.Services.DescribeContext;
+
+namespace Orchard.DynamicForms.Forms {
+    public class PlaceholderForm : Component, IFormProvider {
```
sfmskywalker, line 5: Since the placeholder attribute seems to be such an intrinsic thing to input elements, I think it would be even better (and simpler) to just update the AutoLabel form provider and add the Placeholder textbox shape there. Which would make even more sense if you'd also get rid of `PlaceholderableFormElement` and move the `Placeholder` property to `LabeledFormElement`.

src/Orchard.Web/Modules/Orchard.DynamicForms/Elements/PlaceholderableFormElement.cs
sfmskywalker: Actually, I take it back. Not all form elements deriving from `LabeledFormElement` will necessarily have a placeholder (such as drop down lists). So perhaps you were right in the first place by introducing a new sub class.

src/Orchard.Web/Modules/Orchard.DynamicForms/Forms/PlaceholderForm.cs
sfmskywalker: I take this back as well for the same reason.

sfmskywalker (commented):
The PR looks good to me, but I suggest finding a better word for "Placeholderable". Perhaps "FormElementWithPlaceholder" would be better, despite that it's longer?

src/Orchard.Web/Modules/Orchard.DynamicForms/Elements/PlaceholderableFormElement.cs
mdameer: Thanks @sfmskywalker, I'm agree on both names and see them as appropriate, but let's wait for other suggestions to make the final decision

ViRuSTriNiTy:
I vote for `FormElementWithPlaceholder` as i tend to name my classes the same way when no distinct word is available that can be placed in front of the class name.

mdameer:
Class renamed to `FormElementWithPlaceholder` as requested.

PR #7825, opened 2017-08-21
Log EmailActivity not sending email because null or empty recipients

jardg:
Right now EmailActivity always returns "Done" as his outcome. This is actually not true when the recipients parameter is empty or null. This happens when the Email Activity hasn't got his fields filled by the user. So I think We should detect this case and return as the outcome "Not Done". Also its nice to log this so we can detect the case. 
In my particular case this change is pretty useful because in the company I work We provide worfklows created by recipes to our clients in which they have to fill the email fields of the activity. They sometimes don't notice this so the e-mail alerting them of an important event is not being sent. Neither We do. With the log and the outcome change we can configure log4net or other Activity to alert us so We can call the client to provide the data.


sebastienros:
Can you renamed the alternate outcome "Failed" ?

jardg:
Renamed

Issue #7834, opened 2017-08-29
Unable to override displaytype for default itemcontroller easily


carlwoodhouse:
Hi;

So i have a requirement to display a contentitem differently depending on a value set on one of its part; mostly the ordering of shapes and in some cases alternates. 

When the value isn't set i can achieve all the ordering etc easily in placement just by targetting a display type of 'Detail' - I had hoped to be able to target the placement using a custom displaytype of 'Hub' But have found that unless using your own controller; its impossible to set the displaytype (i think). Due to it being hardcoded to nothing in https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Core/Contents/Controllers/ItemController.cs; line 46; which in turn defaults to "Detail" later down the stack; some way to set this would be super handy.

This is an example of what i'd hoped to do with placement
`  <Match DisplayType="Detail">
    
    <Place Parts_PageTools="MyZone:1" />
    </Match>
   <Match DisplayType="Hub">
    
    <Place Parts_PageTools="MyZone:10" />
  </Match>
 `

I guess i'll be able to achieve what im trying to with some clever alternate manipulation in a shapeprovider; or by overriding the route for the default item controller and using my own; neither of which appeal!

carlwoodhouse:
So i decided to take a different approach as i couldn't decide quite how/when you'd set that displaytype for the controller to know what displaytype to build; really i just needed to be able to customize placement matching.

I've implemented this in my own module by suppressing `ShapePlacementParsingStrategy.cs`  and replacing it with my own extendable version :| also provided a PR for consideration into getting this into orchard.

Issue #4138, opened 2013-11-25
Open up placement file processing API


orchardbot:
@Piedone created:
https://orchard.codeplex.com/workitem/20309

What's now in ShapePlacementParsingStrategy can't be used from the outside at all. My suggestion would be to open up the functionality contained in the class (i.e. the ability to take a PlacementFile object and modify a ShapeTableBuilder object with its declarations) by refactoring it to a separate service.

ShapePlacementParsingStrategy could use this service, together with other, 3rd party implementations (that could, for instance, provide alternate sources of Placement.info files).

Similarly IPlacementFileParser could also work with placement declaration strings, not just virtual paths to Placement.info files.

Issue #3671, opened 2013-06-29
Match Url in placement by Regex (Placement extensibility)


orchardbot:
**neTp9c** created:
https://orchard.codeplex.com/workitem/19843

It does not require a lot of code, but in my opinion it is very useful.

For example I need to show BodyPart only on the first page of the list.

For do this need change BuildPredicate function in Orchard.DisplayManagement.Descriptors.ShapePlacementStrategy.ShapePlacementParsingStrategy class
Add to it
case "PathRegex":
                    return ctx =>
                    {
                        return Regex.IsMatch(VirtualPathUtility.ToAppRelative(
                            HttpContext.Current != null && HttpContext.Current.Request != null ? HttpContext.Current.Request.Url.PathAndQuery : ""), 
                            expression, RegexOptions.IgnoreCase) && predicate(ctx);
                    };

And then use
<Match PathRegex="^~/.*\?.*">

In general, it would be useful to be able to create their own rules for handling placement

PR #7837, opened 2017-08-30
Made shape placement parsing strategy extendable with providers

carlwoodhouse:
Changes the hardcoded nature of Placement.info match nodes by opening it up so you can match placement on custom conditions using new providers.

Includes default match providers for contentpart, contenttype, displaytype and path to replicate the current match types.

This should allow us to possibly close the following 
#4138
#3671

and solves the problems outlined in #7834

### DIFF
--- a/src/Orchard.Tests/DisplayManagement/Descriptors/DefaultShapeTableManagerTests.cs
+++ b/src/Orchard.Tests/DisplayManagement/Descriptors/DefaultShapeTableManagerTests.cs
@@ -247,7 +247,10 @@ namespace Orchard.Tests.DisplayManagement.Descriptors {
 
                 _container.Resolve<TestShapeProvider>().Discover =
                     builder => builder.Describe("Hello").From(TestFeature())
-                                   .Placement(ShapePlacementParsingStrategy.BuildPredicate(c => true, new KeyValuePair<string, string>("Path", path)), new PlacementInfo { Location = "Match" });
+                                   .Placement(ShapePlacementParsingStrategy.BuildPredicate(c => true, 
+                                    new KeyValuePair<string, string>("Path", path), 
+                                    new[] { new PathPlacementParseMatchProvider() }), 
+                                    new PlacementInfo { Location = "Match" });
 
                 var manager = _container.Resolve<IShapeTableManager>();
                 var hello = manager.GetShapeTable(null).Descriptors["Hello"];
--- a/src/Orchard/DisplayManagement/Descriptors/ShapePlacementStrategy/DefaultPlacementParseMatchProviders.cs
+++ b/src/Orchard/DisplayManagement/Descriptors/ShapePlacementStrategy/DefaultPlacementParseMatchProviders.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Linq;
+using System.Web;
+
+namespace Orchard.DisplayManagement.Descriptors.ShapePlacementStrategy {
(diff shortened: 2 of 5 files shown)


carlwoodhouse:
@MatteoPiovanelli-Laser I'm not overly sure i follow? the key is literally the name of the attribute on the match? wouldn't any evaluation for fancy stuff be inside the Match method? 

I might be misunderstanding you :) can you provide an example?

MatteoPiovanelli:
Sure.

In the Gist you posted yesterday as an example you had in your Placement.info
```
<Match Hub="true">
```
And you `Match` method did
```
return contentItem.Has<IClincialArticlePart>() && contentItem.As<IClincialArticlePart>().IsTopicHub;
```

We could make that more generic with reflection and have the key in the placement info be
```
<Match ClincialArticlePart.IsTopicHub="true">
```
Then if we had a `bool CheckKey(ShapePlacementContext context, string key)` method it could check whether the `context.Content.ContentItem` had a `ClincialArticlePart` with a property called `IsTopicHub`.

The base cases that are in place would still check the key against a constant. Most cases would check the key against a constant. Having the `CheckKey` method allows more flexibility.

As I am writing this, I feel I would also pass the key to the `Match` method, for the same reason and to be able to use it in the reflection example.

carlwoodhouse:
That would be pretty powerful; though maybe the efficiency would be a concern as it would have to evaluate all those keys against every match in every placement? regardless of if any match statement actually existed for it.

would you not write it 

```
<Match ClinicalArticlePart="IsTopicHub">
...
</Match>
```

or 

```
<Match Hub="ClinicalArticlePart.IsTopicHub">
...
</Match>
```
Then just parse the expression to do whatever you wish with reflection; this would prevent the logic getting run if there was no actual match statements in any placement

MatteoPiovanelli:
> That would be pretty powerful; though maybe the efficiency would be a concern as it would have to evaluate all those keys against every match in every placement? 

Could you explain what you mean? I see that efficiency would probably be poor. I don't understand "evaluate all those keys against every match in every placement"

carlwoodhouse:
The placement processor basically says here is a `<match`> it will do that for every `<match>` node in every module and theme. it then evaluates to see if it knows what to do with the key; so with many modules/themes and in turn many `<match>` node it would have to run the key logic for each one; so the performance of doing the logic vs is it this constant would be poor. at the moment the Match() method of the providers which contains logic will only run if there has been a `<match>` node in a placement file that matches it's key ....

MatteoPiovanelli:
I get it now.
I would still allow it though. If the `CheckKey` method compares to a constant then it won't really impact performances. If it does something else it might, but then it's the method's "fault", I think, not the interface's.

I am not saying we should have the fancy provider using reflection, as it will be way overkill in 99.9999% of cases (or more). But maybe we'll find a way a clever caching strategy that would make it's performance impact negligible. Maybe it only even tries to parse the key when it starts with a "@" character, and `CheckKey` returns false in every other case without doing anything  more.

sebastienros:
Just have a key like `Predicate` with a constraint inside, instead of matching different keys. Could even be some language like JS or C#.

`Predicate="ClinicalArticlePart.IsTopicHub == true"`

carlwoodhouse:
Agreed it shouldn't match diff keys ^ the above is pretty much possible
with what I've done here already no?

sebastienros:
You confirm nothing is broken by this change?

It's just a new interface, and some code is moved outside, right?

carlwoodhouse:
Correct, I can confirm nothing broken

PR #7847, opened 2017-09-13
Display correct Comments total at /Admin/Blogs. Fixes #7838.

### LINKED ISSUE #7838: Comments total at /Admin/Blogs is always 0
This issue can be seen by either:
- Clicking on "Blog" in the /Admin menu (for an Orchard instance where there is more than one blog)
- Navigating directly to /Admin/Blogs (for an Orchard instance where there's only one blog which results in clicking on "Blog" in the /Admin menu taking you to the posts view for that blog)

The text under the name of the blog will look something like this:

> 0 comments |   Last modified: 7 days ago |  By robertwray

The issue is that, irrespective of the number of comments actually present against individual blog posts, the totaliser will always state "0 comments". I *think* this is due to the fact that `CommentsContainerPartDriver` (from Orchard.Comments) asks the commentService for the number of comments associated with the ContentItem for the *blog*, which is *0*, rather than asking for the number of comments associated with the blog posts associated with the blog.

Unfortunately I'm too new to the Orchard codebase and how it all hangs together to work out how this could be fixed, other than by simply removing the field that indicates the number of comments against the blog.

jchenga:
The problem was due to CommentsContainerPartDriver getting the comments for a blog post.
The fix adds a new method, **GetCommentsForContainer**, in CommentService that returns the comments created for the blog. The part driver then calls that method instead to calculate the number of comments for the container.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Comments/Drivers/CommentsContainerPartDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Comments/Drivers/CommentsContainerPartDriver.cs
@@ -13,7 +13,7 @@ namespace Orchard.Comments.Drivers {
 
         protected override DriverResult Display(CommentsContainerPart part, string displayType, dynamic shapeHelper) {
 
-            var commentsForCommentedContent = _commentService.GetCommentsForCommentedContent(part.ContentItem.Id);
+            var commentsForCommentedContent = _commentService.GetCommentsForContainer(part.ContentItem.Id);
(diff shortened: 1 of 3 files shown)


sebastienros:
It's a breaking change. Either you target your change on dev or you remove the new method and just use `GetComments().Where(c => c.CommentedOnContainer == id);` directly in the driver.

jchenga:
@sebastienros  Out of curiosity, how is the PR a breaking change? 

I'll target my change on dev and create another PR.

Issue #5764, opened 2015-09-08
Enhancing ISearchbuilder to accept array as value field


kumards:
Hi,

The current ISearchBuilder.WithField allows only one value to be supplied as search criterion. Any plans to allow array of values. This will allow conditions like
City IN (New York, Boston) AND ProductCategory IN (Electoronics, Mobile, Accessories)

I would assume these are very common scenarios encountered in most of the applications.

sebastienros:
I assume you can call `WithField` multiple times and it will have the same effect. Two other solutions:
- create an extension method to do it for you
- use lucene query language directly if you want more control over the search query

If you go the extension method way, then we could add it in the same namespace as ISearchBuilder.

kumards:
@sebastienros  extension technique wouldn't work. If you look at the example I gave here, it requires must clause nested with should clause. I tried multiple times ```WithField``` method. It ends up with non nested must and should clauses. That does not give me a desired result.

I think query parameter in ```Parse``` method could be another potential candidate only if it allows defaultField to be empty or null.

sebastienros:
I see, it's a MUST between SHOULD values ... so this needs to be added to the interface.

TwistedWhisper:
I have asked a question similar to this a few weeks back at [http://stackoverflow.com/questions/35596168/orchard-lucene-custom-search-service-with-query-grouping-and-numeric-range-query](url)

Currently the ISearchBuilder is limited to handling linear queries only, it doesn't handle grouping well. By grouping, I mean clauses that involve nested parentheses with combination of `MUST` and/or `SHOULD`. The `ISearchBuilder.Parse()` could be used to write a freestyle query to somehow work around this limitation but it's not without its own limitations. `ISearchBuilder.Parse()` is unable to handle range query.

    // this handles Grouping with no problems, note that you can put "null" as the "defaultField" as long as you specify the fields in the query.
    _searchBuilder.Parse(null, "+(City:New\ York City:Boston) +(ProductCategory:Electronics ProductCategory:Mobile ProductCategory:Accessories)", false).Mandatory();

    // the following code can't handle range query and will return no hits (more info at the link I provided above)
    _searchBuilder.Parse("price", "+([100 TO 150] cost:[100 TO 150]) +(ProductCategory:Electronics)", false).Mandatory();

you're better off using the API

    _searchBuilder.WithinRange("price", 100, 150).Mandatory();

[...]

PR #7850, opened 2017-09-14
Extend ISearchBuilder to support grouping clauses

ryanwawr:
Based on DamienLaw's suggested changes in #5764. Needed clause-grouping functionality for a project I'm developing, so implemented these changes and performed some rudimentary testing. Can confirm that this works for simple groups such as Clause A AND (Clause B OR Clause C). Have not tested nesting groups within each other, but I see no reason why that wouldn't work as well.

Would be nice to see this merged soon, as it significantly expands the range of queries that can be constructed with the ISearchBuilder interface.

Not sure if this is necessary, but to be sure I'm in compliance with the CLA:
Submission containing materials of a third party: DamienLaw. Code was offered as a suggested improvement in #5764, and I am not aware of any licenses restricting its use.


sebastienros:
That looks good, can you just add some unit tests to at least demonstrate the usage?

ryanwawr:
Added two unit tests verifying simple grouped clauses and nested grouped clauses.

PR #7853, opened 2017-09-16
Fixed comments total at /Admin/Blogs is always 0. Based on dev branch (#7838)

jchenga:
Target the fix on dev branch. This PR replaces PR #7847 .

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Comments/Services/CommentService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Comments/Services/CommentService.cs
@@ -60,6 +60,11 @@ namespace Orchard.Comments.Services {
         public Localizer T { get; set; } 
         public ILogger Logger { get; set; }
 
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.Comments/Services/CommentService.cs
```diff
@@ -60,6 +60,11 @@ IMessageService messageService
         public Localizer T { get; set; } 
         public ILogger Logger { get; set; }
 
+        public IContentQuery<CommentPart, CommentPartRecord> GetCommentsForContainer(int id) {
```
sebastienros, line 63: ~~Remove this method and call ` _commentService.GetComments().Where(c => c.CommentedOnContainer == id)` directly in the driver.~~

~~Adding a method in the interface is a breaking change, which can't be shipped in 1.10.x branch.~~

Ok, I realize now that you are targeting dev now.

PR #7861, opened 2017-09-19
Added eagerly loading extensions

jersiovic:
This PR provides extensions for eager loading fields of a type  in content items returned by a ContentQuery or in a list of content items. This helps to avoid the N+1 problem without the need of writting much custom code.
Examples of use of this extensions:
```
//Eagerly loading in a row MediaLibraryPickerFields and TaxonomyFields on each content ítem returned
_contentManager.Query<ProductPart, ProductRecord>(VersionOptions.Published)
                .OrderBy(p => p.Descrip)
                .EagerlyLoadMediaLibraryPickerFields()
                .EagerlyLoadTaxonomyFields().Result

//Eagerly loading in a row Layers containing resulting widgets
_contentManager.Query<WidgetPart, WidgetPartRecord>()
                .WithQueryHints(new QueryHints().ExpandParts<CommonPart>())
                .Where<CommonPartRecord>(x => layerIds.Contains(x.Container.Id))
                .EagerlyLoadContainerContentItems().Result;
```

Next example uses extensión with an IList, in this case we need to provide contentManager as a parameter for the first call to the extensión cause opposite to ContentQuery extensión is not able to get from the IList a content manager.
```
termsList.EagerlyLoadMediaLibraryPickerFields(_contentManager).EagerlyLoadTaxonomyFields();
```

We use EagerlyLoadMediaLibraryPickerFields and EagerlyLoadTaxonomyFields for any query rendered by our custom projection driver for improving the performance of our queries. But I understand this is not convenient for any scenario so I skipped to add sth like that to default projection driver in this PR.
If you see this PR interesting to be approved, in the future we can discuss where in the Admin UI the end user should be able to set which eagerloading extensions need to be used for each query. ¿A new section for eager loading extensions?¿ maybe within each layout? This will allow to improve performance for any query defined by the end user

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
@@ -49,7 +49,7 @@ namespace Orchard.Widgets.Filters {
                 return;
             }
 
-            var widgetParts = _widgetsService.GetWidgets(_layerEvaluationService.GetActiveLayerIds());
+            var widgetParts = _widgetsService.GetWidgetsWithEagerlyLoadedContainers(_layerEvaluationService.GetActiveLayerIds());
 
             // Build and add shape to zone.
             var zones = workContext.Layout.Zones;
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
@@ -52,7 +52,7 @@ namespace Orchard.Widgets.Filters {
 
             var widgetParts = _widgetsService.GetWidgets(_layerEvaluationService.GetActiveLayerIds());
             widgetParts.Select(w => w.ContentItem).ToList()
-                        .EagerlyLoadContainerContentItems(_orchardServices.ContentManager, 1);
+                        .LoadContainerContentItems(_orchardServices.ContentManager, 1);
 
             // Build and add shape to zone.
             var zones = workContext.Layout.Zones;
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
@@ -9,6 +9,7 @@ using Orchard.Mvc.Filters;
 using Orchard.Themes;
 using Orchard.UI.Admin;
 using Orchard.Widgets.Services;
+using Orchard.ContentManagement;
 
 namespace Orchard.Widgets.Filters {
     public class WidgetFilter : FilterProvider, IResultFilter {
@@ -49,7 +50,9 @@ namespace Orchard.Widgets.Filters {
                 return;
             }
 
-            var widgetParts = _widgetsService.GetWidgetsWithEagerlyLoadedContainers(_layerEvaluationService.GetActiveLayerIds());
+            var widgetParts = _widgetsService.GetWidgets(_layerEvaluationService.GetActiveLayerIds());
+            widgetParts.Select(w => w.ContentItem).ToList()
+                        .EagerlyLoadContainerContentItems(_orchardServices.ContentManager, 1);
 
             // Build and add shape to zone.
             var zones = workContext.Layout.Zones;
--- a/src/Orchard/ContentManagement/DefaultContentManager.cs
+++ b/src/Orchard/ContentManagement/DefaultContentManager.cs
@@ -33,7 +33,7 @@ namespace Orchard.ContentManagement {
         private readonly ICacheManager _cacheManager;
         private readonly Func<IContentManagerSession> _contentManagerSession;
         private readonly Lazy<IContentDisplay> _contentDisplay;
-        private readonly Lazy<ITransactionManager> _transactionManager; 
+        private readonly Lazy<ITransactionManager> _transactionManager;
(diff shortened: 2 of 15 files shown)


src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
```diff
@@ -49,7 +49,7 @@ public class WidgetFilter : FilterProvider, IResultFilter {
                 return;
             }
 
-            var widgetParts = _widgetsService.GetWidgets(_layerEvaluationService.GetActiveLayerIds());
+            var widgetParts = _widgetsService.GetWidgetsWithEagerlyLoadedContainers(_layerEvaluationService.GetActiveLayerIds());
```
jersiovic: This prevents n+1 problem caused by next loop that accesses commonPart.container for each widget

sebastienros:
I'd prefer a case by case improvement, where the issue is actually is. And also explaining what kind of data and scenario trigger the issues. Adding eager loading by default might be bad in other cases.

src/Orchard.Web/Modules/Orchard.Widgets/Filters/WidgetFilter.cs
jersiovic: In response to this comment from @sebastienros "I'd prefer a case by case improvement, where the issue is actually is. And also explaining what kind of data and scenario trigger the issues. Adding eager loading by default might be bad in other cases."
I did it here cause noone else consume this method, but I agree it could be used by others. I'm going to move the eagerly loading to the consumer code of this method.

jersiovic:
After last commit this PR doesn't add eagerly loading by default anywhere. 
So, Is it ready for merging it?

jersiovic:
Examples of huge improvement of performance are those pages:

- https://www.sanlamberto2000.com/catalogo Here we load all the terms of a taxonomy eagerly loading their images and its containers that are accessed in the view to be rendered. We do it simply with this line : `termContentItems.EagerlyLoadMediaLibraryPickerFields(_contentManager)
                        .EagerlyLoadTaxonomyFields(loadTermsContainter:true);`
- https://www.sanlamberto2000.com/conservas-precocinados?taxonomy24230=33012 Here we eagerly load taxonomy field of each product and its parent terms to avoid lazy loading in the view for rendering the complete hierarchy of terms of each set of products related to a different term part. We accomplish that using this simply extension: `termContentItems.EagerlyLoadTaxonomyFields(loadTermsContainter:true)`

jersiovic:
@urbanit just in case you review this PR maybe it is useful I paste here some explanations I gave to @sebastienros yesterday after the meeting, so he agreed to try another time to review this PR.

I add a bunch of new files, those are extensions for each kind of field for which you can do eager loading.
The point is you use those extensions only where you want. Those are not used everywhere. In fact, my change adds extensions for three types of fields: taxonomy picker fields, media picker fileds, and container filed. But I only added code to use them in orchard core in one place (in widgets filter) where I only use the Container field eagerly loading.
So I'm using them only where I've seen it is needed.
I provide the other extensions not used in the core because it is common in custom modules you need things like that.
Months ago one guy contacted me asking for this feaute I talked about in an older issue and @Xceno  shows in the PR with an emoji he likes it. So, it looks there is a need.

The taxonomy picker extensión maybe looks a bit cumbersone, but the point was that was hard for us to avoid NHibernate producing queries due to lazy loading,. In a first attempt we skipped lazy loading of a property but later we discover another lazy property that allows to relate a content ítem with taxonomy terms also was doing lazy loading. Till we find we need an extra query to avoid also that lazy loading. That is the readon for the HQL there.
Extensions also have complexity because we pagintate queries to avoid the error due to max length of an sql and because we avoid to eagerly load twice the same content ítem.
Other changes in the core of orchard are to allow to set a value to a lazy property. With original code was not posible to set a value to a lazy property
Finally, another change was to make some private properties public, to skip the need of passing always the content manager by argument to the extensions.

src/Orchard/ContentManagement/DefaultContentManager.cs
```diff
@@ -78,7 +78,11 @@ public class DefaultContentManager : IContentManager {
         public ILogger Logger { get; set; }
 
         public IEnumerable<IContentHandler> Handlers {
-              get { return _handlers.Value; }
+            get { return _handlers.Value; }
+        }
+
+        public ITransactionManager TransactionManager {
```
sebastienros, line 84: Why is it necessary, you can already resolve it.

sebastienros:
Please rename to `LoadXXX` on content item extensions and `IncludeXXX` on Query extensions.

src/Orchard/ContentManagement/DefaultContentManager.cs
jersiovic: I need it in taxonomy extensions https://github.com/OrchardCMS/Orchard/pull/7861/files#diff-bb0c748b3dc7bb68f064a62685877941R35  There the only way I can get it is through an extra parameter on the extension. I avoided that getting the transaction from the content manager.

jersiovic:
After the change we will produce calls that mix `LoadXXX` and `IncludeXXX` extensions like following one, is it ok?
```
termsList.LoadMediaLibraryPickerFields(_contentManager).IncludeTaxonomyFields();
```

Jetski5822:
@jersiovic yo - did you try to apply these to the Menu?

jersiovic:
@Jetski5822 No, I didn't cause I cache menus to avoid performance problems using this module http://gallery.orchardproject.net/Packages/MainBit.Navigation. But i looks a good scenario to harness this eager loading extensions.

PR #7865, opened 2017-09-21
[Fixes #7862] Orchard.Taxonomies, allow autocomplete query to return results when no query term is passed

### LINKED ISSUE #7862: Taxonomy Field no longer displays all possible values
I recently upgraded a client's Orchard CMS installation from 1.9.x to 1.10.2. Everything went smoothly. However, we are missing some functionality that was previously available.

Before, to select a topic (Taxonomy Field), I would click in the Topics field and a box with all available topics would appear. I would then click on the desired topic. To add more topics, I just had to repeat.

Now, however, clicking on the field does nothing. A cursor appears in the box, but no topic window appears...

...until I start typing. But here’s the catch. If I type an ‘a’ only topic names that contain an ‘a’ are shown. Any topic that does not contain an ‘a’ is not shown.

If I know the topic I want, no problem. I simply start typing it, and the topics list will filter itself accordingly and I just need to click the topic to add. Repeat as needed.

The problem with this is that if I don’t know the topic the I want, there’s no way to easily know what topics are available. And if I wasn’t aware that the list was filtered based on the letter I typed, I might assume that a certain topic wasn’t entered into our system.

Is it possible to go back to the old way the Taxonomy Field was set up?

Thanks.

ViRuSTriNiTy:
...to be able to select a taxonomy without the need to enter something into the autocomplete field, see issue #7862


sebastienros:
Does it show everything then by default?

ViRuSTriNiTy:
@sebastienros Yes, but only if you click into the field (set focus to the field).

Issue #1067, opened 2011-01-18
Layer Rules will not allow for multiple Url "value" 


orchardbot:
__ChanceUSC__ created:
https://orchard.codeplex.com/workitem/17235

The Layer Rules do not appear to allow for multiple Url values regardless of the syntax or logical statement. In either event, the following error is presented:

"The rule is not valid: Syntax error: Unexpected token in primary expression (Token &amp;#39;StringLiteral&amp;#39; ("[routevalue]") at position 34)" 

where [routevalue] is the second value passed to Url

PR #7868, opened 2017-09-22
Adds support for multiple urls in urlcondition

andrewcartwright1:
I noticed that the UrlCondition layer rule was limited to a single url and came across an old issue for this: https://github.com/OrchardCMS/Orchard/issues/1067

This adds support for checking each argument provided instead of the first argument.

This allows rules such as `url("~/this", "~/that", "~/who")` can be used and will not cause an exception when trying to save or evaluate.


sebastienros:
Would be nice of you to update the documentation now. Thanks.

andrewcartwright1:
Done: https://github.com/OrchardCMS/OrchardDoc/pull/344 😃

PR #7892, opened 2017-11-10
5408: Dropping static view compilation temporarily to be able support Dynamic Compilation with Roslyn

BenedekFarkas:
Also includes cherrypicked commit 48685b3ed7f3ed4f40d1081f0bc937a4e3b2338f to fix ClickToBuild.cmd.

### DIFF
--- a/ClickToBuild.cmd
+++ b/ClickToBuild.cmd
@@ -1,15 +1,26 @@
-for /f "usebackq tokens=*" %%i in (`lib\vswhere\vswhere -latest -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -property installationPath`) do (
-  set InstallDir=%%i
+@echo off
+
+for /f "usebackq tokens=1* delims=: " %%i in (`lib\vswhere\vswhere -latest -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild`) do (
+  if /i "%%i"=="installationPath" (
+	set InstallDir=%%j
+	if exist "%InstallDir%\MSBuild\15.0\Bin\MSBuild.exe" (
+		echo "Using MSBuild from Visual Studio 2017"
+		set msbuild="%InstallDir%\MSBuild\15.0\Bin\MSBuild.exe"
+		goto build
+	)
+  )
 )
 
-
 FOR %%b in (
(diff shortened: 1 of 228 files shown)


sebastienros:
I believe you are still trying to find a better solution, that might involve fixing the codedomprovider library.

BenedekFarkas:
Yep, already have a proof of concept with a "hacked" piece of code, but I think I'm close to a nice (extensible) solution.

sebastienros:
Blocked on the CodeDomProvider PR to be accepted

sebastienros:
What if a module doesn't have this

```
<compiler language="c#;cs;csharp" extension=".cs" type="Microsoft.CodeDom.Providers.DotNetCompilerPlatform.CSharpCodeProvider, Microsoft.CodeDom.Providers.DotNetCompilerPlatform" warningLevel="4" compilerOptions="/langversion:6 /compilerPath:..\..\bin\roslyn"/>
```

Is it breaking existing modules from the gallery? Or the ones you'd have in your project?

sebastienros:
Can you revert the previous change that is breaking 1.10.x in the meantime, and work on a PR that provides all the changes?

BenedekFarkas:
> What if a module doesn't have this
> ...
> Is it breaking existing modules from the gallery? Or the ones you'd have in your project?

When writing C# 6 or above code in a Razor template in a module that doesn't have this configured, you'll see an error indicating that those language features are not available (but Razor IntelliSense is still operational), which is the same behavior as before. But it's OK in runtime, because Orchard.Web is responsible for that (just like with Dynamic Compilation). I'll go for a test run in Azure with this, but it looks OK locally.

> Can you revert the previous change that is breaking 1.10.x in the meantime, and work on a PR that provides all the changes?

So we should revert what exactly? I'm asking because the current problem is Roslyn Dynamic Compilation vs. Static View Compilation (with the current CodeDomProvider package). We can safely release with one of them, but not both (without the CodeDomProvider fix). I'd vote for Roslyn Dynamic Compilation.

sebastienros:
Merge what you think is right as long as it's not breaking existing modules from the gallery, and we don't require symlinks.

Issue #7890, opened 2017-11-06
Lucene index settings lock file is not properly released and cannot be accessed by application


0liver:
We're running around 20 sites off of a single code base that uses Orchard as its foundation. After every deployment (to all these sites) there is a handful of sites on which we have a locked **.lock** file on one of our indices. 98% of the time it's the same one (Managemen_Latest).

Here's the stacktrace:
```
System.IO.IOException: The process cannot access the file
'[...]\App_Data\Sites\Default\Management_Latest.settings.xml.lock' because it is being used by another process.
(Tried to aquire the index lock file 'Sites/Default/Management_Latest.settings.xml.lock'!)

at System.IO.__Error.WinIOError(Int32, System.String)
at System.IO.FileStream.Init(System.String, System.IO.FileMode, System.IO.FileAccess, Int32, Boolean, System.IO.FileShare, Int32, System.IO.FileOptions, SECURITY_ATTRIBUTES, System.String, Boolean, Boolean, Boolean)
at System.IO.FileStream..ctor(System.String, System.IO.FileMode, System.IO.FileAccess, System.IO.FileShare, Int32, System.IO.FileOptions, System.String, Boolean, Boolean, Boolean)
at System.IO.StreamReader..ctor(System.String, System.Text.Encoding, Boolean, Int32, Boolean)
at System.IO.File.InternalReadAllText(System.String, System.Text.Encoding, Boolean)
at Orchard.FileSystems.LockFile.DefaultLockFileManager.IsLockedImpl(System.String) in C:\TeamCity\buildAgent\work\8eded68158648c8\src\Orchard\FileSystems\LockFile\DefaultLockFileManager.cs:line 64
at Orchard.FileSystems.LockFile.DefaultLockFileManager.TryAcquireLock(System.String, Orchard.FileSystems.LockFile.ILockFile ByRef) in C:\TeamCity\buildAgent\work\8eded68158648c8\src\Orchard\FileSystems\LockFile\DefaultLockFileManager.cs:line 24
at Discoverize.Management.Indexing.IndexingTaskExecutorDz.UpdateIndexBatch(System.String)
at Orchard.Indexing.Services.UpdateIndexScheduler.UpdateIndex(System.String)
[...]

PR #7897, opened 2017-11-21
Fix issue #7890: Lucene index settings lock file

GiuseppeMusso-Laser:
Lucene index settings lock file is not properly released and cannot be accessed by application.
Fix as described in issue #7890 itself.

### DIFF
--- a/src/Orchard/FileSystems/AppData/AppDataFolder.cs
+++ b/src/Orchard/FileSystems/AppData/AppDataFolder.cs
@@ -113,11 +113,33 @@ namespace Orchard.FileSystems.AppData {
 
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times")]
         public void CreateFile(string path, string content) {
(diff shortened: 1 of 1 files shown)


src/Orchard/FileSystems/AppData/AppDataFolder.cs
```diff
@@ -113,13 +113,35 @@ public class AppDataFolder : IAppDataFolder {
...
+            if (File.Exists(filePath)) {
+                File.WriteAllText(filePath, content);
+            } else {
+                WriteAllTextWithTempFileAndMove(filePath, content);
             }
         }
 
+        private static void WriteAllTextWithTempFileAndMove(string path, string content) {
+            // generate a temp filename
+            var tempPath = path + "." + Path.GetRandomFileName();
```
sebastienros: This should be created in the environment temporary folder, just in case they are not deleted correctly and don't pollute the app_data.

GiuseppeMusso-Laser:
I moved the creation of temporary file in temporary folder using Path.GetTempFileName() to ensure a unique name.

PR #7911, opened 2017-12-13
Removed LocalizationPart

### LINKED ISSUE #7721: LocalizationPart is in Page even when Localization Module is not enabled
THis should be easy to fix.

The Default recipe adds LocalizationPart to Page, even though it does not enable the Localization Module.
Might be worth spending a little bit double-checking the recipes, and deciding whther it makes sene to change the migration in the localization module to add a LocalizationPart to Page when the feature is enabled (I think that would not be correct).

MatteoPiovanelli:
Removed LocalizationPart form ContentTypes defined in the Default and Blog recipes, because the Localization feature was not enabled there.

This finally should fix #7721

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Setup/Recipes/blog.recipe.xml
+++ b/src/Orchard.Web/Modules/Orchard.Setup/Recipes/blog.recipe.xml
@@ -26,14 +26,12 @@
             <Page ContentTypeSettings.Draftable="True" TypeIndexing.Indexes="Search">
                 <TitlePart/>
                 <TagsPart />
-                <LocalizationPart />
                 <AutoroutePart />
                 <MenuPart />
             </Page>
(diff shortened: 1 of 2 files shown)


sebastienros:
The issue is that when the feature is enabled then nothing is localizable. Are you ok with that?

MatteoPiovanelli:
I am, the reason being that I feel enabling localization should either make everything localizable, or nothing at all. As it is you would end up with localizable Pages and Blogs, and by default nothing else. There's no real reason why those two ContentTypes should behave differently compared to everything else.

MatteoPiovanelli:
Actually, let me open a separate issue regarding adding LocalizationPart to everything....

sebastienros:
I will ask for a vote on Tuesday.

MatteoPiovanelli:
Was this ever discussed, or did it fall under the radar entirely?

sebastienros:
Nobody cared, so you can do it

PR #7912, opened 2017-12-14
Fix import export of identically named fields

### LINKED ISSUE #7773: Import/export doesn't handle identically-named fields across parts
For [this XML](https://github.com/OrchardCMS/Orchard/files/1138438/export.xml.txt), if I create an item with values for all fields, only the last set of values is exported. I.e. if I export both content and definition, this is what I get:

```xml
  <ContentDefinition>
    <Types>
      <DetailContact ...>
        <AdditionalApplicationsContactPart />
        <AdditionalApprenticeshipProgramContactPart />
        <AdditionalBidsContactPart />
        ...
      </DetailContact>
    </Types>
    <Parts>
      <AdditionalApplicationsContactPart ...>
        <FirstName.TextField DisplayName="First Name" TextFieldSettings.Required="False" />
        ...
      </AdditionalApplicationsContactPart>
      <AdditionalApprenticeshipProgramContactPart ...>
        <FirstName.TextField DisplayName="First Name" TextFieldSettings.Required="False" />
        ...
      </AdditionalApprenticeshipProgramContactPart>
      <AdditionalBidsContactPart ...>
        <FirstName.TextField DisplayName="First Name" TextFieldSettings.Required="False" />
        ...
      </AdditionalBidsContactPart>
      ...
    </Parts>
  </ContentDefinition>
  <Content>
    <DetailContact Id="/alias=test-detail" Status="Published">
      <TextField.FirstName Text="First 3" />
      ...
    </DetailContact>
  </Content>
```

The issue in the above is that the item has fields with three different "first name" values, and only the last is exported. As far as I can see, it works fine in Orchard (editing, viewing).

From @Xceno : 

> I think the problem is right here. `ContentFields` should probably be prefixed with the partname
>
```C#
// Orchard.ContentManagement.Handlers.ExportContentHandler.cs
public XElement Element(string elementName) {
            var element = Data.Element(elementName);
            if (element == null) {
                element = new XElement(elementName);
                Data.Add(element);
            }
            return element;
        }
```
ref: [slack](https://gitter.im/OrchardCMS/Orchard?at=5964c928bf7e6af22cc891c9)

> I've created a blank site with your exact configuration (`DetailContact` with 3 Parts, each having a `FirstName` field). I can confirm the export issue and it's indeed a problem with the fieldname. So `ExportContentHandler.Element(myFieldName)` will always return the existing field in the context and `SetAttributeValue()` overrides it.
>
> A version attribute in the XML could be used for a fallback but I'd rather not introduce a "legacy import/export api".
> I think it should be sufficient enough to check if the currently imported field has a prefix equal to a `ContentPart` name and otherwise proceed the old, "buggy" way.
> Though I haven't tried that out yet.
> The generated XML will look like this btw:
>
```xml
<DetailContact Id="/Identifier=63638ee501024f26a1d2c30da524cdac" Status="Published">
      <AdditionalApplicationsContactPart.TextField.FirstName Text="first value" />
      <AdditionalApprenticeshipProgramContactPart.TextField.FirstName Text="second value" />
      <AdditionalBidsContactPart.TextField.FirstName Text="third value" />
      <CommonPart Owner="/User.UserName=admin" CreatedUtc="2017-07-11T14:30:28.3681873Z" PublishedUtc="2017-07-11T14:30:28.4651858Z" ModifiedUtc="2017-07-11T14:30:28.4581861Z" />
      <IdentityPart Identifier="63638ee501024f26a1d2c30da524cdac" />
    </DetailContact>
```
ref: [slack](https://gitter.im/OrchardCMS/Orchard?at=5964e567bc46472974db6fac)

chanond-w:
Fix #7773


sebastienros:
Can you confirm that it will still work for existing recipes that don't have the prefix? I saw the code but I want to be certain this is what this is doing.

chanond-w:
I've tested by creating a new site by `default` recipe and import `blog` recipe in. It worked as expected.

chanond-w:
@sebastienros I'm sorry I forgot to mentioned you. 😄

PR #7913, opened 2017-12-14
Fixing the FileSystemOutputCache feature

### LINKED ISSUE #6115: File System Output Cache can generate file paths that are too long - can't access the page at all
Depending on the directory structure on the host and the parameters for a particular request, the file system output cache feature can generate file paths that are too long to handle. The result when this happens is an exception and the inability to access the page.

image

The actual exception is below but gets caught and thrown as what you see above: 

```
Path.GetFullPath(mappedPath).StartsWith(Path.GetFullPath(basePath), StringComparison.OrdinalIgnoreCase) 'Path.GetFullPath(mappedPath).StartsWith(Path.GetFullPath(basePath), StringComparison.OrdinalIgnoreCase)' threw an exception of type 'System.IO.PathTooLongException'   bool {System.IO.PathTooLongException}
```

I understand that the technique used for this feature relies on embedding the parameters in the file name. But it would be much better if a check was added to prevent this failure so even if the page can't be cached, it will at least allow access to the non-cached page.

sebastienros:
Fixes #6115

- Caching keys for filenames to prevent too long paths
- Separating metadata from content storage to optimize some scenarios
- Fixing retry logic

TODO: Don't use BinaryFormatter to serialize the output

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Services/FileSystemOutputCacheBackgroundTask.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Services/FileSystemOutputCacheBackgroundTask.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System.IO;
+using System.Linq;
 using Orchard.Caching;
 using Orchard.Environment.Configuration;
 using Orchard.Environment.Extensions;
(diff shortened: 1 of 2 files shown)


sebastienros:
Hey @Jetski5822 ! You should tell if that worked for you so I can merge.

jonvee:
Tested and it fixed encoding issues with long blog post URLs as well as Tags with multi-word keywords

PR #7920, opened 2017-12-18
Fix #7903 adding the appropriate columns to index fields latest values

### LINKED ISSUE #7903: Fields are not indexed if content is not published
At the present a field is indexed only when a content is published. The module which is responsible of indexing fields, is the projections module via FieldIndexPartHandler.OnPublishing. Now I understand why the index is populated only on publishing because we normally use it in queries/projections in the front-end of a site. But if I would like to use it in back-end? 
I've this use case:
2 Content Types Event (Event 1 published, Event 2 unpblished, Event 3 unpblished) and Edition (Spring 2017 published, Summer 2017 unpublished): Event has a Content Picker Field referencing an Edition (Event 1 > Spring 2017, Event 2,3 > Summer 2017).
What I would like to achieve is that when I publish the Edition "Summer 2017" I want to publish Event 2 and 3 too; but I cannot search for those events because the index is not yet there to be searched for and searching the field into infoset is not doable due to perfomance issues when contents are a lot.
The idea is to make the index tables (StringFieldIndexRecord and so on) "versionable" relating them to ContentItemVersionRecord table; in alternative we could index published and latest version.
Is there another way to achieve my goal?

HermesSbicego-Laser:
Fix #7903
- Adds the column LatestValue to all type of FieldIndexRecord
- Adds Indexes to increase performances in queries
- Adds Handler to store LatestValue during Update event
- Adds Service to Get and Set the LatestValue of a field
- Adds heavy projections index tables update to the Upgrade module

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
@@ -16,11 +16,13 @@ namespace Orchard.Projections.Handlers {
         private readonly IFieldIndexService _fieldIndexService;
         private readonly IFieldStorageProvider _fieldStorageProvider;
         private readonly IEnumerable<IContentFieldDriver> _contentFieldDrivers;
+        private readonly IDraftFieldIndexService _draftFieldIndexService;
 
         public FieldIndexPartHandler(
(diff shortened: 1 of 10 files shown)


sebastienros:
Looks like this PR is duplicated

HermesSbicego-Laser:
closed in favor of #7923

Issue #7921, opened 2017-12-19
Improvement: To add the ability to choose to execute query over Published (default) or Latest version


HermesSbicego-Laser:
This is related to #7903
How I imagine this should be implemented:
1. In the QueryPartRecord Add a property VersionScope of type string; and manage it during the query Edit (while edtiing the Query ContentItem)
2. In ProjectionManager.GetContentQueries  alter the logic using this new property or using the PubishedVersion for backward compatibility
3. In all IFieldTypeEditor implementations use LatestValue (see #7920) if the QueryPartRecord.VersionScope is set to "Latest" 

Could it work?

HermesSbicego-Laser:
This could be closed?

PR #7923, opened 2017-12-20
Implements #7921 extend query to support latest versions

HermesSbicego-Laser:
Implements #7921, branched from #7920 cause we need the LatestValue column to be there.
When a query is created/edited it is possible to specify the Content Version to query (Published or Latest).
Based on that choice the HqlQuery will filter 
- the right VersionOption 
- the right column of the `(*)IndexFilterRecord`

Based on that choice the HqlQuery will order
- using the right column of the `(*)IndexFilterRecord`

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
@@ -16,11 +16,13 @@ namespace Orchard.Projections.Handlers {
         private readonly IFieldIndexService _fieldIndexService;
         private readonly IFieldStorageProvider _fieldStorageProvider;
         private readonly IEnumerable<IContentFieldDriver> _contentFieldDrivers;
+        private readonly IDraftFieldIndexService _draftFieldIndexService;
 
         public FieldIndexPartHandler(
             IContentDefinitionManager contentDefinitionManager,
             IRepository<FieldIndexPartRecord> repository,
             IFieldIndexService fieldIndexService,
+            IDraftFieldIndexService draftFieldIndexService,
             IFieldStorageProvider fieldStorageProvider,
             IEnumerable<IContentFieldDriver> contentFieldDrivers) {
             Filters.Add(StorageFilter.For(repository));
@@ -28,7 +30,8 @@ namespace Orchard.Projections.Handlers {
             _fieldIndexService = fieldIndexService;
             _fieldStorageProvider = fieldStorageProvider;
             _contentFieldDrivers = contentFieldDrivers;
-
+            _draftFieldIndexService = draftFieldIndexService;
+            OnUpdated<FieldIndexPart>(Updated);
             OnPublishing<FieldIndexPart>(Publishing);
         }
 
@@ -43,24 +46,56 @@ namespace Orchard.Projections.Handlers {
                 context.Builder.Weld<FieldIndexPart>();
             }
         }
+        private void Updated(UpdateContentContext context, FieldIndexPart fieldIndexPart) {
+            if (context.UpdatingItemVersionRecord.Latest) { // updates projection draft indexes only if it is the latest version
+                foreach (var part in fieldIndexPart.ContentItem.Parts) {
+                    foreach (var field in part.PartDefinition.Fields) {
+
+                        // get all drivers for the current field type
+                        // the driver will describe what values of the field should be indexed
+                        var drivers = _contentFieldDrivers.Where(x => x.GetFieldInfo().Any(fi => fi.FieldTypeName == field.FieldDefinition.Name)).ToList();
+
+                        ContentPart localPart = part;
+                        ContentPartFieldDefinition localField = field;
+                        var membersContext = new DescribeMembersContext(
+                            (storageName, storageType, displayName, description) => {
+                                var fieldStorage = _fieldStorageProvider.BindStorage(localPart, localField);
+
+                                // fieldStorage.Get<T>(storageName)
+                                var getter = typeof(IFieldStorage).GetMethod("Get").MakeGenericMethod(storageType);
+                                var fieldValue = getter.Invoke(fieldStorage, new[] { storageName });
+                                _draftFieldIndexService.Set(fieldIndexPart,
+                                    localPart.PartDefinition.Name,
+                                    localField.Name,
+                                    storageName, fieldValue, storageType);
+                            });
+
+                        foreach (var driver in drivers) {
+                            driver.Describe(membersContext);
+                        }
+                    }
+                }
+            }
+        }
+
 
         public void Publishing(PublishContentContext context, FieldIndexPart fieldIndexPart) {
             foreach (var part in fieldIndexPart.ContentItem.Parts) {
-                foreach(var field in part.PartDefinition.Fields) {
-                    
+                foreach (var field in part.PartDefinition.Fields) {
+
                     // get all drivers for the current field type
                     // the driver will describe what values of the field should be indexed
                     var drivers = _contentFieldDrivers.Where(x => x.GetFieldInfo().Any(fi => fi.FieldTypeName == field.FieldDefinition.Name)).ToList();
-                    
+
                     ContentPart localPart = part;
                     ContentPartFieldDefinition localField = field;
-                    var membersContext = new DescribeMembersContext( 
+                    var membersContext = new DescribeMembersContext(
                         (storageName, storageType, displayName, description) => {
                             var fieldStorage = _fieldStorageProvider.BindStorage(localPart, localField);
 
                             // fieldStorage.Get<T>(storageName)
                             var getter = typeof(IFieldStorage).GetMethod("Get").MakeGenericMethod(storageType);
-                            var fieldValue = getter.Invoke(fieldStorage, new[] {storageName});
+                            var fieldValue = getter.Invoke(fieldStorage, new[] { storageName });
 
                             _fieldIndexService.Set(fieldIndexPart,
                                 localPart.PartDefinition.Name,
--- a/src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using Orchard.ContentManagement;
 using Orchard.ContentManagement.Drivers;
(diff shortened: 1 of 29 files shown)


src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
```diff
@@ -282,5 +281,38 @@ public class Migrations : DataMigrationImpl {
...
+
+            SchemaBuilder.AlterTable("DoubleFieldIndexRecord", table => table.CreateIndex("IX_PropertyName", new string[] { "PropertyName" }));
+            SchemaBuilder.AlterTable("DoubleFieldIndexRecord", table => table.CreateIndex("IX_FieldIndexPartRecord_Id", new string[] { "FieldIndexPartRecord_Id" }));
+
+            SchemaBuilder.AlterTable("DecimalFieldIndexRecord", table => table.CreateIndex("IX_PropertyName", new string[] { "PropertyName" }));
+            SchemaBuilder.AlterTable("DecimalFieldIndexRecord", table => table.CreateIndex("IX_FieldIndexPartRecord_Id", new string[] { "FieldIndexPartRecord_Id" }));
+
+            return 5;
+        }
+        public int UpdateFrom5() {
```
sebastienros: Move to `UpdateFrom4`. And update the `Create` method to create the table correctly including all subsequent changes, then return `5`.

src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
```diff
@@ -282,5 +281,38 @@ public class Migrations : DataMigrationImpl {
...
+            SchemaBuilder.AlterTable("IntegerFieldIndexRecord", table => table
+            .AddColumn<long>("LatestValue"));
+
+            SchemaBuilder.AlterTable("DoubleFieldIndexRecord", table => table
+            .AddColumn<double>("LatestValue"));
+
+            SchemaBuilder.AlterTable("DecimalFieldIndexRecord", table => table
+            .AddColumn<decimal>("LatestValue"));
+
+            //Adds indexes for better performances in queries
```
sebastienros, line 297: Two comments:
- We need to talk about the "necessity" of these indices
- You are reusing the same index name multiple times, is it an issue? should it be unique per database, or no otherwise table prefix would break it?

src/Orchard.Web/Modules/Orchard.Projections/Providers/Filters/ContentFieldsFilter.cs
```diff
@@ -75,7 +75,9 @@ public class ContentFieldsFilter : IFilterProvider {
             var relationship = fieldTypeEditor.GetFilterRelationship(propertyName.ToSafeName());
 
             // generate the predicate based on the editor which has been used
-            Action<IHqlExpressionFactory> predicate = fieldTypeEditor.GetFilterPredicate(context.State);
+            dynamic fullState = context.State;
```
sebastienros, line 78: Why `dynamic` ? You can't access the `VersionScope` property directly?

src/Orchard.Web/Modules/Orchard.Projections/Providers/SortCriteria/ContentFieldsSortCriteria.cs
```diff
@@ -79,11 +79,11 @@ public class ContentFieldsSortCriterion : ISortCriterionProvider {
...
             // apply where clause
             context.Query = context.Query.Where(relationship, predicate);
-            
+
             // apply sort
-            context.Query = ascending 
-                ? context.Query.OrderBy(relationship, x => x.Asc("Value")) 
-                : context.Query.OrderBy(relationship, x => x.Desc("Value"));
+            context.Query = ascending
+                ? context.Query.OrderBy(relationship, x => x.Asc(context.VersionScope == QueryVersionScopeOptions.Latest ? "LatestValue" : "Value"))
```
sebastienros: Maybe create a method in the context to get the field name (based on the scope). I have seen this multiple times.

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
```diff
@@ -168,7 +168,8 @@ public class ProjectionManager : IProjectionManagerExtension {
             foreach (var sortCriterion in queryRecord.SortCriteria.OrderBy(s => s.Position)) {
                 var sortCriterionContext = new SortCriterionContext {
                     Query = groupQuery,
-                    State = FormParametersHelper.ToDynamic(sortCriterion.State)
+                    State = FormParametersHelper.ToDynamic(sortCriterion.State),
+                    VersionScope = queryRecord.VersionScope
```
sebastienros: Instead of duplicating this property in all context objects, can't we just pass the parent Query which already  has the value?

src/Orchard.Web/Modules/Orchard.Projections/Views/EditorTemplates/Parts/QueryPart_Edit.cshtml
```diff
@@ -0,0 +1,8 @@
+﻿@model Orchard.Projections.ViewModels.QueryViewModel
+@using Orchard.Projections;
+
+<fieldset>
+    @Html.LabelFor(m => m.VersionScope, T("Query Version Scope"))
```
sebastienros: "Version Scope" is not very intuitive. "Versions" only seems better. What do we use in Lucene indices?

src/Orchard.Web/Modules/Upgrade/Controllers/ProjectionsController.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (!_featureManager.GetEnabledFeatures().Any(x => x.Id == "Orchard.Fields")) {
+                _orchardServices.Notifier.Warning(T("You need to enable Orchard.Fields in order to migrate current fields."));
+            }
+
+            return View(viewModel);
+        }
+
+        [HttpPost, ActionName("Index")]
+        public ActionResult IndexPOST() {
```
sebastienros: Shouldn't it do it by batch of content items? If there are too many of them this will fail. There are some examples doing it in other upgrade controllers.

src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
```diff
@@ -43,24 +46,56 @@ public class FieldIndexPartHandler : ContentHandler {
                 context.Builder.Weld<FieldIndexPart>();
             }
         }
+        private void Updated(UpdateContentContext context, FieldIndexPart fieldIndexPart) {
+            if (context.UpdatingItemVersionRecord.Latest) { // updates projection draft indexes only if it is the latest version
+                foreach (var part in fieldIndexPart.ContentItem.Parts) {
```
sebastienros: Try to externalize this method, and use a lambda as the differences.

src/Orchard.Web/Modules/Orchard.Projections/Services/DraftFieldIndexService.cs
```diff
@@ -0,0 +1,125 @@
+﻿using System;
+using System.Linq;
+using Orchard.Projections.Models;
+
+namespace Orchard.Projections.Services {
+    public class DraftFieldIndexService : IDraftFieldIndexService {
```
sebastienros, line 6: I understand this is to prevent a breaking change. But I'd prefer to see it merged into the existing service in the dev branch. Can you file an issue for this? I will assign it for 1.11.

src/Orchard.Web/Modules/Orchard.Projections/Handlers/FieldIndexPartHandler.cs
HermesSbicego-Laser: OK, I will try...

src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
HermesSbicego-Laser: Ok, I will threat this PR and the #7920 one as they were the same PR, so I can aggregate UpdateFrom4 and 5

src/Orchard.Web/Modules/Orchard.Projections/Providers/Filters/ContentFieldsFilter.cs
HermesSbicego-Laser: fieldTypeEditor.GetFilterPredicate (2 lines above) method use a dynamic parameter as input. I need that information within IFieldTypeEditor.GetFilterPredicate in order to return the right column (Value or LatestValue) to query.

src/Orchard.Web/Modules/Orchard.Projections/Providers/SortCriteria/ContentFieldsSortCriteria.cs
HermesSbicego-Laser: Ok, i will try

src/Orchard.Web/Modules/Orchard.Projections/Services/DraftFieldIndexService.cs
HermesSbicego-Laser: Ok, i will do it.

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
HermesSbicego-Laser: Do you mean to add the QueryPart to the context?

src/Orchard.Web/Modules/Orchard.Projections/Views/EditorTemplates/Parts/QueryPart_Edit.cshtml
HermesSbicego-Laser: Ok for "Versions", or "Content versions"?

src/Orchard.Web/Modules/Upgrade/Controllers/ProjectionsController.cs
HermesSbicego-Laser: Can you address me to an example where I can copy from? thanks.

src/Orchard.Web/Modules/Orchard.Projections/Migrations.cs
HermesSbicego-Laser: - Necessity of these indexes: do you mean SQL indexes or the "LatestValue" columns? For SQL indexes, I think at least `IX_FieldIndexPartRecord_Id` should be created because it is a foreign key "in fact". The index over PropertyName column is for projections and HqlQueries so I think they should be there. I don't see a good reason why they shouldn't be there
- Same index name: I seem to remember that Indexes name should be unique per table and not per database. At least in SQL Server, so i don't see an issue here. Perhaps it's not the same in MySql or Postgre?

src/Orchard.Web/Modules/Upgrade/Controllers/ProjectionsController.cs
HermesSbicego-Laser: Found it by myself in InfosetController.cs

src/Orchard.Web/Modules/Orchard.Projections/Services/DraftFieldIndexService.cs
HermesSbicego-Laser: opened issue #7925

BenedekFarkas:
Does the code expect that the Query's version scope can only be Published or All? I'm asking because #8111.

HermesSbicego-Laser:
In this PR, the Query's version scope  can be Published or Latest.

HermesSbicego-Laser:
@BenedekFarkas I read a bit the code in your #8112 PR. Adding the filter on the right versionOption of the content is not sufficient. Should be changed other code. In particular where we test the VersionOption in order to choose the right column (Value/LatestValue) to check/order.
see:  
1. https://github.com/OrchardCMS/Orchard/pull/7923/files#diff-2e5f3c3cb7233e791056952d3b891828
2. https://github.com/OrchardCMS/Orchard/pull/7923/files#diff-718cbbbafe30601793f25355e00903f1
3. classes implementing IFieldTypeEditor

BenedekFarkas:
@HermesSbicego-Laser good point, I only saw the addition of LatestValue after submitting my PR - I'll look into it and update accordingly, thanks!

Issue #4540, opened 2014-05-23
Href("~/foo") eats the / when in tenant


orchardbot:
__qt1__ created:
https://orchard.codeplex.com/workitem/20711

Fix:
change
 src\Orchard\Mvc\ViewEngines\Razor\WebViewPage.cs

line 207

from 
                    return base.Href("~/" + _tenantPrefix + path.Substring(2), pathParts);

to
                    return base.Href("~/" + _tenantPrefix + path.Substring(_tenantPrefix==""?2:1), pathParts);

BTW: would it be more convenient to post this micro-patch via fork ?

PR #7930, opened 2018-01-05
7919: The Alias UI links for sites using UrlPrefix are broken

### LINKED ISSUE #7919: The Alias UI links for sites using UrlPrefix are broken
The UrlPrefix is manually added to the alias and root urls  but the helper methods do that automatically so the links are broken

Hazzamanic:
Fixes https://github.com/OrchardCMS/Orchard/issues/7919. Also reverting this commit's changes to the Href method (https://github.com/OrchardCMS/Orchard/commit/9c2ea8067094d9ed02f50bcd372bf445ff2e81ee#diff-3806ff11c9d923fac04777c83296a19e) that overrode this commit https://github.com/OrchardCMS/Orchard/commit/a4894930f054fa9521fb4ea98594dd60fea72152 that fixed this issue https://github.com/OrchardCMS/Orchard/issues/4540


sebastienros:
Looks good, but can someone try it and confirm it fixes the issue? I am worried as the code explicitly handled the prefix.

BenedekFarkas:
@sebastienros tested and slightly adjusted, good to go!

sebastienros (approved):

src/Orchard/Mvc/ViewEngines/Razor/WebViewPage.cs
```diff
@@ -190,11 +190,10 @@ public abstract class WebViewPage<TModel> : System.Web.Mvc.WebViewPage<TModel>,
...
 
-            if (!String.IsNullOrEmpty(_tenantPrefix)
-                && path.StartsWith("~/")  
-                && !CommonLocations.Any(gpp=>path.StartsWith(gpp, StringComparison.OrdinalIgnoreCase))
-            ) { 
-                    return base.Href("~/" + _tenantPrefix + path.Substring(2), pathParts);
+            if (!string.IsNullOrWhiteSpace(_tenantPrefix)
+                && path.StartsWith("~/")
+                && !CommonLocations.Any(gpp => path.StartsWith(gpp, StringComparison.OrdinalIgnoreCase))) {
+                return base.Href("~/" + _tenantPrefix + path.Substring(1), pathParts);
```
AndreaPiovanelli, line 196: Why is the substring start index parameter changed?
Since code gets to line 196 only if path starts with ~, it stripped ~/ in the previous version of the file; now it only strips the ~ character: is it correct keeping the / character?

BenedekFarkas: FYI this is only applied if there is a tenant prefix. e.g., the base URL is `mysite.com/mytenant`. In this case you still need the `/` between the tenant prefix and the rest of the path. Without this change all the Hrefs were broken if you provided a path that started with `~/` when there's a tenant prefix.

Issue #7843, opened 2017-09-09
Media ContentTypes Localization enhancements


HermesSbicego-Laser:
As per Taxonomies, ContentPickerFields and others contents (see #7352) we should consider to enhance localization of media's Content Types (Image, Documents, etc).
By now if you have a multilanguage site and you want to seriously manage SEO and w3c recommendations you have to do a lot of manual tricks in order to manage alternate textes, title, etc... (You have to upload a new media - even if it is the same of the master content - and then you can manage the media alt, title, etc)

I have some ideas on how we could manage it.

A feature "Media Localization Extensions"
1. It adds a LocalizationPart to interested Content Types (I have to go deeper in code to understand where it would be better to add it). 
  - Translating a "Media" should clone the media file too? I think yes, it should. What do you think?
  - In the list of media we should put in evidence the language of that media in order to not mislead the user.
2. It adds a Setting "Localize media on translating" to all MediaLibraryPickerFields (default is true). When a content - which contains a MediaLibraryPickerField - is translated and the setting is true 
  - all the medias already translated will be replaced with their translation
  - all the medias not trasnlated will be cloned, so user can write the translation of textes.
  - The Localization of Content and its medias should stay always in sync

@sebastienros @MatteoPiovanelli-Laser @jersiovic @BenedekFarkas @urbanit What do you think?

MatteoPiovanelli:
A couple comments to add to that.

Regarding cloning the actual files:
I don't know about that. Maybe they are already managed in such a way that there would be no need for that. On top of that, not al media have a corresponding file in storage.

Regarding the fields:
Basically what we did for ContentPickerField, so I don't see any issue there.

sebastienros:
I don't think media files should be cloned. But there needs to be a way to change which file is attached if the translation needs a different one.

sebastienros:
Otherwise I don't see how localization should be different than another content item for media. Some fields should be synced, content should be cloned, ...

BenedekFarkas:
Yeah, I agree, the files themselves should not be cloned (at least initially), so basically the MediaPart.Path property is translation-independent - I think what matters most here is the title (caption), which should also be cloned, but can be changed (translated).
[...]

Issue #7934, opened 2018-01-09
Duplicated media items


HermesSbicego-Laser:
Working on #7843 **branched from dev** I notice a bug in Media Library Picker Field.
repro steps:
1. Add a media library picker field to Page
2. In the field settings check "Allow multiple content items"
3. Create a new Page, Add 1 media 
4. Publish
5. Add a second media
this break the media items in the UI and even in the form: the second media results 2 or more times in the list

HermesSbicego-Laser:
Found the bug, I will work on it and I will PR a fix

PR #7935, opened 2018-01-10
New feature "Orchard.MediaLibrary.LocalizationExtensions"

HermesSbicego-Laser:
Implements what opened with #7843 

"Orchard.MediaLibrary.LocalizationExtensions" feature:
- MediaLocalizationMigrations: Adds LocalizationPart to all Media stereotyped types
- MediaLibraryPickerFieldLocalizationExtensionHandler: Updates the media items within the field according to field settings
- Settings for translation behaviours for MediaLibraryPickerField
- Adds warning on localizing media items
- form filter media by culture

"MediaLibrary" feature:
- Implemented Cloning methods of all medias PartDrivers
- MediaPartHandler: Adds a check to prevent file system media deletion when the file is referenced by other medias
- ClientStorageController: Adds a check to prevent file system media deletion when the file is referenced by other medias

"Localization" feature:
- Adds a logic to propagate returnurl parameter to the redirection url
- ILocalizationService: Adds a new method to set the culture of a content and at the same time the master content of it
- Propagate returnurl parameter if necessary

"Search" feature:
- updated json of media items called via ajax

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
@@ -6,6 +6,7 @@ using Orchard.Localization.Services;
 using Orchard.UI.Notify;
 using System;
 using System.Web.Mvc;
+using System.Linq;
 
 namespace Orchard.Localization.Controllers
 {
@@ -72,6 +73,14 @@ namespace Orchard.Localization.Controllers
             Services.Notifier.Success(T("Successfully cloned. The translated content was saved as a draft."));
 
             var editorRouteValues = _contentManager.GetItemMetadata(contentItemTranslation).EditorRouteValues;
+            // adds request variables of current controller to the new redirect route 
+            // for example the returnUrl parameter
+            foreach (var key in Request.Form.AllKeys.Where(x=> !x.StartsWith("__") && !editorRouteValues.Keys.Contains(x))) {
+                editorRouteValues.Add(key, Request.Form[key]);
+            }
+            foreach (var key in Request.QueryString.AllKeys.Where(x => !x.StartsWith("__") && !editorRouteValues.Keys.Contains(x))) {
+                editorRouteValues.Add(key, Request.QueryString[key]);
+            }
             return RedirectToRoute(editorRouteValues);
         }
     }
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/ClientStorageController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/ClientStorageController.cs
@@ -183,8 +183,18 @@ namespace Orchard.MediaLibrary.Controllers {
                     // Raise update and publish events which will update relevant Media properties
                     _handlers.Invoke(x => x.Updating(new UpdateContentContext(replaceMedia.ContentItem)), Logger);
 
-                    _mediaLibraryService.DeleteFile(replaceMedia.FolderPath, replaceMedia.FileName);
-                    _mediaLibraryService.UploadMediaFile(replaceMedia.FolderPath, replaceMedia.FileName, file.InputStream);
+                    var mediaItemsUsingTheFile = Services.ContentManager.Query<MediaPart, MediaPartRecord>()
+                                                                .ForVersion(VersionOptions.Latest)
+                                                                .Where(x => x.FolderPath == replaceMedia.FolderPath && x.FileName == replaceMedia.FileName)
+                                                                .Count();
+                    if (mediaItemsUsingTheFile == 1) { // if the file is referenced only by the deleted media content, the file too can be removed.
+                        _mediaLibraryService.DeleteFile(replaceMedia.FolderPath, replaceMedia.FileName);
+                    }
+                    else {
+                        // it changes the media file name
+                        replaceMedia.FileName = filename;
+                    }
+                    _mediaLibraryService.UploadMediaFile(replaceMedia.FolderPath, filename, file.InputStream);
                     replaceMedia.MimeType = mimeType;
 
                     _handlers.Invoke(x => x.Updated(new UpdateContentContext(replaceMedia.ContentItem)), Logger);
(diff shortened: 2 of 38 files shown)


src/Orchard.Web/Modules/Orchard.MediaLibrary/Handlers/MediaLibraryPickerFieldLocalizationExtensionHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+                                else {
+                                    _orchardServices.Notifier.Warning(T(
+                                        "{0}: the media item {1} has been removed from the field because culture neutral",
+                                        field.DisplayName,
+                                        mediaItem.As<MediaPart>().FileName));
+                                }
+                            }
+                        }
+                        if (mediaIds.Count > 0) {
+                            field.Ids = mediaIds.Distinct().ToArray();
```
HermesSbicego-Laser: @sebastienros I don't know if we should do distinct() here. Should we allow duplicated media items in the field? I added the distinct here because of #7934, but I can remove it.

src/Orchard.Web/Modules/Orchard.Localization/Controllers/AdminController.cs
```diff
@@ -72,6 +73,14 @@ public class AdminController : Controller {
...
 
             var editorRouteValues = _contentManager.GetItemMetadata(contentItemTranslation).EditorRouteValues;
+            // adds request variables of current controller to the new redirect route 
+            // for example the returnUrl parameter
+            foreach (var key in Request.Form.AllKeys.Where(x=> !x.StartsWith("__") && !editorRouteValues.Keys.Contains(x))) {
+                editorRouteValues.Add(key, Request.Form[key]);
+            }
+            foreach (var key in Request.QueryString.AllKeys.Where(x => !x.StartsWith("__") && !editorRouteValues.Keys.Contains(x))) {
+                editorRouteValues.Add(key, Request.QueryString[key]);
+            }
```
HermesSbicego-Laser, line 83: In media library the edit button adds a "ReturnUrl" QueryString in order to have a redirection to the caller page when saving/cancelling operation.
The translate link has the prop "Unsafe", so it transforms the request from get + querystring in post + body . So I propagate those parameters through redirection translate (which clone) > edit content, in order to be able to manage the right returnurl (and other parameters if they are there). To restrict the logic to only the "redirecturl" parameter, could be less flexible but more safe. @sebastienros what do you prefer?

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/LocalizedMediaController.cs
```diff
@@ -0,0 +1,153 @@
...
+                Shape = Services.ContentManager.BuildDisplay(x.ContentItem, "Thumbnail")
+            }).ToList();
+
+            var viewModel = new MediaManagerMediaItemsViewModel {
+                MediaItems = mediaItems,
+                MediaItemsCount = mediaPartsCount,
+                FolderPath = folderPath
+            };
+            return View(viewModel);
+        }
```
HermesSbicego-Laser, line 99: I have replicated the logic within Orchard.Search, even if I don't like so much the results because filters from different features can't works together. Basically I have not introduced anything worse than it already was, so i ask you if you can merge it anyway. 
But we should consider to change the current approach; I don't know if it is possible but:
- the responsible of acting the filter should be the medialibrary module
- other modules can partecipate in building the where clause of the query (as per projections)
- every change in the filters form/folders navigation fires a medialibrary controller responsible of combining the queries and then exposing the result.
@sebastienros if you agree I will file a new issue describing the problem and proposing the above solution.

sebastienros:
Looks good, waiting for a demo.
And hopefully someone else will try it.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Handlers/MediaPartHandler.cs
```diff
@@ -113,7 +116,13 @@ public class MediaPartHandler : ContentHandler {
 
         protected void RemoveMedia(MediaPart part) {
             if (!string.IsNullOrEmpty(part.FileName)) {
-                _mediaLibraryService.DeleteFile(part.FolderPath, part.FileName);
+                var mediaItemsUsingTheFile = _contentManager.Query<MediaPart, MediaPartRecord>()
+                                                            .ForVersion(VersionOptions.Latest)
```
HermesSbicego-Laser, line 120: Thinking about your @sebastienros comment to query latest and published version here...
As I Know Medias are not draftable, so querying latest should be enough. The only case in which a Media can be in draft, right now, is after a clone/translation and it has not a published version. In all the other cases if a media is published it's the Latest also. If you agree I will keep this code as is.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/ClientStorageController.cs
```diff
@@ -183,8 +183,18 @@ public class ClientStorageController : Controller {
                     // Raise update and publish events which will update relevant Media properties
                     _handlers.Invoke(x => x.Updating(new UpdateContentContext(replaceMedia.ContentItem)), Logger);
 
-                    _mediaLibraryService.DeleteFile(replaceMedia.FolderPath, replaceMedia.FileName);
-                    _mediaLibraryService.UploadMediaFile(replaceMedia.FolderPath, replaceMedia.FileName, file.InputStream);
+                    var mediaItemsUsingTheFile = Services.ContentManager.Query<MediaPart, MediaPartRecord>()
+                                                                .ForVersion(VersionOptions.Latest)
+                                                                .Where(x => x.FolderPath == replaceMedia.FolderPath && x.FileName == replaceMedia.FileName)
+                                                                .Count();
+                    if (mediaItemsUsingTheFile == 1) { // if the file is referenced only by the deleted media content, the file too can be removed.
```
HermesSbicego-Laser, line 190: @sebastienros Same here, Latest should be enough

HermesSbicego-Laser:
@sebastienros , After yesterday demo, consider to merge it in dev. We will cherry pick and deliver it in our production env.

PR #7940, opened 2018-01-16
Recipes description format style

psp589:
I wanted to have a well formatted description of my custom recipes.
This is the only way I've found to accomplish that

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Modules/Views/Admin/Recipes.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Modules/Views/Admin/Recipes.cshtml
@@ -23,8 +23,8 @@
                             @foreach (var recipe in moduleEntry.Recipes) {
                                 <br />
                                 <div>
-                                    <h4>@recipe.Name.CamelFriendly() - @Html.ActionLink(T("Execute").Text, "Recipes", "Admin", new { area = "Orchard.Modules", moduleId = descriptor.Id, name = recipe.Name }, new { itemprop = "UnsafeUrl" })</h4>
-                                    <p style="white-space: pre-line">@(!string.IsNullOrEmpty(recipe.Description) ? recipe.Description : T("No description").ToString())</p>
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Modules/Views/Admin/Recipes.cshtml
```diff
@@ -23,8 +23,8 @@
                             @foreach (var recipe in moduleEntry.Recipes) {
                                 <br />
                                 <div>
-                                    <h4>@recipe.Name.CamelFriendly() - @Html.ActionLink(T("Execute").Text, "Recipes", "Admin", new { area = "Orchard.Modules", moduleId = descriptor.Id, name = recipe.Name }, new { itemprop = "UnsafeUrl" })</h4>
-                                    <p>@(!string.IsNullOrEmpty(recipe.Description) ? recipe.Description : T("No description").ToString())</p>
+                                    <h4>@T(recipe.Name).Text - @Html.ActionLink(T("Execute").Text, "Recipes", "Admin", new { area = "Orchard.Modules", moduleId = descriptor.Id, name = recipe.Name }, new { itemprop = "UnsafeUrl" })</h4>
+                                    <p style="white-space: pre-line">@(!string.IsNullOrEmpty(recipe.Description) ? T(recipe.Description).Text : T("No description").ToString())</p>
```
sebastienros, line 27: 1. Don't translate `recipe.Description`
2. Fix the statement such that we call `T("No description")` without `ToString` or `Text`. Using an `@if` block for instance.

BenedekFarkas:
@sebastienros I can't see the point of the styling changes, any ideas?

BenedekFarkas:
OK, so the styling changes is needed so line breaks are stripped out, so this way you can have a recipe description that is divided into multiple paragraphs. I'm declining the PR as it targets dev, but I committed this directly to 1.10.x (a55cb430e227f69d4a80da06a516086e65b8d2ea) with @psp589 marked as the author.

PR #7943, opened 2018-01-17
Fixes Duplicated media items

### LINKED ISSUE #7934: Duplicated media items

HermesSbicego-Laser:
When updating via js (pickAdnClose) the list of items in media library picker fields updates only the <ul> containing items (with the specific css-class "media-items") not the others <ul>
Fixes #7934

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/MediaLibraryPicker.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/MediaLibraryPicker.cshtml
@@ -34,7 +34,7 @@
     <label @if (required) { <text> class="required" </text>    }>@displayName</label>
     <div class="message message-Warning media-library-picker-message">@T("You need to save your changes.")</div>
     <div class="items media-library-picker" summary="@displayName">
-        <ul>
+        <ul class="media-items">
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/MediaLibraryPicker.cshtml
```diff
@@ -34,7 +34,7 @@
     <label @if (required) { <text> class="required" </text>    }>@displayName</label>
     <div class="message message-Warning media-library-picker-message">@T("You need to save your changes.")</div>
     <div class="items media-library-picker" summary="@displayName">
-        <ul>
+        <ul class="media-items">
```
HermesSbicego-Laser, line 37: @sebastienros instead of a css-class we could use data-name (or another data-* attribute), what is better?

PR #7945, opened 2018-01-18
User data providers and cookie expiration setting

MatteoPiovanelli:
AS mentioned in #6014 I have implemented providers for the UserData string that is used in authentication cookies. Adding this, any extant cookie wil immeditely be considered invalid. Upon first sign in, the new UserData will be used.
I have also made the life span of said cookies configurable (the default is still 30 days).

I'll add comments to the PR to explain/discuss my changes.


src/Orchard.Web/Modules/Orchard.OpenId/Services/OpenIdAuthenticationService.cs
```diff
@@ -19,14 +19,16 @@ public class OpenIdAuthenticationService : IAuthenticationService {
         private readonly IHttpContextAccessor _httpContextAccessor;
         private readonly IMembershipValidationService _membershipValidationService;
         private readonly IEnumerable<IOpenIdProvider> _openIdProviders;
+        private readonly IEnumerable<IUserDataProvider> _userDataProviders;
+        private readonly ISecurityService _securityService;
```
MatteoPiovanelli, line 23: I had to make changes here because of the new services being injected in FormsAuthenticationService

src/Orchard.Web/Modules/Orchard.Users/Services/PasswordChangedDateUserDataProvider.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using Orchard.ContentManagement;
+using Orchard.Security;
+using Orchard.Security.Providers;
+using Orchard.Users.Models;
+
+namespace Orchard.Users.Services {
+    public class PasswordChangedDateUserDataProvider : BaseUserDataProvider {
```
MatteoPiovanelli: With this provider the auth cookies is considered invalid on password changes. 
Because of this, when we change password we are automatically logged out of the tenant (from all clients, including the one from which we changed password), and have to sign in anew. I prefer this to having the user remain logged in, but it is different than the current UX.

src/Orchard/Security/Providers/FormsAuthenticationService.cs
```diff
@@ -178,5 +186,36 @@ public class FormsAuthenticationService : IAuthenticationService {
 
             return cookiePath;
         }
+
+        #region Serialization of UserData Dictionary
```
MatteoPiovanelli, line 292: I wrote this serialization/deserialization to avoid introducing new dependencies in Orchard.Framework.

src/Orchard/Security/Providers/BaseUserDataProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// Tells whether this provider should return true for validation when there is no
+        /// element for it in the UserData dictionary.
+        /// </summary>
+        protected bool DefaultValid;
+
+        /// <summary>
+        /// This is the key that will be used in the UserData dictionary
+        /// </summary>
+        public virtual string Key {
+            get { return GetType().FullName; }
```
MatteoPiovanelli, line 41: This should make the Key unique to each provider, right?

MatteoPiovanelli:
Note:
I'm changing the code a bit so that extant auth cookies arent't automatically invalid

src/Orchard.Web/Modules/Orchard.Users/Services/PasswordChangedDateUserDataProvider.cs
MatteoPiovanelli: Note: in this PR right now (2018/01/18 15.30) I am not firing the LoggedOut event when this happens.

src/Orchard.Web/Modules/Orchard.Users/Views/Account/ChangePassword.cshtml
```diff
@@ -2,6 +2,9 @@
 <h1>@Html.TitleForPage(T("Change Password").ToString()) </h1>
 <p>@T("Use the form below to change your password.")</p>
 <p>@T.Plural("The password can't be empty.", "New passwords are required to be a minimum of {0} characters in length.", (int)ViewData["PasswordLength"])</p>
+@if ((bool)ViewData["InvalidateOnPasswordChange"]) {
+    <p>@T("After changing the password you will be required to login anew.")</p>
```
sebastienros, line 6: typo

src/Orchard.Web/Modules/Orchard.Users/Views/Account/ChangePasswordSuccess.cshtml
```diff
@@ -1,3 +1,6 @@
 ﻿@model dynamic
 <h1>@Html.TitleForPage(T("Change Password").ToString())</h1>
 <p>@T("Your password has been changed successfully.")</p>
+@if ((bool)ViewData["InvalidateOnPasswordChange"]) {
```
sebastienros, line 4: Unnecessary. Users are redirected to the login page, it's obvious they need to use the new one.

src/Orchard.Web/Modules/Orchard.Users/Views/EditorTemplates/Parts/Users.SecuritySettings.cshtml
```diff
@@ -0,0 +1,17 @@
+﻿@model Orchard.Users.Models.SecuritySettingsPart
+
+<fieldset>
+    <legend>@T("Security settings")</legend>
+    <div>
+        <label>@T("Life span of authentication cookies")</label>
+        @Html.TextBoxFor(m => m.AuthCookieLifeSpan, new { @class = "text medium" })
+        @Html.ValidationMessageFor(m => m.AuthCookieLifeSpan)
+        <p class="hint">@T("The time after signing in after which authentication cookies will be invalidated, in standard TimeSpan format (dd.hh:mm:ss).")</p>
```
sebastienros: Don't mention `TimeSpan`, just the format.

sebastienros: For reference https://msdn.microsoft.com/library/1d3t3c61(v=vs.100).aspx

src/Orchard.Web/Modules/Orchard.Users/Views/EditorTemplates/Parts/Users.SecuritySettings.cshtml
```diff
@@ -0,0 +1,17 @@
...
+    </div>
+    <div>
+        @Html.CheckBoxFor(m => m.ShouldInvalidateAuthOnPasswordChanged)
+        <label class="forcheckbox" for="@Html.FieldIdFor(m => m.ShouldInvalidateAuthOnPasswordChanged)">@T("Invalidate authentication cookies on password change")</label>
+        @Html.ValidationMessageFor(m => m.ShouldInvalidateAuthOnPasswordChanged)
+        <p class="hint">@T("Select this flag if you want that changing a user's password will require them to sign in anew from all clients.")</p>
```
sebastienros: `When checked, the user will be disconnected from all clients when their password is changed.`

src/Orchard/Security/Providers/FormsAuthenticationService.cs
```diff
@@ -35,62 +39,32 @@ public class FormsAuthenticationService : IAuthenticationService {
...
 
             Logger = NullLogger.Instance;
 
-            ExpirationTimeSpan = TimeSpan.FromDays(30);
         }
 
         public ILogger Logger { get; set; }
 
-        public TimeSpan ExpirationTimeSpan { get; set; }
+        public TimeSpan ExpirationTimeSpan {
```
sebastienros: This is breaking what users currently have to do to change this value (host.config).
Not sure it's wise to make it editable in the UI. More code, not usually used, accessible to more users, ... lots of reasons to keep it the way it is today.

src/Orchard/Security/Providers/FormsAuthenticationService.cs
```diff
@@ -135,39 +109,114 @@ public class FormsAuthenticationService : IAuthenticationService {
...
-            var userDataName = userDataSegments[0];
-            var userDataTenant = userDataSegments[1];
-
-            try {
-                userDataName = userDataName.FromBase64();
+            var userData = formsIdentity.Ticket.UserData ?? "";
+            var userDataDictionary = new Dictionary<string, string>();
+            
+            if (formsIdentity.Ticket.Version == 3) {
+                var userDataSegments = userData.Split(';');
```
sebastienros, line 134: Just revoke the cookie.

src/Orchard/Security/Providers/FormsAuthenticationService.cs
sebastienros: Json ?

sebastienros: Don't even try, otherwise there is a risk that user data conflict with the format if you are not escaping correctly.

src/Orchard.Web/Modules/Orchard.Users/Services/PasswordChangedDateUserDataProvider.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using Orchard.ContentManagement;
+using Orchard.Security;
+using Orchard.Security.Providers;
+using Orchard.Settings;
+using Orchard.Users.Models;
+
+namespace Orchard.Users.Services {
+    public class PasswordChangedDateUserDataProvider : BaseUserDataProvider {
```
sebastienros, line 10: Does it need to be extensible? Isn't the setting already available in the form authentication service?

src/Orchard/Security/Providers/FormsAuthenticationService.cs
MatteoPiovanelli: It's what I initially thought of doing. However that would have meant that deploying this update to how the cookies are processed would have caused all users to effectively be logged out. 
That would have been jarring in terms of UX, and no LoggedOut event would have been fired for anyone.

Something else I thought would be adding at the end of this method a condition to replace V3 cookies with new ones by releasing it (I'd just call CreateAndAddCookie)

src/Orchard.Web/Modules/Orchard.Users/Services/PasswordChangedDateUserDataProvider.cs
MatteoPiovanelli: Could you clarify what you mean? I don't understand what you are referring to.

src/Orchard.Web/Modules/Orchard.Users/Views/Account/ChangePasswordSuccess.cshtml
MatteoPiovanelli: While trying the feature, I was not being redirected anywhere, so I added this paragraph. Could be the redirect is not in place anymore?

src/Orchard.Web/Modules/Orchard.Users/Views/EditorTemplates/Parts/Users.SecuritySettings.cshtml
MatteoPiovanelli: I went to the URL, but I am not sure what I am looking for in relation with the TimeSpan

src/Orchard.Web/Modules/Orchard.Users/Views/EditorTemplates/Parts/Users.SecuritySettings.cshtml
MatteoPiovanelli: ok

src/Orchard/Security/Providers/FormsAuthenticationService.cs
MatteoPiovanelli: The value for this previously was contant and 30 days. I don't see where it was being read from host.config. Unless writing the form authentication settings there directly overrides anything that is done in code here for the auth cookie.

In this PR, the value is in a setting that is editable from the UI (that is the implementation I put in Orchard.Users, not the default one from Orchard.Framework, even though I realize that close to 100% tenants have Orchard.Users enabled), but here in the service we are getting it from a service, rather than from the setting. It's straightforward to write a service that reads from a config file, or that does something else entirely. DIfferent calls could lead to different values for this.

src/Orchard/Security/Providers/FormsAuthenticationService.cs
MatteoPiovanelli: The way it's written there should be no escaping concerns.
To safely serialize to json, I would feel safer to use JsonSerializer, or Newtonsoft.Json, or something like that. I haven't done that directly to avoid introducing new dependencies in Orchard.Framework. However, if you tell me that is fine, I would go that route.

sebastienros: I think there is already a dependency on it.

sebastienros: > there should be no escaping concerns

Until there is

src/Orchard/Security/Providers/FormsAuthenticationService.cs
sebastienros: The value is assigned automatically by autofac when the service is resolve. It's called "Property Injection". I don't think we need a service for this.

src/Orchard.Web/Modules/Orchard.Users/Services/PasswordChangedDateUserDataProvider.cs
sebastienros: I don't see why we need to be able to extend this. The forms auth logic should be self sufficient, and store in the auth cookie only what it needs. If other modules need to store stuff for the user in a cookie, they are free to create their own.

MatteoPiovanelli: The point is not to store information in the auth cookie per se, but to stamp it with information to be able to invalidate it if/when required by something. In the specific case of PasswordChangedDateUserDataProvider, the idea is that a password change can make all auth cookies, for the user and from any client, invalid. That information we are storing in the cookie (the datetime in this case) is not for the user/client.

src/Orchard/Security/Providers/FormsAuthenticationService.cs
MatteoPiovanelli: Ok. I will use Newtonsoft.Json

src/Orchard/Security/Providers/FormsAuthenticationService.cs
MatteoPiovanelli: I did not know that.
I am trying now to see how to set this up in a per-tenant way using the .config files ("Sites.MyTenant.config", right?)

MatteoPiovanelli: We could have both the services and property injection:

If I moved that `ExpirationTimeSpan` to DefaultSecurityService I could still do Property Injection into there.

Alternatively, in the constructor of `FormsAuthenticationService` I could initialize the value of `ExpirationTimeSpan` to a negative value. Then rather than using the property in the rest of the class, I could have a method like:
```
public TimeSpan GetExpirationTime() {
            if (ExpirationTimeSpan < TimeSpan.Zero) {
                // We have not set expiration from Sites.Tenant.config
                return  _securityService.GetAuthenticationCookieLifeSpan();
            }
            // we have set the expiration from the config file
            return ExpirationTimeSpan;
        }
```
As long as the UI for each of those services clearly stated that having a value in the config file would take precedence, that behaviour would not be obscure.

MatteoPiovanelli: The question, I guess, is more of a "should we" than a "could we" nature, at this point, right?

MatteoPiovanelli: Personally I think there may be value in having the option, for example, to set different `ExpirationTimeSpan` depending on the "source" of the SignIn. For example, if you SignIn from a browser I may set a short validity, but if you SignIn from my app on your phone I may decide that that may be ok with a longer-lived cookie (of course that requires me to identify the source in my service, but that is a separate issue).

src/Orchard/Security/Providers/FormsAuthenticationService.cs
sebastienros: I like the idea of replacing as a good compromise.

src/Orchard.Web/Modules/Orchard.Users/Views/Account/ChangePasswordSuccess.cshtml
sebastienros: Good then.

src/Orchard/Security/Providers/FormsAuthenticationService.cs
MatteoPiovanelli: Another scenario may be:
 - Cookie expires in a bunch of days for admin
 - Cookie expires in a shorter timeframe for "normal" users

sebastienros:
I'll let you do all the changes and compromises that you deem necessary, and trust your judgement as you are the main user of this feature and it's not breaking.
Once you are happy just let me know and I will merge.

MatteoPiovanelli:
Thanks for the trust in my code. 

I think with the last couple commits this feature is good for me.
I removed the service implementations I previously added in Orchard.Users to override the cookie ExpirationTimeSpan. I also wrote the code in FormsAuthenticationService so that it should behave exactly as it does right now (wrt the ExpirationTimeSpan): if autofac injects a value on that property, it will take priority on services trying to affect that; otherwise, the value form services will be used (with the default implementation giving 30days, that is the current default).
I used Newtonsoft.Json to serialize/deserialize the userdata dictionary. I had missed the fact that the dependency on that was already in place.
A call to GetAuthenticatedUser for a request that carries an "old" cookie will cause an upgrade of the cookie: the way it's set up, the new cookie that will be released will have the same expiration as the "old" cookie from the request.

As it is, nothing should break. Let me know what do you think.

sebastienros:
I trust you more than your code.

PR #7959, opened 2018-02-01
Decoupled cache eviction from cache settings handler fixes #7958

### LINKED ISSUE #7958: Orchard.OutputCache shouldn't handle cache invalidation in the settings part handler
At the moment the invalidation of items from orchard's output cache is handled by hooking into event's in the CacheSettingsPartHandler (https://github.com/OrchardCMS/Orchard/blob/master/src/Orchard.Web/Modules/Orchard.OutputCache/Handlers/CacheSettingsPartHandler.cs)

It would be better if these 3 events weere in a separate handler, this would make it easier to suppress these event's if doing custom stuff with cache invalidation without effecting the settingspart. Really these events have nothing to do with the settings.

Can provide a PR if considered a good idea.

carlwoodhouse:


src/Orchard.Web/Modules/Orchard.OutputCache/Orchard.OutputCache.csproj
```diff
@@ -26,6 +26,7 @@
     <IISExpressWindowsAuthentication />
     <IISExpressUseClassicPipelineMode />
     <UseGlobalApplicationHostFile />
+    <Use64BitIISExpress />
```
sebastienros: Can you remove it, unless necessary?

carlwoodhouse:
Gimme 30 mins and I’ll remove; must have been vs2017 for some reason

src/Orchard.Web/Modules/Orchard.OutputCache/Orchard.OutputCache.csproj
carlwoodhouse: removed

Issue #7989, opened 2018-03-07
Information Discovery


MatteoPiovanelli:
Every time we have a security scan ran on our Orchard tenants, we get flagged with warnings related to information discovery. While the risk with those "vulnerabilities" is debatable, I propose a fix to quickly remove some of them.
 
 | Header | Solution |
 | - | - |
 | Server | Can be removed by changing global.asax |
 | X-AspNet-Version | Can be removed by changing web.config |
 | X-AspNetMvc-Version | Can be removed by changing global.asax |
 | X-Generator | Can be removed by changing Orchard.Framework/Owin/Startup.cs |
 | X-Powered-By | Requires a fix from IIS (*) |

I will create a PR momentarily with the changes to remove the first four headers, but let me know whether this is desired by the community. Also, let me know how everyone else is handling this bit.

(*) I played around with trying to remove it from code in the same ways I am taking care of the other headers, but to no avail.

sebastienros:
I want to keep the Orchard one, it pays my salary.

MatteoPiovanelli:
That is fair.

carlwoodhouse:
+1 as it means we can remove some stuff from our transforms ;)

PR #7990, opened 2018-03-07
Remove headers for information discovery

MatteoPiovanelli:
Changes to remove the following headers being flagged as possible Information Discovery Issues:

Server, X-AspNet-Version, X-AspNetMvc-Version, X-Generator

This PR is related to #7989

### DIFF
--- a/src/Orchard.Web/Global.asax.cs
+++ b/src/Orchard.Web/Global.asax.cs
@@ -20,6 +20,7 @@ namespace Orchard.Web {
         }
 
         protected void Application_Start() {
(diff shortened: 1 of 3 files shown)


sebastienros:
Please make it an option, or several, I'd say that you can also already remove some of these from IIS or web.config. The orchard one at least can be an option and enabled by default.

MatteoPiovanelli:
I am not 100% sure how to inject those settings in those classes and keep it Orchard-y.
I'll have to play with it.

Any input is appreciated.

Issue #7991, opened 2018-03-07
Failed Tests


MatteoPiovanelli:
Using VS2015 I just run all tests on O1.10.x. A bunch of them failed, a bunch were skipped.

36 Tests fail, 14 Tests are skipped, 1414 Tests passed.

I haven't checked any yet. I will as soon as I have some time in the next couple of days, as I am trying to set up CI and CD for our systems.

MatteoPiovanelli:
I did a bunch of fixes and now 15 tests are still failing.
Some are from Specs/Orchard.Profile, so they are expected to fail if I simply launch them from Visual Studio's Test Runner.

Then there are some tests that are related to parsing of DateTime from DefaultDateFormatterTests.
These require more fiddling because they may be failing just because my machine has the it-IT culture set as default. I actually checked and the failing test cases are those for the it-IT culture itself, but if I try to run the very same code in the tests form the C# interactive window I get the expected results. I'll have to double check those.

Then in FileSystemStorageProviderTests two tests fail throwing an UnauthorizedAccessException on setup or teardown.

I'll do a PR with the fixes for the other tests in a moment.

sebastienros:
Functional tests must be run from the command line, there is a target for this

MatteoPiovanelli:
Yes, I figured that out by trial and error.

The DefaultDateFormatterTests and FileSystemStorageProviderTests are still a mistery to me. I'll try to sort them out as soon as I can sit with them a bit.

MatteoPiovanelli:
DefaultDateFormatterTests:
This fail because ` var expected = date.ToString(timeFormat, culture);` is wrong, I think.

This is the data for one condition where `FormatTimeTest01` fails:
    
    date = {01/01/1998 00.30.30}
    timeFormat = "HH:mm"
    culture = {it-IT}

The results are:

    target.FormatTime(timeParts, timeFormat) == "00:30"
    date.ToString(timeFormat, culture) == "00.30"

This may have to do with a setting on my machine, I fear. I'll need someone else to confirm this on different machines.

sebastienros:
Tests should be culture independent and work on any machine, or not at least take the current culture into account when expecting specific values.

MatteoPiovanelli:
I agree.
By the looks of things, the wrong string is the one being computed as the expected value, but I can't figure out why the `date.ToString(timeFormat, culture)` would give that result when culture equals my machine's culture, especially since we are passing the timeFormat explicitly.

[...]

PR #7993, opened 2018-03-08
Fix/failed tests

MatteoPiovanelli:
related to #7991 
The code here fixes a bunch of the failing tests. Some are still left to fix:

Some tests that are related to parsing of DateTime from DefaultDateFormatterTests.
These require more fiddling because they may be failing just because my machine has the it-IT culture set as default. I actually checked and the failing test cases are those for the it-IT culture itself, but if I try to run the very same code in the tests form the C# interactive window I get the expected results. I'll have to double check those.

Then in FileSystemStorageProviderTests two tests fail throwing an UnauthorizedAccessException on setup or teardown.

### DIFF
--- a/src/Orchard.Specs/Boolean.feature
+++ b/src/Orchard.Specs/Boolean.feature
@@ -48,7 +48,7 @@ Scenario: Creating and using Boolean fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                | value                        |
-            | Fields[0].BooleanFieldSettings.Hint | Check if the event is active |
+            | Fields[Active].BooleanFieldSettings.Hint | Check if the event is active |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Check if the event is active"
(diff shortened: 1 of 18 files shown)


src/Orchard.Specs/Boolean.feature
```diff
@@ -48,7 +48,7 @@ Scenario: Creating and using Boolean fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                | value                        |
-            | Fields[0].BooleanFieldSettings.Hint | Check if the event is active |
```
sebastienros, line 51: This change is weird ... did you find which commit did it?

MatteoPiovanelli: I did a PR a while back where I changed the names written in the generated html for fields and such, because they would not have the same index when opened in two browser tabs at the same time.

The repro for the issue was to open a content definition on two tabs, update it from one, then try to update it from the other. After saving on the first tab, the order of fields in memory would be different than the one in the html of the second tab, hence saving again there would potentially mess settings up.

I can dig the issue and PR up for more details if you like.

PR #8005, opened 2018-03-29
File formatter

sebastienros:


sebastienros:
@jonvee could you test that?

jonvee:
Just tested and it seems to work fine... No events logged.

PR #7723, opened 2017-06-01
Added locks to autoroute service

MatteoPiovanelli:
This should fix #7720 
I tested the code on my machine, with both ReadUncommited enabled and disabled.

By my tests, it allows correct concurrency in creating/updating AutoroutePart. I tried this by creating pages all with the same title, as well as by creating them with different titles. 
In the latter case, the parts do not lock each other out of the second part of the ProcessPath method.

I would like someone to test this and confirm my results, as I am not experienced enough to be 100% confident about this solution.

Any comment and review are appreciated.

Issue #7720, opened 2017-05-29
Concurrency on AutoroutePart


MatteoPiovanelli:
As per #7708 I am testing concurrency.
The first category of issues I am consistently seeing has to do with AutorouteParts.

The errors, happening at 2 concurrent calls:

```
Orchard.ContentManagement.DefaultContentManager - NoLaser - GenericADOException thrown from IContentHandler by Orchard.Autoroute.Handlers.AutoroutePartHandler
NHibernate.Exceptions.GenericADOException: could not execute query
 [ SELECT this_.Id as Id6240_3_, this_.Number as Number6240_3_, this_.Published as Published6240_3_, this_.Latest as Latest6240_3_, this_.Data as Data6240_3_, this_.ContentItemRecord_id as ContentI6_6240_3_, autoroutep1_.Id as Id6314_0_, autoroutep1_.UseCustomPattern as UseCusto2_6314_0_, autoroutep1_.UseCulturePattern as UseCultu3_6314_0_, autoroutep1_.CustomPattern as CustomPa4_6314_0_, autoroutep1_.DisplayAlias as DisplayA5_6314_0_, autoroutep1_.ContentItemRecord_id as ContentI6_6314_0_, contentite4_.Id as Id6239_1_, contentite4_.Data as Data6239_1_, contentite4_.ContentType_id as ContentT3_6239_1_, contenttyp5_.Id as Id6241_2_, contenttyp5_.Name as Name6241_2_ FROM Orchard_Framework_ContentItemVersionRecord this_ inner join Orchard_Autoroute_AutoroutePartRecord autoroutep1_ on this_.Id=autoroutep1_.Id left outer join Orchard_Framework_ContentItemRecord contentite4_ on this_.ContentItemRecord_id=contentite4_.Id left outer join Orchard_Framework_ContentTypeRecord contenttyp5_ on contentite4_.ContentType_id=contenttyp5_.Id WHERE (autoroutep1_.DisplayAlias is not null and autoroutep1_.DisplayAlias like @p0) and this_.Published = @p1 ]
  Name:cp0 - Value:mytitle%  Name:cp1 - Value:True
[...]

PR #8019, opened 2018-04-09
Proposal for concurrency

MatteoPiovanelli:
Here I propose the ILockingProvider and it base implementation that should work when there is a single server.
In the PR here I use this implementation in AutorouteService and TagService, that fixes the issues with those services discussed in #7708 and #7720.
This PR supersedes #7723.


MatteoPiovanelli:
I will remove the changes for `AutoroutePart` and push the updated `LockingProvider`

MatteoPiovanelli:
Closing this because of work on #8025

Issue #7597, opened 2017-03-02
Exceptions creating a new tenant


MatteoPiovanelli-Laser:
Orchard 1.10.x
Creating a new Tenant with an existing SQL Server db. (SQL Server 2012)
After hitting "Finish Setup" with the Default recipe, the Home page is shows regularly, but I have the following in the Error Log:

```
2017-03-02 14:14:28,351 [5] Orchard.Alias.Implementation.Updater.AliasHolderUpdater - NoLaser - Exception during Alias refresh [https://localhost/Laser.Orchard/NoLaser]
NHibernate.Exceptions.GenericADOException: could not execute query
[ select aliasrecor0_.Id as Id411_, aliasrecor0_.Path as Path411_, aliasrecor0_.RouteValues as RouteVal3_411_, aliasrecor0_.Source as Source411_, aliasrecor0_.IsManaged as IsManaged411_, aliasrecor0_.Action_id as Action6_411_ from Orchard_Alias_AliasRecord aliasrecor0_ where aliasrecor0_.Id>@p0 order by aliasrecor0_.Id asc ]
  Name:p1 - Value:0
[SQL: select aliasrecor0_.Id as Id411_, aliasrecor0_.Path as Path411_, aliasrecor0_.RouteValues as RouteVal3_411_, aliasrecor0_.Source as Source411_, aliasrecor0_.IsManaged as IsManaged411_, aliasrecor0_.Action_id as Action6_411_ from Orchard_Alias_AliasRecord aliasrecor0_ where aliasrecor0_.Id>@p0 order by aliasrecor0_.Id asc] ---> System.Data.SqlClient.SqlException: Il timeout di esecuzione è scaduto. Il periodo di timeout è scaduto prima del completamento dell'operazione oppure il server non risponde. ---> System.ComponentModel.Win32Exception: Tempo di attesa scaduto
   --- Fine della traccia dello stack dell'eccezione interna ---
   in System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   in System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
[...]

PR #8025, opened 2018-04-12
Fix write concurrency issue with Parts Services

Skrypt:
Using AliasService instead of retrieving ContentItem data from database.
Needs testing from @MatteoPiovanelli-Laser

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -5,33 +5,42 @@ using Orchard.ContentManagement;
 using Orchard.ContentManagement.Handlers;
 using Orchard.Data;
 using Orchard.Localization;
+using Orchard.Locking;
 using Orchard.UI.Notify;
 
-namespace Orchard.Autoroute.Handlers {
-    public class AutoroutePartHandler : ContentHandler {
+namespace Orchard.Autoroute.Handlers
+{
+    public class AutoroutePartHandler : ContentHandler
+    {
 
         private readonly Lazy<IAutorouteService> _autorouteService;
         private readonly IOrchardServices _orchardServices;
         private readonly IHomeAliasService _homeAliasService;
+        private readonly ILockingProvider _lockingProvider;
 
         public Localizer T { get; set; }
 
         public AutoroutePartHandler(
             IRepository<AutoroutePartRecord> autoroutePartRepository,
             Lazy<IAutorouteService> autorouteService,
-            IOrchardServices orchardServices, 
-            IHomeAliasService homeAliasService) {
+            IOrchardServices orchardServices,
+            IHomeAliasService homeAliasService,
+            ILockingProvider lockingProvider)
+        {
 
             Filters.Add(StorageFilter.For(autoroutePartRepository));
             _autorouteService = autorouteService;
             _orchardServices = orchardServices;
             _homeAliasService = homeAliasService;
+            _lockingProvider = lockingProvider;
 
             OnUpdated<AutoroutePart>((ctx, part) => CreateAlias(part));
 
-            OnCreated<AutoroutePart>((ctx, part) => {
+            OnCreated<AutoroutePart>((ctx, part) =>
+            {
                 // non-draftable items
-                if (part.ContentItem.VersionRecord == null) {
+                if (part.ContentItem.VersionRecord == null)
+                {
                     PublishAlias(part);
                 }
             });
@@ -44,29 +53,44 @@ namespace Orchard.Autoroute.Handlers {
             OnUnpublished<AutoroutePart>((ctx, part) => RemoveAlias(part));
 
             // Register alias as identity
-            OnGetContentItemMetadata<AutoroutePart>((ctx, part) => {
+            OnGetContentItemMetadata<AutoroutePart>((ctx, part) =>
+            {
                 if (part.DisplayAlias != null)
                     ctx.Metadata.Identity.Add("alias", part.DisplayAlias);
             });
         }
 
-        private void CreateAlias(AutoroutePart part) {
+        private void CreateAlias(AutoroutePart part)
+        {
             ProcessAlias(part);
         }
 
-        private void PublishAlias(AutoroutePart part) {
+        private string LockString {
+            get {
+                return string.Join(".",
+                    _orchardServices.WorkContext?.CurrentSite?.BaseUrl ?? "",
+                    _orchardServices.WorkContext?.CurrentSite?.SiteName ?? "",
+                    "AutoroutePartHandler");
+            }
+        }
+
+        private void PublishAlias(AutoroutePart part)
+        {
             ProcessAlias(part);
 
             // Should it become the home page?
-            if (part.PromoteToHomePage) {
+            if (part.PromoteToHomePage)
+            {
                 // Get the current homepage an unmark it as the homepage.
                 var currentHomePageId = _homeAliasService.GetHomePageId();
-                if(currentHomePageId != part.Id) {
+                if (currentHomePageId != part.Id)
+                {
                     var autoroutePart = part.As<AutoroutePart>();
 
-                    if (autoroutePart != null) {
+                    if (autoroutePart != null)
+                    {
                         autoroutePart.PromoteToHomePage = false;
-                        if(autoroutePart.IsPublished())
+                        if (autoroutePart.IsPublished())
                             _orchardServices.ContentManager.Publish(autoroutePart.ContentItem);
                     }
                 }
@@ -74,44 +98,65 @@ namespace Orchard.Autoroute.Handlers {
                 // Update the home alias to point to this item being published.
                 _homeAliasService.PublishHomeAlias(part);
             }
-            
-            _autorouteService.Value.PublishAlias(part);
+
+            _lockingProvider.Lock(LockString,
+                () => { _autorouteService.Value.PublishAlias(part); });
+
         }
 
-        private void ProcessAlias(AutoroutePart part) {
+        private void ProcessAlias(AutoroutePart part)
+        {
+            // Generate Alias for published content items
             if (!part.HasDraft())
             {
-                // Generate an alias if one as not already been entered.
+                LocalizedString message = null;
+
+                // Generate an alias if one has not already been entered.
                 if (String.IsNullOrWhiteSpace(part.DisplayAlias))
                 {
                     part.DisplayAlias = _autorouteService.Value.GenerateAlias(part);
                 }
 
-                // If the generated alias is empty, compute a new one.
-                if (String.IsNullOrWhiteSpace(part.DisplayAlias))
-                {
-                    _autorouteService.Value.ProcessPath(part);
-                    _orchardServices.Notifier.Warning(T("The permalink could not be generated, a new slug has been defined: \"{0}\"", part.Path));
-                    return;
-                }
+                _lockingProvider.Lock(LockString,
+                    () =>
+                    {
+                    // Lock this portion of the code to prevent race conditions where we are reading the aliases to compute
+                    // stuff for a content, while another content is publishing. 
+                    // If the generated alias is empty, compute a new one.
+                    if (String.IsNullOrWhiteSpace(part.DisplayAlias))
+                        {
+                            _autorouteService.Value.ProcessPath(part);
+                            message = T("The permalink could not be generated, a new slug has been defined: \"{0}\"", part.Path);
+                            return;
+                        }
 
-                // Check for permalink conflict, unless we are trying to set the home page.
-                var previous = part.Path;
-                if (!_autorouteService.Value.ProcessPath(part))
-                    _orchardServices.Notifier.Warning(
-                        T("Permalinks in conflict. \"{0}\" is already set for a previously created {2} so now it has the slug \"{1}\"",
-                                                    previous, part.Path, part.ContentItem.ContentType));
+                    // Check for permalink conflict, unless we are trying to set the home page.
+                    var previous = part.Path;
+                        if (!_autorouteService.Value.ProcessPath(part))
+                            message =
+                                T("Permalinks in conflict. \"{0}\" is already set for a previously created {2} so now it has the slug \"{1}\"",
+                                                            previous, part.Path, part.ContentItem.ContentType);
+                    });
+
+                if (message != null)
+                {
+                    _orchardServices.Notifier.Warning(message);
+                }
             }
         }
 
-        void RemoveAlias(AutoroutePart part) {
+        void RemoveAlias(AutoroutePart part)
+        {
             var homePageId = _homeAliasService.GetHomePageId();
 
             // Is this the current home page?
-            if (part.ContentItem.Id == homePageId) {
+            if (part.ContentItem.Id == homePageId)
+            {
                 _orchardServices.Notifier.Warning(T("You removed the content item that served as the site's home page. \nMost possibly this means that instead of the home page a \"404 Not Found\" page will be displayed. \n\nTo prevent this you can e.g. publish a content item that has the \"Set as home page\" checkbox ticked."));
             }
-            _autorouteService.Value.RemoveAliases(part);
+
+            _lockingProvider.Lock(LockString,
+                () => { _autorouteService.Value.RemoveAliases(part); });
         }
     }
 }
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -79,24 +79,29 @@ namespace Orchard.Autoroute.Handlers {
         }
 
         private void ProcessAlias(AutoroutePart part) {
-            // Generate an alias if one as not already been entered.
-            if (String.IsNullOrWhiteSpace(part.DisplayAlias)) {
-                part.DisplayAlias = _autorouteService.Value.GenerateAlias(part);
-            }
+            if (!part.HasDraft())
+            {
+                // Generate an alias if one as not already been entered.
+                if (String.IsNullOrWhiteSpace(part.DisplayAlias))
+                {
+                    part.DisplayAlias = _autorouteService.Value.GenerateAlias(part);
+                }
 
-            // If the generated alias is empty, compute a new one.
-            if (String.IsNullOrWhiteSpace(part.DisplayAlias)) {
-                _autorouteService.Value.ProcessPath(part);
-                _orchardServices.Notifier.Warning(T("The permalink could not be generated, a new slug has been defined: \"{0}\"", part.Path));
-                return;
-            }
+                // If the generated alias is empty, compute a new one.
+                if (String.IsNullOrWhiteSpace(part.DisplayAlias))
+                {
+                    _autorouteService.Value.ProcessPath(part);
+                    _orchardServices.Notifier.Warning(T("The permalink could not be generated, a new slug has been defined: \"{0}\"", part.Path));
+                    return;
+                }
 
-            // Check for permalink conflict, unless we are trying to set the home page.
-            var previous = part.Path;
-            if (!_autorouteService.Value.ProcessPath(part))
-                _orchardServices.Notifier.Warning(
-                    T("Permalinks in conflict. \"{0}\" is already set for a previously created {2} so now it has the slug \"{1}\"", 
-                                                previous, part.Path, part.ContentItem.ContentType));
+                // Check for permalink conflict, unless we are trying to set the home page.
+                var previous = part.Path;
+                if (!_autorouteService.Value.ProcessPath(part))
+                    _orchardServices.Notifier.Warning(
+                        T("Permalinks in conflict. \"{0}\" is already set for a previously created {2} so now it has the slug \"{1}\"",
+                                                    previous, part.Path, part.ContentItem.ContentType));
+            }
         }
 
         void RemoveAlias(AutoroutePart part) {
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -60,9 +60,9 @@ namespace Orchard.Autoroute.Handlers {
             // Should it become the home page?
             if (part.PromoteToHomePage) {
                 // Get the current homepage an unmark it as the homepage.
-                var currentHomePage = _homeAliasService.GetHomePage(VersionOptions.Latest);
-                if(currentHomePage != null && currentHomePage.Id != part.Id) {
-                    var autoroutePart = currentHomePage.As<AutoroutePart>();
+                var currentHomePageId = _homeAliasService.GetHomePageId();
+                if(currentHomePageId != part.Id) {
+                    var autoroutePart = part.As<AutoroutePart>();
 
                     if (autoroutePart != null) {
                         autoroutePart.PromoteToHomePage = false;
@@ -100,7 +100,7 @@ namespace Orchard.Autoroute.Handlers {
         }
 
         void RemoveAlias(AutoroutePart part) {
-            var homePageId = _homeAliasService.GetHomePageId(VersionOptions.Latest);
+            var homePageId = _homeAliasService.GetHomePageId();
 
             // Is this the current home page?
             if (part.ContentItem.Id == homePageId) {
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -76,31 +76,37 @@ namespace Orchard.Autoroute.Handlers
 
         private void PublishAlias(AutoroutePart part)
         {
-            ProcessAlias(part);
-
-            // Should it become the home page?
-            if (part.PromoteToHomePage)
-            {
-                // Get the current homepage an unmark it as the homepage.
-                var currentHomePageId = _homeAliasService.GetHomePageId();
-                if (currentHomePageId != part.Id)
-                {
-                    var autoroutePart = part.As<AutoroutePart>();
-
-                    if (autoroutePart != null)
-                    {
-                        autoroutePart.PromoteToHomePage = false;
-                        if (autoroutePart.IsPublished())
-                            _orchardServices.ContentManager.Publish(autoroutePart.ContentItem);
-                    }
-                }
-
-                // Update the home alias to point to this item being published.
-                _homeAliasService.PublishHomeAlias(part);
-            }
-
+            // Lock this portion of the code to prevent race conditions where we are reading the aliases to compute
+            // stuff for a content, while another content is publishing. 
+            // If the generated alias is empty, compute a new one.
             _lockingProvider.Lock(LockString,
-                () => { _autorouteService.Value.PublishAlias(part); });
+                () =>
+                {
+                    ProcessAlias(part);
+
+                    // Should it become the home page?
+                    if (part.PromoteToHomePage)
+                    {
+                        // Get the current homepage an unmark it as the homepage.
+                        var currentHomePageId = _homeAliasService.GetHomePageId();
+                        if (currentHomePageId != part.Id)
+                        {
+                            var autoroutePart = part.As<AutoroutePart>();
+
+                            if (autoroutePart != null)
+                            {
+                                autoroutePart.PromoteToHomePage = false;
+                                if (autoroutePart.IsPublished())
+                                    _orchardServices.ContentManager.Publish(autoroutePart.ContentItem);
+                            }
+                        }
+
+                        // Update the home alias to point to this item being published.
+                        _homeAliasService.PublishHomeAlias(part);
+                    }
+
+                    _autorouteService.Value.PublishAlias(part);
+                });
 
         }
 
@@ -117,26 +123,19 @@ namespace Orchard.Autoroute.Handlers
                     part.DisplayAlias = _autorouteService.Value.GenerateAlias(part);
                 }
 
-                _lockingProvider.Lock(LockString,
-                    () =>
-                    {
-                    // Lock this portion of the code to prevent race conditions where we are reading the aliases to compute
-                    // stuff for a content, while another content is publishing. 
-                    // If the generated alias is empty, compute a new one.
-                    if (String.IsNullOrWhiteSpace(part.DisplayAlias))
-                        {
-                            _autorouteService.Value.ProcessPath(part);
-                            message = T("The permalink could not be generated, a new slug has been defined: \"{0}\"", part.Path);
-                            return;
-                        }
+                if (String.IsNullOrWhiteSpace(part.DisplayAlias))
+                {
+                    _autorouteService.Value.ProcessPath(part);
+                    message = T("The permalink could not be generated, a new slug has been defined: \"{0}\"", part.Path);
+                    return;
+                }
 
-                    // Check for permalink conflict, unless we are trying to set the home page.
-                    var previous = part.Path;
-                        if (!_autorouteService.Value.ProcessPath(part))
-                            message =
-                                T("Permalinks in conflict. \"{0}\" is already set for a previously created {2} so now it has the slug \"{1}\"",
-                                                            previous, part.Path, part.ContentItem.ContentType);
-                    });
+                // Check for permalink conflict, unless we are trying to set the home page.
+                var previous = part.Path;
+                if (!_autorouteService.Value.ProcessPath(part))
+                    message =
+                        T("Permalinks in conflict. \"{0}\" is already set for a previously created {2} so now it has the slug \"{1}\"",
+                                                    previous, part.Path, part.ContentItem.ContentType);
 
                 if (message != null)
                 {
--- a/src/Orchard.Specs/Boolean.feature
+++ b/src/Orchard.Specs/Boolean.feature
@@ -48,7 +48,7 @@ Scenario: Creating and using Boolean fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                | value                        |
-            | Fields[0].BooleanFieldSettings.Hint | Check if the event is active |
+            | Fields[Active].BooleanFieldSettings.Hint | Check if the event is active |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Check if the event is active"
@@ -57,7 +57,7 @@ Scenario: Creating and using Boolean fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value |
-            | Fields[0].BooleanFieldSettings.DefaultValue | True  |
+            | Fields[Active].BooleanFieldSettings.DefaultValue | True  |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "checked=\"checked\""
@@ -66,7 +66,7 @@ Scenario: Creating and using Boolean fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value |
-            | Fields[0].BooleanFieldSettings.Optional     | false |
+            | Fields[Active].BooleanFieldSettings.Optional     | false |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
         And I fill in 
--- a/src/Orchard.Specs/Boolean.feature.cs
+++ b/src/Orchard.Specs/Boolean.feature.cs
@@ -157,7 +157,7 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table4.AddRow(new string[] {
-                        "Fields[0].BooleanFieldSettings.Hint",
+                        "Fields[Active].BooleanFieldSettings.Hint",
                         "Check if the event is active"});
 #line 49
         testRunner.And("I fill in", ((string)(null)), table4, "And ");
@@ -174,7 +174,7 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table5.AddRow(new string[] {
-                        "Fields[0].BooleanFieldSettings.DefaultValue",
+                        "Fields[Active].BooleanFieldSettings.DefaultValue",
                         "True"});
 #line 58
         testRunner.And("I fill in", ((string)(null)), table5, "And ");
@@ -191,7 +191,7 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table6.AddRow(new string[] {
-                        "Fields[0].BooleanFieldSettings.Optional",
+                        "Fields[Active].BooleanFieldSettings.Optional",
                         "false"});
 #line 67
         testRunner.And("I fill in", ((string)(null)), table6, "And ");
--- a/src/Orchard.Specs/DateTime.feature
+++ b/src/Orchard.Specs/DateTime.feature
@@ -61,7 +61,7 @@ Scenario: Creating and using Date fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                 | value                       |
-            | Fields[0].DateTimeFieldSettings.Hint | Enter the date of the event |
+            | Fields[EventDate].DateTimeFieldSettings.Hint | Enter the date of the event |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Enter the date of the event"
@@ -70,7 +70,7 @@ Scenario: Creating and using Date fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                    | value    |
-            | Fields[0].DateTimeFieldSettings.Display | DateOnly |
+            | Fields[EventDate].DateTimeFieldSettings.Display | DateOnly |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Event.EventDate.Editor.Date"
@@ -80,7 +80,7 @@ Scenario: Creating and using Date fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                    | value    |
-            | Fields[0].DateTimeFieldSettings.Display | TimeOnly |
+            | Fields[EventDate].DateTimeFieldSettings.Display | TimeOnly |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Event.EventDate.Editor.Time"
@@ -90,8 +90,8 @@ Scenario: Creating and using Date fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                     | value       |
-            | Fields[0].DateTimeFieldSettings.Display  | DateAndTime |
-            | Fields[0].DateTimeFieldSettings.Required | true        |
+            | Fields[EventDate].DateTimeFieldSettings.Display  | DateAndTime |
+            | Fields[EventDate].DateTimeFieldSettings.Required | true        |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Event.EventDate.Editor.Date"
@@ -119,8 +119,8 @@ Scenario: Creating and using Date fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                     | value    |
-            | Fields[0].DateTimeFieldSettings.Display  | DateOnly |
-            | Fields[0].DateTimeFieldSettings.Required | true     |
+            | Fields[EventDate].DateTimeFieldSettings.Display  | DateOnly |
+            | Fields[EventDate].DateTimeFieldSettings.Required | true     |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Event.EventDate.Editor.Date"
@@ -131,8 +131,8 @@ Scenario: Creating and using Date fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                     | value    |
-            | Fields[0].DateTimeFieldSettings.Display  | TimeOnly |
-            | Fields[0].DateTimeFieldSettings.Required | true     |
+            | Fields[EventDate].DateTimeFieldSettings.Display  | TimeOnly |
+            | Fields[EventDate].DateTimeFieldSettings.Required | true     |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Event.EventDate.Editor.Date"
@@ -143,9 +143,9 @@ Scenario: Creating and using Date fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value       |
-            | Fields[0].DateTimeFieldSettings.Display     | DateAndTime |
-            | Fields[0].DateTimeFieldSettings.Editor.Date | 01/31/2016  |
-            | Fields[0].DateTimeFieldSettings.Editor.Time | 10:00 AM    |
+            | Fields[EventDate].DateTimeFieldSettings.Display     | DateAndTime |
+            | Fields[EventDate].DateTimeFieldSettings.Editor.Date | 01/31/2016  |
+            | Fields[EventDate].DateTimeFieldSettings.Editor.Time | 10:00 AM    |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Event.EventDate.Editor.Date"
@@ -190,8 +190,8 @@ Scenario: Creating and using date time fields in another culture
         And I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                     | value       |
-            | Fields[0].DateTimeFieldSettings.Display  | DateAndTime |
-            | Fields[0].DateTimeFieldSettings.Required | true        |
+            | Fields[EventDate].DateTimeFieldSettings.Display  | DateAndTime |
+            | Fields[EventDate].DateTimeFieldSettings.Required | true        |
         And I hit "Save"
     When I go to "Admin/Contents/Create/Event"
         And I fill in 
--- a/src/Orchard.Specs/DateTime.feature.cs
+++ b/src/Orchard.Specs/DateTime.feature.cs
@@ -186,7 +186,7 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table6.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Hint",
+                        "Fields[EventDate].DateTimeFieldSettings.Hint",
                         "Enter the date of the event"});
 #line 62
         testRunner.And("I fill in", ((string)(null)), table6, "And ");
@@ -203,7 +203,7 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table7.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Display",
+                        "Fields[EventDate].DateTimeFieldSettings.Display",
                         "DateOnly"});
 #line 71
         testRunner.And("I fill in", ((string)(null)), table7, "And ");
@@ -222,7 +222,7 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table8.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Display",
+                        "Fields[EventDate].DateTimeFieldSettings.Display",
                         "TimeOnly"});
 #line 81
         testRunner.And("I fill in", ((string)(null)), table8, "And ");
@@ -241,10 +241,10 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table9.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Display",
+                        "Fields[EventDate].DateTimeFieldSettings.Display",
                         "DateAndTime"});
             table9.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Required",
+                        "Fields[EventDate].DateTimeFieldSettings.Required",
                         "true"});
 #line 91
         testRunner.And("I fill in", ((string)(null)), table9, "And ");
@@ -309,10 +309,10 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table13.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Display",
+                        "Fields[EventDate].DateTimeFieldSettings.Display",
                         "DateOnly"});
             table13.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Required",
+                        "Fields[EventDate].DateTimeFieldSettings.Required",
                         "true"});
 #line 120
         testRunner.And("I fill in", ((string)(null)), table13, "And ");
@@ -333,10 +333,10 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table14.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Display",
+                        "Fields[EventDate].DateTimeFieldSettings.Display",
                         "TimeOnly"});
             table14.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Required",
+                        "Fields[EventDate].DateTimeFieldSettings.Required",
                         "true"});
 #line 132
         testRunner.And("I fill in", ((string)(null)), table14, "And ");
@@ -357,13 +357,13 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table15.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Display",
+                        "Fields[EventDate].DateTimeFieldSettings.Display",
                         "DateAndTime"});
             table15.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Editor.Date",
+                        "Fields[EventDate].DateTimeFieldSettings.Editor.Date",
                         "01/31/2016"});
             table15.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Editor.Time",
+                        "Fields[EventDate].DateTimeFieldSettings.Editor.Time",
                         "10:00 AM"});
 #line 144
         testRunner.And("I fill in", ((string)(null)), table15, "And ");
@@ -461,10 +461,10 @@ this.ScenarioSetup(scenarioInfo);
                         "name",
                         "value"});
             table18.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Display",
+                        "Fields[EventDate].DateTimeFieldSettings.Display",
                         "DateAndTime"});
             table18.AddRow(new string[] {
-                        "Fields[0].DateTimeFieldSettings.Required",
+                        "Fields[EventDate].DateTimeFieldSettings.Required",
                         "true"});
 #line 191
         testRunner.And("I fill in", ((string)(null)), table18, "And ");
--- a/src/Orchard.Specs/Enumeration.feature
+++ b/src/Orchard.Specs/Enumeration.feature
@@ -35,7 +35,7 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                       | value   |
-            | Fields[0].EnumerationFieldSettings.Options | Seattle |
+            | Fields[Location].EnumerationFieldSettings.Options | Seattle |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "<option>Seattle</option>"
@@ -57,7 +57,7 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                    | value                    |
-            | Fields[0].EnumerationFieldSettings.Hint | Please select a location |
+            | Fields[Location].EnumerationFieldSettings.Hint | Please select a location |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "Please select a location"
@@ -66,7 +66,7 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value    |
-            | Fields[0].EnumerationFieldSettings.ListMode | Dropdown |
+            | Fields[Location].EnumerationFieldSettings.ListMode | Dropdown |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "select id=\"Event_Location_Value\" name=\"Event.Location.Value\""
@@ -75,7 +75,7 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value       |
-            | Fields[0].EnumerationFieldSettings.ListMode | Radiobutton |
+            | Fields[Location].EnumerationFieldSettings.ListMode | Radiobutton |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "input id=\"Event_Location_Value\" name=\"Event.Location.Value\" type=\"radio\""
@@ -84,7 +84,7 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value   |
-            | Fields[0].EnumerationFieldSettings.ListMode | Listbox |
+            | Fields[Location].EnumerationFieldSettings.ListMode | Listbox |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "select id=\"Event_Location_SelectedValues\" multiple=\"multiple\" name=\"Event.Location.SelectedValues\""
@@ -93,7 +93,7 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value    |
-            | Fields[0].EnumerationFieldSettings.ListMode | Checkbox |
+            | Fields[Location].EnumerationFieldSettings.ListMode | Checkbox |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "input type=\"checkbox\" name=\"Event.Location.SelectedValues\""
@@ -102,7 +102,7 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                        | value |
-            | Fields[0].EnumerationFieldSettings.Required | true  |
+            | Fields[Location].EnumerationFieldSettings.Required | true  |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
         And I hit "Save"
@@ -112,9 +112,9 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                            | value    |
-            | Fields[0].EnumerationFieldSettings.Options      | Seattle  |
-            | Fields[0].EnumerationFieldSettings.ListMode     | Dropdown |
-            | Fields[0].EnumerationFieldSettings.DefaultValue | Seattle  |
+            | Fields[Location].EnumerationFieldSettings.Options      | Seattle  |
+            | Fields[Location].EnumerationFieldSettings.ListMode     | Dropdown |
+            | Fields[Location].EnumerationFieldSettings.DefaultValue | Seattle  |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "selected=\"selected">Seattle"
@@ -123,8 +123,8 @@ Scenario: Creating and using Enumeration fields
     When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                            | value   |
-            | Fields[0].EnumerationFieldSettings.Required     | true    |
-            | Fields[0].EnumerationFieldSettings.ListMode     | Listbox |
+            | Fields[Location].EnumerationFieldSettings.Required     | true    |
+            | Fields[Location].EnumerationFieldSettings.ListMode     | Listbox |
         And I hit "Save"
         And I go to "Admin/Contents/Create/Event"
     Then I should see "required=\"required\""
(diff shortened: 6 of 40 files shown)


MatteoPiovanelli:
Posted my reply here:
https://github.com/OrchardCMS/Orchard/issues/7708#issuecomment-381059724

Skrypt:
After investigating and removing MenuPart and LayoutPart I can successfully run @MatteoPiovanelli-Laser tool and create 2 content items.

Here are the parts that concurrency needs to be fixed that I found so far : 

- ~~AutoroutePart~~
- ~~TagsPart~~
- ~~PublishLaterPart~~
- MenuPart
    Issue when trying to retrieve all website Menus when creating new records concurrently : _menuService.GetMenus()
- ~~LayoutPart
    Issue when trying to retrieve Layout Templates and Elements when creating new records concurrently.~~

## Update

After looking at the LayoutPart there's a lot of DB queries made to the database to retrieve Layout templates and Layout Elements. I think these requests are made every time we save a content item. I believe we should maybe use a ConcurrentDictionary to store them on site start and update when needed instead of retrieving these in the DB. In the actual form, we could try to use the LockingProvider on these if we can make it work as an alternative. But I believe that using the **ConcurrentDictionnary** would also optimize Orchard performances. The same approach could be taken for website menus just to remove the extra query made every time we open New Page in the admin.

## Update 2

After merging LockProvider PR in this PR I've been able to successfully run TagsPart concurrently. The PublishLaterPart had a database query from it's ViewModel which seemed odd to me. I've removed that code but it was there for a purpose. I will need to find/fix/repro the original issue.

## Update 3

I've runned @MatteoPiovanelli-Laser tool using SQL Ce database and I didn't get any database lock even if I didn't fixed the MenuPart and LayoutPart issues.

@sebastienros @sfmskywalker @MatteoPiovanelli-Laser

Skrypt:
This PR includes #8019

Skrypt:
Also I noticed that the LockingProvider will make the UI crash if you are using the command line tool to write concurrently and try to navigate the admin content list at the same time.

MatteoPiovanelli:
Maybe enabling the NHibernate logs would tell us more about the reason why there seem to be no lock on SQLCE.

What kind of crashes are you seeing on the UI? Could it be something else is deadlocking there?

MatteoPiovanelli:
I just launched the ParallelTests.js script on a brand new tenant using this PR as code.
I had to tweak the tests in the collection to parse correctly the output (the div to look into has class message-Success rather than message-Information).
After I did that, I get no apparent deadlock, but the resulting computed DisplayAlias properties are not correct, as some are duplicate. See the screenshot:

image

MatteoPiovanelli:
The screenshost here shows, for the same condition above, also the `AliasRecord` table.
image

Looks like we replaced a race causing a deadlock on one side with a race messing up the data on the other.

Skrypt:
You're not getting any lock issues from MenuPart or LayoutPart? Your tenant uses SQL Server or something else? The AutoroutePart issue is probably just a small fix remaining to do.

Skrypt:
The intermittent issue I'm getting is : "Transaction not connected, or was disconnected" when I'm browsing the website and running concurrent writes. The issue with the AutoroutePart is that it doesn't find a matching Alias while comparing with the RouteCollection. Normally it should at all times; need to find what's happening exactly.

From the MainMenuService here is what I tried : 

```C#
using System;
using System.Linq;
using Orchard.ContentManagement;
using Orchard.Core.Navigation.Models;
using Orchard.Core.Title.Models;
using Orchard.Locking;

namespace Orchard.Core.Navigation.Services
{
    public class MainMenuService : IMenuService
    {
        private readonly IContentManager _contentManager;
        private readonly ILockingProvider _lockingProvider;

        public MainMenuService(IContentManager contentManager, ILockingProvider lockingProvider)
        {
            _contentManager = contentManager;
            _lockingProvider = lockingProvider;
        }

        public IEnumerable<MenuPart> Get()
        {
            return _contentManager.Query<MenuPart, MenuPartRecord>().List();
        }

        public IEnumerable<MenuPart> GetMenuParts(int menuId)
        {
            return _contentManager
                .Query<MenuPart, MenuPartRecord>()
                .Where(x => x.MenuId == menuId)
                .List();
        }

        public IContent GetMenu(string menuName)
        {
            if (string.IsNullOrWhiteSpace(menuName))
            {
                return null;
            }

            return _contentManager.Query<TitlePart, TitlePartRecord>()
                .Where(x => x.Title == menuName)
                .ForType("Menu")
                .Slice(0, 1)
                .FirstOrDefault();
        }

        public IContent GetMenu(int menuId)
        {
            return _contentManager.Get(menuId, VersionOptions.Published);
        }

        public MenuPart Get(int menuPartId)
        {
            return _contentManager.Get<MenuPart>(menuPartId);
        }

        public IContent Create(string name)
        {

            if (string.IsNullOrWhiteSpace(name))
            {
                throw new ArgumentNullException(name);
            }

            ContentItem result = null;

            var lockString = string.Join(".",
                "MainMenuService.Create",
                name);

            _lockingProvider.Lock(lockString,
                () => {
                    result = _contentManager.Create("Menu");
                    result.As<TitlePart>().Title = name;
                });

            return result;
        }

        public void Delete(MenuPart menuPart)
        {
            _contentManager.Remove(menuPart.ContentItem);
        }

        public IEnumerable<ContentItem> GetMenus()
        {
            return _contentManager.Query().ForType("Menu").Join<TitlePartRecord>().OrderBy(x => x.Title).List();
        }
    }
}
```

Though it's failing on GetMenus() method which is trying to retrieve data from :  

image

So the issue is that NHibernate locks the Orchard_Framework_ContentItemRecord table every time it does a CREATE and we can't use the _contentManager.Query() to retrieve data while it is doing so. I found out that we could use the repository data instead and it will work concurrently. Example : 

https://github.com/OrchardCMS/Orchard/blob/9e98c8f6e673f367a41f0230d0deca388cc1dde0/src/Orchard.Web/Modules/Orchard.Tags/Services/TagService.cs#L48

Skrypt:
Here is what Dapper does : https://github.com/StackExchange/Dapper#limitations-and-caveats
Somehow I think in Orchard Core this type of issues should be easy to fix "if needed" since we have a cache that comes from Dapper as a ConcurrentDictionnary.

MatteoPiovanelli:
I had removed the `LayoutPart`, and was getting no issue from `MenuPart`. My logs were completely clean.

I always use SQL Server, because that is the "realistic" scenario for us, meaning that is what we also have on our production servers.

That Autoroute issue (duplicated DisplayAlias) is something I saw a bunch of times in different conditions, and, in my cases, had to do with reading and then getting a race on write without a check there. Might be that locking there would fix this instance. I'll check as soon as I have a chance to put my hands on the code again.

As you said, I feel many if not most of the issues have to do with the locks on ContentItemRecord (and I add ContentItemVersionRecord and ContentTypeRecord) intrinsically generated by the transactions there.

Skrypt:
Nhibernate has a documentation on how to handle optimistic concurrency but I think that it's never been taken into consideration since in 01 the DB was structured as a documentDB. The LockingProvider should work if we use it on top of my changes on the AutoroutePart. Somehow the LockingProvider should be doing the same without needing to add a Version field on each table like NHibernate ask for handling optimistic concurrency. I've just not been successfull with the LockingProvider so far; maybe there's something I'm doing wrong or a setup on SQL Server that is different than yours ... I've enabled TCP/IP transport and Named Pipes locally maybe there's something different that I'm missing.

MatteoPiovanelli:
Using a `LockingProvider` will probably not be enough to solve the issue from `AutorouteService` as it is, because:
 - In `ProcessPath`, the call to `GetSimilarPaths(string)` fetches the paths from `_aliasService.List()`, that fetches from `_aliasStorage.List()`.
 - In `ProcessPath`, after some computation, we set `part.DisplayAlias = newPath;`, and that writes on the db in the `AutoroutePartRecord`. We cannot be writing to the `AliasStorage` here, because we may be computing the `DisplayAlias` for a draft version of a `ContentItem`.
 - Only in `PublishAlias(AutoroutePart)` the new `DisplayAlias `is added to the `AliasStorage`.

The version of `GetSimilarPaths(string)` currently live also only fetches the Published versions (because that is the default for `DefaultContentQuery`), however it does so from the same table where `ProcessPath` also writes, so it should be possible (in principle) to lock there and avoid a race resulting in duplicate `DisplayAlias` values. The issue there arises, I think, more from the fact that even though that code may be locked, that does not prevent other code from taking their own locks on the records that are bound by default on all `DefaultContentQuery` instances. 
It could be argued that the `DisplayAlias` only really matters for Pulished ContentItems, however that is often not true in practice, because a common scenario we see form our clients is that they would create several related contents as drafts, with links among them written down using the `DisplayAlias`, and then publish them all at a later time. If the `DisplayAlias` changed on publish, all the links on those contents would be dead.

Note on the other hand that the version of `GetSimilarPaths(string)` in this PR appears to work with no issue where we are not attempting to concurrently create contents with the same title.

Skrypt:
Ok, so I made a quick test and saved a Page as a draft and when I want to navigate this page I get the standard controller/action url. So the AutoroutePart should only be generating Alias for Published content items. Though, right now we still run the full ProcessPath methods even if an AutoroutePart is coming from a content item that is a draft.

MatteoPiovanelli:
I think that draft versions need to have a DisplayAlias. The reason is the scenario I briefly described above, where our clients create the drafts for a bunch of contents, adding links among them by explicitly typing the anchor tags in, and then publish them all together.
Of course, the live version of Orchard gives no guarantee that the DisplayAlias will be the same after a draft is published, and if the publication is invoked by a process (e.g. a PublishLaterPart) the content creator will likely see no warning of that.

I'll add a few comments on the code of the PR.

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -60,9 +60,9 @@ public class AutoroutePartHandler : ContentHandler {
             // Should it become the home page?
             if (part.PromoteToHomePage) {
                 // Get the current homepage an unmark it as the homepage.
-                var currentHomePage = _homeAliasService.GetHomePage(VersionOptions.Latest);
-                if(currentHomePage != null && currentHomePage.Id != part.Id) {
-                    var autoroutePart = currentHomePage.As<AutoroutePart>();
+                var currentHomePageId = _homeAliasService.GetHomePageId();
+                if(currentHomePageId != part.Id) {
+                    var autoroutePart = part.As<AutoroutePart>();
```
MatteoPiovanelli: This changes completely what the handler is doing. The live version is removing the `PromoteToHomePage` flag from the current homepage, rather than from the AutoroutePart being handled right now.

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -79,28 +79,33 @@ public class AutoroutePartHandler : ContentHandler {
         }
 
         private void ProcessAlias(AutoroutePart part) {
-            // Generate an alias if one as not already been entered.
-            if (String.IsNullOrWhiteSpace(part.DisplayAlias)) {
-                part.DisplayAlias = _autorouteService.Value.GenerateAlias(part);
-            }
+            if (!part.HasDraft())
```
MatteoPiovanelli: Looking at the implementation of `HasDraft(this IContent)`, it returns `true` if the content being processed is not the published version, or if it is published but not latest. I point this out just to make sure this is what you intended. 
However, I am not convinced we should not give a DisplayAlias to drafts, even though that will not go in the AliasStorage, as that contains only the published ones (since they are the ones we actually need to use as aliases for content).

MatteoPiovanelli: Note also that this does not prevent the creation of duplicate DisplayAlias values for concurrent creation of published or non-draftable content items

MatteoPiovanelli:
I have no idea if I can push to this PR, but I figured out a thing that sort of works:
I saw no deadlocks in my tests, and no duplicate aliases, but the aliases "look bad" when I try to generate stuff from multiple concurrent content creations with the same title. Suppose that from the base alias "esta" (the result of `AutorouteService.GenerateAlias(AutoroutePart)`) I already have `DisplayAlias` "esta-8"; suppose also I am trying to publish concurrently 3 different contents, whose alias on update has been computed as "esta-9". Publishing will lead to the following three aliases, one for each content: "esta-9", "esta-9-2", "esta-9-3".

I wrote the modified `AutoroutePartHandler` I am currently proposing to a PR to the branch this PR is from (autoroute-concurrency)

Skrypt:
Yes feel free to push on this PR if it's the best one. You can also merge what you need on your PR Alternatively I don't mind 😉. I'll check why the Alias are like these but I'm pretty sure it's a single condition that I changed that is wrong. I'm trying at the same time to remove these database reads that are not neglectable on global perfs. For the Draft aliases I'm not against as long as we have also a strategy for cleaning them if the Drafts are deleted or published.

MatteoPiovanelli:
I don't think it's worth to store draft aliases anywhere. I was also talking about this with @HermesSbicego-Laser. The thing is, if we start storing and considering draft aliases, we end up having a hard time "replacing" a content (i.e. reusing an alias by deleting a content and creating a different one with the same alias). If we keep things as they are, we may have both drafts with the same DisplayAlias (that will not be used by the AliasService to reroute, so a duplicate here would not cause any real issue), and a DisplayAlias changing at the moment a content is published.

I think the second situation (that is both what is live on Orchard and what will be in place with the changes here) is safer, because it may only really cause issues when there are contents with the same result from `AutorouteService.GenerateAlias(AutoroutePart)`: that happens rarely enough for contents created by back office users, and is hardly an issue for content generated automatically.

MatteoPiovanelli:
I cannot push here, so I opened the PR #8028 targeting the branch of this one

MatteoPiovanelli:
By the way, I think all the changes we are introducing so far can easily go to 1.10.x

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
Skrypt: I think that yes if a Content Item HasDraft, even if there's a published version of the item, when we save it as a draft then we should not need to ProcessAlias() since we will eventually do if we publish that draft later on. Also that "already published" content item would have an Alias associated with it. Therefore, even if the title of the content item would have changed on that draft we should not alter the Alias of the currently published item. I think we should keep the draft url's as they are for now. This would be a separate PR "Add Alias to draft" and we would need to have a different route for them something like /draft/slug instead of /content/slug because they should not alter the currently published live url's.

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
Skrypt: I will make more tests but I've not had issues with editing/publishing/creating a home page.

Skrypt:
I added #8028 to this PR.

After testing the concurrency issue is gone. Though, I found out an issue with the Alias versions that are not always getting created properly. That's because in the AutoroutePartHandler PublishAlias() method when we run concurrently, sometimes ProcessAlias(part) will get executed twice before adding the actual Alias to the RouteCollection by doing _homeAliasService.PublishHomeAlias(part) later on that same method.

The issue that @MatteoPiovanelli-Laser had is probably related to the fact that he has already a lot of Managed Alias in it's database from previous tests that didn't get deleted. I needed to clean my Orchard_Alias_AliasRecord table from those before testing.

I'm targeting dev branch because that's what I'm used to do normally. I guess it will get merged later on in 1.10.x. Since this is a sensible PR I prefer having it in dev so that we get feedback from it first.

Skrypt:
Ok the Alias version is fixed with the last commit.

Skrypt:
The only concurrency issue I'm still getting with this PR is on the MenuPart. I tried different things with the LockingProvider and had no success so far. Actually on a fresh install the concurrency issue is not happening. Though when I restart the debugger it starts to happen back. I've looked in the Navigation_MenuPartRecord table and there's a lot of Null menus that gets created from the concurrency tests. @MatteoPiovanelli-Laser could you confirm that my tests are the same as yours? I've used what comes from your repository directly.

Here also is a bug reported in the log from the installation on the dev and this PR branch :  

```
2018-04-23 01:47:41,123 [11] Orchard.Recipes.Services.RecipeManager - Default - INFO Executing recipe 'Default'.
2018-04-23 01:47:41,147 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Enqueuing recipe step 'Feature'.
2018-04-23 01:47:41,172 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Enqueuing recipe step 'ContentDefinition'.
2018-04-23 01:47:41,188 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Enqueuing recipe step 'Settings'.
2018-04-23 01:47:41,202 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Enqueuing recipe step 'Migration'.
2018-04-23 01:47:41,218 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Enqueuing recipe step 'Command'.
2018-04-23 01:47:41,237 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Enqueuing recipe step 'ActivateShell'.
2018-04-23 01:47:41,442 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe steps.
2018-04-23 01:47:41,454 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe step 'Feature'.
2018-04-23 01:47:41,462 [11] Orchard.Recipes.Services.RecipeStepExecutor - Default - INFO Executing recipe step 'Feature'.
2018-04-23 01:47:41,473 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Executing recipe step 'Feature'.
2018-04-23 01:47:41,484 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO FeatureStep: Enabling features: Orchard.Blogs;Orchard.Comments;Orchard.Tags;Orchard.Alias;Orchard.Autoroute;TinyMce;Orchard.MediaLibrary;Orchard.ContentPicker;Orchard.PublishLater;Orchard.Resources;Orchard.Widgets;Orchard.ContentTypes;Orchard.Scripting;Orchard.Scripting.Lightweight;PackagingServices;Orchard.Packaging;Orchard.Projections;Orchard.Fields;Orchard.OutputCache;Orchard.Taxonomies;Orchard.Workflows;Orchard.Layouts;Orchard.Layouts.Tokens;TheThemeMachine
2018-04-23 01:47:41,577 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Finished executing recipe step 'Feature'.
2018-04-23 01:47:41,607 [11] Orchard.Recipes.Services.RecipeScheduler - Default - INFO Scheduling next step of recipe.
2018-04-23 01:47:43,631 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe steps.
2018-04-23 01:47:43,644 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe step 'ContentDefinition'.
2018-04-23 01:47:43,655 [11] Orchard.Recipes.Services.RecipeStepExecutor - Default - INFO Executing recipe step 'ContentDefinition'.
2018-04-23 01:47:43,665 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Executing recipe step 'ContentDefinition'.
2018-04-23 01:47:43,674 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO ContentDefinitionStep: Importing content type 'Page'.
2018-04-23 01:47:43,776 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO ContentDefinitionStep: Importing content type 'BlogPost'.
2018-04-23 01:47:43,801 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO ContentDefinitionStep: Importing content part 'BodyPart'.
2018-04-23 01:47:43,817 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Finished executing recipe step 'ContentDefinition'.
2018-04-23 01:47:43,831 [11] Orchard.Recipes.Services.RecipeScheduler - Default - INFO Scheduling next step of recipe.
2018-04-23 01:47:43,981 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe steps.
2018-04-23 01:47:43,993 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe step 'Settings'.
2018-04-23 01:47:44,000 [11] Orchard.Recipes.Services.RecipeStepExecutor - Default - INFO Executing recipe step 'Settings'.
2018-04-23 01:47:44,008 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Executing recipe step 'Settings'.
2018-04-23 01:47:44,133 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Finished executing recipe step 'Settings'.
2018-04-23 01:47:44,146 [11] Orchard.Recipes.Services.RecipeScheduler - Default - INFO Scheduling next step of recipe.
2018-04-23 01:47:44,286 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe steps.
2018-04-23 01:47:44,300 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe step 'Migration'.
2018-04-23 01:47:44,311 [11] Orchard.Recipes.Services.RecipeStepExecutor - Default - INFO Executing recipe step 'Migration'.
2018-04-23 01:47:44,321 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Executing recipe step 'Migration'.
2018-04-23 01:47:44,414 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Finished executing recipe step 'Migration'.
2018-04-23 01:47:44,426 [11] Orchard.Recipes.Services.RecipeScheduler - Default - INFO Scheduling next step of recipe.
2018-04-23 01:47:44,573 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe steps.
2018-04-23 01:47:44,593 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe step 'Command'.
2018-04-23 01:47:44,608 [11] Orchard.Recipes.Services.RecipeStepExecutor - Default - INFO Executing recipe step 'Command'.
2018-04-23 01:47:44,622 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Executing recipe step 'Command'.
2018-04-23 01:47:44,638 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: layer create Default /LayerRule:"true" /Description:"The widgets in this layer are displayed on all pages"
2018-04-23 01:47:44,880 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: layer create Authenticated /LayerRule:"authenticated" /Description:"The widgets in this layer are displayed when the user is authenticated"
2018-04-23 01:47:44,906 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: layer create Anonymous /LayerRule:"not authenticated" /Description:"The widgets in this layer are displayed when the user is anonymous"
2018-04-23 01:47:44,953 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: layer create Disabled /LayerRule:"false" /Description:"The widgets in this layer are never displayed"
2018-04-23 01:47:44,976 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: layer create TheHomepage /LayerRule:"url '~/'" /Description:"The widgets in this layer are displayed on the home page"
2018-04-23 01:47:44,999 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: site setting set baseurl
2018-04-23 01:47:45,010 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: menu create /MenuName:"Main Menu"
2018-04-23 01:47:45,037 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: page create /Slug:"welcome-to-orchard" /Title:"Welcome to Orchard!" /Path:"welcome-to-orchard" /Homepage:true /Publish:true /UseWelcomeText:true
2018-04-23 01:48:15,226 [11] Orchard.Alias.Implementation.Updater.AliasHolderUpdater - Default - ERROR Exception during Alias refresh
NHibernate.Exceptions.GenericADOException: could not execute query
[ select aliasrecor0_.Id as Id409_, aliasrecor0_.Path as Path409_, aliasrecor0_.RouteValues as RouteVal3_409_, aliasrecor0_.Source as Source409_, aliasrecor0_.IsManaged as IsManaged409_, aliasrecor0_.Action_id as Action6_409_ from Orchard_Alias_AliasRecord aliasrecor0_ where aliasrecor0_.Id>@p0 order by aliasrecor0_.Id asc ]
  Name:p1 - Value:0
[SQL: select aliasrecor0_.Id as Id409_, aliasrecor0_.Path as Path409_, aliasrecor0_.RouteValues as RouteVal3_409_, aliasrecor0_.Source as Source409_, aliasrecor0_.IsManaged as IsManaged409_, aliasrecor0_.Action_id as Action6_409_ from Orchard_Alias_AliasRecord aliasrecor0_ where aliasrecor0_.Id>@p0 order by aliasrecor0_.Id asc] ---> System.Data.SqlClient.SqlException: Le délai d'exécution a expiré. Le délai d'attente s'est écoulé avant la fin de l'opération ou le serveur ne répond pas. ---> System.ComponentModel.Win32Exception: Dépassement du délai d’attente
   --- Fin de la trace de la pile d'exception interne ---
   à System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   à System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   à System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   à System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   à System.Data.SqlClient.SqlDataReader.TrySetMetaData(_SqlMetaDataSet metaData, Boolean moreInfo)
   à System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   à System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   à System.Data.SqlClient.SqlDataReader.get_MetaData()
   à System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption)
   à System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, Boolean inRetry, SqlDataReader ds, Boolean describeParameterEncryptionRequest)
   à System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry)
   à System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method)
   à System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior, String method)
   à System.Data.SqlClient.SqlCommand.ExecuteDbDataReader(CommandBehavior behavior)
   à System.Data.Common.DbCommand.System.Data.IDbCommand.ExecuteReader()
   à NHibernate.AdoNet.AbstractBatcher.ExecuteReader(IDbCommand cmd)
   à NHibernate.Loader.Loader.GetResultSet(IDbCommand st, Boolean autoDiscoverTypes, Boolean callable, RowSelection selection, ISessionImplementor session)
   à NHibernate.Loader.Loader.DoQuery(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies, IResultTransformer forcedResultTransformer)
   à NHibernate.Loader.Loader.DoQueryAndInitializeNonLazyCollections(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies, IResultTransformer forcedResultTransformer)
   à NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
   --- Fin de la trace de la pile d'exception interne ---
   à NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
   à NHibernate.Loader.Loader.ListIgnoreQueryCache(ISessionImplementor session, QueryParameters queryParameters)
   à NHibernate.Loader.Loader.List(ISessionImplementor session, QueryParameters queryParameters, ISet`1 querySpaces, IType[] resultTypes)
   à NHibernate.Loader.Hql.QueryLoader.List(ISessionImplementor session, QueryParameters queryParameters)
   à NHibernate.Hql.Ast.ANTLR.QueryTranslatorImpl.List(ISessionImplementor session, QueryParameters queryParameters)
   à NHibernate.Engine.Query.HQLQueryPlan.PerformList(QueryParameters queryParameters, ISessionImplementor session, IList results)
   à NHibernate.Impl.SessionImpl.List(IQueryExpression queryExpression, QueryParameters queryParameters, IList results)
   à NHibernate.Impl.AbstractSessionImpl.List(IQueryExpression queryExpression, QueryParameters parameters)
   à NHibernate.Impl.AbstractQueryImpl2.List()
   à NHibernate.Linq.DefaultQueryProvider.ExecuteQuery(NhLinqExpression nhLinqExpression, IQuery query, NhLinqExpression nhQuery)
   à NHibernate.Linq.DefaultQueryProvider.Execute(Expression expression)
   à NHibernate.Linq.DefaultQueryProvider.Execute[TResult](Expression expression)
   à Remotion.Linq.QueryableBase`1.GetEnumerator()
   à System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   à System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   à System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   à Orchard.Alias.Implementation.Storage.AliasStorage.List(Expression`1 predicate) dans E:\Repositories\Orchard\src\Orchard.Web\Modules\Orchard.Alias\Implementation\Storage\AliasStorage.cs:ligne 141
   à Orchard.Alias.Implementation.Updater.AliasHolderUpdater.Refresh() dans E:\Repositories\Orchard\src\Orchard.Web\Modules\Orchard.Alias\Implementation\Updater\AliasHolderUpdater.cs:ligne 32
2018-04-23 01:48:46,012 [11] Orchard.Alias.Implementation.Updater.AliasHolderUpdater - Default - ERROR Exception during Alias refresh
NHibernate.Exceptions.GenericADOException: could not execute query
[ select aliasrecor0_.Id as Id409_, aliasrecor0_.Path as Path409_, aliasrecor0_.RouteValues as RouteVal3_409_, aliasrecor0_.Source as Source409_, aliasrecor0_.IsManaged as IsManaged409_, aliasrecor0_.Action_id as Action6_409_ from Orchard_Alias_AliasRecord aliasrecor0_ where aliasrecor0_.Id>@p0 order by aliasrecor0_.Id asc ]
  Name:p1 - Value:0
[SQL: select aliasrecor0_.Id as Id409_, aliasrecor0_.Path as Path409_, aliasrecor0_.RouteValues as RouteVal3_409_, aliasrecor0_.Source as Source409_, aliasrecor0_.IsManaged as IsManaged409_, aliasrecor0_.Action_id as Action6_409_ from Orchard_Alias_AliasRecord aliasrecor0_ where aliasrecor0_.Id>@p0 order by aliasrecor0_.Id asc] ---> System.Data.SqlClient.SqlException: Le délai d'exécution a expiré. Le délai d'attente s'est écoulé avant la fin de l'opération ou le serveur ne répond pas. ---> System.ComponentModel.Win32Exception: Dépassement du délai d’attente
   --- Fin de la trace de la pile d'exception interne ---
   à System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   à System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   à System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   à System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   à System.Data.SqlClient.SqlDataReader.TrySetMetaData(_SqlMetaDataSet metaData, Boolean moreInfo)
   à System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   à System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   à System.Data.SqlClient.SqlDataReader.get_MetaData()
   à System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption)
   à System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, Boolean inRetry, SqlDataReader ds, Boolean describeParameterEncryptionRequest)
   à System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry)
   à System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method)
   à System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior, String method)
   à System.Data.SqlClient.SqlCommand.ExecuteDbDataReader(CommandBehavior behavior)
   à System.Data.Common.DbCommand.System.Data.IDbCommand.ExecuteReader()
   à NHibernate.AdoNet.AbstractBatcher.ExecuteReader(IDbCommand cmd)
   à NHibernate.Loader.Loader.GetResultSet(IDbCommand st, Boolean autoDiscoverTypes, Boolean callable, RowSelection selection, ISessionImplementor session)
   à NHibernate.Loader.Loader.DoQuery(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies, IResultTransformer forcedResultTransformer)
   à NHibernate.Loader.Loader.DoQueryAndInitializeNonLazyCollections(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies, IResultTransformer forcedResultTransformer)
   à NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
   --- Fin de la trace de la pile d'exception interne ---
   à NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
   à NHibernate.Loader.Loader.ListIgnoreQueryCache(ISessionImplementor session, QueryParameters queryParameters)
   à NHibernate.Loader.Loader.List(ISessionImplementor session, QueryParameters queryParameters, ISet`1 querySpaces, IType[] resultTypes)
   à NHibernate.Loader.Hql.QueryLoader.List(ISessionImplementor session, QueryParameters queryParameters)
   à NHibernate.Hql.Ast.ANTLR.QueryTranslatorImpl.List(ISessionImplementor session, QueryParameters queryParameters)
   à NHibernate.Engine.Query.HQLQueryPlan.PerformList(QueryParameters queryParameters, ISessionImplementor session, IList results)
   à NHibernate.Impl.SessionImpl.List(IQueryExpression queryExpression, QueryParameters queryParameters, IList results)
   à NHibernate.Impl.AbstractSessionImpl.List(IQueryExpression queryExpression, QueryParameters parameters)
   à NHibernate.Impl.AbstractQueryImpl2.List()
   à NHibernate.Linq.DefaultQueryProvider.ExecuteQuery(NhLinqExpression nhLinqExpression, IQuery query, NhLinqExpression nhQuery)
   à NHibernate.Linq.DefaultQueryProvider.Execute(Expression expression)
   à NHibernate.Linq.DefaultQueryProvider.Execute[TResult](Expression expression)
   à Remotion.Linq.QueryableBase`1.GetEnumerator()
   à System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   à System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   à System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   à Orchard.Alias.Implementation.Storage.AliasStorage.List(Expression`1 predicate) dans E:\Repositories\Orchard\src\Orchard.Web\Modules\Orchard.Alias\Implementation\Storage\AliasStorage.cs:ligne 141
   à Orchard.Alias.Implementation.Updater.AliasHolderUpdater.Refresh() dans E:\Repositories\Orchard\src\Orchard.Web\Modules\Orchard.Alias\Implementation\Updater\AliasHolderUpdater.cs:ligne 32
2018-04-23 01:48:46,356 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: menuitem create /MenuPosition:"0" /MenuText:"Home" /Url:"~/" /MenuName:"Main Menu"
2018-04-23 01:48:46,455 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: widget create MenuWidget /Title:"Main Menu" /RenderTitle:false /Zone:"Navigation" /Position:"1" /Layer:"Default" /Identity:"MenuWidget1" /MenuName:"Main Menu"
2018-04-23 01:48:46,525 [11] Orchard.Recipes.Services.RecipeExecutionLogger - Default - INFO CommandStep: Executing command: theme activate "The Theme Machine"
2018-04-23 01:48:46,539 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Finished executing recipe step 'Command'.
2018-04-23 01:48:46,557 [11] Orchard.Recipes.Services.RecipeScheduler - Default - INFO Scheduling next step of recipe.
2018-04-23 01:48:46,769 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe steps.
2018-04-23 01:48:46,789 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe step 'ActivateShell'.
2018-04-23 01:48:46,807 [11] Orchard.Recipes.Services.RecipeStepExecutor - Default - INFO Executing recipe step 'ActivateShell'.
2018-04-23 01:48:46,822 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Executing recipe step 'ActivateShell'.
2018-04-23 01:48:46,841 [11] Orchard.Recipes.Providers.RecipeHandlers.RecipeExecutionStepHandler - Default - INFO Finished executing recipe step 'ActivateShell'.
2018-04-23 01:48:46,938 [11] Orchard.Recipes.Services.RecipeScheduler - Default - INFO Scheduling next step of recipe.
2018-04-23 01:48:47,142 [11] Orchard.Recipes.Services.RecipeStepQueue - Default - INFO Dequeuing recipe steps.
2018-04-23 01:48:47,168 [11] Orchard.Recipes.Services.RecipeStepExecutor - Default - INFO No more recipe steps left to execute.
2018-04-23 01:48:47,191 [11] Orchard.Recipes.Services.RecipeScheduler - Default - INFO All recipe steps executed; restarting shell.

```

MatteoPiovanelli:
- I am closing #8028 and #8019 because they are both merged here, and wat was done there is more advanced here.
 - That error you see while creating the tenant is not caused by the changes here, but is a pre-existing issue. See #7597.
 - With Page items defined as per the Default recipe, running the ParallelTests.js script I get errors on the first iteration (2 concurrent creations) on `MainMenuService.GetMenus()` and `MenuPartDriver.Editor(MenuPart, Object)` for MenuPart, and on `LayoutManager.GetTemplates()`. After this I have only one of the Page ContentItems created correctly, but two records have been added to alias record (in the screenshot, AliasRecord is the table on the top, AutoroutePartRecord is at the bottom):

image

 - If I remove MenuPart and LayoutPart from the definition of Page everything looks right. 

Currently, the issues I see with concurrency from this PR are:
 - Something with MenuPart
 - Something with LayoutPart
 - AliasRecord gets created even though the content creation failed and everything else rolled back.

MatteoPiovanelli:
Starting from a fresh tenant with the Default recipe, I removed LayoutPart from the definition of Page. The screnshot shows the initial contents of AliasRecord, AutoroutePartRecord and MenuPartRecord (top to bottom):

image

The Postman collection called in ParallelTests.js sends the default values for the MenuPart:
```
{
    "key": "MenuPart.OnMenu",
    "value": "false",
    "type": "text",
    "enabled": true
},
{
    "key": "MenuPart.CurrentMenuId",
    "value": "8",
    "type": "text",
    "enabled": true
},
{
    "key": "MenuPart.MenuText",
    "value": "",
    "type": "text",
    "enabled": true
}
```

After launching the ParallelTests.js script, and failing at 2 concurrent creations with those errors related to MenuPart, the content of those three tables from above is:

image

Note the last record in MenuPartRecord, corresponding to the Page that was created, has MenuText = NULL and MenuPosition = NULL: this is not (should not be) an issue, because it the same as if we were to create a single Page content through the UI and not flag the "Show on a menu" checkbox.

MatteoPiovanelli:
I tried reverting back to the live version of MenuPartDriver, and I get the same. The same deadlock in calling `MainMenuService.GetMenus()`, as well as the alias added to AliasRecord for the Page that has not been created.

Skrypt:
Ok, thanks @MatteoPiovanelli-Laser. Could you try, from a fresh install, remove only MenuPart on the Page Content Type and run these tests. If it works then we would have the same results. For some reason when a specific Part fails it makes the LockingProvider fail and therefore makes all the other parts fail too.

Also, the problem with the second Alias getting created is because I've changed the LockingProvider to encapsulate the get and create methods together. We would need to add a Try / Catch in that lock to prevent creating the Alias if we fail getting the second item "I think".

For the MenuPart issue the only thing I can think of right now is that the DB read method in that module is made in it's Driver Editor method(). Maybe that's what makes it fail. The method is called 4 times per page creation request.

MatteoPiovanelli:
Is `LayoutPart` not crashing in your tests? It crashes in mine. If I remove both LayoutPart and MenuPart, on a fresh tenant, I see no other errors creating Page ContentItems.

The methods from `ILockingProvider` take additional parameters to handle exceptions. They are two `Action<Exception>`: one handles exceptions while holding the lock, the other handles them after releasing the lock. I will look in more detail at what is going on when the Alias is published to understand if handling the failed creation there is the right way to go, or if things should stop elsewhere.
I will also try to store the GetMenus() in a Lazy object to see if that helps.

Skrypt:
From a fresh install without restarting VS and removing MenuPart from page content type I had no issue.

sebastienros:
If I was to call `GetMenus` from an action, with concurrent calls, would it crash?

MatteoPiovanelli:
The fact LayoutPart is causing issue on my end and not on yours is something else we'll need to figure out...

@sebastienros I don't think so, not necessarily. I think the crash is due to us calling that while also writing. I think it's the writes to ContentItemRecord and/or ContentItemVersionRecord. Just calling the `GetMenus` method by itself, even several times concurrently, is basically what's done by opening a bunch of `Create/Page`. That never seemed to cause issues on my end.

MatteoPiovanelli:
From `GetMenus()`, changing from 

    _contentManager.Query().ForType("Menu").Join<TitlePartRecord>().OrderBy(x => x.Title).List();

to

    _contentManager.Query().ForType("Menu").List();

Does not change the error we are seeing (except of course it changes the query being logged), as further evidence of the issues being related to `ContentItemRecord`, `ContentItemVersionRecord` or `ContentTypeRecord`

MatteoPiovanelli:
I was looking at `DefaultContentQuery`. The issue I was seeing with the `LayoutPart` failin on its `GetTemplates()` probably has to do with the code here (that comment is in the source):
```
private int GetContentTypeRecordId(string contentType) {
	return _cacheManager.Get(contentType + "_Record", true, ctx => {
		ctx.Monitor(_signals.When(contentType + "_Record"));

		var contentTypeRecord = _contentTypeRepository.Get(x => x.Name == contentType);

		if (contentTypeRecord == null) {
			//TEMP: this is not safe... ContentItem types could be created concurrently?
			contentTypeRecord = new ContentTypeRecord { Name = contentType };
			_contentTypeRepository.Create(contentTypeRecord);
		}

		return contentTypeRecord.Id;
	});
}
```

MatteoPiovanelli:
I'll post here some code I am testing that seems to fix things for the MenuService:

```
using System;
using System.Linq;
using Orchard.Caching;
using Orchard.ContentManagement;
using Orchard.Core.Navigation.Models;
using Orchard.Core.Title.Models;

namespace Orchard.Core.Navigation.Services {
    public class MainMenuService : IMenuService {
        private readonly IContentManager _contentManager;
        private ICacheManager _cacheManager;
        private ISignals _signals;

        public MainMenuService(IContentManager contentManager,
            ICacheManager cacheManager,
            ISignals signals) {
            _contentManager = contentManager;
            _cacheManager = cacheManager;
            _signals = signals;
        }

        public IEnumerable<MenuPart> Get() {
            return _contentManager.Query<MenuPart, MenuPartRecord>().List();
        }

        public IEnumerable<MenuPart> GetMenuParts(int menuId) {
            return _contentManager
                .Query<MenuPart, MenuPartRecord>()
                .Where( x => x.MenuId == menuId)
                .List();
        }

        public IContent GetMenu(string menuName) {
            if(string.IsNullOrWhiteSpace(menuName)) {
                return null;
            }

            return _contentManager.Query<TitlePart, TitlePartRecord>()
                .Where(x => x.Title == menuName)
                .ForType("Menu")
                .Slice(0, 1)
                .FirstOrDefault();
        }

        public IContent GetMenu(int menuId) {
            return _contentManager.Get(menuId, VersionOptions.Published);    
        }

        public MenuPart Get(int menuPartId) {
            return _contentManager.Get<MenuPart>(menuPartId);
        }

        public IContent Create(string name) {
            
            if(string.IsNullOrWhiteSpace(name)) {
                throw new ArgumentNullException(name);
            }
            
            var menu = _contentManager.Create("Menu");
            menu.As<TitlePart>().Title = name;

            _signals.Trigger("MainMenuService.AllMenus");

            return menu;
        }

        public void Delete(MenuPart menuPart) {
            _contentManager.Remove(menuPart.ContentItem);
        }



        private IEnumerable<ContentItem> AllMenus {
            get {
                var fromCache = FromCache();
                if (fromCache.Item1.AddMinutes(2) < DateTime.Now) {
                    // give 2 minutes lifetime to cache
                    _signals.Trigger("MainMenuService.AllMenus");
                    fromCache = FromCache();
                }
                return fromCache.Item2;
            }
        }
        private Tuple<DateTime, IEnumerable<ContentItem>> FromCache() {
            return _cacheManager
                    .Get<string, Tuple<DateTime, IEnumerable<ContentItem>>>("MainMenuService.AllMenus", true, ctx => {
                        ctx.Monitor(_signals.When("MainMenuService.AllMenus"));

                        return Tuple.Create(DateTime.Now,
                            _contentManager.Query().ForType("Menu").Join<TitlePartRecord>().OrderBy(x => x.Title).List());
                    });
        }

        public IEnumerable<ContentItem> GetMenus() {
            return AllMenus; // _contentManager.Query().ForType("Menu").Join<TitlePartRecord>().OrderBy(x => x.Title).List();
        }
    }
}
```

Again, I can't push directly on this branch, so @Skrypt could you please paste it in an see if things work for you as well?
Basically, I am trying to cache the menus for two minutes, so that concurrent calls are prevented from all trying to go fetch from db at the same time.

Skrypt:
The concurrency issue with the MenuPart seems to be fixed by using cache. From a fresh install I confirm that I still get an error from the LayoutPart which is really strange since it was working last week. The LockingProvider doesn't return consistent results; I'm not sure why. That's why I wanted to see if @MatteoPiovanelli-Laser had some insight about it. I'm pretty sure the best approach is using a cache or a singleton for these records. Also consider using a ConcurrentDictionary.

Skrypt:
For the MenuPart. What I meant on thursday meeting is that if we use the repository instead of the contentManager we should be able to retrieve the content item and it's related title. Here with the ContentManager we can do a join Query but I'm not sure if we can do this also with the repository. If we can do this; using the repository normally should fix the issue instead of using a cache on that ContentManager query.

Also, generally, the LockingProvider doesn't seem to fix the concurrency issue. In the AutoroutePart I used it to lock together ProcessAlias(part) and _homeService.PublishHomeAlias(part) since we want that both be completed before moving to the next concurrent process. Though, I'm not sure this is the best way to handle this. I need to make tests by making the PublishAlias method Async and see if I can make it work without the LockingProvider.

Skrypt:
related : https://github.com/OrchardCMS/Orchard/issues/8040

MatteoPiovanelli:
I still think the reason we are seeing the locks we are seeing has to do with the tables that are joined in every `DefaultContentQuery`, since those are being updated on every create or update call.
The `ILockingProvider` used at the level of drivers and services as we have been using it, cannot fix that, because it pretty much only can do what a `lock(obj) { }` can do (albeit with a slightly different management of Exceptions). The point of the `ILockingProvider` has always been to have an interface for locks that would allow a multi-server implementation without having to change the places where it's used. For example, in the fix we propose here for `TagService`, had we simply used `lock(string.Intern(lockString)) { }` we would have had a solution that only works for requests being handled in the same server.

Because of both points above, `ILockingProvider` is only suitable for those situations where `lock(obj) {}` would help, and those situations are those where we are using `IRepository`, because those are the calls where we are not immediately joining `ContentItemRecord` and `ContentItemVersionRecord`.

However, my fear as I write this is that there is still a chance those would mess up when we are doing calls over the `IRepository` while at the same time getting from the same records with a `DefaultContentQuery` somewhere else.

MatteoPiovanelli:
Also, my two cents on using a cache vs a sigleton vs ConcurrentDictionary:
I went with usign a cache because I see clearly how that would still work in a multi-server scenario. I would not know how to do the same with a Singleton or ConcurrentDictionary.

Skrypt:
If we use the repository for only retrieving records I think it's fine. Here the solution would be to just do calls on repository records to get the data we need to look against. The cache is fine, the ConcurrentDictionnary is just more optimized for concurrency performance but here in this scenario I don't think it's necessary.

But also I think @sebastienros is going to check if it's not an issue with the ContentManager itself. I personnally think it comes from NHibernate.

MatteoPiovanelli:
I am all for using IRepository wherever we can, as I also think that will help prevent concurrency issues.

I am not sure here for menus though, since the `GetMenus()` is looking for all the published ContentItems of type "Menu" and sorting them by TitlePart.title. Even adding a repository, e.g. one that contains the menu's title and Id, to not have breaking changes we would need to get all the ContentItems. On top of that, we are checking for the permissions on the items from `GetMenus()` with `IAuthorizationService.TryCheckAccess(Permission, IUser, IContent)`: this calls `Orchard.Core.Navigation.Security.AuthorizationEventHandler.Adjust(CheckAccessContext)` that creates the dynamic permissions based on the ContentItem's metadata. This means we actually need the whole ContentItem objects.

The `DefaultCacheManager` here is using a ConcurrentDictionary behind the scenes in the `DefaultCacheHolder`. What I meant earlier was that if we don't use that directly, but rather use an higher level abstraction such as going through the ICacheManager, I think it will be easier to write code that supports extending the solution to multiple servers.
I wish that the `Get(this ICacheManager, TKey, bool, Func<AcquireContext<TKey>, TResult>)` was not an extension method though, because then we would be able to inject the `ILockingService` to lock, rather than using `lock(key) { }`. Here's the code for that method from the current code base, with my added comments:
```
    public static class CacheManagerExtensions {
        public static TResult Get<TKey, TResult>(this ICacheManager cacheManager, TKey key, bool preventConcurrentCalls, Func<AcquireContext<TKey>, TResult> acquire) {
            if (preventConcurrentCalls) {
                // I think taking the lock like this only works if we assume that we are on a single server
                // instance, and the key passed is correctly a reference object, and is interned if a string.
                // Neither condition on the key is enforced by the code.
                lock(key) {
                    return cacheManager.Get(key, acquire);
                }
            }
            else {
                return cacheManager.Get(key, acquire);
            }
        }
    }
```

MatteoPiovanelli:
tl;dr: Subjecting AutoroutePart to the same tests I did in #8040 yields the same results: the deadlocks appear to be on LCK_M_S involving ContentItemVersionRecord, and specifically its primary key.

I am testing stuff regarding AutoroutePart like I did in #8040 for MenuPart.
On a fresh tenant, from branch 1.10.x. Page has CommonPart, TitlePart, AutoroutePart.

Full error log from an attempt to create two concurrent Page contentItems
```
2018-05-11 09:16:43,171 [12] Orchard.ContentManagement.DefaultContentManager - DevNoLaser - ERROR GenericADOException thrown from IContentHandler by Orchard.Autoroute.Handlers.AutoroutePartHandler
NHibernate.Exceptions.GenericADOException: could not execute query
[ SELECT this_.Id as Id11_2_, this_.Number as Number11_2_, this_.Published as Published11_2_, this_.Latest as Latest11_2_, this_.Data as Data11_2_, this_.ContentItemRecord_id as ContentI6_11_2_, contentite1_.Id as Id7_0_, contentite1_.Data as Data7_0_, contentite1_.ContentType_id as ContentT3_7_0_, contenttyp4_.Id as Id13_1_, contenttyp4_.Name as Name13_1_ FROM Orchard_Framework_ContentItemVersionRecord this_ inner join Orchard_Framework_ContentItemRecord contentite1_ on this_.ContentItemRecord_id=contentite1_.Id left outer join Orchard_Framework_ContentTypeRecord contenttyp4_ on contentite1_.ContentType_id=contenttyp4_.Id WHERE contentite1_.Id = @p0 and this_.Published = @p1 ]
  Name:cp0 - Value:13  Name:cp1 - Value:True
[SQL: SELECT this_.Id as Id11_2_, this_.Number as Number11_2_, this_.Published as Published11_2_, this_.Latest as Latest11_2_, this_.Data as Data11_2_, this_.ContentItemRecord_id as ContentI6_11_2_, contentite1_.Id as Id7_0_, contentite1_.Data as Data7_0_, contentite1_.ContentType_id as ContentT3_7_0_, contenttyp4_.Id as Id13_1_, contenttyp4_.Name as Name13_1_ FROM Orchard_Framework_ContentItemVersionRecord this_ inner join Orchard_Framework_ContentItemRecord contentite1_ on this_.ContentItemRecord_id=contentite1_.Id left outer join Orchard_Framework_ContentTypeRecord contenttyp4_ on contentite1_.ContentType_id=contenttyp4_.Id WHERE contentite1_.Id = @p0 and this_.Published = @p1] ---> System.Data.SqlClient.SqlException: Transaction (Process ID 59) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
   in System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   in System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   in System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   in System.Data.SqlClient.SqlDataReader.TryHasMoreRows(Boolean& moreRows)
   in System.Data.SqlClient.SqlDataReader.TryReadInternal(Boolean setTimeout, Boolean& more)
   in System.Data.SqlClient.SqlDataReader.Read()
   in NHibernate.Driver.NHybridDataReader.Read()
   in NHibernate.Loader.Loader.DoQuery(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies, IResultTransformer forcedResultTransformer)
   in NHibernate.Loader.Loader.DoQueryAndInitializeNonLazyCollections(ISessionImplementor session, QueryParameters queryParameters, Boolean returnProxies, IResultTransformer forcedResultTransformer)
   in NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
   --- Fine della traccia dello stack dell'eccezione interna ---
   in NHibernate.Loader.Loader.DoList(ISessionImplementor session, QueryParameters queryParameters, IResultTransformer forcedResultTransformer)
   in NHibernate.Loader.Loader.ListIgnoreQueryCache(ISessionImplementor session, QueryParameters queryParameters)
   in NHibernate.Impl.SessionImpl.List(CriteriaImpl criteria, IList results)
   in NHibernate.Impl.CriteriaImpl.List(IList results)
   in NHibernate.Impl.CriteriaImpl.List[T]()
   in Orchard.ContentManagement.DefaultContentManager.GetManyImplementation(QueryHints hints, Action`2 predicate) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 370
   in Orchard.ContentManagement.DefaultContentManager.Get(Int32 id, VersionOptions options, QueryHints hints) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 168
   in Orchard.ContentManagement.DefaultContentManager.Get(Int32 id, VersionOptions options) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 131
   in Orchard.ContentManagement.DefaultContentManager.Get(Int32 id) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 127
   in Orchard.Autoroute.Services.AutorouteService.GenerateAlias(AutoroutePart part)
   in Orchard.Autoroute.Handlers.AutoroutePartHandler.ProcessAlias(AutoroutePart part)
   in Orchard.Autoroute.Handlers.AutoroutePartHandler.CreateAlias(AutoroutePart part)
   in Orchard.Autoroute.Handlers.AutoroutePartHandler.<.ctor>b__7_0(UpdateContentContext ctx, AutoroutePart part)
   in Orchard.ContentManagement.Handlers.ContentHandler.InlineStorageFilter`1.Updated(UpdateContentContext context, TPart instance) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\Handlers\ContentHandler.cs:riga 206
   in Orchard.ContentManagement.Handlers.StorageFilterBase`1.Orchard.ContentManagement.Handlers.IContentStorageFilter.Updated(UpdateContentContext context) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\Handlers\StorageFilterBase.cs:riga 78
   in Orchard.ContentManagement.Handlers.ContentHandler.Orchard.ContentManagement.Handlers.IContentHandler.Updated(UpdateContentContext context) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\Handlers\ContentHandler.cs:riga 361
   in Orchard.ContentManagement.DefaultContentManager.<>c__DisplayClass52_0.<UpdateEditor>b__1(IContentHandler handler) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 708
   in Orchard.InvokeExtensions.Invoke[TEvents](IEnumerable`1 events, Action`1 dispatch, ILogger logger) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\InvokeExtensions.cs:riga 17
2018-05-11 09:16:43,474 [12] NHibernate.Event.Default.AbstractFlushingEventListener - DevNoLaser - ERROR Could not synchronize database state with session
NHibernate.StaleStateException: Batch update returned unexpected row count from update; actual row count: 0; expected: 1
   in NHibernate.AdoNet.Expectations.VerifyOutcomeBatched(Int32 expectedRowCount, Int32 rowCount)
   in NHibernate.AdoNet.SqlClientBatchingBatcher.DoExecuteBatch(IDbCommand ps)
   in NHibernate.AdoNet.AbstractBatcher.ExecuteBatchWithTiming(IDbCommand ps)
   in NHibernate.AdoNet.AbstractBatcher.ExecuteBatch()
   in NHibernate.AdoNet.AbstractBatcher.PrepareBatchCommand(CommandType type, SqlString sql, SqlType[] parameterTypes)
   in NHibernate.Persister.Entity.AbstractEntityPersister.Update(Object id, Object[] fields, Object[] oldFields, Object rowId, Boolean[] includeProperty, Int32 j, Object oldVersion, Object obj, SqlCommandInfo sql, ISessionImplementor session)
   in NHibernate.Persister.Entity.AbstractEntityPersister.UpdateOrInsert(Object id, Object[] fields, Object[] oldFields, Object rowId, Boolean[] includeProperty, Int32 j, Object oldVersion, Object obj, SqlCommandInfo sql, ISessionImplementor session)
   in NHibernate.Persister.Entity.AbstractEntityPersister.Update(Object id, Object[] fields, Int32[] dirtyFields, Boolean hasDirtyCollection, Object[] oldFields, Object oldVersion, Object obj, Object rowId, ISessionImplementor session)
   in NHibernate.Action.EntityUpdateAction.Execute()
   in NHibernate.Engine.ActionQueue.Execute(IExecutable executable)
   in NHibernate.Engine.ActionQueue.ExecuteActions(IList list)
   in NHibernate.Engine.ActionQueue.ExecuteActions()
   in NHibernate.Event.Default.AbstractFlushingEventListener.PerformExecutions(IEventSource session)
2018-05-11 09:16:43,490 [12] Orchard.ContentManagement.DefaultContentManager - DevNoLaser - ERROR StaleStateException thrown from IContentHandler by Orchard.Autoroute.Handlers.AutoroutePartHandler
NHibernate.StaleStateException: Batch update returned unexpected row count from update; actual row count: 0; expected: 1
   in NHibernate.AdoNet.Expectations.VerifyOutcomeBatched(Int32 expectedRowCount, Int32 rowCount)
   in NHibernate.AdoNet.SqlClientBatchingBatcher.DoExecuteBatch(IDbCommand ps)
   in NHibernate.AdoNet.AbstractBatcher.ExecuteBatchWithTiming(IDbCommand ps)
   in NHibernate.AdoNet.AbstractBatcher.ExecuteBatch()
   in NHibernate.AdoNet.AbstractBatcher.PrepareBatchCommand(CommandType type, SqlString sql, SqlType[] parameterTypes)
   in NHibernate.Persister.Entity.AbstractEntityPersister.Update(Object id, Object[] fields, Object[] oldFields, Object rowId, Boolean[] includeProperty, Int32 j, Object oldVersion, Object obj, SqlCommandInfo sql, ISessionImplementor session)
   in NHibernate.Persister.Entity.AbstractEntityPersister.UpdateOrInsert(Object id, Object[] fields, Object[] oldFields, Object rowId, Boolean[] includeProperty, Int32 j, Object oldVersion, Object obj, SqlCommandInfo sql, ISessionImplementor session)
   in NHibernate.Persister.Entity.AbstractEntityPersister.Update(Object id, Object[] fields, Int32[] dirtyFields, Boolean hasDirtyCollection, Object[] oldFields, Object oldVersion, Object obj, Object rowId, ISessionImplementor session)
   in NHibernate.Action.EntityUpdateAction.Execute()
   in NHibernate.Engine.ActionQueue.Execute(IExecutable executable)
   in NHibernate.Engine.ActionQueue.ExecuteActions(IList list)
   in NHibernate.Engine.ActionQueue.ExecuteActions()
   in NHibernate.Event.Default.AbstractFlushingEventListener.PerformExecutions(IEventSource session)
   in NHibernate.Event.Default.DefaultAutoFlushEventListener.OnAutoFlush(AutoFlushEvent event)
   in NHibernate.Impl.SessionImpl.AutoFlushIfRequired(ISet`1 querySpaces)
   in NHibernate.Impl.SessionImpl.List(CriteriaImpl criteria, IList results)
   in NHibernate.Impl.CriteriaImpl.List(IList results)
   in NHibernate.Impl.CriteriaImpl.List[T]()
   in Orchard.ContentManagement.DefaultContentManager.GetManyImplementation(QueryHints hints, Action`2 predicate) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 370
   in Orchard.ContentManagement.DefaultContentManager.Get(Int32 id, VersionOptions options, QueryHints hints) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 168
   in Orchard.ContentManagement.DefaultContentManager.Get(Int32 id, VersionOptions options) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 131
   in Orchard.ContentManagement.DefaultContentManager.Get(Int32 id) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 127
   in Orchard.Autoroute.Services.AutorouteService.GenerateAlias(AutoroutePart part)
   in Orchard.Autoroute.Handlers.AutoroutePartHandler.ProcessAlias(AutoroutePart part)
   in Orchard.Autoroute.Handlers.AutoroutePartHandler.PublishAlias(AutoroutePart part)
   in Orchard.Autoroute.Handlers.AutoroutePartHandler.<.ctor>b__7_2(PublishContentContext ctx, AutoroutePart part)
   in Orchard.ContentManagement.Handlers.ContentHandler.InlineStorageFilter`1.Published(PublishContentContext context, TPart instance) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\Handlers\ContentHandler.cs:riga 218
   in Orchard.ContentManagement.Handlers.StorageFilterBase`1.Orchard.ContentManagement.Handlers.IContentStorageFilter.Published(PublishContentContext context) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\Handlers\StorageFilterBase.cs:riga 98
   in Orchard.ContentManagement.Handlers.ContentHandler.Orchard.ContentManagement.Handlers.IContentHandler.Published(PublishContentContext context) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\Handlers\ContentHandler.cs:riga 385
   in Orchard.ContentManagement.DefaultContentManager.<>c__DisplayClass35_0.<Publish>b__2(IContentHandler handler) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 397
   in Orchard.InvokeExtensions.Invoke[TEvents](IEnumerable`1 events, Action`1 dispatch, ILogger logger) in C:\Sviluppo\Orchard.Sources.Github\Orchard\src\Orchard\InvokeExtensions.cs:riga 17

```

Just the query (@p0 = 13; @p1= True):
```
SELECT 
	this_.Id as Id11_2_, 
	this_.Number as Number11_2_, 
	this_.Published as Published11_2_, 
	this_.Latest as Latest11_2_, 
	this_.Data as Data11_2_, 
	this_.ContentItemRecord_id as ContentI6_11_2_, 
	contentite1_.Id as Id7_0_, 
	contentite1_.Data as Data7_0_, 
	contentite1_.ContentType_id as ContentT3_7_0_, 
	contenttyp4_.Id as Id13_1_, 
	contenttyp4_.Name as Name13_1_ 
FROM 
	Orchard_Framework_ContentItemVersionRecord this_ 
	inner join Orchard_Framework_ContentItemRecord contentite1_ on this_.ContentItemRecord_id=contentite1_.Id 
	left outer join Orchard_Framework_ContentTypeRecord contenttyp4_ on contentite1_.ContentType_id=contenttyp4_.Id 
WHERE 
	contentite1_.Id = @p0 
	and this_.Published = @p1
```
This is looking for the item with Id == 13, that does not exist in the db after the processes fail, meaning it has likely been "removed" when the transaction rolled back (hence the other two errors in the log, where nhibernate expected to find 1 item but found none).

Using that query on SQLQueryStress (with id = 12, corresponding to a content that exists) and launching the stored procedure sp_BlitzFirst I see no issue being reported.
Then I tried launching the stored procedure while repeatedly launching the ParallelTests.js script. Just like in #8040, this shows long waits on LCK_M_S.

Launching sp_BlitzLock i see that all the deadlocks for the creations here involved only ContentItemVersionRecord and its primary key.

MatteoPiovanelli:
I think there are two classes of issues we are dealing with here:

 1. The shared locks in the db messing things up (as found with the stored procedures from the references @sebastienros  discovered). These issues are likely specific to MS SQL Server. As discussed in #8040, I think that a possible solution to that could be using hilo identifiers, to prevent locks on the whole tables. Hopefully, we wouldn't end up having locks on the table used by the hilo algorithm, because we would not be creating new rows there, except when new tables are added.

 2. Concurrency issues in the code, like what we have in the live branches (1.10.x or dev) for TagPart and AutoroutePart. These issues are at a higher level, and require code changes in the specific drivers, handlers, or services involved, for example with code locks or caches.

I think that an example case for the first issue is what we see for MenuPart; the simplest example of the second issue is TagPart, and another one is the body of the GetContentTypeRecordId method. Of course, sometimes both issues affect a single thing, like it would appear to be the case for AutoroutePart.

I think that issues of the second kind can sometimes be isolated and tackled independently on solutions to the first, because the first kind of issue should not exist when using a different DB, e.g. SQLCE.

MatteoPiovanelli:
Note that on a fresh tenant using SQL CE, from 1.10.x, the only concurrency issue I see creating pages (without changing the definition of the ContentType) is the one for LayoutPart that likely has to do with GetContentTypeRecordId.

MatteoPiovanelli:
ref #8052 since it seems to fix some concurrency issues

src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -44,69 +53,109 @@ public class AutoroutePartHandler : ContentHandler {
...
+                    ProcessAlias(part);
+
+                    // Should it become the home page?
+                    if (part.PromoteToHomePage)
+                    {
+                        // Get the current homepage an unmark it as the homepage.
+                        var currentHomePageId = _homeAliasService.GetHomePageId();
+                        if (currentHomePageId != part.Id)
+                        {
+                            var autoroutePart = part.As<AutoroutePart>();
```
MatteoPiovanelli, line 94: We still need to fix this, because `autoroutePart` here should be the pre-existing home page, but the code as it is in this PR sets it to the same part that we are trying to update and set as homepage.

MatteoPiovanelli:
I'll create a branch:
 - Based on dev (I'd rather use 1.10.x)
 - Cherry pick #8052 as a base
 - pick from here the changes that are not related to Menus: ILockingProvider and tags
 - [at that stage here autoroute should still not work]
 - pick the changes from here related to autoroute, to find the working solution

MatteoPiovanelli:
The NOLOCK stuff form #8052 MAY mean we do not need to go through alias stuff, but can go back to querying "normally". I'll check that as I create the new branch with all the fixes.

PR #8028, opened 2018-04-19
AutoroutePartHandler using ILockingProvider

MatteoPiovanelli:
AutoroutePartHandler using ILockingProvider to reduce the chance of deadlock, while keeping DisplayAlias values unique for published content.
I figured I could do a PR to the branch the other PR is from, rather than just create a Gist for you to copy from.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -1,10 +1,11 @@
-﻿using System;
+using System;
 using Orchard.Autoroute.Models;
 using Orchard.Autoroute.Services;
 using Orchard.ContentManagement;
 using Orchard.ContentManagement.Handlers;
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -54,6 +58,15 @@ public class AutoroutePartHandler : ContentHandler {
             ProcessAlias(part);
         }
 
+        private string LockString {
+            get {
+                return string.Join(".",
+                    _orchardServices.WorkContext?.CurrentSite?.BaseUrl ?? "",
+                    _orchardServices.WorkContext?.CurrentSite?.SiteName ?? "",
+                    "AutoroutePartHandler");
```
Skrypt, line 66: Locking per Url and SiteName ... good idea 👍 
I started looking at async await methods but this does the trick 😄

MatteoPiovanelli:
Closing this because of work on #8025

Issue #7885, opened 2017-10-25
Tenant reset does not delete all tables


MarcoViglione-Laser:
If I reset a tenant through the multitenancy page it only deletes the database tables created by the currently enabled modules.
If I enabled a module and then disabled it, the tables it created will not be deleted by a reset and will remain inside the database.
This can be a problem, because if I try to enable again those modules after doing another setup of the resetted tenant I receive an error, caused by the fact that the migrations try to create again some tables that already exist.

sebastienros:
It's new to me that table are deleted. I thought it was only handled by each module if they implement `Uninstall()`

mdameer:
@sebastienros, I checked the current implementation and found that it trying to get the table names to be dropped through NHibernate 'ClassMappings' object, which will contains only mappings for enabled modules record classes, so we need to find another way to bring all table names with support for all database engines which supported by Orchard.

sebastienros:
Or implement the Uninstsall methods so that when a module is uninstalled it correctly removes the tables and indices it created. Then the remaining tables would be all registered already, and the remaining module could just also delete their own tables.

HermesSbicego-Laser:
I think there is a difference between uninstalling a module and resetting a tenant. If I uninstall a module I don't want to drop the tables too, beacause **I will loose all datas** of the uninstalled module and **of the modules depending from it too**!!!; So I think it's a bad idea to drop tables implementing the Uninstall method. 
Different are the needs when I want to reset a tenant, because I want to drop all the tables, for both installed and uninstalled modules.
@sebastienros @mdameer what do you think?

sebastienros:
But we'd need to remember that a module was once installed to be able to delete the tables. But I could agree with you, and then only on reset we would look at the migrations and detect which ones need to be uninstalled, not using the enabled modules, but actually what migrations were executed. Doable.

mdameer:
@HermesSbicego-Laser, I'm totally agree with you, so I suggest to find a solution to load the database metadata, then dropping all tables will be very simple,  I'm trying to find a way to implement this solution, but the main problem is that we are dealing with different database types.

@sebastienros, regarding the uninstalling mechanism, I think it should be a user choice to delete or keep the tables which was created by the module, what do you think?

sfmskywalker:
[...]

PR #8031, opened 2018-04-21
Query database schema instead of Nhibernate configuration to get tenant tables list

mdameer:
Issue #7885


sebastienros:
Shouldn't it use table prefix somewhere?

mdameer:
@sebastienros, this code will read table names from database schema, so the prefix will be included with table names like the previous implementation, also the following line will remove the prefix before run the drop command through SchemaBuilder: `schemaBuilder.DropTable(schemaBuilder.RemoveDataTablePrefix(tableName));`

sebastienros:
This code reads all the tables, a database can contain tables for many tenants, partitioned with a prefix, lie `Site1_ContentItemRecord`, `Site2_ContentItemRecord`, ... 
Then it removes as many chars as the prefix has, so it can even create wrong table names, and if you are deleting a tenant without prefix, it will delete all the tables, even if they have prefixes.

sebastienros:
So this code only works if you have a single tenant in the db.

PR #8038, opened 2018-04-27
Update ContentPickerFieldDriver.cs

### LINKED ISSUE #8037: Query could not filter content items based on contentpicker field
1. Append a contentpicker field file to a contenttype, set the 
2. Create a contentitem of this type and assigne a content to that contentpicker field. let's say the id of content assigned to contentpicker is 10
3. Create a query and use the contentpicker field filter as the query filters. And set the filter value is 10
4. Preview this query. The query returned no results. It should return the content item which of contentpicker field's ids value is 10

HengzheLi:
try to fix #8037

Not sure if that is the right place to fix this issue. But when I apply my update. The query works


HengzheLi:
my mistake. close it

PR #8041, opened 2018-05-02
[8039]Update QueryElementDriver.cs

### LINKED ISSUE #8039: The selected query in QueryElement is lost when import from an exported recipe
1. Create a page with LayoutPart
2. Drop a QueryElement into this layout
3. Select a query
4. Export this page into a recipe
5. Import the exported recipe into another tenant.
The query of QueryElement selected is lost

HengzheLi:
fixes #8039
Works in my side. 
Hopes it does not involve any side effect

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/QueryElementDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/QueryElementDriver.cs
@@ -174,5 +174,28 @@ namespace Orchard.DynamicForms.Drivers {
             var runtimeValue = element.RuntimeValue;
             return runtimeValue != null ? runtimeValue.Split(new[] { ',', ';' }, StringSplitOptions.RemoveEmptyEntries) : Enumerable.Empty<string>();
         }
+        
+        protected override void OnExporting(Query element, ExportElementContext context) {
+            base.OnExporting(element, context);
+            if (element.QueryId == null)
+                return;
+            var queryIdentityPart = _contentManager.Get<IdentityPart>(element.QueryId.Value);
+            context.Element.Data.Add("QueryElementIdentity", queryIdentityPart.Identifier);
+        }
+
+        protected override void OnImported(Query element, ImportElementContext context) {
+            base.OnImported(element, context);
+            if (!context.Element.Data.Keys.Contains("QueryElementIdentity"))
+                return;
+            var queryIdentifier = context.Element.Data["QueryElementIdentity"];
+            var queryIdentityPart = _contentManager.Query<IdentityPart>()
+                .Where<IdentityPartRecord>(record => record.Identifier == queryIdentifier).List().FirstOrDefault();
+            if (queryIdentityPart == null)
+                return;
+            element.QueryId = queryIdentityPart.ContentItem.Id;
+            if (!context.Element.Data.Keys.Contains("QueryId"))
+                return;
+            context.Element.Data["QueryId"] = element.QueryId.ToString();
+        }
     }
 }
--- a/src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/QueryElementDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/QueryElementDriver.cs
@@ -180,7 +180,7 @@ namespace Orchard.DynamicForms.Drivers {
             if (element.QueryId == null)
                 return;
             var queryIdentityPart = _contentManager.Get<IdentityPart>(element.QueryId.Value);
-            context.Element.Data.Add("QueryElementIdentity", queryIdentityPart.Identifier);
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/QueryElementDriver.cs
```diff
@@ -174,5 +174,28 @@ public QueryElementDriver(IFormsBasedElementServices formsServices, IProjectionM
             var runtimeValue = element.RuntimeValue;
             return runtimeValue != null ? runtimeValue.Split(new[] { ',', ';' }, StringSplitOptions.RemoveEmptyEntries) : Enumerable.Empty<string>();
         }
+        
+        protected override void OnExporting(Query element, ExportElementContext context) {
+            base.OnExporting(element, context);
+            if (element.QueryId == null)
+                return;
+            var queryIdentityPart = _contentManager.Get<IdentityPart>(element.QueryId.Value);
```
sfmskywalker: You should rely on `_contentManager.GetItemMetadata().Identity` instead - there are potentially other handlers that contribute to a content's identity.

Please check the import/export implementation of the `ProjectionElementDriver`, which solves the same issue that you're solving here:

https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.Layouts/Drivers/ProjectionElementDriver.cs

Thanks!

HengzheLi: Thanks for the review. Updated the file.

sfmskywalker: The following line:

`var queryIdentityPart = _contentManager.Get<IdentityPart>(element.QueryId.Value);`

Although it will work if `QueryPart`'s content type also includes the `IdentityPart`, the more robust way would be to not rely on that part to be present (since technically, it shouldn't rely on IdentityPart for its identity - any module could provide its identity). So how about do this instead:

`var queryPart = _contentManager.Get<QueryPart>(element.QueryId.Value);`

sfmskywalker: I think it looks good otherwise.

HengzheLi: Yes, you are right. Applied it.

sfmskywalker (approved):

src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/QueryElementDriver.cs
```diff
@@ -174,5 +174,29 @@ public QueryElementDriver(IFormsBasedElementServices formsServices, IProjectionM
...
+                return;
+            context.Element.Data.Add("QueryElementIdentity", _contentManager.GetItemMetadata(queryContentItem).Identity.ToString());
+        }
+
+        protected override void OnImported(Query element, ImportElementContext context) {
+            base.OnImported(element, context);
+            if (!context.Element.Data.Keys.Contains("QueryElementIdentity"))
+                return;
+            var queryIdentifier = context.Element.Data["QueryElementIdentity"];
+            var queryContentItem = context.Session.GetItemFromSession(queryIdentity).Content;
```
MatteoPiovanelli, line 193: Looks like queryIdentity is not defined.

src/Orchard.Web/Modules/Orchard.DynamicForms/Drivers/QueryElementDriver.cs
```diff
@@ -174,5 +174,29 @@ public QueryElementDriver(IFormsBasedElementServices formsServices, IProjectionM
...
+            if (queryIdentityPart == null)
```
MatteoPiovanelli, line 194: queryIdentityPart is not defined either

HengzheLi: arrh, it's my fault. Just edit file in Github editor. I will fix it.

And I think if I should override the OnImportCompleted method. It seems that it is more fit for this scenario

sfmskywalker: @li0803 will you provide a PR?

HengzheLi: Just fixed this. But seems there is conflict and I need to merge dev to my branch and there are huge changes to be merged as my original branch is based on 1.10.2.  :(
any sugesstion

HengzheLi: submit a PR #8057

Issue #8040, opened 2018-05-01
[Concurrency] Menu query doesn't handle concurrency


sebastienros:
```
_contentManager.Query().ForType("Menu").Join<TitlePartRecord>().OrderBy(x => x.Title).List())
```

In MainMenuService.

sebastienros:
For reference while trying to understand why this query is locking:
https://www.brentozar.com/archive/2018/04/how-entity-framework-and-nhibernate-can-cause-poison-resource_semaphore-waits/
http://sqlapproach.blogspot.com/2012/08/active-processes.html

MatteoPiovanelli:
tl;dr: The issue we have seems to not be the one indicated in the links.

I am trying those tests.

I am using the 1.10.x branch, and a fresh tenant. I try to create contents of type Page. The definition of this ContentType has been changed, so it only contains: CommonPart, TitlePart, MenuPart. I did that because we know there is a concurrency issue with the MenuPart, and I did not want to have other stuff making things even more complex.

Concurrent creation of two items fails, as we knew it would. Here's the full error from the logs:

```
2018-05-10 09:06:01,454 [38] Orchard.ContentManagement.Drivers.Coordinators.ContentPartDriverCoordinator - DevNoLaser - ERROR GenericADOException thrown from IContentPartDriver by Orchard.Core.Navigation.Drivers.MenuPartDriver
NHibernate.Exceptions.GenericADOException: could not execute query
[...]

PR #8052, opened 2018-05-11
Added interceptor to add "WITH (NOLOCK)" to specific tables

MatteoPiovanelli:
See discussion in #8040

This seems to fix the concurrency issues on the query for GetMenus().
Might fix something else too, but I don't know what kind of issues it may introduce.

I based this off 1.10.x, out of habit, but it's trivial to cherry-pick the commit for dev.

### DIFF
--- a/src/Orchard/Data/NoLockInterceptor.cs
+++ b/src/Orchard/Data/NoLockInterceptor.cs
@@ -17,7 +17,7 @@ namespace Orchard.Data {
 
             _shellSettings = shellSettings;
             // allow injecting through autofac config.
-            AllTableNames = "Orchard_Framework_ContentItemVersionRecord, Orchard_Framework_ContentItemRecord, Title_TitlePartRecord";
+            AllTableNames = "Orchard_Framework_ContentItemVersionRecord, Title_TitlePartRecord";
             // TODO: add providers that would inject tablenames and move the autofac injection to one of them
         }
 
--- a/src/Orchard/Data/NoLockInterceptor.cs
+++ b/src/Orchard/Data/NoLockInterceptor.cs
@@ -5,11 +5,17 @@ using System.Text;
 using System.Threading.Tasks;
 using NHibernate;
 using NHibernate.SqlCommand;
+using Orchard.Environment.Configuration;
 
 namespace Orchard.Data {
     public class NoLockInterceptor : EmptyInterceptor, ISessionInterceptor {
 
-        public NoLockInterceptor() {
+        private readonly ShellSettings _shellSettings;
+
+        public NoLockInterceptor(
+            ShellSettings shellSettings) {
+
+            _shellSettings = shellSettings;
             // allow injecting through autofac config.
             AllTableNames = "Orchard_Framework_ContentItemVersionRecord, Orchard_Framework_ContentItemRecord, Title_TitlePartRecord";
             // TODO: add providers that would inject tablenames and move the autofac injection to one of them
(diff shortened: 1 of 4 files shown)


MatteoPiovanelli:
I am quite sure this is missing handling table prefixes

MatteoPiovanelli:
Note that adding `Orchard_Autoroute_AutoroutePartRecord` to the list of tables breaks the functionality of alias/autoroute, just like using ReadUncommitted: even though there is no lock, we end up having duplicate aliases.

MatteoPiovanelli:
I am thinking that maybe this NOLOCK should be added to every ContentPartVersionRecord, because on principle those should never be updated, but rather a new one will be created each time an item is updated.. However AutoroutePartRecord is one, and I already saw that doing this breaks that functionality. Perhaps we should still do it, and handle that in its own code.

src/Orchard/Data/NoLockInterceptor.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            _shellSettings = shellSettings;
+            // allow injecting through autofac config.
+            AllTableNames = "Orchard_Framework_ContentItemVersionRecord, Title_TitlePartRecord";
+            // TODO: add providers that would inject tablenames and move the autofac injection to one of them
+        }
+
+        public List<string> TableNames {
+            get {
+                return AllTableNames
```
sebastienros: The split can be done once.

sebastienros: And we need it to be extensible from other modules.

Skrypt:
The issue with the AutoroutePart table is happening with the PR changes I made or without them?

src/Orchard/Data/NoLockInterceptor.cs
MatteoPiovanelli: I agree with both comments. This PR was to "test the waters" and see if anyone had any issue with using "NOLOCK" like that.

MatteoPiovanelli:
@Skrypt the issue here was happening without any changes from that other PR. I wanted first to see what would be fixed by using this interceptor, since in testing things for menupart I saw the deadlocks were happening on the shared locks taken on the primary keys of ContentItemVersionRecord and TitlePartRecord. If we agree this way to hanlde some of the issues is fine and does not introduce any weird regression, I'll add here the ILockingProvider and everything else we need from #8025

MatteoPiovanelli:
My latest changes make the list of table names extensible

MatteoPiovanelli:
tl;dr: I think that a combination of this PR and #8025 could fix most of the concurrency issues we have been seeing.

This PR does not fix every concurrency issue we have seen, but I think it may solve a class of those.
Specifically:

| *Part under test* | Result |
| - | - |
| MenuPart | The issues we saw when creating these concurrently is fixed with this PR, because it removes the lock contention we were having on ContentItemVersionRecord and TitlePartRecord. |
| TagsPart | The code in this PR does not fix the fact that we may create duplicate tags. This is easily fixed with a lock (ILockingProvider) as we did in another PR. |
| AutoroutePart | I saw deadlocks on the shared lock on the primary key for AutoroutePartRecord. Maybe that table should be added here (as I mentioned earlier, it may be that we should add all ContentPartVersionRecords), but adding it lead to duplicate aliases, so I need to try this together with the stuff from the PR @skrypt did. |
| LayoutPart | GetContentTypeRecordId is still unable to handle concurrency as it is. I think this could easily be fixed with a lock (ILockingProvider) |


I got the results discussed above by attempting to concurrently create Page contentItems, where the definition for the ContentType had TitlePart, CommonPart, and the part under test.

PR #8054, opened 2018-05-15
Concurrency fixes

MatteoPiovanelli:
@Skrypt @sebastienros 
Just to summarize things as I put together all our fixes:

 - With the NOLOCK, without changing anything in the Autouroute Module, we see deadlocks.
 - Adding an `INoLockTableProvider` that adds "Orchard_Autoroute_AutoroutePartRecord" I see deadlocks on "Orchard_Framework_ContentItemRecord". These generally require more than 2 concurrent creations to pop up (usually 3 are enough, so it's not a massive improvement)
 - Adding "Orchard_Framework_ContentItemRecord" to the tables (in `DefaultNoLockTableProvider`), I see no more deadlocks (I tried up to 20 concurrent creations). However there are duplicate aliases, so it's time to pull in the solutions @skrypt has been developing here with the LockingProvider.
 - I inserted the LockingProvider in `AutoroutePartHandler` as was initially done in #8025 . That seems to fix the issue with duplicate aliases.

To find what the deadlocks are on, I use the sp_BlitzLock procedure from the links @sebastienros gave us.
All the tests were done with Page contents, on a fresh tenant, after removing LayoutPart.

This is a synthesis of the stuff from the following:
#7720 
#8019 
#7708
#8025 
#8028 
#8040 
#8052 
(I am sure I am forgetting something)

As it is, by my tests, it fixes the concurrency issues we were seeing on MenuPart, TagsPart, AutoroutePart.
It does not fix the issue for LayoutPart: that is likely to be in `DefaultContentQuery.GetContentTypeRecordId` and I did not want to introduce an `ILockingProvider` there yet.

### DIFF
--- a/src/Orchard.Tests.Modules/Tags/Services/TagsServiceTests.cs
+++ b/src/Orchard.Tests.Modules/Tags/Services/TagsServiceTests.cs
@@ -11,6 +11,7 @@ using Orchard.ContentManagement.Handlers;
 using Orchard.ContentManagement.Records;
 using Orchard.Data;
 using Orchard.Environment;
+using Orchard.Locking;
 using Orchard.Security;
 using Orchard.Tags.Handlers;
 using Orchard.Tags.Models;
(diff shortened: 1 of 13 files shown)


Skrypt:
So this does not include using only RouteCollections for Aliases. I'll redo that PR separately as a perf improvement. If the issue is only getting the current home page record I think I can find a solution.

MatteoPiovanelli:
@skrypt exactly. With the combination of the other fixes, we didn't need to go through the alias stuff for concurrency. As you said, it will be a performance improvement, but I chose to keep it separate so we would not mix more stuff than strictly required for the fix.

PR #8061, opened 2018-05-23
Update AdminController.cs

kbreliver:
fix bugs: remove elements outside the “foreach” snippet


sebastienros:
Can you explain what issue it's fixing?

Issue #8070, opened 2018-06-17
DELETE method is not implemented in WebRequestActivity


HengzheLi:
There is DELETE option in WebRequestActivity edit form. But the DELETE method is not implemented in activity's Execute method.

PR #8071, opened 2018-06-17
[#8070]Update WebRequestActivity.cs

HengzheLi:
Add http DELETE method implement in WebRequestActivity


sfmskywalker:
Looks good to me!

Issue #8030, opened 2018-04-19
Non-creatable contents are actually still creatable


HermesSbicego-Laser:
Despite their setting (creatable is set to false), non-creatable contents are wrongly still creatable.
From what I can see, I found 4 cases:
1. Accessing directly the `Create Action`
2. Accessing directly the `Clone Action`
3. Filtering in the `SummaryAdmin` for a non-creatable content the create button still permit to create that `ContentType`
4. In the SummaryAdmin the `Clone ActionLInk` is there and permit to clone the content.

I think this is an inconrrect behaviour: a non-creatable content should neither be creatable nor clonable.
The only doubt is for translations: should we permit to translate a non-creatable content? I think yes, because it is a translation not a real new creation...

sebastienros:
This is just to hide it from the New menu section.

HermesSbicego-Laser:
Yes, you are right, by now is the only effect and that is the point. But my suggestion was to have a coherent option (I would like not a different one, since a setting exists) even if more restrictive.

sebastienros:
You can use permissions, as they are also checked before rendering the links to create items. We do that with content permission. Have you tried to remove the permission from a role, that should work?

HermesSbicego-Laser:
That was an option we considered but there is no a "create content" permission.
I see three ways to achieve the feature.
1. To use the "creatable" settings in order to hide create button in contents list and hide clone link in summaryadmin, as well in the "new" menu. We can keep routes actives (as is, without any check against the setting) for backward compliance.
2. Add a new permission for each securable content named "create new content". And check over that permission where we need it.
3. Both.

I think these can be 2 useful features in some business context...

We can start working suddenly on point 1, if you agree. Then point 2, later.

sebastienros:
The effect of Creatable should not be changed.

Instead you should ensure that the ways to create items, as states in your first comment, should be secured by ensure the permissions are correctly enforced.

HermesSbicego-Laser:
Although I cannot imagine why the effect of Creatable should not be changed, we will work on a solution providing a new permission `Create new content`: we will provide that permission for all securable contents, as per the `edit content` permission. If a role has the `Create new content` permission will also have the `edit own` permission too, otherwise after the first save the user will see an Unauthorized exception.

sebastienros:
Ok for a new Create permission

PR #8099, opened 2018-08-10
fix: only creatable content types can be created

csutorasr:
Before this change if I set a content type to uncreatable you could create it with the link.

So e.g. in Page content type creatable checkbox is unchecked, but you could create with link: `/Admin/Contents/Create/Page`.

### DIFF
--- a/src/Orchard.Web/Core/Contents/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Core/Contents/Controllers/AdminController.cs
@@ -242,6 +242,9 @@ namespace Orchard.Core.Contents.Controllers {
             if (string.IsNullOrEmpty(id))
                 return CreatableTypeList(containerId);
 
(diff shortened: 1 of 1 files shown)


HermesSbicego-Laser:
Hi guys sorry if I join in the comments but some time.ago I did some stuff around the "Content Creation" with this PR #8048 merged in 1.10.x (don't know why this is not on dev).
During the discussion #8030 was decided to not change the behaviour of "Creatable" (https://github.com/OrchardCMS/Orchard/issues/8030#issuecomment-385022166) but to create a permission for that. Perhaps that PR could be merged in dev to acheive what you have in mind?

sebastienros:
I think we need to merge 1.10.x to dev

BenedekFarkas:
I'm closing this PR, because it's not a bug, nor a security issue. The hint of the Creatable setting was a bit misleading though, so I updated that (cf821e1fdcb7e9e5ed0d0d73a88576538ca1e133), otherwise there are permissions (CreateContent) that the controllers will check to determine whether the user can create items of that type or not.

Issue #8100, opened 2018-08-22
Removing wrong aliases


d-lan1:
When content items are created and an autoroute part is present, it allows you to add an alias url. 

When you publish the content item the autoroute part is created is saved in the table Orchard_Autoroute_AutoroutePartRecord with the alias url being saved in the  table Orchard_Alias_AliasRecord - this table only contains aliases of published content items. This is because when an item is unpublished, removed or destroyed the alias gets removed. 

When a new content item version is created a new autoroute part is created and added to the relevant table. The alias however is deleted from the its table and the new alias takes its place even if it had the same value as there is no versioning in this table. 

If you add a brand new content item it will allow you to use the same url aliases that were used in unpublished (older) versions of other content items. 

My issue appears when running code such as:

var items = _contentManager.Query(VersionOptions.AllVersions, "item").List();
             foreach (var i in items)
            {
                _contentManager.Remove(i);
            }

The Remove() calls the remove handler on each content part within the content item. 

For the autoroute part this calls RemoveAlias() which deletes any record in the alias table with a url matching that of the autoroute, regardless of whether the content item on which the remove() was called is published or not.

This is problematic as another page may now use an alias with the same url which would get removed. 

This causes 404s on published pages which consequentially need republishing.

The solution is to only attempt to remove an alias if the content item being removed autoroute is published.

PR #8101, opened 2018-08-22
[8100] Removing wrong aliases

d-lan1:
See issue  https://github.com/OrchardCMS/Orchard/issues/8100

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -43,9 +43,9 @@ namespace Orchard.Autoroute.Handlers {
             OnPublished<AutoroutePart>((ctx, part) => PublishAlias(part));
 
             // Remove alias if destroyed, removed or unpublished
-            OnRemoving<AutoroutePart>((ctx, part) => RemoveAlias(part));
-            OnDestroyed<AutoroutePart>((ctx, part) => RemoveAlias(part));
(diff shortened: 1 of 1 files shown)


sebastienros:
@HermesSbicego-Laser are you ok with that?

HermesSbicego-Laser:
I agree that there is an issue here, but I do not agree with the proposed fix.
Instead of changing the `RemoveAlias` method adding a boolean, I would suggest to pass the AutoroutePart of the `published content`  (and not the autoroute of the `current` one, the `latest` using the admin UI ) to RemoveAliases method `_autorouteService.Value.RemoveAliases(part);`.
If we opt for this solution, the call to the handlers should be moved to their "ing" methods (removing, unpublishing, detroying).
More, with the code proposed by @1996dylanriley this case will not be managed:
1. Create a page (alias=page) and publish it
2. Modify the alias (alias=page2) and Save it
3. Delete/remove/unpublish the content. 
The alias will still be there in the table.

d-lan1:
@sebastienros will you be implementing the changes suggested by @HermesSbicego-Laser ? Thanks guys.

sebastienros:
I won't, I'll wait for a contributor to do it.

HermesSbicego-Laser:
@1996dylanriley, we have not so much time in this period so we can't implement it right now. By the way, it should be a simple fix so if you write the code we can test and validate it in our environments... If you need some code snippets to work on, feel free to ask.

BenedekFarkas:
We closed the #8100 issue for 1.10.3 (because #8150 was merged), but to be sure I tried to repro the problem and it seems it's working OK, so I'm closing this.

PR #8106, opened 2018-09-07
AuditTrail summary view error when title was empty.

protoncm:
I was working on Audit Trail so far. And i found a error when no title defined on content item.

Current code is check for `null` but not for empty string. the error will occur at part of  code line here (Argument Exception)

```
Html.ItemEditLink(title, contentItemId), eventPastTense)
```

where title can not null and empty.

so i just add statement for make sure `title` can not empty


jean:
@agriffard is anything blocking merging this? Was it fixed elsewhere?

Issue #8108, opened 2018-09-10
ProjectionManager can return duplicate results when using multiple filter groups


jimasp:
There is a bug in ProjectionManager relating to filter groups.
This applies to filter groups which produce duplicate results and you are using paging.

To further explain, with easy steps to reproduce:

Imagine a scenario where you have say 20 blog posts.
You want to create a projection with a pager that shows _6 blog posts per page_.

When setting up your Query, you want to have 2 filter groups for your blogs i.e. an OR scenario, these filters are not exclusive (i.e. internally they may produce duplicate results).
For example, one filter may say blogs that have a tag of "news" or blogs that have a tag of "cabbages". So a blog tagged with both would be in both sets.

For the sake of simplicity and testing though, let's just set up your filter groups so they are guaranteed to produce 100% duplicates, eg:
ContentType="BlogPost" OR ContentType="BlogPost"

You will find that on the second page of the projection you get zero results. You will also find you get the wrong number of pages (double) in the pager.

Now let's look at the code in ProjectionManager.GetContentItems():

```
// aggregate the result for each group query
foreach (var contentQuery in GetContentQueries(queryRecord, queryRecord.SortCriteria.OrderBy(sc => sc.Position), tokens)) {
    contentItems.AddRange(contentQuery.Slice(skip, count));  
}
```

as we have a page size of 6, this would return 12 content items (6 from each group)
note the Slice being applied.
a little further down in the code we have:

```
var ids = contentItems.Select(c => c.Id).ToArray();
var groupQuery = _contentManager.HqlQuery().Where(alias => alias.Named("ci"), x => x.InG("Id", ids));
```

this returns 6 content items as in fact they are all duplicates.

following on from that, we have:

`return groupQuery.Slice(skip, count);`

Now, that would work fine on the first page, as skip will be 0, and count will be 6.

But on the second page of the pager, you will have skip of 6 and count of 6.

Which gives you 0 results.

I'm guessing this should be changed to:
`return groupQuery.Slice(0, count);`

Duplicate results inside the group filters will also cause the pager number of pages to be wrong due to the TotalItemCount being incorrect. This is effectively set via ProjectionManager.GetCount():
```
// aggregate the result for each group query
return GetContentQueries(queryRecord, Enumerable.Empty<SortCriterionRecord>(), tokens)
    .Sum(contentQuery => contentQuery.Count());
```
So this needs to be changed to something more like this:
```
[...]

PR #8114, opened 2018-09-14
#8108 bug in projection manager for dupes from group filters

jimasp:


sebastienros:
Can you give us some number for perf before and after the change on a scenario that is impacted by this change? And also on a scenario that shout not be impacted (no groups).

jimasp:
I ran a very quick + basic test.

I created a query based with Content Type "Any" as a filter.
I also created a Group query with Content Type "Any" OR Content Type "Any".

I then put a StopWatch start in ProjectionPartDriver like so:
```
  var metaData = Services.ContentManager.GetItemMetadata(part.ContentItem);
  Stopwatch stopwatch = new Stopwatch();
  stopwatch.Start();
```

And added Logger code to output the Elapsed.TotalMilliSeconds:
```
if (part.Record.DisplayPager) {
    var contentItemsCount = _projectionManager.GetCount(query.Id, part) - part.Record.Skip;
    contentItemsCount = Math.Max(0, contentItemsCount);
    pagerShape.TotalItemCount(contentItemsCount);
}
Logger.Debug("ProjectionDriver Elapsed:" + stopwatch.Elapsed.TotalMilliseconds);
```

Such that both `_projectionManager.GetContentItems()` and `_projectionManager.GetCount(query.Id, part)` are contained in the timer.

I disabled SysCache, and linked up a projection page with a pager to the test, with the setting for "Maximum number of items which can be queried at once" set to unlimited, and 10 items per page.

I loaded up an unrelated page in the site to start up Orchard and removed all scheduled tasks so unrelated db operations did not interfere with the test.
Then I loaded up the Projection Page 20 times with Ctrl+F5.

Tests run with 18365 (published) content items in the database:

With PR changes:
With Group: mean average - 120ms (range: 115 - 127ms)
Without Group: mean average - 97ms (range: 95 - 132ms)

Without PR changes:
With Group: mean average - 105ms (range: 98 - 148ms)
Without Group: mean average - 85ms (range: 81 - 90ms)

These times are from a debug build running on a fast dev PC.

sebastienros:
@BenedekFarkas what did we decide the next step is with this PR? Did you want to test it?

BenedekFarkas:
Yes, sorry, it's on my list, but haven't had time to do it yet.

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
```diff
@@ -120,9 +120,12 @@ public class ProjectionManager : IProjectionManagerExtension {
...
 
-            // aggregate the result for each group query
-            return GetContentQueries(queryRecord, Enumerable.Empty<SortCriterionRecord>(), tokens)
-                .Sum(contentQuery => contentQuery.Count());
+            var contentQueries = GetContentQueries(queryRecord, Enumerable.Empty<SortCriterionRecord>(), tokens);
+
+            if (queryRecord.FilterGroups.Count <= 1)
+                return contentQueries.Sum(contentQuery => contentQuery.Count());
+            else 
+                return contentQueries.SelectMany(contentQuery => contentQuery.ListIds()).Distinct().Count();
```
BenedekFarkas: I have a fetish for the ternary-operator, to be honest, but besides that, I think this would be a bit cleaner:

```suggestion
            return queryRecord.FilterGroups.Count > 1 ?
                contentQueries.SelectMany(contentQuery => contentQuery.ListIds()).Distinct().Count() :
                contentQueries.Sum(contentQuery => contentQuery.Count());
```

I added this comment with the suggestion feature for the better syntax highlighting, but you won't be able to correctly apply it directly, because this change corresponds to multiple lines in your code.

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
```diff
@@ -120,9 +120,12 @@ public class ProjectionManager : IProjectionManagerExtension {
                 tokens.Add("Content", part.ContentItem);
             }
 
-            // aggregate the result for each group query
-            return GetContentQueries(queryRecord, Enumerable.Empty<SortCriterionRecord>(), tokens)
-                .Sum(contentQuery => contentQuery.Count());
+            var contentQueries = GetContentQueries(queryRecord, Enumerable.Empty<SortCriterionRecord>(), tokens);
+
+            if (queryRecord.FilterGroups.Count <= 1)
```
BenedekFarkas:
```suggestion
```

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
```diff
@@ -120,9 +120,12 @@ public class ProjectionManager : IProjectionManagerExtension {
...
+                return contentQueries.Sum(contentQuery => contentQuery.Count());
```
BenedekFarkas:
```suggestion
```

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
```diff
@@ -120,9 +120,12 @@ public class ProjectionManager : IProjectionManagerExtension {
...
+            else
```
BenedekFarkas:
```suggestion
```

src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
BenedekFarkas: @jimasp I added a suggestion for each (4) line involved, so you can commit them in a single batch.

BenedekFarkas:
@sebastienros
I'm closing this: @jimasp's performance numbers are really promising and the change isn't huge. I also tested it with a site that has multiple thousands of published content items (the client is already running this code in production). I had some code styling suggestions which I committed to Jim's branch, so it's included in the PR.

Issue #8007, opened 2018-04-01
Update target framework to 4.8


Jetski5822:
By doing this you allow us to use the majority of the netstandard packages.

sebastienros:
We need to know what is the minimum version installed by Windows Update. Right now we use 4.5.2 as we are sure it's pushed by WU.

carlwoodhouse:
i think 4.5.2 is the minimum req version by wu still. though 2016/win10+ come with 4.6.1 pre-installed.

that been said, give 4.6.2 so i can use the http2 features ;p

BenedekFarkas:
@sebastienros I'd like to circle back to the choice of upgrading to 4.6.1 as discussed on the PR: I think the argument for 4.6.1 was mainly about whether it's included in Windows Update. Based what I found in the [version and dependency table](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies), here's my take on the issue:

Orchard 1.10 shipped in March of 2016 with .NET 4.5.2 (which was released almost 2 years earlier), but .NET 4.5.2 is not included in any OS versions by default, so at some point you had to install it manually (I'm not sure if it was automatic with Windows Update, but it might've been).

.NET versions above 4.6.1 offer a lot of improvements in different areas (many are unrelated though), but some of them can still be significant performance, security, and/or quality of life improvements for Orchard-based development. In the meanwhile, we're also looking at NuGet package updates in #7998, so I think it's a good chance to give Orchard a nice refresh.

.NET 4.8 is not yet supported by Azure App Service for example, so I'd recommend 4.7.2:  It has basically the same compatibility as 4.5.2 and is included in Windows 10 and Server versions older than a year now (Windows 10/Server 1803).

sebastienros:
You are the boss

sebastienros:
> .NET 4.8 is not yet supported by Azure App Service for example, so I'd recommend 4.7.2

Since yesterday 4.8 is on Windows Update so I assume all machines should have it. even App Service

BenedekFarkas:
Nice, that's even better! I don't see it yet in the configuration for an App Service (still 4.7), but I guess it's just a matter of time. AWS Elastic Beanstalk seems to support 4.8 already.

mwentz:
With TLS 1.0 being deprecated and TLS 1.2 becoming the required standard, this ticket should be prioritized. It is recommended that applications use at least 4.7:
[...]

PR #8126, opened 2018-10-06
Upgrade to .NET 4.6.1

aaronamm:
- Upgrade to .NET 4.6.1 which can be use with .NET Standard library 2.0
- PowerShell script tool to upgrade the .NET framework to all projects and

image

### DIFF
--- a/src/Orchard.Azure.Tests/Orchard.Azure.Tests.csproj
+++ b/src/Orchard.Azure.Tests/Orchard.Azure.Tests.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Orchard.Azure.Tests</RootNamespace>
     <AssemblyName>Orchard.Azure.Tests</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <FileUpgradeFlags>
     </FileUpgradeFlags>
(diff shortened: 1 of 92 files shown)


aaronamm:
@sebastienros  @BenedekFarkas  I just found for using C# 6 feature, you don't need to upgrade to a newer version of .NET. However, I will let you decide if you want to keep this PR or not because PowerShell script to change targeting .NET still useful.

Thanks.

BenedekFarkas:
Related issue: #8007, which mentions that 4.6.2 would also be nice for HTTP/2 support.

@sebastienros judging from the .NET documentation 4.6 would sufficient for that - on the other hand is there a reason not to upgrade to 4.6.2 or even higher? Hosting could be an issue if they are behind with the Windows or Windows Server version, right?

zakaluka:
According to this (https://docs.microsoft.com/en-us/dotnet/framework/get-started/system-requirements), as long as the server has Windows Server 2008 R2, they can go up to .NET Framework 4.7.2 (maybe 4.7.3 also?).

aaronamm:
@BenedekFarkas  @sebastienros  Sorry for late reply since I was on a vacation.
 I am happy to help to update to any versino.
Please let me the your decision. 
For me,  it would be nice, if we can start using a class library that supports only .NET standard library 2.0 up. 

Making a .NET standard library is a great idea to maintain only one code base and works to all platform.
It is very useful for opensource source projects.

BenedekFarkas:
Hey, sorry for replying just now!

@aaronamm the PS script is really nice - I was doing these kinds of bulk updates with search and replace in Notepad++. :)
Maybe we can improve it a bit by processing the XML, but it might not be necessary. The concept, however is nice as we can do the same for LangVersion as well for example.

@zakaluka thanks for looking into the framework compatibility - I think we should be OK with 4.7.2 (so we support from Windows Server 2008 R2 onwards), but I'd like @sebastienros' opinion on it.

aaronamm:
@BenedekFarkas  I think I can do it.
I will convert the PowerShell script to process XML in a PowerShell way and update the project to .NET 4.7.2.

jean:
@aaronamm did you ever get round to that 4.7.2 update?

sebastienros:
4.6.1 approved

PR #8127, opened 2018-10-06
update User migration to follow a migration convention

aaronamm:
update User migration to follow a migration convention

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Migrations.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Migrations.cs
@@ -20,15 +20,9 @@ namespace Orchard.Users {
                     .Column<string>("RegistrationStatus", c => c.WithDefault("Approved"))
                     .Column<string>("EmailStatus", c => c.WithDefault("Approved"))
                     .Column<string>("EmailChallengeToken")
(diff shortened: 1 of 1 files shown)


sebastienros:
Why do you think it's wrong? It looks totally fine to me. On a new setup the Create will be executed, and no other step will as it returns 5. If it was to return 1 it would execute 2, 3, 4, 5 which would be wrong.

aaronamm:
Dear @sebastienros 

Thank you for your comment.
I understand your opinion that my current PR is incorrect and can cause an issue.
However, having first migration return 5 and ignore the following migration look like we are rewriting the history.

For migration, a class should follow the pattern which is "UpdateDateFromN return N+1" and keep a number of methods growing.

If we do this, what will happen when we are going to have version 6.
Should we edit the first method and return 6?

Therefore, I have updated my changes and force push my commit.

Thanks.

sebastienros:
Yes, when you create version 6 we "should" rewrite the Create too. The idea is that `Create` is only called for brand new installs, which mean if they can use a shortcut then they should. Let's say each migration adds and removes some columns? What would be the point of doing it for new installs, let's just create the table with the correct columns if we know if a new install.

PR #8128, opened 2018-10-08
upgrade microsoft.net.compilers to 2.4.0

aaronamm:
Upgrade all projects that have Microsoft.Net.Compilers to version 2.4.0

### DIFF
--- a/src/Orchard.Web/Core/Orchard.Core.csproj
+++ b/src/Orchard.Web/Core/Orchard.Core.csproj
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Microsoft.Net.Compilers.1.3.2\build\Microsoft.Net.Compilers.props" Condition="Exists('..\..\packages\Microsoft.Net.Compilers.1.3.2\build\Microsoft.Net.Compilers.props')" />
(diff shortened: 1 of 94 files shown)


sebastienros:
@BenedekFarkas related to your changes?

aaronamm:
@sebastienros  Thanks for helping update.

BenedekFarkas:
@sebastienros yes! Unfortunately, I think it would be easier if we rejected the PR (sorry @aaronamm), because this is just one part of the upgrade process of the Dynamic Compilation changes we already made on 1.10.x. Btw 1.10.x includes Microsoft.Net.Compilers 2.9.0 (the current latest version), 2.4.0 is a year old.
So we'll upgrade dev to 2.9.0 as well soon.

aaronamm:
@BenedekFarkas Do you want me help update `dev` branch to 2.9.0? I am happy to help. 
Not sure about your plan, will you merge `1.10.x` back to `dev` to get the upgrade. My PR just want to make everything consistency. ^^

BenedekFarkas:
@aaronamm yep, 1.10.x (or actually, probably the commit on 1.10.x that will be marked as the 1.10.3 release) will be merged to dev (and btw the minor version branch is more or less regularly merged into the major version branch), so most of this work is already done (on 1.10.x), there's only a couple more modules on dev that aren't on 1.10.x.

On the other hand, `Microsoft.Net.Compilers` is needed anymore separately (it's completely removed on 1.10.x now), because `Microsoft.CodeDom.Providers.DotNetCompilerPlatform` ships with it included (since version 2.0.0 I think).
1.10.x has the 2.0.1 version, so that includes the 2.10 compiler package (but a newer one has been released recently, so probably DotNetCompilerPlatform will have a new release soon as well).

So thanks and sorry, I'll close this PR, but we can coordinate on other updates (such as the other thread I just commented to).

aaronamm:
Thank you ^^

PR #8144, opened 2018-12-05
Fixes #8143 - CultureNeutral Settings not saved…

### LINKED ISSUE #8143: Cannot save CultureNeutral Setting in a field within a part
When you create a part and you define a field within it the CultureNeutral Settings is not correctly set
(dev)

HermesSbicego-Laser:
The CultureNeutral Settings of a field defined withina Part was not saved on settings update


sebastienros:
Fixes #8143

PR #8154, opened 2019-01-14
Fix Breadcrumb when ContentItem without TitlePart.

Patchareeya:
When ContentItem using Orchard.Lists doesn't have TitlePart, Html.ItemDisplayText(content) return null and show error on website.
So, I using `ContentType` name instead title.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Lists/Views/Breadcrumbs.ContentItem.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Lists/Views/Breadcrumbs.ContentItem.cshtml
@@ -4,7 +4,7 @@
     var breadcrumbs = new List<dynamic>();
     var containerAccessor = (Func<IContent, IContent>) Model.ContainerAccessor;
     do {
-        var displayText = content.Id > 0 ? (Html.ItemDisplayText(content) != null) ? Html.ItemDisplayText(content).ToString() : T("{0}", content.ContentItem.TypeDefinition.DisplayName).ToString() : T("New {0}", content.ContentItem.TypeDefinition.DisplayName).ToString();
+        var displayText = content.Id > 0 ? (Html.ItemDisplayText(content) != null) ? Html.ItemDisplayText(content).ToString() : content.ContentItem.TypeDefinition.DisplayName : T("New {0}", content.ContentItem.TypeDefinition.DisplayName).ToString();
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Lists/Views/Breadcrumbs.ContentItem.cshtml
```diff
@@ -4,10 +4,10 @@
     var breadcrumbs = new List<dynamic>();
     var containerAccessor = (Func<IContent, IContent>) Model.ContainerAccessor;
     do {
-        var displayText = content.Id > 0 ? Html.ItemDisplayText(content).ToString() : T("New {0}", content.ContentItem.TypeDefinition.DisplayName).ToString();
+        var displayText = content.Id > 0 ? (Html.ItemDisplayText(content) != null) ? Html.ItemDisplayText(content).ToString() : T("{0}", content.ContentItem.TypeDefinition.DisplayName).ToString() : T("New {0}", content.ContentItem.TypeDefinition.DisplayName).ToString();
```
sebastienros: Replace `T("{0}", content.ContentItem.TypeDefinition.DisplayName)` by `content.ContentItem.TypeDefinition.DisplayName`. 

No translation is needed for a raw string.

BenedekFarkas:
@sebastienros I agree with the concept of the change (showing content type name if display name is not available - there are other places in Orchard where the same logic is already applied), but I think we can take this one step further:
ContentsHandler already sets default ContentItemMetadata values (routes) if they are null, so we can do the same with display text:
```
if (string.IsNullOrWhiteSpace(context.Metadata.DisplayText))
    context.Metadata.DisplayText = context.ContentItem.ContentType;
```

This would fix the issue above and also a lot of other cases, where some text (e.g. a notification) or logic expects a display text, such as when moving an item between two containers, but neither the item being moved, nor the containers have TitlePart:
` was moved from  to .`
would change to:
`Test was moved from List to List.`
Which is still strange, but at least it's not broken.

What do you think? I'd add this in 1.10.x.

sebastienros:
@BenedekFarkas there can still be some cases where it wouldn't work, but they should not happen in practice. Ok with your suggestion then.

BenedekFarkas:
Closing this PR, because a more generic fix for this kind of problem is applied in 026f81fd16afdd6ebbeddabdc26e1658e8ceded8.

PR #8167, opened 2019-03-08
Custom Forms module enhanced

MarcoViglione-Laser:
Added editing capabilities and improved permission handling


sebastienros:
Does anyone who uses the CustomForms module want to add a comment on this feature?

Issue #6782, opened 2016-04-15
Add parameters for field indexing


sebastienros:
Right now we can set any field to be indexed. 
We need to be able to define these options too:
- Stored (default false)
- Analyzed (default true)
- RemoveTags (default false)

We could also be able to define the name in the index, though right now it works fine as it's `[contenttype]-[fieldname]'.

paynecrl97:
I have the changes ready for this, but I no longer seem to be able to push to this repo 😞 . Can someone help me out and add me to the correct team please?

paynecrl97:
I've opened a PR that adds these three options to every field. There's a slight problem though:

Those three options are not always valid depending on the type of data that the field contains. e.g. remove tags is not valid on a boolean field.

I also noticed that the existing IsIndexed setting is displayed for all fields, but it only affects Infoset fields e.g. it's there for the Taxonomy field, but the taxonomy field handles its own indexing and so does not respect the configuration on the field.


I think the solution to both of these problems would be to allow fields to describe what indexing configuration they support, but that's a separate piece of work IMO.

PR #8173, opened 2019-03-22
Adds ability to configure field indexing options

paynecrl97:
For #6782 

I made analyze and remove tags set to true by default to avoid breaking changes (as this is current behaviour)


sebastienros:
Fixes #6782

Issue #7179, opened 2016-09-30
Orchard.AuditTrail - Encoding Issues With Content Titles


rtpHarry:
On main index:

![image](https://cloud.githubusercontent.com/assets/1038062/19003835/88e7c77c-874a-11e6-8c62-2394b1a24a55.png)

And in detail:

![image](https://cloud.githubusercontent.com/assets/1038062/19003850/986ab150-874a-11e6-8674-afa85b1b2129.png)

rtpHarry:
And in filtered view when you click Audit Trail link from, for example, blog post list view:

![image](https://cloud.githubusercontent.com/assets/1038062/19003900/d7334d0c-874a-11e6-934f-e9d85ee09d36.png)

rtpHarry:
@Skrypt this one isn't admin theme related

PR #8178, opened 2019-04-03
Fix title encoding for edit content item, Fix encoding for Audittrail events

petrdvorak1:
Edited content items have title with html entities for some chars, we use Html.Raw(typeDisplayName) for fix this issue.

Audittrail has bad encoding in Audittrail list for ContentItem name, fix this with add false params to: Html.ItemDisplayText(contentItem, false)

### DIFF
--- a/src/Orchard.Web/Core/Contents/Views/Admin/Edit.cshtml
+++ b/src/Orchard.Web/Core/Contents/Views/Admin/Edit.cshtml
@@ -2,7 +2,7 @@
     var typeDisplayName = Model.ContentItem.TypeDefinition.DisplayName;
     var pageTitle = T("Edit Content");
     if (!string.IsNullOrWhiteSpace(typeDisplayName)) {
-        pageTitle = T("Edit {0}", typeDisplayName);
+        pageTitle = T("Edit {0}", Html.Raw(typeDisplayName));
     }
 
     Layout.Title = pageTitle;
--- a/src/Orchard.Web/Modules/Orchard.AuditTrail/Views/AuditTrailEvent-Content.SummaryAdmin.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.AuditTrail/Views/AuditTrailEvent-Content.SummaryAdmin.cshtml
@@ -1,20 +1,16 @@
-@using Orchard.AuditTrail.Helpers
+﻿@using Orchard.AuditTrail.Helpers
 @using Orchard.AuditTrail.Services.Models
 @using Orchard.ContentManagement
 @{
     var descriptor = (AuditTrailEventDescriptor)Model.Descriptor;
     var eventData = (IDictionary<string, object>) Model.EventData;
     var eventPastTense = descriptor.Name.Text.ToLower();
-    var contentItemId = (int?)Model.ContentItemId;
+    var contentItemId = (int)Model.ContentItemId;
     var contentItem = (ContentItem) Model.ContentItem;
     var eventVersionNumber = eventData.Get<int>("VersionNumber");
     var isPublishedEvent = eventData.Get<bool>("Published");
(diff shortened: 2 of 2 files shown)


sebastienros:
Can you provide some repro steps for each issue?

petrdvorak1:
1. Create new type named: Články - seznam (this name is important), Id: articlelist, with Title part
2. Create new content item for: Články - seznam (publish)  with title: Seznam článků
3. You can see bad encoding in page title
image
4. Select Audit Trail from admin menu
5. You can see bad encoding in Summary column 
image

6. there is similar isssue: https://github.com/OrchardCMS/Orchard/issues/7179

src/Orchard.Web/Modules/Orchard.AuditTrail/Views/AuditTrailEvent-Content.SummaryAdmin.cshtml
```diff
@@ -1,37 +1,34 @@
-@using Orchard.AuditTrail.Helpers
+﻿@using Orchard.AuditTrail.Helpers
 @using Orchard.AuditTrail.Services.Models
 @using Orchard.ContentManagement
 @{
     var descriptor = (AuditTrailEventDescriptor)Model.Descriptor;
     var eventData = (IDictionary<string, object>) Model.EventData;
     var eventPastTense = descriptor.Name.Text.ToLower();
-    var contentItemId = (int?)Model.ContentItemId;
+    var contentItemId = (int)Model.ContentItemId;
```
sebastienros, line 8: Please don't change that and the `.Value` after. It was because some old instances have corrupted data with null values.

src/Orchard.Web/Modules/Orchard.AuditTrail/Views/AuditTrailEvent-Content.SummaryAdmin.cshtml
```diff
@@ -1,37 +1,34 @@
...
     var contentItem = (ContentItem) Model.ContentItem;
     var eventVersionNumber = eventData.Get<int>("VersionNumber");
     var isPublishedEvent = eventData.Get<bool>("Published");
-    var itemDisplayText = contentItem != null ? Html.ItemDisplayText(contentItem) : default(IHtmlString);
-    var title = itemDisplayText != null ? itemDisplayText.ToString() : eventData.Get<string>("Title");
-    if (string.IsNullOrEmpty(title))
-    {
-        title = "[untitled]";
-    }
+    var itemDisplayText = contentItem != null ? Html.ItemDisplayText(contentItem, false) : default(IHtmlString);
```
sebastienros, line 12: This is wrong, because you are not encoding the value, and it's correctly encoded only if it's rendered by ItemEditLink lower. But there is a case where it's passed to `Html.Raw`, so in this case it needs to be encoded.

The solution is to not use ItemEditLink but ItemEditUrl and build the <a> yourself.

src/Orchard.Web/Core/Contents/Views/Admin/Edit.cshtml
```diff
@@ -2,7 +2,7 @@
     var typeDisplayName = Model.ContentItem.TypeDefinition.DisplayName;
     var pageTitle = T("Edit Content");
     if (!string.IsNullOrWhiteSpace(typeDisplayName)) {
-        pageTitle = T("Edit {0}", typeDisplayName);
+        pageTitle = T("Edit {0}", Html.Raw(typeDisplayName));
```
sebastienros, line 5: Are you sure it's necessary? The display name should not have been encoded by anything.

BenedekFarkas:
@petrdvorak1 can you please take care of the review feedback and the CLA?

sebastienros:
This PR keeps being seen as updated in GH

PR #8184, opened 2019-04-10
Fix NRE in DefaultHQLQuery

MatteoPiovanelli:
In the InSubquery method Criterion was not being set. This caused NRE when processing a WHERE predicate that used that method.


sebastienros:
These are merged in 1.10.x because you still use it, and because it can be. But we'll have to merge it to dev, and also it might not be release as a 1.10.4 at there might not be one. Next release is supposed to be 1.11.

/cc @BenedekFarkas

PR #8186, opened 2019-04-16
performance fix for autoroute check

GiuseppeMusso-Laser:
Hi all, this is a proposal to avoid unnecessary loops on autoroute part records.
I think it has no negative side effects but I don't know very well autoroute parts.
Is there someone who can check if this solution is good?
Thanks!

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs
@@ -206,21 +206,26 @@ namespace Orchard.Autoroute.Services {
         }
 
         public bool ProcessPath(AutoroutePart part) {
-            var pathsLikeThis = GetSimilarPaths(part.Path).ToArray();
+            var parthWithSamePath = _contentManager.Query<AutoroutePart, AutoroutePartRecord>()
+                .Where(p => p.DisplayAlias != null && p.DisplayAlias.Equals(part.Path, StringComparison.InvariantCultureIgnoreCase)
+                && p.ContentItemRecord.Id != part.Id)
+                .Count();
+            if (parthWithSamePath > 0) {
+                var pathsLikeThis = GetSimilarPaths(part.Path).ToArray();
 
-            // Don't include *this* part in the list
-            // of slugs to consider for conflict detection.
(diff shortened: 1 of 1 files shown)


Skrypt:
This is quite sensible code there. It should be reviewed carefully ; we never know what small changes could break in there.

```C#
var parthWithSamePath = _contentManager.Query<AutoroutePart, AutoroutePartRecord>()
                .Where(p => p.DisplayAlias != null && p.DisplayAlias.Equals(part.Path, StringComparison.InvariantCultureIgnoreCase)
                && p.ContentItemRecord.Id != part.Id)
                .Count();
            if (parthWithSamePath > 0) {
```

Just make sure that you don't get any issue when trying to create multiple content items with the tool you guys created. This could potentially create database locks.

src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs
```diff
@@ -206,21 +206,26 @@ public class AutorouteService : Component, IAutorouteService {
...
-            pathsLikeThis = pathsLikeThis.Where(p => p.ContentItem.Id != part.ContentItem.Id).ToArray();
-
-            if (pathsLikeThis.Any()) {
-                var originalPath = part.Path;
-                var newPath = GenerateUniqueSlug(part, pathsLikeThis.Select(p => p.Path));
-                part.DisplayAlias = newPath;
-
-                if (originalPath != newPath)
-                    return false;
+            var parthWithSamePath = _contentManager.Query<AutoroutePart, AutoroutePartRecord>()
```
sebastienros, line 209: How is that better? Can you explain what makes it faster on lots of records?

Also is this query done only on database, or does it load everything and then filters in memory?

GiuseppeMusso-Laser: I apologize for the late reply.
GetSimilarPaths performs a query to search all paths starting with the input path. 
In few cases it can extract many records. 
I experienced this problem when I updated my home page whose path was "". 
Because every path starts with "", GetSimilarPaths extracts all paths on my tenant (about 4.000 paths in my case). 
This operation takes a bit of time.
My suggestion is to check for the existence of a path that equals the input path before calling GetSimilarPaths. 
If no such path exists, there is no need to calculate a new path.
So, if my autoroute patterns generate different paths for my content, I get a performance improvement.
Anyway I don't know autoroute management in depth and maybe I missed something...

PR #8193, opened 2019-05-03
[8192] Fix permissions check for taxonomies, terms and widgets during translations

### LINKED ISSUE #8192: Orchard.Localization: Translations tests against wrong permissions
A user without `EditContent permission` if try to translate a Content not set as `Securable` (Taxonomy, Term, ...) cannot translate the content beacause unauthorized (the Orchard.Localization.Controllers.AdminController tests the `EditContent Permission`).
Ex:
- Set two cultures to a new tenant ("it-IT", "en-US") 
- Create a localized taxonomy ("categorie","categories")
- Create a role EditorForTerms and assign it to a new user: with the "Manage terms", "Access Admin Panel" permission
- Log in with the new created user
- Create a new term "term 1 it" culture it-IT
- Translate the "term 1 it"

You will obtain an Unauthorized message.

HermesSbicego-Laser:
Partially (only for taxonomies and terms) fix #8192

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Orchard.Taxonomies.csproj
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Orchard.Taxonomies.csproj
@@ -135,6 +135,7 @@
     <Compile Include="Handlers\LocalizedTaxonomyPartHandler.cs" />
     <Compile Include="LocalizedTaxonomyMigration.cs" />
     <Compile Include="Models\MoveTermsContext.cs" />
+    <Compile Include="Security\AuthorizationEventHandler.cs" />
     <Compile Include="Services\ITaxonomyExtensionsService.cs" />
     <Compile Include="Services\ITaxonomySource.cs" />
     <Compile Include="Services\LocalizedTaxonomySource.cs" />
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Security/AuthorizationEventHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Security/AuthorizationEventHandler.cs
@@ -1,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web;
+using Orchard.ContentManagement;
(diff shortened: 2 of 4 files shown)


HermesSbicego-Laser:
@sebastienros I don't know if it is the right way to proceed, so I stop for now before modifying too many modules.
Basically, so proceeding, every module creating its own `ContentTypes` (not securable and potentially translatable) and having its own `controllers` and `permissions` should implement its own `IAuthorizationServiceEventHandler` in order to adjust permissions (this is the case of `Taxonomies`, `Terms`, `Widgets`, `Menus`, `Media`, `Blog`, `BlogPost`, ... how many other?)
It would be fine to have a more general way to make this kind of check, what do you think... is there an alternative approach or should I continue in this way?

HermesSbicego-Laser:
As discussed during meeting I tried to remove unnecessary `permission` adjustments. 
However I found some troubles to do what you asked for, especially for `Terms`.
I removed all adjustments except for `Edit`, but when I try to create a new Term the `Publish `button is missing. 
So i remembered why I added the adjustements for all Permissions (`Edit, Create, Publish,` etc...): it's because the `Editor Views` use some Core Shapes (like Content.PublishButton.cshtml, Content.SaveButton,etc...) that test against the generic/standard `Content Permissions`.
Moreover, the `Terms` they have not a static `ContentType` since the `ContentType` name depends on `taxonomy` name (i.e. if we have a  taxonomy "categories" its ContentType will be "categories Term") so we cannot use alternates like Content-Term.PublishButton.cshtml.
More, I cannot see a real problem on "remapping" permissions for those contents/functionality that have their own permissions; on the contrary, I see only advantages. If somewhere the code tests Content Permissions for a Taxonomy or a Term or a Widget or a Menu (more generically for a Content that have  its own functionality/permission written in a module), since there is a more specific permission this one will be used: and it seems to me to be more correct.
In the light of these considerations I ask to re-evaluate the submitted PR. 
@sebastienros I am willing to talk about it and to optimize what is written to avoid "too much code", but I think it is correct to adjust all permissions where there is the possibility of mapping them one by one.

PR #8009, opened 2018-04-02
[8008]Workflow404cssfiles

HengzheLi:
fixes #8008

Issue #8008, opened 2018-04-02
404 error in workflow edit page for WorkflowsActivity-*.css files


HengzheLi:
1. go to workflow list page
2. edit or create one
3. F12
4. 404 error on WorkflowsActivity-*.css files

sebastienros:
Duplicate of https://github.com/OrchardCMS/Orchard/issues/7873

BenedekFarkas:
Fixed in #8201.

PR #8201, opened 2019-05-15
Fixing that Orchard.Workflows requires styles that don't exist

### LINKED ISSUE #7873: Orchard.Workflows requires styles that not exists
In the ResourceManifest.cs file of the Orchard.Workflows module there is a foreach that creates a DefineStyle for each existing activity, searching their corresponding css in the Orchard.Workflows/Styles folder.
If that file does not exist there will be a missing reference error in the page where we edit a workflow.

The procedure should check if the file exists before adding the requirement and add it only if there already is a css with the corresponding name. Maybe it should also search in the styles folder of the module where the activity is defined.

barthamark:
Fixes #7873 and #8008 .

Replaces PR #8009 where the source repository doesn't exist anymore.


src/Orchard.Web/Modules/Orchard.Workflows/ResourceManifest.cs
```diff
@@ -8,23 +9,36 @@
...
+                var basePath = VirtualPathUtility.AppendTrailingSlash("~/Modules/" + assemblyName + "/styles/");
+                var resourceName = "WorkflowsActivity-" + activity.Name;
+                var filename = resourceName.HtmlClassify() + ".css";
+
+                if (File.Exists(_hostEnvironment.MapPath(basePath) + filename)) {
+                    resourceNames.Add(resourceName);
+
+                    manifest.DefineStyle(resourceName)
+                        .SetUrl(filename)
+                        .SetDependencies("WorkflowsAdmin");
```
BenedekFarkas: Minor code styling: We can write this in one line.

src/Orchard.Web/Modules/Orchard.Workflows/ResourceManifest.cs
```diff
@@ -8,23 +9,36 @@
...
             manifest.DefineStyle("WorkflowsAdmin").SetUrl("orchard-workflows-admin.css").SetDependencies("~/Themes/TheAdmin/Styles/Site.css");
 
-            var activityNames = _activitiesManager.Value.GetActivities().Select(x => "WorkflowsActivity-" + x.Name).ToArray();
+            var resourceNames = new List<string>();
 
-            foreach (var activityName in activityNames) {
-                manifest.DefineStyle(activityName).SetUrl(activityName.HtmlClassify()).SetDependencies("WorkflowsAdmin");
+            foreach (var activity in _activitiesManager.Value.GetActivities()) {
+                var assemblyName = activity.GetType().Assembly.GetName().Name;
+                var basePath = VirtualPathUtility.AppendTrailingSlash("~/Modules/" + assemblyName + "/styles/");
```
BenedekFarkas: It's technically possible for a Theme to define Workflow activities, but we won't find them this way, because "Modules" is wired in.

BenedekFarkas (changes requested):

BenedekFarkas (approved):

Issue #7925, opened 2017-12-27
Merge DraftFieldIndexService.cs into FieldIndexService.cs


HermesSbicego-Laser:
As requested by @sebastienros , after #7923, we need to merge DraftFieldIndexService.cs into FieldIndexService.cs. (to be targeted 1.11)

PR #8205, opened 2019-05-15
Merge DraftFieldIndexService.cs into FieldIndexService.cs #7925 (Lombiq Technologies: ORCH-199)

domonkosgabor:


sebastienros:
LGTM

BenedekFarkas (changes requested):
@domonkosgabor please make the following changes:

1. Change the if-else structures to switch-case in FieldIndexService. Since FieldIndexRecordVersionOptions is an enum, we could support further values (for example QueryVersionScope supports Draft-only too).
2. Simplify the return statements, as they duplicate a lot of code between the case execution paths.
3. Also apply any code lens recommendations (e.g. "string" instead of "String"), including C# 7 language features.
4. Resolve the merge conflict.

BenedekFarkas (approved):

Issue #8196, opened 2019-05-10
CommonPart(Version)Record query performance problems


BenedekFarkas:
There are no indexes on CommonPart(Version)Record columns, which can cause performance issues and query timeout with a couple thousand content items in common scenarios, such as simply displaying the content list at "/Admin/Contents/List". This listing can order items by Created/Modified/Published date, so at least those columns should have an index.

sebastienros:
Good point

BenedekFarkas:
Fixed in #8210

PR #8210, opened 2019-05-21
Adding indexes to CommonPart(Version)Record tables #8196 (Lombiq Technologies: ORCH-193)

eszterkoczak:


src/Orchard.Web/Core/Common/Migrations.cs
```diff
@@ -112,5 +120,22 @@ public class Migrations : DataMigrationImpl {
             SchemaBuilder.AlterTable("CommonPartVersionRecord", table => table.AddColumn<string>("ModifiedBy", command => command.Nullable()));
             return 5;
         }
+
+        public int UpdateFrom5() {
+
```
BenedekFarkas: @eszterkoczak unnecessary empty line.

src/Orchard.Web/Core/Common/Migrations.cs
```diff
@@ -112,5 +120,22 @@ public class Migrations : DataMigrationImpl {
...
+            SchemaBuilder.AlterTable(nameof(CommonPart), table => {
```
BenedekFarkas: @eszterkoczak this should reference CommonPartRecord, just like in Create.

BenedekFarkas (changes requested):

PR #8213, opened 2019-05-24
Added a filter for projections over StringFields

MatteoPiovanelli:
Added a filter for projections over StringFields that allows returning all items when the provided value is empty.
The goal for this (base on an actual use case we have):
We want to do a "ContainsAny" query over the selected Ids on a ContentPickerField, and return all Items if no "filter" is selected.


MatteoPiovanelli:
Note:
This PR was based off the code in the 1.10.3 tag/release

MatteoPiovanelli:
Note 2:
We made this because it's a somewhat common thing we need to use, and we often had to work around this (and the projection manager looks to me to be kinda broken when there are more than 1 filter group).

PR #8214, opened 2019-05-24
Fix/8209 Taxonomies Terms

### LINKED ISSUE #8209: Taxonomies and Terms
Some issues I am seeing in terms and taxonomies on the dev branch (pasting here from gitter):

I was checking stuff related to optimizations, specifically wrt Terms and Taxonomies. I noticed that in the dev branch TermPart has a FullWeight property, introduced in #7709 .
Using that property as it is now in dev reverses the ordering of terms in a taxonomy when compared to the situation in the 1.10.x branch. In 1.10.x terms (on the same level) are ordered by decreasing weight. In dev they are ordered by increasing weight.
Additionally, in the code from dev TermPart.Name is not used in ordering terms, while it is being used as a "backup" criterion in 1.10.x when two terms have identical paths and weights.
Both these differences may be somewhat breaking in terms of experience for our users.

used to be htat terms were sorted in memory, based on decreasing weight and then alphabetical order of title
in dev a string is built and stored in the db to "map" the tree structure
it only uses weight, and doing a sql's orderby there results in an increasing weight order

and I have issues with the Post Editor in the TermPartDriver
it's doing TryUpdateModel, then doing further updates no matter what the result.
if it sets that to null, a method later in the update will throw an exception, because it will use that string as a paramenter for a string.IndexOf
```
string parentOldFullWeight = termPart.FullWeight == null ? termPart.FullWeight : "";
```
parentOldFullWeight will be null if termPart.FullWeight is null, empty otherwise

with the driver as it is:
Starting from
parent.FullWeight = 000009.13/
​child.FullWeight = 000009.13/000007.158/

Update parent.

Now you have:
parent.FullWeight = 000009./
​child.FullWeight = 000009./000009.13/000007.158/

I'll try to tackle all that, as well as a few refactorings and optimizations.

MatteoPiovanelli:
fixes #8209

The way FullWeight used to be computed resulted in sometimes wrong values.
The FullWeight was not computed the same way in different places in the code.
The FullWeight ordering was not consistent with the ordering from 1.10.x.
The FullWeight ordering did not account for the Name property of TermParts.

This code is based off and targeted to the dev branch.

(I also took the chance to clean up and refactor some code that bothered me)


sebastienros:
The upgrade code is from 1.10.x to dev?

@BenedekFarkas can you get it tested?

MatteoPiovanelli:
It should work to upgrade from 1.10 to dev as well as to upgrade dev to what is in this pr

MatteoPiovanelli:
Any news on this?
We've been running this on 40+ tenants for a few days, and so far we haven't had any issue with it. It's been boring to run the upgrade step on all of those, but apparently nothing broke yet.

BenedekFarkas:
Sorry @MatteoPiovanelli-Laser, reviewing/testing now!

src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TermPartDriver.cs
```diff
@@ -94,23 +94,9 @@ public class TermPartDriver : ContentPartDriver<TermPart> {
...
-            updater.TryUpdateModel(termPart, Prefix, null, null);
-            StringBuilder fullWeightBuilder = new StringBuilder();
-            string parentOldFullWeight = termPart.FullWeight == null ? termPart.FullWeight : "";
-            TermPart containerTerm = termPart;
-
-            for (int i = 0; i < termPart.Path.Count(x => x == '/') - 1; i++) {
-                containerTerm = containerTerm.Container.As<TermPart>();
-                fullWeightBuilder.Insert(0, containerTerm.Weight.ToString("D6") + "." + containerTerm.Id.ToString() + "/");
+            if (updater.TryUpdateModel(termPart, Prefix, null, null)) {
+                //termPart.FullWeight = _taxonomyService.ComputeFullWeight(termPart);
```
BenedekFarkas: We can just call TryUpdateModel without the "if", because the inner code is commented out.

src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TermPartDriver.cs
```diff
@@ -163,20 +149,23 @@ public class TermPartDriver : ContentPartDriver<TermPart> {
...
             }
             if (createFullWeigth) {
-                part.FullWeight = part.FullWeight + part.Weight.ToString("D6") + "." + part.Id + "/";
+                part.FullWeight = _taxonomyService.ComputeFullWeight(part);
             }
         }
 
         protected override void Cloning(TermPart originalPart, TermPart clonePart, CloneContentContext context) {
-            clonePart.Count = originalPart.Count;
+            // arguably, copying the container/parent should be done elsewhere,
```
BenedekFarkas: _A_rguably

src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermPartHandler.cs
```diff
@@ -17,6 +18,7 @@ public class TermPartHandler : ContentHandler {
             T = NullLocalizer.Instance;
             Filters.Add(StorageFilter.For(repository));
             OnInitializing<TermPart>((context, part) => part.Selectable = true);
+
```
BenedekFarkas: Empty line should be after the closing braces.

src/Orchard.Web/Modules/Orchard.Taxonomies/Migrations.cs
```diff
@@ -108,5 +108,22 @@ public class Migrations : DataMigrationImpl {
...
+        public int UpdateFrom9() {
+            SchemaBuilder.AlterTable("TermPartRecord", table => {
+                table.AlterColumn("FullWeight", column => {
+                    column.WithType(DbType.String);
+                    column.WithLength(1023);
+                });
+                table.CreateIndex("IDX_FullWeight", "FullWeight");
+            });
+            return 10;
+        }
```
BenedekFarkas, line 134: UpdateFrom8 and UpdateFrom9 should be merged and also merged into Create (unfortunately, Create is stuck at 5 at the moment).

src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
```diff
@@ -30,7 +30,7 @@ public class TaxonomyService : ITaxonomyService {
         private readonly ShellSettings _shellSettings;
         private readonly IShellDescriptorManager _shellDescriptorManager;
 
-        private readonly HashSet<int> _processedTermParts = new HashSet<int>(); 
+        private readonly HashSet<int> _processedTermParts = new HashSet<int>();
```
BenedekFarkas: "_processedTermPartIds" shows better what it's used for.

src/Orchard.Web/Modules/Upgrade/Controllers/TaxonomyController.cs
```diff
@@ -79,22 +79,26 @@ public class TaxonomyController : Controller {
             return RedirectToAction("Index");
         }
 
+        private const int BATCH = 50;
```
BenedekFarkas: Should be "BatchSize". Btw class-level variables should be at the top before the ctor, so they are more visible.

BenedekFarkas (changes requested):
Did a quick code review and aside from a few minor details, it looks OK to me, but it's also a lot (can't really comment on ProcessPath and ComputeFullWeight).
Can you please show me some use-cases I should test?

src/Orchard.Web/Modules/Orchard.Taxonomies/Migrations.cs
MatteoPiovanelli: I think I split them apart like this because I needed to drop the index in a different method than the one where I altered the column.

src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
MatteoPiovanelli: I think I only changed the new line here, by reformatting the file, because I don't remember using this. I agree with you though

MatteoPiovanelli:
I will look into the proposed changes. 
Let me know if there is any way I can also improve the comments around the code I added to make what's being done more clear.
Regarding examples:
The best way to test this, I think is:
 - Start from a 1.10.x tenant. Create a taxonomy there, whose terms are organized in a few levels, and with different weights. Some terms on a same level in the hierarchy should have the same weight to be ordered alphabetically.
 - "move" the tenancy to the current dev branch here. Run the upgrade process there to upgrade the FullWeight. In most cases the terms will not be ordered correctly. IN some cases something may actually break (I don't remember the exact conditions to break the current code off the top of my head see the issue #8209 for a bit more context).
 - merge the code form this PR in, run the upgrade again. The terms should be correctly ordered again.

As a side, the code as is in this PR gives a significant performance boost in the cases where a taxonomy is very large. We had a taxonomy with several hundred terms, and every time we tried to look at it opening the back office page would take several minutes. The FullWeight as computed in this PR allows proper ordering and pagination, so we don't have to do it in memory.

src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TermPartDriver.cs
MatteoPiovanelli: done

src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TermPartDriver.cs
MatteoPiovanelli: done

src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermPartHandler.cs
MatteoPiovanelli: done

MatteoPiovanelli:
I updated the PR. I think that answers the points raised in the review so far.
I also cleaned up the code in the upgrade module and separated this upgrade for the terms from the old one for Orchard 1.7.

sebastienros:
Waiting for @BenedekFarkas to approve the changes, thanks.

BenedekFarkas:
Yes, sorry, I'll take care of it hopefully this week.

sebastienros:
@BenedekFarkas

BenedekFarkas (approved):

BenedekFarkas:
Sorry it took this long - tested with Matteo's instructions and looks good to me!

PR #8216, opened 2019-05-25
Extending automatic Workflows CSS discovery with caching (extends #8201)

### LINKED ISSUE #7873: Orchard.Workflows requires styles that not exists

barthamark:
Fixes #7873 and #8008 .

Replaces PR #8009 where the source repository doesn't exist anymore.

Extends PR #8201 with caching


BenedekFarkas:
@sebastienros caching added to the ResourceManifest in Orchard.Workflows as discussed during the previous community meeting.

PR #8224, opened 2019-06-18
Contents: Adding redirects to overridden routes in AdminController and ItemController

HermesSbicego-Laser:


BenedekFarkas:
@HermesSbicego-Laser I pushed a commit with the following changes:

- RouteOptions enum moved to ContentItemMetadata as ContentItemRoute.
- Added ToRouteString extension method to RouteValueDictionary.
- Custom route detection moved to a base controller, so it can be used by ItemController too, because it might also apply to Display (if the type doesn't have AutoroutePart).
- Removed the "Remove" custom route detection, because you can't redirect to POST.
- Replaced "Editor" custom route detection with "Admin" (Admin defaults to Editor if it doesn't exist).

One thing I'm not sure about: Currently the redirect happens before permission checks, so if a custom route is present, the permission check needs to happen there. I think it would be more secure if the permission check happened before redirecting (it's safer that way and in case the custom route doesn't check permissions for some reason, you can call that directly if you want to bypass permission checks - but I really feel like that only happens by mistake, in which case it's better check here before redirecting).

HermesSbicego-Laser:
I will test it next week.
Testing permission before redirect in some cases we would receive an unauthorized even if we actually have the right  custom permission. My doubt is that in some cases it could be too restrictive. It should be tested with some custom roles having access to only some contents or features, not having the edit all content or view all content ones.

HermesSbicego-Laser:
> * Replaced "Editor" custom route detection with "Admin" (Admin defaults to Editor if it doesn't exist).

I don't think to understand. Replacing "Editor" with "Admin" prevent redirection working, for example for "Taxonomies". So I Will add the case in ControllerBase switch (see commit below). With that change it seems good to me. 
**My only doubt is** that using a base class we loose the ability to have dynamic DI/IoC in the base class, cause it changes signature of the base constructor...
Adding an IDependency in the base constructor forces changes into AdminController and ItemController too, is there another way to prevent that?

BenedekFarkas:
> > * Replaced "Editor" custom route detection with "Admin" (Admin defaults to Editor if it doesn't exist).
> 
> I don't think to understand. Replacing "Editor" with "Admin" prevent redirection working, for example for "Taxonomies". So I Will add the case in ControllerBase switch (see commit below). With that change it seems good to me.

Hmm, you're right, if you explicitly check for Editor, that fallback won't work, thanks for fixing that.

> **My only doubt is** that using a base class we loose the ability to have dynamic DI/IoC in the base class, cause it changes signature of the base constructor...
> Adding an IDependency in the base constructor forces changes into AdminController and ItemController too, is there another way to prevent that?

Can you give me an example for that? Controllers don't need to be dynamic that way, so I don't see any problem with this - we do this routinely with Controllers that do similar things. But if it's an issue, we can just use a helper class instead.

I think it's ready for merge, so it will be in 1.10.4 and 1.11.

HermesSbicego-Laser:
> Can you give me an example for that? Controllers don't need to be dynamic that way, so I don't see any problem with this - we do this routinely with Controllers that do similar things. But if it's an issue, we can just use a helper class instead.
> > Mine was only a doubt. If you do it routinely, there is no problem for us.

> I think it's ready for merge, so it will be in 1.10.4 and 1.11.
> > Very good.

BenedekFarkas:
@sebastienros if you agree with the changes, then it's good to go!

src/Orchard.Web/Core/Contents/Controllers/AdminController.cs
```diff
@@ -13,20 +13,19 @@
...
 using Orchard.UI.Navigation;
 using Orchard.UI.Notify;
-using Orchard.Settings;
 using Orchard.Utility.Extensions;
-using Orchard.Localization.Services;
 
 namespace Orchard.Core.Contents.Controllers {
     [ValidateInput(false)]
-    public class AdminController : Controller, IUpdateModel {
+    public class AdminController : ContentControllerBase, IUpdateModel {
```
sebastienros, line 28: I hate inherited controllers. I much prefer reusing the code.

BenedekFarkas: What's wrong with it? Reusing the code is exactly the point of inheritance in this case and I would consider this logic specific to this feature (these two controllers).

Like I said above, this could also be in a helper class, but all other controllers in Orchard (e.g. the ones in Orchard.Blogs) are specific to a limited set of content types (usually one), so they wouldn't need to redirect when called with an Id that doesn't match the content type they are responsible for, just return a 404. And I would consider this also true for "3rd-party" controllers too, at least I haven't seen (or written) any that would try to redirect instead of 404.

PR #8226, opened 2019-06-18
Added a filter for projections over StringFields that allows returnin…

### LINKED ISSUE #8225: Return all items by projection in ContainsAll if provided value is empty
Need a filter for projections over StringFields that allows returning all items when the provided value is empty.
Like the filter ContainsAnyIfProvided i need a ContainsAllIfProvided filter
I will make a pr to do this.

LorenzoFrediani-Laser:
Added a filter for projections over StringFields that allows returning all items when the provided value is empty.

Add a ContainsAllifProvided filter to get All items if no "filter" is selected
this solve the #8225 issue

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/StringFilterForm.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/StringFilterForm.cs
@@ -53,6 +53,10 @@ namespace Orchard.Projections.FilterEditors.Forms {
                         Value = Convert.ToString(StringOperator.ContainsAnyIfProvided),
                         Text = T("Contains any word (if any is provided)").Text
                     });
+                    f._Operator.Add(new SelectListItem {
+                        Value = Convert.ToString(StringOperator.ContainsAllIfProvided),
+                        Text = T("Contains all words (if any is provided)").Text
+                    });
 
                     return f;
                 };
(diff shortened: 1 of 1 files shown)


BenedekFarkas:
I see this is a follow-up to #8213. Actually, I think this should be a completely separate option for all types of data, not just strings: A common checkbox for all FieldTypeEditors that (when checked/enabled) does not run the filter if the provided value is empty, which can be quite common if you use a Token (e.g. filter by a query string parameter). I did something similar in [Lombiq.Projections](https://github.com/Lombiq/Lombiq-Projections) (NullSafeStringFieldTypeEditor), but it only applies to strings and is a separate filter.
To have this generalized could be a nice improvement to this feature, what do you think?

@LorenzoFrediani-Laser if I understand correctly, this would also cover your case, because ContainsAll and ContainsAny already exists, so you'd have to just update the filters with this option enabled.

MatteoPiovanelli:
@BenedekFarkas I think what you are proposing makes sense for dev perhaps? I would merge this PR here in 1.10.x

BenedekFarkas:
That only works with those two specific cases. What I'm suggesting above is a simple, generic solution that can be applied to any filter (and when applied to a filter, it works regardless of what the rest of its configuration is).
I'll give it try and see if that indeed works for you too as expected.

BenedekFarkas:
Please see #8781.

BenedekFarkas:
Closing this as #8781 covers the desired functionality: Use the `ContainsAll` value for `Operator` and the new checkbox (`IgnoreFilterIfValueIsEmpty`).

Issue #8227, opened 2019-06-18
Add button to publish contentitem on customform


LorenzoFrediani-Laser:
On CustomForms add the feature to add a publish button with custom label

PR #8228, opened 2019-06-18
Add Customforms publish button option with settings for button label

LorenzoFrediani-Laser:
Add a publish button to customforms with a custom label
#8227

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.CustomForms/Migrations.cs
+++ b/src/Orchard.Web/Modules/Orchard.CustomForms/Migrations.cs
@@ -64,6 +64,12 @@ namespace Orchard.CustomForms {
 
             return 5;
         }
+        public int UpdateFrom5() {
+            SchemaBuilder.AlterTable("CustomFormPartRecord", table => table.AddColumn<bool>("SavePublishContentItem"));
+            SchemaBuilder.AlterTable("CustomFormPartRecord", table => table.AddColumn<string>("PublishButtonText"));
+            return 6;
+        }
+
 
         public void Uninstall() {
             ContentDefinitionManager.DeleteTypeDefinition("CustomForm");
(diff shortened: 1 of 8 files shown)


src/Orchard.Web/Modules/Orchard.CustomForms/Migrations.cs
```diff
@@ -64,6 +64,12 @@ public class Migrations : DataMigrationImpl {
 
             return 5;
         }
+        public int UpdateFrom5() {
+            SchemaBuilder.AlterTable("CustomFormPartRecord", table => table.AddColumn<bool>("SavePublishContentItem"));
```
sebastienros: Default value to the previous default?

src/Orchard.Web/Modules/Orchard.CustomForms/Views/EditorTemplates/Parts.CustomForm.Fields.cshtml
```diff
@@ -42,7 +47,11 @@
     @Html.TextBoxFor(m => m.CustomFormPart.SubmitButtonText, new { @class = "text large", @Value = (Model.CustomFormPart.SubmitButtonText ?? @T("Submit").Text) })
     <span class="hint">@T("The text of the submit button.")</span>
 </fieldset>
-
+<fieldset>
+    @Html.LabelFor(m => m.CustomFormPart.PublishButtonText)
+    @Html.TextBoxFor(m => m.CustomFormPart.PublishButtonText, new { @class = "text large", @Value = (Model.CustomFormPart.PublishButtonText ?? @T("Publish").Text) })
```
sebastienros, line 52: Will it be double encoded? try with some html. If it is then use a tag directly instead of the `TextBoxFor` helper, and change the other ones in this view.

LorenzoFrediani-Laser: I have tried with special char and all work fine

src/Orchard.Web/Modules/Orchard.CustomForms/Migrations.cs
LorenzoFrediani-Laser: ok, i have make a new commit

Issue #8239, opened 2019-07-11
Media Library delete folder could delete media from neighbor folders.


AndreyLuzinov:
https://github.com/OrchardCMS/Orchard/blob/89c562c50216498ecf3b0a87d7ca8832005b4cdf/src/Orchard.Web/Modules/Orchard.MediaLibrary/Services/MediaLibraryService.cs#L98

Example:
1. You have 2 folders: Folder1 and Folder11
2. Delete folder Folder1.
3. Files from Folder11 will also be removed because you cannot search sub-folders with StartsWith function.

AndreyLuzinov:
This issue related to DeleteFolder and RenameFolder methods.

sebastienros:
Nasty one, it needs to be fixed ASAP

PR #8240, opened 2019-07-12
fixed issue #8239

GiuseppeMusso-Laser:
The use of StartsWith is limited to a path ending with directory separator ("\" on Windows systems) to avoid unwanted deletion/rename in neighbor folders.


GiuseppeMusso-Laser:
The use of StartsWith is limited to a path ending with directory separator (backslash on Windows systems) to avoid unwanted deletion/rename in neighbor folders.

Issue #8238, opened 2019-07-10
Some Orchard.Core.Contents Views should use CreateRouteValues of metadata to generate the url


HermesSbicego-Laser:
In Orchard.Core.Contents, CreatableTypeList.cshtml and List.cshtml should use ContentManager.GetItemMetadata to generate the right url for creating a content.

PR #8241, opened 2019-07-12
Use CreateRouteValues for creation routes

MarcoViglione-Laser:
Issue #8238

### DIFF
--- a/src/Orchard.Web/Core/Contents/Views/Admin/CreatableTypeList.cshtml
+++ b/src/Orchard.Web/Core/Contents/Views/Admin/CreatableTypeList.cshtml
@@ -1,5 +1,26 @@
-@{ Layout.Title = T("Create New Content").ToString(); }
+@using Orchard;
+@using Orchard.ContentManagement;
+@using Orchard.Core.Contents;
+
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Core/Contents/Views/Admin/CreatableTypeList.cshtml
```diff
@@ -1,5 +1,26 @@
-@{ Layout.Title = T("Create New Content").ToString(); }
+@using Orchard;
+@using Orchard.ContentManagement;
+@using Orchard.Core.Contents;
+
+@{
+    IOrchardServices _orchardServices;
+    WorkContext.TryResolve<IOrchardServices>(out _orchardServices);
+
+    Layout.Title = T("Create New Content").ToString();
```
sebastienros: Can you set `T("Create New Content")` without the `ToString` ? There might be double encoding issues otherwise.

MarcoViglione-Laser: Done.

Issue #8252, opened 2019-07-26
Partially Hidden Checkbox on Workflow Admin Page


cmsilv:
The column header checkbox on the Workflow Definitions page of the admin panel is misaligned and partially hidden. As indicated in my screenshot below.
![image](https://user-images.githubusercontent.com/16928588/61976439-ca56af80-afb9-11e9-89d3-80870a0fc90f.png)

cmsilv:
I created pull request #8253 to resolve this issue.

PR #8253, opened 2019-07-26
Fix misaligned checkboxes on workflow admin page

cmsilv:
Remove unnecessary html class to fix misaligned check boxes on workflow admin page


sebastienros:
Fixes #8252

Issue #8262, opened 2019-08-09
MenuWidget Performances


MatteoPiovanelli:
We have been giving a look to some possible performance optimizations, and we noticed that one of the things that ends up having a significant impact in the time it takes the server to respond (assuming no-cache) is the generation of menus.

I am currently preparing a bunch of changes to propose that should make them a bit faster (roughly 10-20% faster based on what I am testing right now)

MatteoPiovanelli:
I am closing this issue because further tests in release mode show that the proposed changes have no actual impact on performances.

PR #8264, opened 2019-08-09
Issues/8262 menu widget performance

MatteoPiovanelli:
This is a code proposal related to #8262 

By the tests I made, it reduce the time to BuildDisplay for MenuWidgets by 10%-20%.

I think more testing is required to make sure nothing breaks (I don't think anything should be broken by this code) and to evaluate actual improvements in different conditions.

### DIFF
--- a/src/Orchard.Web/Core/Navigation/Drivers/MenuWidgetPartDriver.cs
+++ b/src/Orchard.Web/Core/Navigation/Drivers/MenuWidgetPartDriver.cs
@@ -41,42 +41,26 @@ namespace Orchard.Core.Navigation.Drivers {
         }
 
         protected override DriverResult Display(MenuWidgetPart part, string displayType, dynamic shapeHelper) {
-            return ContentShape( "Parts_MenuWidget", () => {
-                var menu = _menuService.GetMenu(part.MenuContentItemId);
-                
-                if(menu == null) {
-                    return null;
-                }
+            if (part.Breadcrumb) {
+                return ContentShape("Parts_MenuWidget", () => {
+                    var menu = _menuService.GetMenu(part.MenuContentItemId);
 
-                var menuName = menu.As<TitlePart>().Title.HtmlClassify();
-                var currentCulture = _workContextAccessor.GetContext().CurrentCulture;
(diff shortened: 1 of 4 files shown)


src/Orchard.Web/Core/Navigation/Drivers/MenuWidgetPartDriver.cs
```diff
@@ -41,42 +41,29 @@ public class MenuWidgetPartDriver : ContentPartDriver<MenuWidgetPart> {
...
-                    menuItems = menuItems.Where(x => !x.Content.Has<BreadcrumbMenuItemPart>());
-                }
-
-                var localized = new List<MenuItem>();
-                foreach(var menuItem in menuItems) {
-                    // if there is no associated content, it as culture neutral
-                    if(menuItem.Content == null) {
-                        localized.Add(menuItem);
-                    }
+            var menu = _menuService.GetMenu(part.MenuContentItemId);
```
sebastienros, line 44: Having this code oustide of the lambda will make it run more often than necessary. You might not see it while you are using pages that will render the menu or a single content item.

MatteoPiovanelli: Yes, I see it will run more often than strictly necessary. In my tests however the code in the lambda was the code taking the longest. Splitting it up in the two lambdas for the two cases (breadcrumb and "normal" menu) and moving these common parts out of the lambdas entirely gave that 10%-20% improvement I mentioned. 
While the GetMenu is cached already (as it is calling a ContentManager.Get), I can see the value in caching the remaining operations here for a request.

MatteoPiovanelli: Wait, I was wrong. I was seeing that the lambda was slower than it should because it wasn't being optimized by the compiler (I was testing in Debug mode). Let me get back on this.

MatteoPiovanelli:
After further testing in release mode, the proposed changes seem to yield no gain in performances whatsoever. I am closing this PR as useless.

PR #8275, opened 2019-09-03
Projections: "RewriteOutput" bool property of a Layout Property migrated to a tokenized condition text, fixes #6928

### LINKED ISSUE #6928: To have a rule in queries at property level
It will be very interesting to have a rule field in queries at property level to don't show data for some content items.
For example I show results of a query in different rows. Through a custom property the user can see a button for delete each item. How I show them contents that they can own or not I only want to show delete button for those contents that they own.

I have implemented it. Here it is an screenshot showing how I use it:
image

I'm sending a pull request

BenedekFarkas:
This is a modified version of #6929 (that fixes #6928) that migrates the RewriteOutput (bool) property of a Layout Property to RewriteOutputCondition (tokenized condition text).


sebastienros:
GTG

Issue #8286, opened 2019-11-20
Setting "Try to translate" == false not effective on localized taxonomy field


GiuseppeMusso-Laser:
Hi all.
I experienced a problem with localized taxonomies in dev branch.
In a custom module, I need to translate a content item **programmatically**, without opening it in edit on the web page.
My original content item is localized and it has a localized taxonomy field with setting "Try to translate the selected terms" **unchecked**.
When I translate my original content item, my cloned content item keeps all the terms selected in the original content item, despite its "Try to translate" setting.

I found a simple solution in the LocalizedTaxonomyFieldHandler and I am going to make a PR on it.
When the BuildEditor is first created just after cloning and TryToLocalize is false, I simply remove all terms associated with the taxonomy field.

Does anyone see some problems in doing that?

sebastienros:
> "Try to translate the selected terms" unchecked, my cloned content item keeps all the terms selected in the original content item

It sounds like this is exactly what should happen with this unchecked. And when it's checked you then ask the new taxonomy to point to each term's translation instead.

If you expect the cloned item to have no terms at all, then it would be another options in the settings.
Unless I didn't understand your issue.

GiuseppeMusso-Laser:
I get your point.
I'll review my code because I think I misunderstood localization settings of taxonomy fields.
Thanks!

PR #8287, opened 2019-11-20
Issue/8286 apply settings in taxonomy filed translation

GiuseppeMusso-Laser:
This is my proposal to fix issue #8286.
When the BuildEdito is first created just after cloning and TryToLocalize is false, I simply remove all terms associated with the taxonomy field.


GiuseppeMusso-Laser:
Sorry, I used the wrong branch.

PR #8289, opened 2019-11-21
Issue/8284

### LINKED ISSUE #8284: Performance Issue: GetSiteSettings
Tagging @sebastienros @HermesSbicego-Laser @BenedekFarkas 

We have been doing some performance work on our Orchard installations lately.
As part of that I've been profiling a copy of one of our tenants that I have on my dev box.
The process I was profiling was the repeated access (a few hundred times) to a cached page (a version from my local tenant of https://www.anfiteatromorenicoivrea.it/erbaluce-di-caluso).

Roughly 48% CPU time is used in Orchard.Core.Settings.Service.SiteService.GetSiteSettings() on this line:

    return _contentManager.Get<ISite>(siteId, VersionOptions.Published);

That is of course a `WorkContext.CurrentSite` being called.

Going deeper, I found that the line where that gets called most times "uses" about 51% CPU time and is the following:

    var sslSettings = _wca.GetContext().CurrentSite.As<SslSettingsPart>();

From Orchard.SecureSocketsLayer.Services.StrictTransportSecurityMiddlewareProvider.GetOwinMiddlewares().

My first instinct here would be to cache that `CurrentSite`, but I would not feel 100% positive on when to evict that cache because of updates to any setting. A simpler fix would be to cache the `SslSettingsPart`: we would easily know to evict that with an handler on that ContentPart.

Would anyone validate that these results are not artifacts of how my box and my tenant are set up?
Do you see any drawback to the approach I propose as a possible solution?

On further testing on pages that are not cached, the "contribuition" of the call to `GetSiteSettings()` to the used CPU seems to drop significantly, to about 1%.

MatteoPiovanelli:
Fixes #8284

Tested locally Starting off Orchard 1.10.x as well as on our custom configuration.

The changes to `NavigationManager` improve performances on menu-heavy frontends, as well as a little bit on the backoffice (because they also cache some of the operations that are done for the admin menu there)

The other changes can really only be measured when serving cached content, as those operations are not that relevant compared to those required to generate frontend views. When serving stuff out of cache however those changes combined gave around 30% quicker responses on my a couple different test configurations.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Handlers/SslSettingsPartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Handlers/SslSettingsPartHandler.cs
@@ -3,12 +3,24 @@ using Orchard.Data;
 using Orchard.ContentManagement.Handlers;
 using Orchard.Localization;
 using Orchard.SecureSocketsLayer.Models;
+using Orchard.Caching;
 
 namespace Orchard.SecureSocketsLayer.Handlers {
     public class SslSettingsPartHandler : ContentHandler {
-        public SslSettingsPartHandler() {
+        private readonly ISignals _signals;
+
+        public SslSettingsPartHandler(ISignals signals) {
+
+            _signals = signals;
+
             T = NullLocalizer.Instance;
+
             Filters.Add(new ActivatingFilter<SslSettingsPart>("Site"));
+
+            // Evict cached content when updated, removed or destroyed.
+            OnPublished<SslSettingsPart>((context, part) => Invalidate(part));
+            OnRemoved<SslSettingsPart>((context, part) => Invalidate(part));
+            OnDestroyed<SslSettingsPart>((context, part) => Invalidate(part));
         }
 
         public Localizer T { get; set; }
@@ -21,6 +33,11 @@ namespace Orchard.SecureSocketsLayer.Handlers {
                 Id = "Ssl",
                 Position = "2"
             });
+        }
+
+        private void Invalidate(SslSettingsPart content) {
+            _signals.Trigger($"SslSettingsPart_{content.Id}");
+            _signals.Trigger("SslSettingsPart_EvictAll");
         }
     }
 }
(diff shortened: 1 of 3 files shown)


sebastienros:
LGTM

@BenedekFarkas are you ok with it? If you don't answer I will merge myself.

MatteoPiovanelli:
I was thinking about this PR earlier under the shower and I think I am not correctly handling the dashboard/backoffice menu. I should add an evict on enable/disable of a feature, but I am not sure that would be enough.
Better to not cache the backoffice menu, by adding an AdminFilter check

MatteoPiovanelli:
Regarding admin side menus:
I still don't feel like it's trivial to cover all conditions, but:
 - On changing the Listable property of a ContentType, the menu gets updated correctly.
 - On Creating a new ContentType, the menu gets updated correctly
 - (basically, whenever we post for the creation or edit of a ContentType, as far as I can tell, that menu cache gets evicted correctly)
 - When I Enable or Disable a feature, the cache gets evicted, regardless of the fact that the feature may or may not change the menu.
That's all "out of the box", without having to explicitly write anything.

However, AdminMenuPart does not cause an update of that value. That means there is at least this one case that is not handled. I have to assume there will be more.
Hence I updated the PR to prevent caching of admin menus.

strongus:
I merged this fix into my project it dramatically improved the speed of the Admin side of the site. Thanks so much for this! The slowness was driving me nuts.

MatteoPiovanelli:
I found some issues with the approach we took for caching frontend menu items.

In summary:
`GetSources(string)` and `GetSource(IContent)`:
The results of these two methods depend on their input parameters as well as a bunch of injected IDependencies. Even though we are sure that the cache would be cleared if the IEnumerable of the IDependencies changed, we cannot be sure that none of those depends on settings or properties that would not reset/evict the cache upon changing. As a consequence, we cannot simply and safely cache the results of these methods. That would mean that a cache for a menu would have to monitor all possible changes to all its items. Since there is no easy way to do that, we decided to rollback the changes we were proposing here for Navigation.

The other changes (caching of SslSettingPart and SiteSelector) still stand.

src/Orchard.Web/Modules/Orchard.SecureSocketsLayer/Handlers/SslSettingsPartHandler.cs
```diff
@@ -22,5 +34,10 @@ public class SslSettingsPartHandler : ContentHandler {
                 Position = "2"
             });
         }
+
+        private void Invalidate(SslSettingsPart content) {
+            _signals.Trigger($"SslSettingsPart_{content.Id}");
```
sebastienros, line 39: I will keep it but I think it's not necessary. And I don't see it used anywhere else either.

PR #8292, opened 2019-11-25
ReCaptchaPartDriver always returns a shape

### LINKED ISSUE #8291: ReCaptchaPartDriver does not return ShapeType
`Orchard.AntiSpam.Models.ReCaptcha` works by using the `Editor` methods (GET and POST) of its driver to display the Recaptcha and invoke Google's validation when the content is posted.

The `Editor(ReCaptchaPart part, dynamic shapeHelper)` is supposed to not show the ReCaptcha when the content is being edited on the BackOffice. However, the way the `Editor(ReCaptchaPart part, dynamic shapeHelper)` is written right now, it will return `null` for the backoffice, meaning that we cannot change the SHape's placement in the "Edit Placement" section of managing a "ContentType" that has a `ReCaptchaPart`.

It's an easy fix to move the test in the shape's delegate, so that the `Editor` method always returns a shape (even when that shape will turn out to be empty), and any supporting feature that wants to know ShapeTypes will then have the information it requires.

I am sure there are other places where the `Editor` method behaves like that, and in most of them the fix will be easy as easy as the one I am proposing here. I am not looking for all of them right now, but I am proposing to fix `ReCaptchaPart` because it came up in a recent upgrade we have done to some of our tenants.

MatteoPiovanelli:
Fixes #8291 
We changed when ReCaptchaPartDriver verifies whether it's being executed on the Admin side, moving the test from the driver's execution to the shape's, so that the ShapeType is available to other features and services that process the driver's output (e.g. for placement)

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.AntiSpam/Drivers/ReCaptchaPartDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.AntiSpam/Drivers/ReCaptchaPartDriver.cs
@@ -35,14 +35,16 @@ namespace Orchard.AntiSpam.Drivers {
         public Localizer T { get; set; }
         public ILogger Logger { get; set; }
         protected override DriverResult Editor(ReCaptchaPart part, dynamic shapeHelper) {
-            var workContext = _workContextAccessor.GetContext();
-
-            // don't display the part in the admin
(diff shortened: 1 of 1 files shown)


MatteoPiovanelli:
I will update the PR to also fix #7747

PR #8296, opened 2019-12-01
Upgrade to net 461

aaronamm:
Upgrade all projects to .NET 4.6.1
Please refer to the original PR https://github.com/OrchardCMS/Orchard/pull/8126 that I have been closed because I can't pushed to it.

Main changes:
- Upgrade to .NET 4.6.1 which can be use with .NET Standard library 2.0
- For custom class library can target .NET standard that can be use in both .NET Core and .NET Framework.
- PowerShell script tool to upgrade the .NET framework to all projects

image

I know that we should update NuGet package to the latest version after we have upgraded .NET Framework version.
I have tried upgrade all NuGet package and there are many breaking changes.

These are some packages that I have tried to rollback to minimize coding chanages

- Nhibernate 4.0.1.4000
- Fluent Nhibernate 2.0.3.0
- Lesi.Collections4.0.1.4000 
- NUnit 2.5.10.11092
- Autofac 3.5.2
- Autofac.Configuration 3.3.0
- Castle.Core 3.3.1

However, there some other packages e.g. Azure library also need to be rollback, etc. to make all projects compile without any errors.

Therefore, I think we can leave packages.config have targeFramework="net452" for now.
Update many packages at once and causes many breaking change seem risky and need a lot of regression tests.

FYI, after this PR get merged. I am happy to help upgrade NHibernate to the later version.
Thanks.

### DIFF
--- a/src/Orchard.Azure.Tests/Orchard.Azure.Tests.csproj
+++ b/src/Orchard.Azure.Tests/Orchard.Azure.Tests.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Orchard.Azure.Tests</RootNamespace>
     <AssemblyName>Orchard.Azure.Tests</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <FileUpgradeFlags>
     </FileUpgradeFlags>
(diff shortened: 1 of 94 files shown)


aaronamm:
@sebastienros 
You have mentioned that 4.6.1 approved, please could you check this PR to make sure it is ready to merged.

@BenedekFarkas I have updated the PowerShell script to process XML of csproj file.
However, I chose LINQ to XML not XPath because it is mentioned in this document.
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/comparison-of-xpath-and-linq-to-xml#performance-differences.
If we have more complex queries later, we can get benefit from LINQ to XML.

sebastienros:
@BenedekFarkas if you think the changes are correct, I am ok to merge.
Does it also mean we can update some dependencies to newer versions?

aaronamm:
@sebastienros  I think we can update nuget packages to the newer version but the main benefit is that we can now start using a library that only target .NET Standard.

For some libraries, we can't update them to the latest because of the breaking changes.
I am thinking to help on updating NHibernate to the latest version after we get this PR merged.

Thanks.

sebastienros:
@BenedekFarkas

aaronamm:
@BenedekFarkas  👍

sebastienros:
Based on this issue, the web.config files should have been updated too. https://github.com/OrchardCMS/Orchard/issues/8325

PR #8298, opened 2019-12-02
removed useless separators in EnumerationField

GiuseppeMusso-Laser:
Removed useless separators when SelectedValues sets Value property in EnumerationField because they can generate incorrect behaviour durion field edit.
Scenario: let us have an EnumerationField which is rendered by a drop-down list.
If you set the selected value of this field programmatically by means of SelectedValues property
to the value "MyValue", the Value property is set to ";MyValue;". 
When you edit this field, the drop-down list is not set to "MyValue" because it gets ";MyValue;" as the selected value and ";MyValue;" does not match any of the options in the drop-down list.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Fields/Fields/EnumerationField.cs
+++ b/src/Orchard.Web/Modules/Orchard.Fields/Fields/EnumerationField.cs
@@ -26,7 +26,7 @@ namespace Orchard.Fields.Fields {
                     Value = String.Empty;
                 }
                 else {
-                    Value = Separator + string.Join(Separator.ToString(), value) + Separator;
+                    Value = string.Join(Separator.ToString(), value);
                 }
             }
         }


sebastienros:
I don't think this is the correct fix. The semi-colons were placed intentionally such that we can do some queries in SQL for instance, and find strings like `;foo;`, even if it's the single value.

I found multiple issues however:
1- https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Fields/Views/EditorTemplates/Fields/Enumeration.Edit.cshtml#L17

This should not use `.Value` as it can contain `;` (which you found), and it should use `.SelectedValues.FirstOrDefault()` instead. Which will extract the `;` from the stored value.

2-
https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.Fields/Drivers/EnumerationFieldDriver.cs#L42

In this driver, the default value will not contain the semi-colon, so we need to use .SelectedValues also. If it works, it's just luck.

BenedekFarkas:
See #8789 instead.

PR #8301, opened 2019-12-11
Making messages of register.cshtml consistent with the User settings

### LINKED ISSUE #8300: Make some messages of the shape register.cshtml consistent with the User settings
In the dev branch there is the ability to configure password strength using 3 new settings:
```
EnablePasswordUppercaseRequirement
EnablePasswordLowercaseRequirement
EnablePasswordNumberRequirement
```
Filling the registration form, a new User does not know that the password should contain some uppercase/lowercase letters or numbers because there is no message informing him about the requirement.

So my suggestion is to add a message informing the user about the password requirements, as well as for the password length message.
The question is: could there be a security issue related to the change?

HermesSbicego-Laser:
Fixes #8300

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Views/Register.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Users/Views/Register.cshtml
@@ -14,8 +14,7 @@
         additionalRequirementsMessage.Add(T("one special character").Text);
     }
     if (additionalRequirementsMessage.Count() > 0) {
-        requirementsMessage += T(" and must contain at least ");
-        requirementsMessage += string.Join(", ", additionalRequirementsMessage);
+        requirementsMessage = T("{0} and must contain at least {1}", requirementsMessage, string.Join(", ", additionalRequirementsMessage)).Text;
     }
     requirementsMessage += ".";
 }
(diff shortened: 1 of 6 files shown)


src/Orchard.Web/Modules/Orchard.Users/Views/Register.cshtml
```diff
@@ -1,32 +1,53 @@
-﻿<h1>@Html.TitleForPage(T("Create a New Account").ToString())</h1>
+﻿@{
+    var requirementsMessage = T.Plural("The password can't be empty", "Passwords are required to be a minimum of {0} characters in length", (int)ViewData["PasswordLength"]).Text;
+    var additionalRequirementsMessage = new List<string>();
+    if ((bool)ViewData["LowercaseRequirement"]) {
+        additionalRequirementsMessage.Add(T("one lowercase letter").Text);
```
sebastienros, line 5: Can you ensure the Text property is not html encoded, otherwise it would result in double encoding in the output. To try it just replace some text with html, it should be encoded only once.

sebastienros:
LGTM as long as there is no double encoding

src/Orchard.Web/Modules/Orchard.Users/Views/Register.cshtml
HermesSbicego-Laser: I can confirm that T("").Text does not html encode the string.
image
image

HermesSbicego-Laser: Looking better into the code I noticed we need to have similar changes somewhere else.
I will integrate this PR soon.

HermesSbicego-Laser: Done with 2 new commits

HermesSbicego-Laser:
It should be good to be merged in dev.

PR #8303, opened 2019-12-19
Fix/8302 manage pagination blogpostarchive

### LINKED ISSUE #8302: BlogPostController does not support pagination

ElenaRepository:
fixes #8302


sebastienros:
My only concern is that it's a breaking change on the service interface. Either target dev or try not to add this to the interface for now.

ElenaRepository:
I close this pr because it has already been merged into dev

PR #8308, opened 2020-01-17
Remove ArchiveLaterTask if content item is deleted

andrewcartwright1:
Replication Steps:
- Enable Archive Later module
- Add Archive later part to a content definiton (e.g Page)
- Create a new content (e.g Page) and set a archive later date
- Examine the [dbo].[Scheduling_ScheduledTaskRecord] table and observe that task is there.
- Delete the created content
- Examine the [dbo].[Scheduling_ScheduledTaskRecord] table and observe that task is still there.

By not removing the task when the content is deleted, when the task runs an exception is thrown inside `UnpublishingTaskHandler Process()` and the task will persist in the database causing an endless cycle of errors.

Noticed that this is already being done for the PublishLaterPart: https://github.com/OrchardCMS/Orchard/blob/1.10.x/src/Orchard.Web/Modules/Orchard.PublishLater/Handlers/PublishLaterPartHandler.cs#L17


MatteoPiovanelli (changes requested):
I agree with this change, but I would argue the task should also be removed `OnDestroying` of the content

PR #8310, opened 2020-01-23
Tabs and Cards to better organize parts and fields

### LINKED ISSUE #8309: Placement with tabs: issues and upgrades
(I am collecting here a few topics since we already have a PR ready that covers them all and it's quicker to have a single issue)

Some issues and limits of shape placement:
- It's possible to define tabs in placements. However, the way tabs are "written" to the html output is hardcoded. It should support alternates.
- Tab names can't have whitespace in them: it won't be displayed as such. Alternates allow fixing that as needed, without having to worry about finding a general solution right now that also will not break anyone's code and hacks.
- Within a tab, there is no way to further "group" shapes. Bootstrap does that with cards. Cards should allow alternates as well, and in general behave similarly to tabs (in terms of setting them up for and using them in pages and editor views)

In our PR we will add cards to placement, handled in a similar way as tabs. Both tabs and cards are "generated" using shapes, so we can create alternates of them.
We update the "Edit Placement" functionality of ContentTypes to handle both for editor views.

(cc: @HermesSbicego-Laser who has come up with the feature and developed it)

MatteoPiovanelli:
- Introduce Card concept within Placement.info that allows to organize parts and fields in cards/groups
- Creates shapes for Tab and Card in order to allow alternates
- Update "Edit Placement" feature to use cards.

When defined in a placement "file", tabs still use the '#' separator. Cards use the '%' separator.

This fixes #8309

### DIFF
--- a/src/Orchard.Web/Core/Shapes/CoreShapes.cs
+++ b/src/Orchard.Web/Core/Shapes/CoreShapes.cs
@@ -31,11 +31,11 @@ namespace Orchard.Core.Shapes {
         private readonly IResourceFileHashProvider _resourceFileHashProvider;
 
         public CoreShapes(
-            Work<WorkContext> workContext, 
+            Work<WorkContext> workContext,
             Work<IResourceManager> resourceManager,
             Work<IHttpContextAccessor> httpContextAccessor,
             Work<IShapeFactory> shapeFactory,
-            IResourceFileHashProvider resourceHashProvider
+            IResourceFileHashProvider resourceHashProvider 
             ) {
             _workContext = workContext;
             _resourceManager = resourceManager;
@@ -47,7 +47,7 @@ namespace Orchard.Core.Shapes {
         }
 
         public Localizer T { get; set; }
-        public dynamic New { get { return _shapeFactory.Value;  } }
+        public dynamic New { get { return _shapeFactory.Value; } }
 
         public void Discover(ShapeTableBuilder builder) {
             // the root page shape named 'Layout' is wrapped with 'Document'
@@ -57,7 +57,7 @@ namespace Orchard.Core.Shapes {
                 .OnCreating(creating => creating.Create = () => new ZoneHolding(() => creating.New.Zone()))
                 .OnCreated(created => {
                     var layout = created.Shape;
-                    
+
                     layout.Head = created.New.DocumentZone(ZoneName: "Head");
                     layout.Body = created.New.DocumentZone(ZoneName: "Body");
                     layout.Tail = created.New.DocumentZone(ZoneName: "Tail");
@@ -112,7 +112,7 @@ namespace Orchard.Core.Shapes {
                     string contentType = null;
                     int level = menuItem.Level;
                     if (menuItem.Content != null) {
-                        contentType = ((IContent) menuItem.Content).ContentItem.ContentType;
+                        contentType = ((IContent)menuItem.Content).ContentItem.ContentType;
                     }
 
                     menuItem.Metadata.Alternates.Add("MenuItemLink__level__" + level);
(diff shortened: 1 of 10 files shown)


HermesSbicego-Laser:
after-for-github

src/Orchard.Web/Core/Title/Placement.info
```diff
@@ -7,7 +7,7 @@
     -->
     <Place Parts_Title_Edit="Content:before.5"/>
     <Match DisplayType="Detail">
-        <Place Parts_Title="Header:5"/>
```
sebastienros: Why change the default zone of the Title ? Not related to this PR I think, and quite a breaking change for themes.

src/Orchard.Web/Core/Shapes/CoreShapes.cs
```diff
@@ -811,6 +814,54 @@ string PagerId
         public void DefinitionTemplate(HtmlHelper Html, TextWriter Output, string TemplateName, object Model, string Prefix) {
             RenderInternal(Html, Output, "DefinitionTemplates/" + TemplateName, Model, Prefix);
         }
+        [Shape]
+        public void Tab(dynamic Display, IGrouping<string, dynamic> Tab, TextWriter Output) {
```
sebastienros, line 818: And now we can customize it, good (still necessity of an admin theme)

src/Orchard.Web/Modules/Orchard.ContentTypes/Controllers/AdminController.cs
```diff
@@ -142,17 +142,42 @@ public class AdminController : Controller, IUpdateModel {
...
-                .GroupBy(x => x.PlacementInfo.GetTab())
-                .ToDictionary(x => x.Key, y => y.ToList());
-
-            var content = grouped.ContainsKey("") ? grouped[""] : new List<DriverResultPlacement>();
-            var listPlacements = grouped.Values.SelectMany(e => e).ToList();
-
-            grouped.Remove("");
+                .GroupBy(g => g.PlacementInfo.GetTab() + "%" + g.PlacementInfo.GetCard())
+                .Select(x =>
+                    new Card {
```
sebastienros, line 164: This code should be next to `GetTab()`, not in the controller. And I have the feeling that this code is more complex than it should be.

sebastienros (commented):
I think it's a good idea. The only missing concept would be to set the width of a card with a grid system. Which could be defined in a custom property to store the value for each named card. Maybe from the editor and the placement file. But that can be done in a separate PR.

Do you want to share a demo at a meeting?

src/Orchard.Web/Core/Shapes/CoreShapes.cs
MatteoPiovanelli: We have a couple of those here https://github.com/LaserSrl/Laser.Orchard.Platform/tree/master/src/Themes . Not sure that they would cover all possible uses in the community, because they are developed for our own custom needs.

src/Orchard.Web/Core/Title/Placement.info
MatteoPiovanelli: oops... We were testing things and forgot to revert this. doing it.

src/Orchard.Web/Modules/Orchard.ContentTypes/Controllers/AdminController.cs
MatteoPiovanelli: I my latest commit I added a bunch of comments to clarify what we are doing there. We The code can be simplified slightly (e.g. by removing the `Select` statements and have the `GroupBy` statements return the objects) but For now I kept it like this for clarity. Could you clarify the changes you would like here?

MatteoPiovanelli:
Can we demo this quickly on Tuesday 28th?

MatteoPiovanelli:
we actually found a small js bug on deleting. Why don't I fix it and demo the thing on Thursday?

sebastienros:
Whenever you want. Remember we are now using Teams, the link is on the main site.

MatteoPiovanelli:
looking for it

PR #8311, opened 2020-01-23
Tweak TermsPartHandler to skip unpublished terms

Lawyerson:
When a term gets unpublished, rather than deleted, TermsPartHandler will spit out a  "given key was not present in the dictionary" error any time a linked content item requests its terms. Terms can be unpublished when their content type is listable.


MatteoPiovanelli (approved):
Looks good to me

Issue #8317, opened 2020-02-04
Add the approve/disable button to the user edit


ElenaRepository:
We need to have the approve / disable button in the user edit shape.

I thought it could:
- create a driver, as for the publish later, that clicks on approves triggers the approval
- modify the edit action in the admin controller to add the approval button

Which do you think is the most optimal solution? Do you have other ideas?

PR #8321, opened 2020-02-11
Add button manage approve disable in edit user

ElenaRepository:
#8317 

Added approve / disable in user edit to conveniently manage functionality directly during user modification


src/Orchard.Web/Modules/Orchard.Users/Drivers/UserApprovePartDriver.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Laser.Orchard.StartupConfig.ApproveUserExtension.Services;
```
sebastienros: Not yet a Laser project

ElenaRepository: Sorry I was wrong, I made changes to the files.

PR #8329, opened 2020-02-24
It makes images recognizable in the admin even if they are white over a transparent background

### LINKED ISSUE #8328: MediaLibrary: PNG not visible if white over transparent bg
If we upload a white png with a transparent background, it results to be NOT 
recognizable in Media Library and in the Media Library Picker Field.
download

HermesSbicego-Laser:
fixes #8328, adding in the admin a white/grey squares background to containers of media of type "image"
Before
before-download

After
download


sebastienros:
Why not use a static file instead so it doesn't have to be duplicated in the css?
It's not a huge issue, just a suggestion.

HermesSbicego-Laser:
Intead of base64? It was just for consistency with other images used in that css... I will change it...

sebastienros:
No it's ok then

sebastienros:
No, it's ok then

Issue #8318, opened 2020-02-04
Problem with scope in translations:


diegomarguerettaz:
If I want to translate the string "hello" in Italian and such string is contained in the class A, I add in the files . po the translation:
msgctxt "...A"
msgid "Hello"
msgstr "Ciao".
Now, if I have class B inheriting from A, when I try to call the translation from B, orchard can’t find the translation as it looks for it with scope B and not with scope A. It would be good thing if, when it doesn’t find the translation with scope B, try searching for such a translation with scope A. An example in Orchard.Framework can be found in Classes DefaultOrchardCommandHandler and FeatureCommands.

To make this fix you should engage:
- In the LocalizationModule, add to the dictionary of localizers both the one with the scope of the concrete class, and the one with the scope of the abstract class.
- From the concrete scope, one must be able to go back to the abstract one, perhaps saving the thing somewhere (additional dictionary of cache), or putting it directly in the objects of class Text.
- LocaliLocaliLocaliManager Manager Manager has parentalscope information to attempt a fallback if it is not located... "Final" fallbacks must remain as they are now (empty scope and then do not translate)

@MatteoPiovanelli-Laser

MatteoPiovanelli:
**To test this issue**
 - Build the solution in debug mode.
 - Within Visual Studio, set a debug point on https://github.com/OrchardCMS/Orchard/blob/ffaa98724af14bc32860519f05b4a10383b28b2c/src/Orchard/Commands/DefaultOrchardCommandHandler.cs#L18
 - Set another debug point on https://github.com/OrchardCMS/Orchard/blob/ffaa98724af14bc32860519f05b4a10383b28b2c/src/Orchard.Web/Modules/Orchard.Modules/Commands/FeatureCommands.cs#L62
 - Open the command line utility Orchard.exe.
 - From Visual Studio, attach the debugger to the process for Orchard.exe.
 - In the command line tool, execute the command `feature enable Orchard.CodeGeneration`. You may use any feature name.
 - From both debug points, verify (watch) the scope that is set within the Localizer `T`.

**What we have**
In both cases the scope will be the same, ending with "FeatureCommands".

The net result of this issue is that to see the correct localization for the strings that are written in `DefaultOrchardCommandHandler.cs` we would have to add them to po files once for every concrete implementation of that abstract class.

**Likely cause**
[...]

PR #8333, opened 2020-03-12
8318 localization multiple scopes

diegomarguerettaz:
Issue #8318: 
Managing multiple scopes in localization concurrent dictionary. Target dev beacause i changed the signature:
ILocalizedStringManager
System.Tuple<string, string> GetLocalizedString(IEnumerable<string> scopes, string text, string cultureName);


sebastienros:
Can you rebased your PR on dev also, otherwise the changeset contains too many things

Issue #7928, opened 2018-01-04
Passing html attributes with an underscore to ItemDisplayLink do not get converted into dashes


Hazzamanic:
Generally in razor you can replace the dash in an html attribute with an underscore and it will be converted into a dash when rendered. However, when using the Orchard helper ItemDisplayLink, this does not work. 

`@Html.ItemDisplayLink("View on main site", content, new { data_amp_donotadjust = "" })`

It would seem the following method is to blame, where the htmlAttributes parameter is wrapped in a RouteValueDictionary as it is passed to the ActionLink helper from MVC.

```
public static MvcHtmlString ItemDisplayLink(this HtmlHelper html, string linkText, IContent content, object htmlAttributes = null) {
            var metadata = content.ContentItem.ContentManager.GetItemMetadata(content);
            if (metadata.DisplayRouteValues == null)
                return null;

            return html.ActionLink(
                NonNullOrEmpty(linkText, metadata.DisplayText, "view"),
                Convert.ToString(metadata.DisplayRouteValues["action"]),
                metadata.DisplayRouteValues,
                new RouteValueDictionary(htmlAttributes));
        }
```

Removing the RouteValueDictionary does indeed solve the problem but it also then adds unwanted parameters onto the url:

`<a data-amp-donotadjust="" href="/Contents/Item/Display/83?Count=4&amp;Keys=System.Collections.Generic.Dictionary%602%2BKeyCollection%5BSystem.String%2CSystem.Object%5D&amp;Values=System.Collections.Generic.Dictionary%602%2BValueCollection%5BSystem.String%2CSystem.Object%5D">View on main site</a>`

sebastienros:
I think it has already been reported, might be wrong. This should be fixed IMO.

Hazzamanic:
So it doesn't work when I change 

> new RouteValueDictionary(htmlAttributes) 

to just htmlAttributes because it changes the method signature making metadata.DisplayRouteValues an object, hence all the weird stuff on the url. We could convert the htmlAttributes object into a Dictionary ourselves and do the _ to - conversion but seems a bit silly since .net should do that for us. Happy to do that or another fix if you have any ideas?

sebastienros:
> since .NET should do that for us

I will ask what component in ASP.NET is currently doing it

MatteoPiovanelli:
hello

`new RouteValueDictionary(htmlAttributes)`

should be replaced with a call to the static method:
[...]

PR #8337, opened 2020-03-25
Issue #7928 Failure to convert ItemDisplayLink from underscore to dashes

ElenaRepository:
#7928

we replaced the code as suggested by @MatteoPiovanelli-Laser  and tested it in this way the parameter behave correctly


sebastienros:
Beautiful, thanks

PR #8349, opened 2020-04-15
Fixes errors during Indexing

### LINKED ISSUE #8348: Errors during indexing
We found in our logs 2 errors during the Indexing process (Module Orchard.Indexing):
1. NRE on indexing DateTimeField
2. NHibernate.ObjectNotFoundException: No row with the given identifier exists[Orchard.ContentManagement.Records.ContentItemRecord#148] during Rebuild an index.

We will work on it and we will propose a PR

HermesSbicego-Laser:
fixes #8348 

IndexingTaskExecutor:
 - wrapped with a try/catch the ContentManager.Get that throws an error when ContentItemVersionRecord has not a ContentItemRecord reference (sometimes happens on Database)

DescribeMembersContext:
 - Adds a new constructor having a ContentPartFieldDefinition parameter, needed because DateTimeField.DateTime property try to access its own PartFieldDefinition.Settings that is null when called by Orchard.ContentManagement.Handlers.Enumerate called by InfosetFieldIndexingHandler.OnIndexing (see below)

InfosetFieldIndexingHandler:
 - Creates a DescribeMembersContext having the ContentPartFieldDefinition defined

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
+++ b/src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
@@ -20,8 +20,7 @@ namespace Orchard.Indexing.Services {
     /// This class is synchronized using a lock file as both command line and web workers can potentially use it,
     /// and singleton locks would not be shared accross those two.
     /// </remarks>
-    public class IndexingTaskExecutor : IIndexingTaskExecutor, IIndexStatisticsProvider
-    {
+    public class IndexingTaskExecutor : IIndexingTaskExecutor, IIndexStatisticsProvider {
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.Indexing/Services/IndexingTaskExecutor.cs
```diff
@@ -160,13 +159,22 @@ public class IndexingTaskExecutor : IIndexingTaskExecutor, IIndexStatisticsProvi
                         .OrderBy(versionRecord => versionRecord.Id)
                         .Take(ContentItemsPerLoop)
                         .ToList()
+                        .Select(versionRecord => {
```
sebastienros, line 162: The Select should be before the ToList

Issue #8354, opened 2020-04-17
Performance issues


HermesSbicego-Laser:
We are experiencing performance issues using admin panel with more than 2 users.
SQL profiler tells that `select shelldescr0_.Id as Id38_, shelldescr0_.SerialNumber as SerialNu2_38_ from Settings_ShellDescriptorRecord shelldescr0_ where shelldescr0_.Id is not null` is fired hundreds of times.
We have a fix ready for  that, but we will go deeper in order to find a way to have orchard performing better.

MatteoPiovanelli:
Expanding on this, even with the improvements from #8355, the database was being hit for the lists of ShellFeatureRecord and ShellParameterRecord on every request. Running sql traces, those queries were noticeable. So we wrote and tested a version of ShellDescriptorManager that caches all that. This will target dev rather than 1.10.x because we used ILockingProvider for an application lock around the part of code that handles updating the records and evicting the cache.

MatteoPiovanelli:
closing this (evennthough it is part of a "series" of issues and PR that is still ongoing) because the specific thing discussed here has been dealt with

PR #8355, opened 2020-04-17
Prevent Settings_ShellDescriptorRecord to be hit many times

HermesSbicego-Laser:
#8354 
- prevent Settings_ShellDescriptorRecord to be hit many times during a http request


src/Orchard.Web/Core/Settings/Descriptor/ShellDescriptorManager.cs
```diff
@@ -51,7 +51,10 @@ public class ShellDescriptorManager : Component, IShellDescriptorManager {
         }
 
         private ShellDescriptorRecord GetDescriptorRecord() {
-            return _shellDescriptorRepository.Get(x => x != null);
+            if (_shellDescriptorRecord == null) {
```
sebastienros, line 54: Any call to Update should reset the private field then.

HermesSbicego-Laser: I thought it wasn't necessary because UpdateShellDescriptor gets the ShellDescriptorRecord from GetDescriptorRecord, so I supposed every operation on it affects the private property too. I did test enabling and disabling features and it seems to work as expected.

PR #8358, opened 2020-04-23
Prevent loosing contentId when Model has errors

HermesSbicego-Laser:


src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
```diff
@@ -103,11 +103,15 @@ public class ItemController : Controller, IUpdateModel {
                 return new HttpUnauthorizedResult();
 
             var model = _contentManager.BuildEditor(contentItem);
-
+            var routeUrl = Url.RouteUrl(_contentManager.GetItemMetadata(form).DisplayRouteValues);
```
sebastienros: shouldn't it be added to the RouteValues instead of altering the url directly? This way it would conform to the route.

HermesSbicego-Laser: Yes, a new commit is in.

HermesSbicego-Laser:
Can this be merged?

PR #8364, opened 2020-05-05
Fix/8354 caching

MatteoPiovanelli:
As proposed in #8354 we are caching the ShellDescriptorRecord and all related records to prevent all that from being fetched on every request.

### DIFF
--- a/src/Orchard.Web/Core/Settings/Descriptor/ShellDescriptorManager.cs
+++ b/src/Orchard.Web/Core/Settings/Descriptor/ShellDescriptorManager.cs
@@ -33,6 +33,10 @@ namespace Orchard.Core.Settings.Descriptor {
             _lockingProvider = lockingProvider;
             _cacheManager = cacheManager;
             _signals = signals;
+
+            _lockString = string.Join(".",
+                _shellSettings["Name"] ?? "",
+                "ShellDescriptorManager");
         }
 
         public ShellDescriptor GetShellDescriptor() {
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Core/Settings/Descriptor/ShellDescriptorManager.cs
```diff
@@ -50,47 +62,84 @@ public class ShellDescriptorManager : Component, IShellDescriptorManager {
...
 
-        public void UpdateShellDescriptor(int priorSerialNumber, IEnumerable<ShellFeature> enabledFeatures, IEnumerable<ShellParameter> parameters) {
-            ShellDescriptorRecord shellDescriptorRecord = GetDescriptorRecord();
-            var serialNumber = shellDescriptorRecord == null ? 0 : shellDescriptorRecord.SerialNumber;
-            if (priorSerialNumber != serialNumber)
-                throw new InvalidOperationException(T("Invalid serial number for shell descriptor").ToString());
+        private string _lockString = "";
+        private string LockString {
+            get {
+                if (string.IsNullOrWhiteSpace(_lockString)) {
```
sebastienros: Can be set once in the constructor. Can also be simplified (just tenant name + "ShellDescriptorManager").

sebastienros (approved):
Do the simplification of the lockstring and merge

sebastienros:
Thanks

Issue #8376, opened 2020-05-13
NoLockInterceptor does not set the hint for subqueries


MatteoPiovanelli:
The parser for NoLockInterceptor that is on dev does not set the `WITH(NOLOCK)` hint for the tables it should when that is within a subquery.

I have a PR for this, but it uses a regex.

Issue #8336, opened 2020-03-14
Problem in NoLockInterceptor Implementation


armanforghani:
@sebastienros , @MatteoPiovanelli-Laser 
I have a simple query as below:

```
var query = session.Query<ContentItemVersionRecord>().Where(v => v.Published)
                .Join(session.Query<UserPartRecord>(), v => v.ContentItemRecord.Id, u => u.Id, (v, u) => u)
                .Join(session.Query<UserRolesPartRecord>(), user => user.Id, userRole => userRole.UserId, (user, userRole) => new { user, userRole })
                .Join(session.Query<RoleRecord>(), x => x.userRole.Role, role => role, (x, role) => new { User = x.user, Role = role })
                .Where(x => x.User.Id == id);
```
Previously this query was running correctly. After pulling of latest commits in ```dev``` branch, running this query results in the following error message:

```
An exception of type 'NHibernate.Exceptions.GenericADOException' occurred in NHibernate.dll but was not handled in user code
Additional information: could not execute query
```

I think the Implementation of NoLockInterceptor causes this issue because it modifies SQL query and adds ```WITH (NOLOCK)``` in the wrong place. 

For example, in this case, it results the following SQL:

```
select 
	userpartre1_.Id as Id187_,
	userpartre1_.UserName as UserName187_,
	userpartre1_.Email as Email187_,
	userpartre1_.NormalizedUserName as Normaliz4_187_,
	userpartre1_.Password as Password187_,
	userpartre1_.PasswordFormat as Password6_187_,
	userpartre1_.HashAlgorithm as HashAlgo7_187_,
	userpartre1_.PasswordSalt as Password8_187_,
	userpartre1_.RegistrationStatus as Registra9_187_,
	userpartre1_.EmailStatus as EmailSt10_187_,
	userpartre1_.EmailChallengeToken as EmailCh11_187_,
	userpartre1_.CreatedUtc as CreatedUtc187_,
	userpartre1_.LastLoginUtc as LastLog13_187_,
	userpartre1_.LastLogoutUtc as LastLog14_187_,
	userpartre1_.LastPasswordChangeUtc as LastPas15_187_ 
from 
	Orchard_Framework_ContentItemVersionRecord contentite0_,
	WITH (NOLOCK) Orchard_Users_UserPartRecord userpartre1_,
	Orchard_Roles_UserRolesPartRecord userrolesp2_,
	Orchard_Roles_RoleRecord rolerecord3_ 
where 
[...]

PR #8377, opened 2020-05-13
Fix/8376 nolock subqueries

MatteoPiovanelli:
Fix for #8376 
Using a Regex, we can find all subqueries and process each of them, using the results to rebuild the entire query in the end.
This does not help with #8336 because the method that is doing the parsing is still the same.

### DIFF
--- a/src/Orchard/Data/NoLockInterceptor.cs
+++ b/src/Orchard/Data/NoLockInterceptor.cs
@@ -1,9 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Text.RegularExpressions;
-using System.Threading.Tasks;
 using NHibernate;
 using NHibernate.SqlCommand;
 using Orchard.Data.Providers;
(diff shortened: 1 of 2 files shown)


src/Orchard/Data/NoLockInterceptor.cs
```diff
@@ -46,56 +45,226 @@ public class NoLockInterceptor : EmptyInterceptor, ISessionInterceptor {
...
+                // see whether we have anything to add the NOLOCK hint to
+                var tableNamesForQuery =
+                    TableNames.Where(tn => sql.IndexOfCaseInsensitive(tn) >= 0);
+                if (tableNamesForQuery.Any()) {
+                    // Modify the sql to add hints
+                    // the sql may contain substrings, and we want to also process them
+                    var sqlString = sql.ToString();
+                    // https://docs.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions?redirectedfrom=MSDN#balancing-group-definitions
+                    string regexPattern = @"^[^\(\)]*(((?'Open'\()[^\(\)]*)+((?'Close-Open'\))[^\(\)]*)+)*(?(Open)(?!))$";
+                    var matches = Regex.Match(sqlString, regexPattern);
```
sebastienros: Use precompiled regex

MatteoPiovanelli:
Done. Merging into dev now. Thanks for reviewing this

PR #8384, opened 2020-05-27
Process ignored paths while being aware of RequestUrlPrefix

### LINKED ISSUE #8383: OutputCache: Ignored Urls
The list of ignored urls in the cache settings (/Admin/Orchard.OutputCache) requires to write the urls down with the RequestUrlPrefix. 
This has two main drawbacks that I can think of off the top of my head:
1. I can't copy-paste the alias from an AutoroutePart if my tenant has a RequestUrlPrefix
2. In case I want to replicate a configuration across several tenants, I have to take care and fix all those urls manually.

I propose a simple fix to test for the url by stripping the RequestUrlPrefix if it's there. A PR is coming soon because the code is simpler than I am able to explain it.

MatteoPiovanelli:
Fixes #8383 

If there is a ReequestUrlPrefix, and the url we are processing begins with it, try to remove it. But also try to keep it for retrocompatibility with current configurations.

basically:
RequestUrlPrefix: "foo"
Request Url: "/foo/bar"

We test whether we should ignore either "/bar" or "/foo/bar".
The second test in this example is to ensure that anything that works before this change, keeps working afterwards.

If there is no RequestUrlPrefix, nothing changes compared to the current.

I took the liberty of caching the urls to ignore, evicting that cache whenever the cache for the settings as a whole is also evicted.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -497,8 +497,9 @@ namespace Orchard.OutputCache.Filters {
                     ?? (_ignoredRelativePaths = _cacheManager.Get($"{CacheSettings.CacheKey}_IgnoredUrls", true, ContextBoundObject => {
                         ContextBoundObject.Monitor(_signals.When(CacheSettings.CacheKey));
                         return CacheSettings.IgnoredUrls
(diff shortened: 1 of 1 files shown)


sebastienros (approved):

MatteoPiovanelli:
Should I merge this myself?

sebastienros:
Yes, if we approve, since you have merge commits. Use Squash/Merge. We mentioned it at our last meeting.

Issue #8386, opened 2020-06-08
support hierarchy of tenants


MatteoPiovanelli:
Orchard as is doesn't support the following scenario:

Have an Orchard Multitenancy at www.mymultitenancy.com
Create a tenant Foo that you want to respond to www.foo.com
Create a tenant Bart that you want to respond to www.foo.com/bar

The way "redirection" is handled among the shells, a call to www.foo.com/bar is intercepted by Foo, while we want it to be intercepted by Bar. The same goes for www.foo.com/bar/*.

To make that possible, we propose to mark Foo with a boolean IsBaseHost that gives it a "lower priority" when trying to figure out what tenant should handle a request.
This can be handled in a way that breaks nothing, since that bool would default to false when no corresponding setting is found for a tenant, meaning that tenant would, after introducing the changes we propose, still behave exactly like before.

This would allow from a single Orchard Install to have more than one "base", each with potentially several "sub-tenants".

A PR is coming soon.

Care will have to be taken when using this setting:
In the example above, tenant Foo at www.foo.com cannot have its own content whose alias is 'bar', because the tenant Bar would respond to requests for it. This is the opposite of what happens currently.

sebastienros:
https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore/OrchardCore/Shell/RunningShellTable.cs

MatteoPiovanelli:
https://github.com/OrchardCMS/OrchardCore/blob/54f9334aec344eb4fd387ea0db6d1a9a86aedee6/test/OrchardCore.Tests/Shell/RunningShellTableTests.cs#L126

@sebastienros 
This is the test you pointed out to me yesterday. I've run the same test for Orchard, and it passes. However, in Orchard, if you do `table.Add(settingsG)` before `table.Add(settingsA)` the test fails: this is the condition we were seeing in the case of our application.
I think it should still work in OCore, thanks to how `_shellsByHostAndPrefix` is built. Since `ImmutableDictionary` does not seem to be available in .NET 4.5.2 (the current framework version for Orchard.Framework), I'll figure out a work around. It may be enough to keep values in `_shellsByHost` sorted so that the shells with a RequestUrlPrefix are tested first.

PR #8387, opened 2020-06-08
Upgrade host resolution

MatteoPiovanelli:
implements #8386 
I am proposing this for 1.10.x rather than dev because this introduces no breaking change in the way things are done: as it is, everything that is working right now, will keep working the same, but the new setting, when set to true, allows reversing some logic in selecting the tenant/shell that should respond to a request.

### DIFF
--- a/src/Orchard.Tests/Orchard.Framework.Tests.csproj
+++ b/src/Orchard.Tests/Orchard.Framework.Tests.csproj
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\packages\NUnitTestAdapter.2.3.0\build\NUnitTestAdapter.props" Condition="Exists('..\packages\NUnitTestAdapter.2.3.0\build\NUnitTestAdapter.props')" />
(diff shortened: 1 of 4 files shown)


MatteoPiovanelli:
@sebastienros  I made the changes I commented about on the issue. I am not merging this in yet because they are not the same we discussed yesterday. The result is the same, but I had to change a the test because it wasn't correctly testing for the conditions we have.

MatteoPiovanelli:
Is this confirmed for me to merge in 1.10.x?

src/Orchard.Tests/Orchard.Framework.Tests.csproj
```diff
@@ -459,4 +462,10 @@
   <Target Name="AfterBuild">
     <CallTarget Targets="CopySqlCeBinaries" />
   </Target>
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
```
sebastienros: Why is this change necessary?

sebastienros (approved):

src/Orchard.Tests/Orchard.Framework.Tests.csproj
MatteoPiovanelli: it's not. VS had added those lines when I was updating the test adapters for the tests to work properly with VS 2017. I removed it in the latest push

PR #8393, opened 2020-06-16
Fixes Boolean Conversion error

### LINKED ISSUE #8392: LogOn.cshtml: Html.CheckBox() throws exception if ModelState contains non-convertable value
When some malicious user try to programmatically logon to a tenant or simply it modifies the value of "rememberMe" input checkbox with 1/-1 we log an error.
The error is: 
System.InvalidOperationException: The parameter conversion from type 'System.String' to type 'System.Boolean' failed. See the inner exception for more information. ---> System.FormatException: 1 is not valid for Boolean.

It seems to be a MVC problem as shown there https://github.com/aspnet/AspNetWebStack/issues/233
For the moment (waiting for a low level MVC fix) in order to prevent spamming our logs I will remove the value from ModelState in the controller before the view is rendered if ModelState has errors. 
See related PR.

HermesSbicego-Laser:
Fixes #8392 removing "remeberMe" from ModelState when it is in error state: the Html.CheckBox() in LogOn.cshtml view throws exception if ModelState contains non-convertable value

### DIFF
--- a/src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
+++ b/src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
@@ -1,9 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Web.Mvc;
-using System.Xml;
-using System.Xml.Linq;
-using Orchard.Mvc.ModelBinders;
 
 namespace Orchard.Mvc.ModelBinders {
     public class BooleanBinderProvider : IModelBinderProvider, IModelBinder {
--- a/src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
+++ b/src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
@@ -17,7 +17,7 @@ namespace Orchard.Mvc.ModelBinders {
         public object BindModel(ControllerContext controllerContext, ModelBindingContext bindingContext) {
             var value = false;
             var requestBooleanValue = controllerContext.HttpContext.Request[bindingContext.ModelName].Split(',')[0]; //Html.CheckBox and Html.CheckBoxFor return "true,false" string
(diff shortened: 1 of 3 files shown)


MatteoPiovanelli:
Is this confirmed for me to merge in 1.10.x?

sebastienros:
I don't really agree with the change. Could you try to create a custom Model Value provider to detect this, and make the form invalid before it throws an exception? My concern is that this can happen on any page with a bool, and I don't want us to handle this case every time someone is bothered by the generated log.

HermesSbicego-Laser:
@sebastienros please can you verify if the last 2 commits are the correct way to manage that kind of situation?

src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+                                                           ModelBinder = this,
+                                                           Type = typeof(bool)
+                                                       }
+                         };
+        }
+
+        public object BindModel(ControllerContext controllerContext, ModelBindingContext bindingContext) {
+            var value = false;
+            var requestBooleanValue = controllerContext.HttpContext.Request[bindingContext.ModelName].Split(',')[0]; //Html.CheckBox and Html.CheckBoxFor return "true,false" string
+            if (!bool.TryParse(requestBooleanValue.Split(',')[0], out value)) {
```
sebastienros: I think `requestBooleanValue.Split(',')[0]` is unnecessary here as it was already split in the previous statement

HermesSbicego-Laser: Yes, you are right. I Removed it; is it mergeable now? Thanks.

sebastienros:
Can you confirm that the value can't be null or empty and thus prevent `.Split(',')[0]` from throwing?

HermesSbicego-Laser:
I Added a null check in order to prevent Split from throwing

PR #8398, opened 2020-06-30
Update DistributedLockService.cs

### LINKED ISSUE #8399: Continuous lock in multi-server setup causes "Failed to acquire lock" error
I recently resolved an issue where the lock wasn't being removed from the dictionary see #8394 

I've now discovered another issue where if there are two servers that attempt to run the same task at the same time one of them ends up with a continuous lock which keeps logging the following error:

`Failed to acquire lock '{0}' within the specified timeout ({1})`

We have a multi-server setup on our live environment and we are seeing over 20,000 of these messages.

See this line: https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L49

Here is an example scenario:

**Server A**
Task `Orchard.AuditTrail.Services.AuditTrailTrimmingBackgroundTask` runs and acquires a lock with a database record in `[Orchard_Framework_DistributedLockRecord]`

**Server B**
Meanwhile Task `Orchard.AuditTrail.Services.AuditTrailTrimmingBackgroundTask` also runs and enters the lock via `Monitor.Enter` here: https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L77

However since Server A already has a database record lock in `[Orchard_Framework_DistributedLockRecord]` the query to check to see if a lock record exists and has expired returns a record since Server A has acquired a lock: https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L144

Therefore it moves to the `else if` statement https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L170 and eventually it returns `false`

As a result the lock on the object on this line never gets a chance to exit: https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L77  

I have a proposed fix here: https://github.com/OrchardCMS/Orchard/pull/8398

What do you think @jtkech  @sebastienros?

imranazad:
Fixes #8399

### DIFF
--- a/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
+++ b/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
@@ -23,7 +23,6 @@ namespace Orchard.Tasks.Locking.Services {
         private readonly ShellSettings _shellSettings;
         private readonly Dictionary<string, DistributedLock> _locks;
         private readonly TimeSpan _defaultRepeatInterval;
-        private bool _anotherMachineHasTheLock;
 
         public DistributedLockService(
             IApplicationEnvironment applicationEnvironment,
@@ -36,7 +35,6 @@ namespace Orchard.Tasks.Locking.Services {
             _applicationEnvironment = applicationEnvironment;
             _locks = new Dictionary<string, DistributedLock>();
             _defaultRepeatInterval = TimeSpan.FromMilliseconds(500);
-            _anotherMachineHasTheLock = false;
         }
 
         public bool TryAcquireLock(string name, TimeSpan? maxValidFor, TimeSpan? timeout, out IDistributedLock dLock) {
@@ -48,8 +46,7 @@ namespace Orchard.Tasks.Locking.Services {
                     return true;
                 }
                 
-                if(!_anotherMachineHasTheLock)
-                   Logger.Warning("Failed to acquire lock '{0}' within the specified timeout ({1}).", name, timeout);
+                Logger.Debug("Failed to acquire lock '{0}' within the specified timeout ({1}).", name, timeout);
             }
             catch (Exception ex) {
                 Logger.Error(ex, "Error while trying to acquire lock '{0}'.", name);
@@ -115,9 +112,6 @@ namespace Orchard.Tasks.Locking.Services {
                             _locks.Add(monitorObj, dLock);
                             return true;
                         }
-                        
-                        if(_anotherMachineHasTheLock)
-                            Monitor.Exit(monitorObj);
                         return false;
                     });
 
@@ -126,7 +120,9 @@ namespace Orchard.Tasks.Locking.Services {
 
                         if (throwOnTimeout)
                             throw new TimeoutException(String.Format("Failed to acquire lock '{0}' within the specified timeout ({1}).", internalName, timeout));
-
+                        else
+                            Monitor.Exit(monitorObj);
+                        
                         return null;
                     }
                 }
@@ -177,8 +173,7 @@ namespace Orchard.Tasks.Locking.Services {
                     // Existing lock was for correct machine name => lock record exists.
                     Logger.Debug("Found a valid record for lock '{0}' and current local machine name '{1}'.", internalName, environmentIdentifier);
                     hasLockRecord = true;
-                }  else if (record.MachineName != environmentIdentifier)
-                        _anotherMachineHasTheLock = true;
+                }
             });
 
             return hasLockRecord;
--- a/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
+++ b/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
@@ -23,6 +23,7 @@ namespace Orchard.Tasks.Locking.Services {
         private readonly ShellSettings _shellSettings;
         private readonly Dictionary<string, DistributedLock> _locks;
         private readonly TimeSpan _defaultRepeatInterval;
+        private bool _anotherMachineHasTheLock;
 
         public DistributedLockService(
             IApplicationEnvironment applicationEnvironment,
@@ -35,6 +36,7 @@ namespace Orchard.Tasks.Locking.Services {
             _applicationEnvironment = applicationEnvironment;
             _locks = new Dictionary<string, DistributedLock>();
             _defaultRepeatInterval = TimeSpan.FromMilliseconds(500);
+            _anotherMachineHasTheLock = false;
         }
 
         public bool TryAcquireLock(string name, TimeSpan? maxValidFor, TimeSpan? timeout, out IDistributedLock dLock) {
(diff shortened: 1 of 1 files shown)


jtkech:
As said i would need to focus more on it and try it, but i trust you and you will be able to try it ;)

First, in the context of the related issue where it is called by the audit trail bg task we have `maxValidFor = 1 hour`, `timeout = TimeSpan.Zero`, and as it is done trhough `TryAcquireLock()`, we have `throwOnTimeout = false`.

I don't think that doing `Monitor.Exit()` in the context of the `RepeatUntilTimeout()` (here executed once as `timeout = TimeSpan.Zero`) is the right place, i think it needs to be fixed after in the `if (!success) {` block. If `throwOnTimeout` is true we throw an exception so we go to the enclosing `catch` that does a `Monitor.Exit()` so that's okay, **but if `throwOnTimeout` is false, as in your case and i think this is the path that we missed, we still need to do explicitly a `Monitor.Exit()`**.

Could you try it?

Note: Maybe an little issue with the record `ValidUntilUtc` (here 1 hour),  as i can see it may expire while it is still in use ...

jtkech:
So maybe we only need to add the 2 lines below

    if (!success) {
        Logger.Debug("Record for lock '{0}' could not be created for current machine within the specified timeout ({1}).", internalName, timeout);

        if (throwOnTimeout)
            throw new TimeoutException(String.Format("Failed to acquire lock '{0}' within the specified timeout ({1}).", internalName, timeout));
        else // here the 2 added lines
            Monitor.Exit(monitorObj);

        return null;
    }

imranazad:
Haha @jtkech thank you for having faith in me! :-) I wasn't happy with my proposed fix as I knew something was off about it. Your suggestion works a treat! And I can confirm the lock is now released! :-) 

I tested it albeit manually:

1. I ran the task and hit a break point here: https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L144

2. I then inserted the following record:
```
  INSERT INTO  [Orchard_Framework_DistributedLockRecord] (
      [Name]
      ,[MachineName]
      ,[CreatedUtc]
      ,[ValidUntilUtc]
	  ) VALUES ('DistributedLock:Default:Orchard.AuditTrail.Services.AuditTrailTrimmingBackgroundTask', 'IMT-000', '2001-03-20 18:45:12.000', '2020-07-01 21:55:12.000')
```

3. I then continued execution from the break point which led to this line which returns false https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L177

4. I then further continued execution until it hit the new else statement to exit the lock: https://github.com/OrchardCMS/Orchard/pull/8398/files#diff-f5d641271fe12d6f2c0f9373dc2c7786R123

5. I then waited for the task to run again and then hit a break point on this line to ensure and verify the lock enters successfully and that the previous lock was indeed exited: https://github.com/OrchardCMS/Orchard/blob/5b5eda6ae36dc2d7906599882165ef1e985cdc8c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L77

One additional change I have made is to change the log level from `warning` to `debug` for this line otherwise despite the fix this error will keep being logged as the lock will still return null where another machine has acquired a lock on the same task: https://github.com/OrchardCMS/Orchard/pull/8398/files#diff-f5d641271fe12d6f2c0f9373dc2c7786L49

Please do let me know your thoughts. If you're happy with the changes can you please do the honours of a code review and approve them and then I'll ask @sebastienros if he is happy to merge in as we are keen to get this change released for our live systems as we're being inundated with these log messages.

Many thanks

jtkech:
@imranazad okay cool

So, as a reminder, the issue was that if we can't acquire a lock in a given timeout, if `throwOnTimeout` is true the lock is freed by a `Monitor.Exit()` called by the enclosing catch, but if `throwOnTimeout` is false the lock was not freed. So subsequent `Monitor.TryEnter()` were always returning false.

So with this fix, normally you will have a lot less logged warnings, but when used in the context of `TryAcquireLock()` and, as with the audit trail bg task, with a `timeout = TimeSpan.Zero` you will still have some warnings messages, in fact each time the lock is doing its job.

So i agree, in the context of `TryAcquireLock()` we can use `Logger.Debug()` in place of `Logger.Warning()`.

jtkech (approved):

imranazad:
> @imranazad okay cool
> 
> So, as a reminder, the issue was that if we can't acquire a lock in a given timeout, if `throwOnTimeout` is true the lock is freed by a `Monitor.Exit()` called by the enclosing catch, but if `throwOnTimeout` is false the lock was not freed. So subsequent `Monitor.TryEnter()` were always returning false.
> 
> So with this fix, normally you will have a lot less logged warnings, but when used in the context of `TryAcquireLock()` and, as with the audit trail bg task, with a `timeout = TimeSpan.Zero` you will still have some warnings messages, in fact each time the lock is doing its job.
> 
> So i agree, in the context of `TryAcquireLock()` we can use `Logger.Debug()` in place of `Logger.Warning()`.

@jtkech  yep that is spot on and thanks for the prompt response.

Regarding 
>
> Maybe an little issue with the record ValidUntilUtc (here 1 hour), as i can see it may expire while it is still in use 

That's a good point, I didn't get a chance to test this - I've acknowledged it here maybe someone else reading this can volunteer to take a look at it as a separate issue.

@sebastienros could you please take a look at merging this in?

src/Orchard/Tasks/Locking/Services/DistributedLockService.cs
```diff
@@ -121,7 +120,9 @@ public class DistributedLockService : Component, IDistributedLockService {
 
                         if (throwOnTimeout)
                             throw new TimeoutException(String.Format("Failed to acquire lock '{0}' within the specified timeout ({1}).", internalName, timeout));
-
+                        else
+                            Monitor.Exit(monitorObj);
```
sebastienros, line 124: Shouldn't it be released even if `throwOnTimeout` is true?

imranazad: @sebastienros the enclosing catch releases it further below see here: https://github.com/OrchardCMS/Orchard/blob/af0f7157eddd469427b0fe67848b07f024a2152c/src/Orchard/Tasks/Locking/Services/DistributedLockService.cs#L133

jtkech: @sebastienros yes, when an exception is thrown, explicitly or not, this is the enclosing catch that does a `Monitor.Exit()`, so we should not do it before if `throwOnTimeout` is true as we throw explicitly an exception, but we should do it if `throwOnTimeout` is false, i think we missed this path when the `throwOnTimeout` option has been added

PR #8412, opened 2020-09-07
Removed changes to model bindings

MatteoPiovanelli:
This is the same as reverting 1.10.x to commit 868ce12ad63de121703222ad86afdb739d3e864c

We found that those changes caused weird failures in several unexpected places. Some had to do with the fact that the HttpContext does not implement the getter for the Request property. We tried to sidestep that (with a try-catch) but that opened a bunch of other mysterious issues.

More testing is required to use the model binding like this. One of the unexpected side effects was that no ContentType appeared configured as Creatable or Listable anymore.

I did not revert 1.10.x directly to the commit before merging the changes I am reverting, to keep the history on the branch.


MatteoPiovanelli:
@sebastienros I will merge this so 1.10.x is in a stable state. I will then reopen this same PR to keep track of it and the fact that we should perhaps roll back 1.10.x to the state above

PR #8413, opened 2020-09-09
New version of Boolean Binder Provider

MatteoPiovanelli:
This should finally fix the issue for the binding of boolean values

### DIFF
--- a/src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
+++ b/src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
@@ -17,7 +17,7 @@ namespace Orchard.Mvc.ModelBinders {
         public object BindModel(ControllerContext controllerContext, ModelBindingContext bindingContext) {
             var value = false;
             try {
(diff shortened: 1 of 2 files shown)


src/Orchard/Mvc/ModelBinders/BooleanBinderProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+                                                           ModelBinder = this,
+                                                           Type = typeof(bool)
+                                                       }
+                         };
+        }
+
+        public object BindModel(ControllerContext controllerContext, ModelBindingContext bindingContext) {
+            var value = false;
+            try {
+                value = (bool)bindingContext.ValueProvider.GetValue(bindingContext.ModelName).ConvertTo(typeof(bool));
```
sebastienros: I assume this is for the potential null ref I mentioned. Usually I use something like `Convert.ToBoolean( /* something that can be null */)`

```suggestion
                value = Convert.ToBoolean(bindingContext.ValueProvider.GetValue(bindingContext.ModelName));
```

sebastienros:
Feel free to merge whatever version you prefer

Issue #8414, opened 2020-09-15
NHibernate error saving a very long text in LayoutRecord


LorenzoFrediani-Laser:
The error occurs trying to save a long text in a LayoutRecord.
To reproduce the error, add a new Query contentitem with Layouts of type Raw with display mode set to Properties and than  add a long text on Append field.

Saving the Query contentitem NHibernate Exception occurs.
HibernateException: The length of the string value exceeds the length configured in the mapping/parameter.

PR #8415, opened 2020-09-15
fix record mapping nhibernate

LorenzoFrediani-Laser:
fix for issue #8414


sebastienros:
Is it a breaking change? What is the db type before and after?

sebastienros (approved):

PR #8420, opened 2020-09-25
Add FromName and ReplyTo properties to SmtpSettingsPart

aaronamm:
Refer to :
- https://stackoverflow.com/a/14555043/1872200 and
- https://www.postmastery.com/list-unsubscribe-header-critical-for-sustained-email-delivery/

I think adding from name, reply to and List-Unsubscribe can help passing an email spam filter.

Therefore, I have updated SmtpSettingsPart to include FromName, ReplyTo and ListUnsubscribe properties.
In addition, I also updated SmtpMessageChannel to make sure everything work correctly.

BTW, I have changed naming convention of fields to not use _ underscore and use a this keyword.
I know it is more convenient to use underscore but I think it is not .NET naming convention.

Hopefully, this PR is valid and will get approved. I am happy to update this PR from your feedback.

Thanks.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Email/Controllers/EmailAdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Email/Controllers/EmailAdminController.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Web.Hosting;
 using System.Web.Mvc;
 using Orchard.ContentManagement;
 using Orchard.Email.Models;
@@ -12,12 +11,12 @@ using Orchard.UI.Admin;
 namespace Orchard.Email.Controllers {
     [Admin]
     public class EmailAdminController : Controller {
-        private readonly ISmtpChannel _smtpChannel;
-        private readonly IOrchardServices _orchardServices;
+        private readonly ISmtpChannel smtpChannel;
+        private readonly IOrchardServices orchardServices;
 
         public EmailAdminController(ISmtpChannel smtpChannel, IOrchardServices orchardServices) {
-            _smtpChannel = smtpChannel;
-            _orchardServices = orchardServices;
+            this.smtpChannel = smtpChannel;
+            this.orchardServices = orchardServices;
             T = NullLocalizer.Instance;
         }
 
@@ -29,14 +28,13 @@ namespace Orchard.Email.Controllers {
             ILogger logger = null;
             try {
                 var fakeLogger = new FakeLogger();
-                var smtpChannelComponent = _smtpChannel as Component;
-                if (smtpChannelComponent != null) {
+                if (smtpChannel is Component smtpChannelComponent) {
                     logger = smtpChannelComponent.Logger;
                     smtpChannelComponent.Logger = fakeLogger;
                 }
 
                 // Temporarily update settings so that the test will actually use the specified host, port, etc.
-                var smtpSettings = _orchardServices.WorkContext.CurrentSite.As<SmtpSettingsPart>();
+                var smtpSettings = orchardServices.WorkContext.CurrentSite.As<SmtpSettingsPart>();
 
                 smtpSettings.FromAddress = testSettings.FromAddress;
                 smtpSettings.FromName = testSettings.FromName;
@@ -48,18 +46,19 @@ namespace Orchard.Email.Controllers {
                 smtpSettings.UseDefaultCredentials = testSettings.UseDefaultCredentials;
                 smtpSettings.UserName = testSettings.UserName;
                 smtpSettings.Password = testSettings.Password;
+                smtpSettings.ListUnsubscribe = testSettings.ListUnsubscribe;
 
                 if (!smtpSettings.IsValid()) {
                     fakeLogger.Error("Invalid settings.");
                 }
                 else {
-                    _smtpChannel.Process(new Dictionary<string, object> {
+                    smtpChannel.Process(new Dictionary<string, object> {
                         {"Recipients", testSettings.To},
                         {"Subject", T("Orchard CMS - SMTP settings test email").Text}
                     });
                 }
 
-                if (!String.IsNullOrEmpty(fakeLogger.Message)) {
+                if (!string.IsNullOrEmpty(fakeLogger.Message)) {
                     return Json(new { error = fakeLogger.Message });
                 }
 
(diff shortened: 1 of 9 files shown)


src/Orchard.Web/Modules/Orchard.Email/Controllers/EmailAdminController.cs
```diff
@@ -12,12 +11,12 @@
 namespace Orchard.Email.Controllers {
     [Admin]
     public class EmailAdminController : Controller {
-        private readonly ISmtpChannel _smtpChannel;
-        private readonly IOrchardServices _orchardServices;
+        private readonly ISmtpChannel smtpChannel;
```
sebastienros: Can you revert all the names you changed?

src/Orchard.Web/Modules/Orchard.Email/Views/EditorTemplates/Parts/SmtpSettings.cshtml
```diff
@@ -65,6 +72,11 @@
             </span>
         </div>
     </div>
+    <div>
+        <label for="@Html.FieldIdFor(m => m.ListUnsubscribe)">@T("List-Unsubscribe header")</label>
```
sebastienros, line 76: Is that something that mail clients understand natively, or just a field that can be used in a template?

sebastienros (commented):
Can you also check that the workflow activities don't need to be updated to support these new fields?

src/Orchard.Web/Modules/Orchard.Email/Controllers/EmailAdminController.cs
aaronamm: @sebastienros Okay. Since it is naming convention that Orchard team use, I will follow the standard.

src/Orchard.Web/Modules/Orchard.Email/Views/EditorTemplates/Parts/SmtpSettings.cshtml
aaronamm: List-Unsubscribe is a STMP header that a mail client can understand natively.
https://www.postmastery.com/list-unsubscribe-header-critical-for-sustained-email-delivery/
It can help an email not go to junk.

Additional information:
https://blog.mailtrap.io/list-unsubscribe-header/#What_is_the_list-unsubscribe_header

aaronamm:
> Can you also check that the workflow activities don't need to be updated to support these new fields?

I am not familiar with Orchard.WorkFlow area.
However after I search carefully, I did not find any usage of SMTP setting here.
image


There are two main places that use the setting.
- EmailMessagingChannel I don't update it because it is obsoleted.
- SmtpMessageChannel the main file that I changed.

aaronamm:
@sebastienros  Fixed all from your feedback, thank you so much.

src/Orchard.Web/Modules/Orchard.Email/Services/EmailMessagingChannel.cs
```diff
@@ -53,7 +53,7 @@ public class EmailMessagingChannel : IMessagingChannel {
                 smtpClient.EnableSsl = smtpSettings.EnableSsl;
                 smtpClient.DeliveryMethod = SmtpDeliveryMethod.Network;
 
-                context.MailMessage.From = new MailAddress(smtpSettings.Address);
+                context.MailMessage.From = new MailAddress(smtpSettings.FromAddress);
```
sebastienros, line 56: This is an important breaking change, since all existing sites will lose this setting. Maybe a migration step would be good to move the old value to the new properties.

aaronamm: @sebastienros  That's a good point, sorry for my mistake. I will continue the work on a migration.
This should be check list item for a contribution note.

aaronamm: Update a migration with a test cases result.
However, there is an interesting issue that should fix by other PR.
The existing attribute do not remove from XML after SmtpSettingsPart get updated.
It is still there. I think a serialize XML of saving StmpSettingsPart should be responsible for this.
Therefore, I don't remove it manually.

Could you please check the data after migration in the test case 1.

**Test case 1**
SmtpSettingPart **with existing value** before migration
```
<Data>
    <SiteSettingsPart SiteSalt="fdb3ec855e864eba91d2a3bdfa3c620a" SiteName="orchard-pr" PageTitleSeparator=" - " SiteTimeZone="SE Asia Standard Time" UseFileHash="true" SuperUser="admin" SiteCulture="en-US" PageSize="20" BaseUrl="http://localhost:61745" />
    <CacheSettingsPart VaryByQueryStringIsExclusive="true" />
    <CommentSettingsPart ModerateComments="true" />
    <ThemeSiteSettingsPart CurrentThemeName="TheThemeMachine" />
    <SmtpSettingsPart Address="admin@orchard.com" Host="localhost" Port="25" EnableSsl="false" RequireCredentials="false" UseDefaultCredentials="false" UserName="" Password="" />
</Data>
```

SmtpSettingPart after migration
```
<Data>
    <SiteSettingsPart SiteSalt="fdb3ec855e864eba91d2a3bdfa3c620a" SiteName="orchard-pr" PageTitleSeparator=" - " SiteTimeZone="SE Asia Standard Time" UseFileHash="true" SuperUser="admin" SiteCulture="en-US" PageSize="20" BaseUrl="http://localhost:61745" />
    <CacheSettingsPart VaryByQueryStringIsExclusive="true" />
    <CommentSettingsPart ModerateComments="true" />
    <ThemeSiteSettingsPart CurrentThemeName="TheThemeMachine" />
    <SmtpSettingsPart Address="admin@orchard.com" Host="localhost" Port="25" EnableSsl="false" RequireCredentials="false" UseDefaultCredentials="false" UserName="" Password="" FromAddress="admin@orchard.com" />
</Data>
```

**Test case 2**
SmtpSettingPart **without existing value** before migration 
```
<Data>
    <SiteSettingsPart SiteSalt="fdb3ec855e864eba91d2a3bdfa3c620a" SiteName="orchard-pr" PageTitleSeparator=" - " SiteTimeZone="SE Asia Standard Time" UseFileHash="true" SuperUser="admin" SiteCulture="en-US" PageSize="20" BaseUrl="http://localhost:61745" />
    <CacheSettingsPart VaryByQueryStringIsExclusive="true" />
    <CommentSettingsPart ModerateComments="true" />
    <ThemeSiteSettingsPart CurrentThemeName="TheThemeMachine" />
</Data>
```

SmtpSettingPart after migration
```
<Data>
    <SiteSettingsPart SiteSalt="fdb3ec855e864eba91d2a3bdfa3c620a" SiteName="orchard-pr" PageTitleSeparator=" - " SiteTimeZone="SE Asia Standard Time" UseFileHash="true" SuperUser="admin" SiteCulture="en-US" PageSize="20" BaseUrl="http://localhost:61745" />
    <CacheSettingsPart VaryByQueryStringIsExclusive="true" />
    <CommentSettingsPart ModerateComments="true" />
    <ThemeSiteSettingsPart CurrentThemeName="TheThemeMachine" />
</Data>
```

Here is admin panel after a database migration.
image


Sent test email result

image

aaronamm:
@sebastienros  Fixed all your PR feedbacks.
Thanks.

MatteoPiovanelli:
@HermesSbicego-Laser @ElenaRepository 
Tagging us here to review our email activities/workflows

Issue #8437, opened 2020-12-09
Event for workflows for the first update after a ContentItem's creation


MatteoPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/6e410a321d01014ba8c0ef69f48224ba3e3de6f4/src/Orchard.Web/Modules/Orchard.Workflows/Handlers/WorkflowContentHandler.cs#L41

https://github.com/OrchardCMS/Orchard/blob/6e410a321d01014ba8c0ef69f48224ba3e3de6f4/src/Orchard.Web/Modules/Orchard.Workflows/Activities/ContentActivity.cs#L9

A common requirement we see in our applications is to have something happen when a new item (of a specific type) has finished its first update. We had to hack this together several times, and I was thinking to add a workflow event (with corresponding activity) for that condition.

The idea would in the WorkflowContentHandler, to memorize `OnCreated` the ids of the ContentItems we are creating. If in the `OnUpdated` we have one of those ids, then we fire the new `ContentFirstUpdated` event.

It shares a possible issue with the existing `ContentUpdated`: the ContentItem in the context may not actually have that version be created, in case some validation step fails.

PR #8438, opened 2020-12-10
Added an event activity for workflows that activates on the first Upd…

MatteoPiovanelli:
Implements the feature described in #8437 
Added an event activity for workflows that activates on the first Updated event after creation of a ContentItem.
We memorize the ids of ContentItems that went through the "Created" event to know which ones have been created.
We memorize the ids of ContentItems that went through the "Updated" event to prevent firing the "ContentFirstUpdated" event several times in case Updated is invoked multiple times.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Workflows/Handlers/WorkflowContentHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Workflows/Handlers/WorkflowContentHandler.cs
@@ -6,8 +6,16 @@ using Orchard.Workflows.Services;
 namespace Orchard.Workflows.Handlers {
 
     public class WorkflowContentHandler : ContentHandler {
+        // Used to memorize the ids of ContentItems for which we go through the
+        // OnCreated handler.
+        private List<int> _createdItems;
+        // Used to memorize the ids of ContentItems for which we go through the
+        // OnUpdated handler.
+        private List<int> _updatedItems;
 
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Workflows/Handlers/WorkflowContentHandler.cs
```diff
@@ -6,8 +6,16 @@
...
     public class WorkflowContentHandler : ContentHandler {
+        // Used to memorize the ids of ContentItems for which we go through the
+        // OnCreated handler.
+        private List<int> _createdItems;
+        // Used to memorize the ids of ContentItems for which we go through the
+        // OnUpdated handler.
+        private List<int> _updatedItems;
 
         public WorkflowContentHandler(IWorkflowManager workflowManager) {
+            _createdItems = new List<int>();
```
sebastienros: Hashset is better suited here

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Workflows/Handlers/WorkflowContentHandler.cs
MatteoPiovanelli: You are right. Let me update that.

sebastienros:
It was approved, you could have merged without waiting

PR #8458, opened 2021-02-18
1.10.x into dev

MatteoPiovanelli:
Merged 1.10.x into dev.

@sebastienros @HermesSbicego-Laser 
We need to verify the approach we chose for solving conflicts with the migrations for CommonPartRecord:
https://github.com/LaserSrl/Orchard/blob/799b3a920d3c1134a87e152cc79a83d7714275af/src/Orchard.Web/Core/Common/Migrations.cs#L186
basically we considered the version from 1.10.x "right" and had the conflicting migrations from dev follow it. This is a breaking change for dev.


MatteoPiovanelli:
@sebastienros  I fixed what we discussed yesterday for that migration, so I'll go ahead and merge this into dev.
I will also close the PRs regarding the new version of log4net, since the upgrade of that is included here.

PR #8461, opened 2021-03-04
Prevents throwing exception when Href is null

HermesSbicego-Laser:
Somehow in 1 of our tenants the Href is null. 
This prevent throwing an exception in case the Href property is null

### DIFF
--- a/src/Orchard/UI/Navigation/NavigationHelper.cs
+++ b/src/Orchard/UI/Navigation/NavigationHelper.cs
@@ -124,7 +124,7 @@ namespace Orchard.UI.Navigation {
                     string appPath = currentRequest.ApplicationPath ?? "/";
                     string requestUrl = currentRequest.Path.StartsWith(appPath) ? currentRequest.Path.Substring(appPath.Length) : currentRequest.Path;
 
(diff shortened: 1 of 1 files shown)


src/Orchard/UI/Navigation/NavigationHelper.cs
```diff
@@ -124,7 +124,7 @@ public static class NavigationHelper {
                     string appPath = currentRequest.ApplicationPath ?? "/";
                     string requestUrl = currentRequest.Path.StartsWith(appPath) ? currentRequest.Path.Substring(appPath.Length) : currentRequest.Path;
 
-                    string modelUrl = menuItem.Href.Replace("~/", appPath);
+                    string modelUrl = menuItem.Href?.Replace("~/", appPath);
```
sebastienros: It's incomplete, the next statement will then throw.

HermesSbicego-Laser: Right, next commit should prevent it.

sebastienros (approved):

sebastienros:
I approved, so it means you can merge it if you want.

Issue #8469, opened 2021-04-23
Cached pages don't work with forms and their RequestVerificationToken when navigation is authenticated


HermesSbicego-Laser:
Because OutputCache stores the RequestVerificationToken for authenticated users also, cached pages fails to POST data.

sebastienros:
I remember how we had decided to handle XSRF tokens, by actually not caching any page that generates one. Then anything with forms would not be cached. The solution we have was to issue forms using ajax such that we could still cache most of the page.

I don't know what it's not applied anymore, but that might be a better solution? If it's possible to see that the outgoing cookie is set. Can you confirm the logic? And that it still works with comment forms?

HermesSbicego-Laser:
fixed via #8470

PR #8470, opened 2021-04-23
Bypass cache for XSRF Tokens

HermesSbicego-Laser:
solves #8469

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
@@ -29,7 +29,7 @@ using Orchard.Utility.Extensions;
 namespace Orchard.OutputCache.Filters {
     public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, IDisposable {
         private const string REQUEST_VERIFICATION_TOKEN_BEACON_TAG = "<$request-verification-token-beacon-for-hidden-field />";
-        private const string REQUEST_VERIFICATION_TOKEN_PATTERN = "<\\s*input\\s*.*name\\s*=\\s*\"__RequestVerificationToken\"\\s*.*value\\s*=\\s*\"(?<value>[\\w-]*)\"\\s*/>";
+        private const string REQUEST_VERIFICATION_TOKEN_INVARIANT_TAG = "<input name=\"__RequestVerificationToken\"";
         private static string _refreshKey = "__r";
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.OutputCache/Filters/OutputCacheFilter.cs
```diff
@@ -588,6 +608,44 @@ public class OutputCacheFilter : FilterProvider, IActionFilter, IResultFilter, I
...
+            // Because of the __RequestVerificationToken hidden field vary by the user, before caching the output, we need to replace the 
+            // __RequestVerificationToken hidden field with a "beacon" text that will be replaced before rendering the page
+            // with a fresh new __RequestVerificationToken hidden field.
+            // What we do is to replace every <input name="__RequestVerificationToken " value="{the-value}" /> with
+            //                                <$request-verification-token-beacon-for-hidden-field />
+            if (PreventCachingRequestVerificationToken()) {
+                string patternRegion = "<\\s*input\\s*.*name\\s*=\\s*\"__RequestVerificationToken\"\\s*.*value\\s*=\\s*\"" +
+                                       "(?<value>[\\w-]*)" +
+                                       "\"\\s*/>";
+                RegexOptions regexOptions = RegexOptions.Multiline | RegexOptions.IgnoreCase | RegexOptions.Compiled;
```
sebastienros: Try not to use a regex, it's dangerous. Or at least define a small timeout value. Also before using the regex, so a simple IndexOf to ensure there is a potential match first.

HermesSbicego-Laser: The new commit adds indexof check and a regex timeout as suggested. I modified the regex replace with a static one in order to be more performant. 
In our dev environment:
With the regex object it performs a replace in 20.68 millisenconds
With the static regex it performs a replace in 4.23 millisenconds

sebastienros: But do you really need a regex? Why not simply search for `<input name="__RequestVerificationToken"` since it's generated by some well-known code. ? If you search for spaces between `name` and `=` why not seach for other attributes before `name`? Because you are making an assumption about what is generating this input, and the same assumption lets you search for the exact string I suggested, removing the need of a regex.

HermesSbicego-Laser: I removed regex as suggested, see latest commit.

sebastienros (approved):

Issue #8471, opened 2021-04-30
YamlDotNet v. 9.1.3 missing in nuget packages


HermesSbicego-Laser:
For some mysterious reason YamlDotNet is no more available as nuget package: so it cannot be downloaded or restored and fresh new build of Orchard solution does not compile. We will upgrade it to Latest stable (11.1.1) in a soon PR.

PR #8472, opened 2021-04-30
Upgrades YamlDotNet from 9.1.3 to 11.1.1

HermesSbicego-Laser:
because 9.1.3 is not more available as nuget package


sebastienros:
The package is still available, it's just not listed. So it will work but unlisting it means it's not recommended to use.

sebastienros (approved):

sebastienros:
Fixes #8471

PR #8474, opened 2021-05-01
Deploy to Azure App Service with GitHub Actions

aaronamm:
# Changes:
- Add GitHub Actions workflow for deploying to Azure App Service
- Move .editorconfig to root of the project

**Here is the reason why I want to add `deploy-to-azure-app-service.yml` to the project.**
- Deployment with traditional Kudu can take 25 - 30 minutes which is quite slow.

image

- For Kudu deployment it requires many files: 
  - .deployment
  - deploy.ps1
  - DeploymentUtility.psm1  
- Build script should be more declarative but our PowerShell scripts are more imperative.
- I sometime found these PowerShell scripts are not reliable when building Node.js projects because it fails sometime which shouldn't happened. 
- Kudu deployment does not support **Yarn Workspaces** which is very useful for managing Node.js packages in each Orchard modules. This is due to the file permission issue.
- I think right now Microsoft try to promote GitHub Actions more. As you can see, it is a default build provider option when you select GitHub as a source of a project.
image
- When deploy with GitHub Actions, it takes around 7 minutes.
image

# How to use GitHub Actions in this PR
Please refer to README at the top of **deploy-to-azure-app-service.yml** file.

# When will the workflow start
- This GitHub workflow will run when `pushing code to a main branch &  AZURE_WEBAPP_NAME defined`
- It will try build Yarn Workspace packages if exist.

# Deployment result
- Here is a website deployed with GitHub Actions workflow in this PR
- https://orchard-cms-pr.azurewebsites.net/
- To log in admin panel, user
  - username: admin
  - password: 12345Abc% 
 - GitHub Actions log https://github.com/aaronamm/Orchard/actions

### DIFF
--- a/.github/workflows/deploy-to-azure-app-service.yml
+++ b/.github/workflows/deploy-to-azure-app-service.yml
@@ -1,0 +1,127 @@
+# How to use this GitHub Actions workflow
+
+# - Enable GitHub Action in your repository .
+# - Create Azure App Service code publish with .NET 4.8
+# - Add required GitHub Secrets:
+#   - NAME: AZURE_WEBAPP_NAME
+#     VALUE: Your Azure App Service name (only name, not a full URL)
+#   - NAME: AZURE_WEBAPP_PUBLISH_PROFILE
+#     VALUE: A content of downloaded publish profile from your App Service in Azure portal
+#     *** Before downloading a publish profile, please make sure you have set a configuration WEBSITE_WEBDEPLOY_USE_SCM = true.
+# - Push code to a main branch.
+# - GitHub Actions will automatic build and deploy the project to Azure App Service.
+
+name: Deploy Orchard CMS (.NET Framework) to Azure App Service
+
+# This workflow is triggered when pushing to a specific branch.
+on:
+  push:
+    branches:
+      - main
+
+env:
+  # Set MSBuild version to use. "[ or ]" means inclusive and "( or )" means exclusive.
+  # You can find release version at https://github.com/dotnet/msbuild/releases
+  MSBUILD_VERSION: "[16.9,17.0)"
+
+  NUGET_VERSION: 5.x # Set NuGet version to use.
+  NODE_VERSION: 10.x # Set Node.js version to use.
+
+  SOLUTION_PATH: src/Orchard.sln
+  # For Azure deployment, we use a custom MS Build at the root of the project.
+  PROJECT_PATH: Orchard.proj
+  PRECOMPILED_DIR: build/precompiled
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set supportedAppService variable
+        id: setvar
+        run: |
+          if [ -z "${{ secrets.AZURE_WEBAPP_NAME }}" ] # test -z return True if the length of String is zero.
+          then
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -0,0 +1,128 @@
+# How to use this GitHub Actions workflow
```
sebastienros, line 1: I think this should be a file available somewhere in the repos or the documentation that you copy to your own github workflows. Otherwise there won't be any visibility, and it's like it's a workflow for the current repos.

.github/workflows/deploy-to-azure-app-service.yml
```diff
@@ -0,0 +1,128 @@
...
+          yarn install
+          yarn workspaces run build
+
+      - name: Setup NuGet version ${{ env.NUGET_VERSION }}
+        # Find a release version. https://github.com/NuGet/setup-nuget/releases/
+        uses: NuGet/setup-nuget@v1
+        with:
+          nuget-version: ${{ env.NUGET_VERSION }}
+
+      - name: Restore Nugetet packages
```
sebastienros: Typo "NuGet"

.github/workflows/deploy-to-azure-app-service.yml
aaronamm: @sebastienros  I agree with your idea. 

However, I tried to make this PR does the same behavior as Kudu script which we have.
This works flow will run automatically if 
- You have defined AZURE_WEBAPP_NAME secret
- You push code to a `main` branch

I still prefer to keep the workflow file in `.github` folder and auto deploy if it matches a condition.
Would you be okay if I add a new section "how to deploy Orchard to Azure App Service" in the main README.md file
https://github.com/OrchardCMS/Orchard/blob/dev/README.md?

If you any other idea, please let me know. I would like to have this PR merged because it will be useful for those who need to maintain Orchard 1 project.

.github/workflows/deploy-to-azure-app-service.yml
aaronamm: @sebastienros  Fixed, thanks.

.github/workflows/deploy-to-azure-app-service.yml
sebastienros: I still disagree with having this file in the .github folder by default. It should only be in some samples/docs folder, and explained for devs to customize it and put it there if they want it. Most users don't use GH as their CI for Orchard. It a very custom flow that is nice but you are part of a few doing it. If you make the file available and some doc on how to use it, then it would be nice to add a pointer to it in the README for sure.

aaronamm: @sebastienros Thank you for your reply. I will update the PR as your suggested.

BenedekFarkas: I agree that it shouldn't be in this repo directly (we have CI workflows for this repo since then and it would appear there), so I'll close this PR. Please open a new one if you'd like to add this to the docs.
Also check out https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/workflows/deploy-orchard1-to-azure-app-service.yml, we use this for our own O1 applications, but you can fully customize it through parameters.

Issue #8480, opened 2021-05-31
Flag “Exclude children” for Taxonomies


ElenaRepository:
Hi,

We found a problem with the management of the Has Term filter in the query, with the steps below to check the error.

After creating a taxonomy and attaching it to a content, we select 2 child terms, one which has other child terms within it

![image](https://user-images.githubusercontent.com/50620495/120204738-70fcde80-c229-11eb-829b-d2cd4bd6aa73.png)

Let us create the query by selecting ‘Is all of’ and the two terms we have previously selected, without selecting relatives.

![image](https://user-images.githubusercontent.com/50620495/120204787-7ce8a080-c229-11eb-9d9b-8a16b71b017b.png)

The query will answer with ‘no outcome’, this because currently all the selected terms and their children are taken, so in this case the filter will search the items which contain "child a" "a 1" "child b"

Adding a flag as “Automatically excludes children terms in filtering”, that with default to false continues to work as it does now, instead of true, it filters only for selected terms and not for children.

HermesSbicego-Laser:
closed via #8481

PR #8481, opened 2021-05-31
flag exclude children for taxonomies

ElenaRepository:
#8480

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Projections/TermsFilterForms.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Projections/TermsFilterForms.cs
@@ -48,8 +48,13 @@ namespace Orchard.Taxonomies.Projections {
                                 Id: "operator-is-all-of", Name: "Operator",
                                 Title: T("Is all of"), Value: "1"
                                 )
-                            )
-                        );
+                            ),
+                        _ExcludeChildren: Shape.Checkbox(
+                            Id: "ExcludeChildren", Name: "ExcludeChildren",
+                            Title: T("Automatically exclude children terms in filtering"),
+                            Value: "true"
+                        )
+                    );
 
                     foreach (var taxonomy in _taxonomyService.GetTaxonomies()) {
                         f._Terms.Add(new SelectListItem { Value = String.Empty, Text = taxonomy.Name });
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Projections/TermsFilterForms.cs
```diff
@@ -48,8 +48,13 @@ public class TermsFilterForms : IFormProvider {
                                 Id: "operator-is-all-of", Name: "Operator",
                                 Title: T("Is all of"), Value: "1"
                                 )
-                            )
-                        );
+                            ),
+                        _ExcludeChildren: Shape.Checkbox(
+                            Id: "ExcludeChildren", Name: "ExcludeChildren",
+                            Title: T("Automatically exclude children terms in filtering"),
+                            Value: "true"
```
sebastienros, line 55: If this is `true` by default then isn't it a breaking change compared to the previous behaviour? If so then it should be false by default.

ElenaRepository: The true value is correct, it is not the initial value of the checkbox, but is used by the mvc binding to assign the correct value to the property if the checkbox is checked.

sebastienros: If I go to the TermsFilterForm, the checkbox will be ticked by default, right? So `ExcludeChildren.Value` will then be set to `true` if I save the form without changing anything. Then `bool.TryParse(context.State.ExcludeChildren?.Value, out bool excludeChildren)` will return `true`, and 

```
                    if (!excludeChildren)
                        allChildren.AddRange(_taxonomyService.GetChildren(term));
```

won't be executed, which was not the case before this PR

HermesSbicego-Laser: Sorry if I join the discussion, but I think there was a misunderstanding. The `Shape.Checkbox(...)` code renders this kind of html 
`<input id="IncludeChildren" name="IncludeChildren" type="checkbox" value="true" />`
the attribute responsible of thicking the checkbox is the `checked` attribute not the `value` one. 
So this PR does not change the previous behaviour and it's not breaking. @sebastienros

sebastienros: OK, thanks

sebastienros (approved):

PR #8487, opened 2021-07-15
Cloning doesn't overwrite identity

### LINKED ISSUE #8486: Cloning changes identity of original contentItem
https://github.com/OrchardCMS/Orchard/blob/7131970e91d19a4ac5ec92b10fae885e50bd1fb5/src/Orchard.Web/Core/Common/Handlers/IdentityPartHandler.cs#L13

That line is assigning the identity of the original contentItem. It should do that for the clone. PR incoming.

MatteoPiovanelli:
Fixes #8486


sebastienros (approved):
Seems weird we have never seen that one. If there is an existing unit test for cloned items that would be great to add one for this just to be sure it's doing the correct thing.

MatteoPiovanelli:
I opened a separate issue to add the relevant tests

PR #8495, opened 2021-08-05
Revert "Cloning doesn't overwrite identity"

MatteoPiovanelli:
Reverts OrchardCMS/Orchard#8487

Those changes should have gone on the dev branch rather than 1.10.x. My mistake.


MatteoPiovanelli:
@sebastienros I am merging this in 1.10.x to revert my earlier change because that broke compilation

PR #8496, opened 2021-08-05
Cloning doesn't overwrite identity

### LINKED ISSUE #8486: Cloning changes identity of original contentItem

MatteoPiovanelli:
Fixes #8486


sebastienros:
This is the same change that was reverted?

MatteoPiovanelli:
Yes, this is the change that I had mistakenly merged in 1.10.x

sebastienros (approved):

Issue #8489, opened 2021-07-22
Projection part default value


ElenaRepository:
It is often the case that we need all projections of a some type to have common settings. 
In order to ease this operation, we have decided to add a setting to the projectionPart that allows us to overcome this difficulty.
In the setting, it would be useful to have not only the default query, which is already evaluated during the creation phase, but also the "static" parts such as 
- Items to display 
- Offset 
- Maximun items 
- Suffix 
- Show pager

in order to have a good starting situation for content creation. It would also be useful for these values to be able to choose to have them edited or not.
It might also be useful to have a filter which, in the case of a lot of queries, allows only those queries appropriate to the type of content to be selected.
Finally, this setting will definitely require import-export management.

MatteoPiovanelli:
I'm closing this because the corresponding PR has been merged

PR #8497, opened 2021-08-05
Feature/8489 projection default settings

HermesSbicego-Laser:
Adds default values to ProjectionPart, as proposed in #8489

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/Models/LayoutRecord.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Models/LayoutRecord.cs
@@ -9,6 +9,7 @@ namespace Orchard.Projections.Models {
         }
 
         public virtual int Id { get; set; }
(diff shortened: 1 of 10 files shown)


src/Orchard.Web/Modules/Orchard.Projections/Models/LayoutRecord.cs
```diff
@@ -9,6 +9,7 @@ public class LayoutRecord {
         }
 
         public virtual int Id { get; set; }
+        public virtual string GUIdentifier { get; set; }
```
sebastienros, line 12: What is that?

ElenaRepository: While implementing the setting, we noticed that an identifier for the query layouts was missing. The Guid is needed to correctly manage the import and export of the setting.

sebastienros (approved):

Issue #8509, opened 2021-09-15
Different cache key by role


ElenaRepository:
We wanted to propose a feature to enable for managing front-end edit content.

From our tests, we noticed that in the case of active caching.

User A has the role of Administrator
User B has the role of Editor,Moderator

In the case where user A logs in, and immediately after user B logs in, user B will see all the front-end edit links of user A

If user B clicks to edit the content, the message "there is no permission to edit the content" will be correctly displayed.

As a fix we can add a flag to the outputcache settings. If the flag is active add the user role to the cache key, this way we can have a role specific cache. Cache eviction can be done when editing or deleting a role by referring to the role added to the cache key this way if permissions are changed a new cache key is created.

sebastienros:
Caching and authenticated users don't work well together. There is an option in OC to enable caching for authenticated users but it means that all users will share the cache, so in your case this option is not compatible. So either you need to disable the option and lose cache for authenticated users, or have it disabled dynamically by your feature when some "edit links" are rendered. Check if there is an extensibility point in the caching filter to do that dynamically and not just with the setting.

MatteoPiovanelli:
I renamed the issue to better explain what we are proposing.

The issue we are seeing is in the case where authenticated users, depending on their roles, need to see different "stuff" in the frontend. One example of such cases is the one described by Elena., but there are others: suppose for instance that there is a Layer that is active only for a specific role (e.g. "VIP") per https://github.com/OrchardCMS/Orchard/blob/d51418b2c11d4f3c02a8440e49e983b9bf53e374/src/Orchard.Web/Modules/Orchard.Roles/Conditions/RoleCondition.cs#L12
Users with the "VIP" role would have to see different things on their pages than users without it.

One way to do it would be to implement
https://github.com/OrchardCMS/Orchard/blob/10963275c355c23a6e67f365add4d0888770d36f/src/Orchard.Web/Modules/Orchard.OutputCache/ICachingEventHandler.cs#L6
The KeyGenerated implementation would basically add a list of the current user's roles to the cache key.
We would do this in a new feature in Orchard.Roles. This new feature will need to depend on Orchard.Roles and Orchard.OutputCache, at least.
The feature would also need to handle evicting cached content when the permissions for a role change.
(There are likely several other conditions where authenticated cached content should be evicted, but those are likely out of scope and hard to detect reliably without basically not caching anything.)
[...]

PR #8510, opened 2021-09-17
Different cache key by role

ElenaRepository:
added role in the cache key as proposed #8509


MatteoPiovanelli:
@sebastienros We put the feature in the OutputCache Module. In principle, it could go in the Orchard.Roles module (changing only dependencies). Let us know which one you prefer.

src/Orchard.Web/Modules/Orchard.OutputCache/Handlers/CacheByRoleHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+using Orchard.OutputCache.Models;
+using Orchard.OutputCache.Services;
+using Orchard.Roles.Events;
+using System;
+using System.Collections;
+using System.Linq;
+
+namespace Orchard.OutputCache.Handlers {
+    [OrchardFeature("Orchard.OutputCache.CacheByRole")]
+    public class CacheByRoleHandler : IRoleEventHandler {
```
sebastienros: I'd suggest instead to create a hash based on the permissions of the role, and cache this hash. Then the ICachingEventHandler can just use this value in the key. If the permissions change, the key change, and the old cache becomes obsolete after some time. That's how we do these things in OC.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+            _authenticationService = authenticationService;
+        }
+
+        public void KeyGenerated(StringBuilder key) {
+            var currentUser = _authenticationService.GetAuthenticatedUser();
+            if (currentUser != null) {
+                var roles = currentUser.As<UserRolesPart>().Roles;
+                if (roles.Any()) {
+                    // append roles in alphabetical order
+                    key.Append("UserRoles=" +
```
sebastienros: If you are given a `StringBuilder`, don't do you own concatenations.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Linq;
+using System.Text;
+using System.Web.Mvc;
+using Orchard.ContentManagement;
+using Orchard.Environment.Extensions;
+using Orchard.Mvc.Filters;
+using Orchard.Roles.Models;
+using Orchard.Security;
+
+namespace Orchard.OutputCache.Filters {
```
sebastienros, line 10: I'd rather see these files in the Roles module so they don't take a direct dependency on the module. Or try to get the roles without the UserRolesPart.

MatteoPiovanelli: Looks to me all the records that would let us compute Roles and Permissions from the User are in Orchard.Roles. Even going around UserRolesPart, the "clean" way would probably be using IRepository<UserRolesPartRecord>, and possibly Join that with other stuff from the Orchard.Roles module if we want to get the Permissions (see the comment about hashing a list of those).
On the other hand, if we moved this feature in Orchard.Roles, that would create a dependency there from Orchard.OutputCache, where the ICachingEventHandler is defined.
Orchard.Roles is one of those "Core" features that cannot be disabled from the UI. That is the main reason we set the dependency up in this direction in this PR rather than the other way around.

MatteoPiovanelli: Adding to that:
The functionality would live in its own feature either way, so enabling Orchard.OutputCache or Orchard.Roles or both would not cause it to be enabled by default.

src/Orchard.Web/Modules/Orchard.OutputCache/Handlers/CacheByRoleHandler.cs
MatteoPiovanelli: I like that alternative, because it would prevent the need to evict explicitly. I have a few considerations:

- From the User, to get all Permissions for all their roles we would need to go UserRolesPartRecord -> RoleRecord (this has the Role Name, which we may need anyway) -> RolesPermissionsRecord ->PermissionRecord (out of this I think we'd need Name and FeatureName). 
- The Name of the Role may be needed to include the list of roles in the CacheKey anyway, to handle code that specifically checks for Roles rather than permissions (e.g. https://github.com/OrchardCMS/Orchard/blob/3336be6c779e0e4ca017c0cf3ff2630f5071e3a7/src/Orchard.Web/Modules/Orchard.Roles/Activities/IsInRoleActivity.cs or the rules for layers). Having the portion of the CacheKey based on the permissions would still be enough to prevent the need for explicit eviction.
- I would use, for each Permission, something like FeatureName.Name, to reduce the change of duplicates or weird corner cases.

MatteoPiovanelli: @sebastienros  could please point out the bit in OC where you are doing what you describe. At a cursory search I did not find it, and I don't know that software well enough to reliably find it by myself.
The file I found that looks more relevant to this is https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore/OrchardCore.Infrastructure/Cache/CacheContextProviders/RolesCacheContextProvider.cs

I was looking for an ICacheContextProvider that that dealt with Permissions and I don't see it.

ElenaRepository:
Cache evict removed for the following reasons:
- added roles and permissions in the cache key (hashed)
- added anonymous users management for cache 
- added, for authenticated users, specific key building if user is "Site Owner" (because it has no specific permissions, but roles only)
We tried different implementations and picked the best one performance-wise.

src/Orchard.Web/Modules/Orchard.OutputCache/Handlers/CacheByRoleHandler.cs
sebastienros: We are not doing it for roles, it's just the technique of using hashes as part of the key so that it doesn't need to be evicted. Using a sliding expiration (or at least some absolute one) is necessary so the values don't stay cached forever.

MatteoPiovanelli: The code in this is only contributing to the cachekey. Expiration of the cache entries is controlled elsewhere, configurable through the settings of the OutputCache already.

MatteoPiovanelli:
I'll summarize the open questions regarding this PR here, for my clarity.

The added code is its own feature in the Orchard.OutputCache Module.
Wherever this feature is, it needs dependencies on the following dlls:
- Orchard.OutputCache, because it implements an interface defined there.
- Orchard.Roles, because it queries on records defined there to find roles and permissions.
In module.txt it has dependencies on these features:
- Orchard.OutputCache, because that's where the implemented filter would be used. This isn't strictly necessary: removing it would mean the filter is useless/never invoked.
- Orchard.Roles, because we need the records from its migrations to exist.
Since Orchard.Roles is a Core feature, we decided to have the new Orchard.OutputCache.CacheByRole feature in the Orchard.OutputCache module, to avoid adding a dependency to the Orchard.OutputCache project in the Orchard.Roles project. As a conseguence, the Orchard.OutputCache project now depends on the Orchard.Roles project.
Is this ok @sebastienros ?

The added code does not affect the expiration of whatever values in output cache it is involved with. It contributes to the cachekey, but those cached pages will expire following their own rules, configured elsewhere. That logic is already in place in the Orchard.OutputCache feature.

I am not 100% sure on using GetHashCode() to hash the strings used as keys. It's fast, so that's a big plus. And we aren't persisting it anywhere, so that's fine too. On application restart the OutputCache is going to be empty, so even if the result changes it's won't be a problem. I am just slightly concerned about the possibility of collisions (https://docs.microsoft.com/en-us/archive/blogs/ericlippert/socks-birthdays-and-hash-collisions). Do you have a better recommendation?

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
```diff
@@ -0,0 +1,151 @@
...
+            _rolesPermissionsRepo = rolesPermissionsRepo;
+            _permissionRepo = permissionRepo;
+        }
+
+        public void KeyGenerated(StringBuilder key) {
+            List<UserPermission> userRolesPermissions = new List<UserPermission>();
+            IQueryable<UserPermission> userRolesPermissionsQuery = Enumerable.Empty<UserPermission>().AsQueryable();
+            IQueryable<UserPermission> permissionsQuery = Enumerable.Empty<UserPermission>().AsQueryable();
+
+            var currentUser = _authenticationService.GetAuthenticatedUser();
```
sebastienros, line 40: That's a lot of queries. How often are they executed?

MatteoPiovanelli: The queries are being executed every request if the feature is active. Two are executed for authenticated users, and one for anonymous users.
We did a few hundred tests and this code did not seem to impact the page load time significantly.
The results of some of those queries (e.g. the ones getting permissions for "Authenticated" and "Anonymous" roles) could on principle be cached themselves, but since they didn't seem to be time critical we opted to not go through the hassle of having to evict them, because there's a lot of stuff that may affect those permissions.

MatteoPiovanelli:
@sebastienros  What do you think about this PR?

sebastienros (approved):

Issue #8516, opened 2021-12-03
Extend Users and  Roles capabilities and settings


HermesSbicego-Laser:
Starting from dev, we will work on adding these capabilities to Orchard.Users and Orchard.Roles Module:
1. Set a User to forcely change its own password at next LogOn;
2. Prevent reuse of last n password on password change; 
3. Suspend accounts that didn't logon within last n days;
3.1. the SiteOwner will never be suspended;
3.2. a flag over specific Users will prevent suspension for them;
4. Resend a confirmation e-mail if an "email unconfirmed" user try to re-register;
5. Have a dynamic permission to Assign roles : taking the idea from "Assign Role - {0}" implementation of OrchardCore https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore/OrchardCore.Roles.Core/CommonPermissions.cs

sebastienros:
> Resend a confirmation e-mail if an "email unconfirmed" user try to re-register

Should do some rate limiting for this one.

MatteoPiovanelli:
> Have a dynamic permission to Assign roles : taking the idea from "Assign Role - {0}" implementation of OrchardCore

Working on that now.
Accessing a User's Edit view requires the ManageUsers Permission. The "Assign Role - {0}" permission naturally wouldn't imply that, so I would add a specific view for that functionality. Accessing that would require a way to inject a link to that action in the table of users:
![image](https://user-images.githubusercontent.com/19283945/150520646-a87238ac-1dca-4c45-84c1-2d4a0042b957.png)

MatteoPiovanelli:
I think we merged everything proposed relating to this issue, so I'll close it.

PR #8523, opened 2021-12-23
extend users and roles capabilities

MatteoPiovanelli:
Implement a bunch of things proposed in #8516 

- Option to force a user to change their password on their next logon.
- Prevent reuse of recent N passwords.
- Task to suspend inactive users.
  - Providers to prevent user suspension based on a flag or on specific roles.
- Action allowing a user to resend their confirmation email.
  - Rate limiting on this would require discussion because there are several possible "conditions" to limit rate (e.g. send max N emails in a given time; wait at least a given timespan before sending further email...).
- Flag to "protect" a user from having to change their expired password.

I'm creating the PR so that we can discuss the way @HermesSbicego-Laser and I implemented this. @sebastienros please review this and involve whoever you feel is necessary for the discussion.


MatteoPiovanelli:
to be fair, this should be merged after #8521 and #7944

sebastienros (approved):

Issue #8525, opened 2022-01-04
Updating to JQuery 3.6.0


HermesSbicego-Laser:
For dev branch, we will update JQuery to its latest stable release (3.6.0.)

HermesSbicego-Laser:
Working on this, we found the the gulp pipeline is broken. 
what we see installing it, it's that package has lots of vulnerabilities (2 low, 10 moderate, 19 high, 4 critical) and some dependencies are deprecated.
installing new versions and new dependencies results in a not working gulp.js file.
So the question is: should we use gulp anymore?
We suggest and we will work on porting the gulp pipeline to this VS extension https://marketplace.visualstudio.com/items?itemName=MadsKristensen.BundlerMinifier, https://www.talkingdotnet.com/gulp-is-no-longer-default-choice-asp-net-core/
We use it in our solutions and it works as expected.

Skrypt:
If I remember correctly, Orchard used to use BundlerMinifier and got migrated to the gulp pipeline because we could fine-tune it more. There are also dependencies that only need to be copied without getting bundled so the gulp pipeline can also do that.

Moving back to the BundlerMinifier won't fix the deprecated dependencies issue either. So, I believe this would be more regression than anything. I would try to fix the npm packages and dependencies before anything else. Because in the end, that's what's needed.

HermesSbicego-Laser:
Thanks @Skrypt for clarifying why gulp was choosen.
So, we will try to update and keep gulp pipeline. Any help will be appreciated.

MatteoPiovanelli:
@Skrypt 
Based also on your comments in #8522, what would you recommend?
We think the pipeline should be updated to a state where it works "out of the box", and it possibly easy to set up for new and custom modules/themes.

Skrypt:
What would need to be done is to use the latest LTS version of Node.js + npm and update the packages.json files up until no package deprecated are found. After that, test things manually in OC to make sure the components are still working. Because it can "transpile" without necessarily working.

That's why I said earlier that normally using the Node.js version that was used back then to transpile everything should be the easiest way to make it transpile and work.

Skrypt:
Also as part of being able to do reproducible builds, I think we should document the version of Node.js used to transpile the Gulp pipeline to avoid having this situation again in a few years. The recommendation is good for Orchard Core too.

PR #8531, opened 2022-01-18
Upgrade/8525 j query3.6 bootstrap4.6

HermesSbicego-Laser:
Upgrade to jQuery 3.6, jQueryUI 1.13 and Boostrap 4.6.1
From our tests everything seems to work correctly, if someone can take a look into this, they are welcome.
We will continue to test it during next days anyway


src/Gulpfile.js
```diff
@@ -237,7 +237,7 @@ function buildJsPipeline(assetGroup, doConcat, doRebuild) {
         ))
         .pipe(plumber())
         .pipe(gulpif(generateSourceMaps, sourcemaps.init()))
-        .pipe(typescript(typeScriptOptions))
+        .pipe(typescript(typeScriptOptions)) // typescript does not support javascript modules
```
Skrypt, line 240: Typescript should support javascript modules out of the box. I would need to see an example of what you are refering to. An alternative is to build the modules assets with whatever works and to use the gulp pipeline to copy over the bundled assets. If you want to automate that build another option is to call for example Webpack from Gulp.

src/Orchard.Web/Modules/Orchard.Layouts/Views/EditorTemplates/LayoutEditor.cshtml
```diff
@@ -26,9 +26,6 @@
     Style.Include("LayoutEditor.css", "LayoutEditor.min.css");
     Script.Include("LayoutDesignerHost.js");
 
-    // The grid system.
-    Style.Include("default-grid.css");
```
Skrypt, line 30: This removes probably extensibility.

sebastienros: Is that related to the `Parent: Row` thing? People use it on the front-end for custom themes.

MatteoPiovanelli: This is only in the editor shape for layouts.
I think for front-end you are referring to what's in TheThemeMachine theme? That's a separate css file (with the same name and perhaps same content) that is included in the layout there.

Here in the backoffice editor for the layout, the content of that stylesheet "conflicts" with bootstrap's grid system. I would argue that the stylesheet itself should also be removed from Orchard.Layouts.
In case someone would want to extend styling for this editor, as it is, they would have to go in the sources and change that stylesheet, and at that point one might as well add a `Style.Include("my-new.css");` line in LayoutEditor.cshtml.

src/Gulpfile.js
MatteoPiovanelli: I am talking specifically about gulp-typescript here, because I'm not familiar enough with typescript to make blanket statements.
Typescript does support javascript modules, as long as there also are other dependencies. I mean to say that the transpilation step for modules that are written in .js files results in javascript that depends on other libraries.
For example, it could create javascript that depends on system.js, or amd.js, but doesn't seem to be able to produce javascript that stands on its own. The specific library the js will depend on is "selected" by a property in typeScriptOptions, but the library itself is not automatically included (it would be hard to do so at this stage), so the end result is that the resulting js fails to execute.

src/Orchard.Web/Modules/Orchard.Layouts/Views/EditorTemplates/LayoutEditor.cshtml
sebastienros: Should not be removed from Orchard.Layouts, because that's the default to use when it's not themed. If there is a conflict on the admin, then we need to ensure that all existing classes are also in bootstrap. Or change the names of the classes to they don't collide.

src/Gulpfile.js
Skrypt: Have you tried updating gulp-typescript?
"gulp-typescript": "3.1.4",

Because as you can see here :  

https://github.com/OrchardCMS/OrchardCore/blob/5310bdb49797e9db02f3107ea9c0feb270e92346/package.json#L54

We are using a newer version. And also, when you say that it produces a non-working javascript file maybe that's because we were back then using a lower version of TypeScript too. So, one thing that could be done here is to try updating TypeScript to the latest version and try to transpile manually these modules assets. Then, if it works, update the gulp-typescript version to the latest version too, and see if that works.

We would need to update the Documentation about the Gulp pipeline so that we know which exact version everything was built with including the TypeScript version used on the system it was built with.

Skrypt: Basically it should support TypeScript 2.2.0

https://github.com/ivogabe/gulp-typescript/pull/457/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R83

BenedekFarkas:
It sucks that we left this PR here, but what do you think about upgrading to 3.7 now too?

HermesSbicego-Laser:
I comment on this PR in order to resume it and try to finalize the work.
@BenedekFarkas @Skrypt @sebastienros how could we proceed to make this PR part of the dev branch?

BenedekFarkas:
I'd prefer to merge it with the affected libraries updated to the latest stable version (I can help with that). We'll also need some time to test it in a client project (although that's already using jQuery 3.6 for custom components). But this really is a much-needed update!

HermesSbicego-Laser:
@BenedekFarkas can we plan a dedicated meeting to bring this PR to 1.10.x or dev? In particular we are not able to make the gulp pipeline working properly see #8525.

BenedekFarkas:
> @BenedekFarkas can we plan a dedicated meeting to bring this PR to 1.10.x or dev? In particular we are not able to make the gulp pipeline working properly see #8525.

Yes, sounds good! Sometime next week works for me, I'll check out the corresponding issue(s)/PR(s) and get back to you!

MatteoPiovanelli:
@BenedekFarkas I'll ping you on Gitter so we can sync for a call

BenedekFarkas:
Sorry, I don't really use Gitter - BTW on the topic of communications, if you and/or your team use Discord, then I'd prefer that, and we can set up a channel there just for preparing for 1.11.

HermesSbicego-Laser:
> Sorry, I don't really use Gitter - BTW on the topic of communications, if you and/or your team use Discord, then I'd prefer that, and we can set up a channel there just for preparing for 1.11.

Hi @BenedekFarkas I can propose you to join us on Discord next tuesday at 11AM Italian time.
Please contact me at hermes.sbicego(at)laser-group.com for sharing details

BenedekFarkas:
@HermesSbicego-Laser I had to update jQueryUI to 1.13.2 on dev earlier, because some admin UI features were broken, see 8dec61baaba9828ac02214f4dde68e2506b9a8ff. Can you please revert the jQueryUI changes in this PR and merge from dev? If you can't work on it now, then I can do that too (but I'll need to push the changes into a new branch in the Orchard repo).

BenedekFarkas:
I'll continue working on this in #8848.

PR #8535, opened 2022-01-26
Roles management

MatteoPiovanelli:
Point 5 of #8516 
I took the logic from what's in Orchard Core.

I added a ViewUsers permission in Orchard.Users to access the backoffice list of users on the site. This way even if an operator doesn't have ManageUsers they may still be able to see the list for other functionalities (such as what's implemented in this PR).

I added the ability for providers to inject additional links in the "Actions" column of the backoffice list of users, while the "original" actions still depend on the ManageUsers Permission. The net result from this PR is:
image

I added, in OrchardRoles, dynamic permissions allowing to assign/remove specific roles from users. I also added an action allowing access to that functionality from a separate page than the user's Edit page, to prevent mismatched permissions from causing weird situations.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Users/Services/IUserManagementActionsProvider.cs
+++ b/src/Orchard.Web/Modules/Orchard.Users/Services/IUserManagementActionsProvider.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Web.Mvc;
+using Orchard.Security;
+
+namespace Orchard.Users.Services {
+    public interface IUserManagementActionsProvider : IDependency {
+        // Using a delegate so implementations don't have to build/figure out
+        // their own HtmlHelper
+        IEnumerable<Func<HtmlHelper, MvcHtmlString>> UserActionLinks(IUser user);
+    }
+}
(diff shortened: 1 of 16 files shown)


src/Orchard.Web/Modules/Orchard.Users/Services/IUserManagementActionsProvider.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Web.Mvc;
+using Orchard.Security;
+
+namespace Orchard.Users.Services {
+    public interface IUserManagementActionsProvider : IDependency {
+        // Using a delegate so implementations don't have to build/figure out
+        // their own HtmlHelper
+        IEnumerable<Func<HtmlHelper, MvcHtmlString>> UserActionLinks(IUser user);
```
sebastienros, line 13: UI composition, usually we use drivers for that

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Users/Services/IUserManagementActionsProvider.cs
MatteoPiovanelli: Fair. However the Action implementations of this are used in is building its own viewmodel (for its own view). I've looked at using drivers, but the "cleanest" way I found for that would have required me to rewrite a lot of Orchard.Users' AdminController so the list of existing users was generated more like the list of contents (basically, displaying SummaryAdmin views for each user, rather than a flat "custom" table). Not that we shouldn't get on that someday...

MatteoPiovanelli: if/when we do, implementations of this interface can be called in one of those drivers, or we could get rid of them and replace them with drivers.

sebastienros (approved):

Issue #8537, opened 2022-02-14
Cloning of widgets inside a layout


AndreaPiovanelli:
When I clone a layout, the widgets inside it aren't really cloned, but a reference to the original widget is saved.
This causes a breaking issue when editing or deleting the child widget, which gets edited or deleted across multiple layouts (because they have the same id).
Case scenario:
- Creation of Layout widget 1 (with id 1), with child widget 1 (with id 2)
- Cloning of Layout widget 1
- The cloned content is Layout widget 2 (id = 3), with child widget 1 inside it (id = 2!)
To avoid this problem, I need to edit the ImportElement routine in order to create a brand new widget (forcing its cloning) only when the original container is different from the clone container (this condition ensures it's a cloning operation and not a standard import).

The line of code I'm working on is:
https://github.com/OrchardCMS/Orchard/blob/88a79cb7688775856596d57a16fa09ab631967e0/src/Orchard.Web/Modules/Orchard.Widgets/Layouts/Providers/WidgetElementHarvester.cs#L188

PR #8538, opened 2022-02-14
Layout elements cloning

AndreaPiovanelli:
Created a new element (cloning the old one) when session widget has a different container than the widget in the import context (to ensure actual cloning of the elements and not just adding a reference to the old element from the original container).

Referenced issue: #8537

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Layouts/Providers/WidgetElementHarvester.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Layouts/Providers/WidgetElementHarvester.cs
@@ -13,6 +13,7 @@ using Orchard.Layouts.Framework.Drivers;
 using Orchard.Layouts.Framework.Elements;
 using Orchard.Layouts.Framework.Harvesters;
 using Orchard.Layouts.Helpers;
+using Orchard.Layouts.Models;
 using Orchard.Mvc.Html;
 using Orchard.Security;
 using Orchard.Widgets.Layouts.Elements;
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Widgets/Layouts/Providers/WidgetElementHarvester.cs
```diff
@@ -185,6 +186,18 @@ public class WidgetElementHarvester : Component, IElementHarvester {
                 return;
 
             var widget = context.Session.GetItemFromSession(widgetIdentity);
+
+            // I need to create a new widget and save the new one.
+            // This is to avoid the fact the very same element ending up in multiple layouts, causing issues when e.g. deleting a LayoutWidget of a cloned ContentItem (which would delete the elements of multiple layouts).
+            // I only do this when the container of the original element is different then the container of the cloned element, to ensure doing it when cloning elements and avoid doing the same when importing content.
+            var cp = widget.As<ICommonPart>();
```
sebastienros, line 193: I don't think this should be done here. What about in the Layout handlers when it's cloned instead?

AndreaPiovanelli: I used the WidgetElementHarvester because this issue is specific to widget elements into a layout. When layout is cloned, I had to force the cloning of its child widgets, too, because new layout needed to reference its own widgets and not the original ones.
The issue triggered while importing widget elements: when I cloned a layout (L1) with a widget element (W1), the resulting clone was a layout (L2) with widget element (W1). W1 kept referencing L1 as its container, but both L1 and L2 had W1 in their own child elements. 
For this reason I believe WidgetElementHarvester ImportElement is the right place to apply the patch.

AndreaPiovanelli: @sebastienros any news on this?
Thank you,
Andrea

src/Orchard.Web/Modules/Orchard.Widgets/Layouts/Providers/WidgetElementHarvester.cs
```diff
@@ -185,6 +186,18 @@ public class WidgetElementHarvester : Component, IElementHarvester {
                 return;
 
             var widget = context.Session.GetItemFromSession(widgetIdentity);
+
+            // I need to create a new widget and save the new one.
```
sebastienros: Avoid using "I" in comments. Prefer the past form: "A new widget needs to be created and saved".
Then read all the comments you have ever written and see these "I"s in most of them ;)

sebastienros (approved):

Issue #8541, opened 2022-03-15
Fixed password history policy in back office user password changes


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/f2a8450d9093471efb3143d099c71e075a9b848b/src/Orchard.Web/Modules/Orchard.Users/Drivers/UserPartPasswordDriver.cs#L38

When I change the password of a user from back office, password history policy fails because the password requirement check is after new password is set for the user. For this reason, password change always fails (because new password is already into forbidden password list when executing PasswordHistoryService.GetLastPasswords ("current" password is appended to the results).

PR #8542, opened 2022-03-15
8541 password history

AndreaPiovanelli:
Corrected password history policy enforcement for user password changes in back office.

#8541


Skrypt:
Is it not something that was done separately. Because it is not implied that if you edit a user that you want to change its password?

AndreaPiovanelli:
> Is it not something that was done separately. Because it is not implied that if you edit a user that you want to change its password?

Yes, it is: to edit passwords in the back office you need to activate the proper feature ("User password editor") through the "modules" menu. Password history policy was not working properly in that situation, as described in the linked issue #8541

sebastienros (approved):

PR #8545, opened 2022-03-16
Fix/tinymce medialibrary

### LINKED ISSUE #8544: MediaLibrary Plugin for TinyMCE only works on some editors
https://github.com/OrchardCMS/Orchard/blob/88a79cb7688775856596d57a16fa09ab631967e0/src/Orchard.Web/Modules/TinyMce/Scripts/plugins/medialibrary/plugin.js#L20

The way the javascript is written there, this plugin only works when the editor is loaded for a content that is being edited on the "main" editor for contents (actually, for a subset of editors whose controllers have specific paths). For example, this plugin won't work for a BodyPart in a Term, but it will work in a BlogPost and a Page.

I will push a PR momentarily with a proposal to fix this.

MatteoPiovanelli:
fixes #8544


MatteoPiovanelli:
I think I created this off the wrong branch...

PR #8551, opened 2022-03-23
Fixes #8550

### LINKED ISSUE #8550: ManageUsersInRole Permissions
https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore/OrchardCore.Users.Core/CommonPermissions.cs

In line with the developments we are doing to upgrade the capabilities for user management, we propose to port the `ManageUsersInRole` Permission from OCore.

I have a PR for it incoming.

MatteoPiovanelli:
Added permissions allowing management of users belonging to specific roles only based on what's implemented in Orchard Core.
See Here:
https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore/OrchardCore.Users.Core/CommonPermissions.cs

and Here:
https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Users/Services/UserAuthorizationHandler.cs

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Roles/Security/ManageUserByRolePermissions.cs
+++ b/src/Orchard.Web/Modules/Orchard.Roles/Security/ManageUserByRolePermissions.cs
@@ -1,0 +1,64 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Orchard.Data;
+using Orchard.Environment.Extensions.Models;
+using Orchard.Roles.Constants;
+using Orchard.Roles.Models;
+using Orchard.Security.Permissions;
+using UserPermissions = Orchard.Users.Permissions;
+
+namespace Orchard.Roles.Security {
+    public class ManageUserByRolePermissions : IPermissionProvider {
+        private readonly IRepository<RoleRecord> _roleRepository;
+
+        public virtual Feature Feature { get; set; }
+
+        private static readonly Permission ManageUsersInRoleTemplate =
+            new Permission {
+                Description = "Manage Users in Role - {0}",
+                Name = "ManageUsersInRole_{0}",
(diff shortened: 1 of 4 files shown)


src/Orchard.Web/Modules/Orchard.Roles/Security/ManageUserByRolePermissions.cs
```diff
@@ -0,0 +1,64 @@
...
+            };
+        }
+
+
+        private IEnumerable<Permission> GetManageUsersInRolePermissions() {
+            var allRoleNames = _roleRepository.Table
+                .Select(r => r.Name)
+                .ToList()
+                // Never have to manage explicitly Anonymous or Authenticated roles
+                .Except(SystemRoles.GetSystemRoles());
```
sebastienros, line 46: Reason?

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Roles/Security/ManageUserByRolePermissions.cs
MatteoPiovanelli: The roles from the constant `SystemRoles.GetSystemRoles()` are never explicitly managed anywhere else.
Anonymous: It makes no sense to have a permission to manage "Anonymous" users, since they are not going to be there to manage.
Authenticated: The way we've implemented things here, we've followed the same idea we found implemented in Orchard Core: If you can manage users in any of the other "non system" roles, you can manage "authenticated" basic" users, so we don't have an explicit permission for that role.

sebastienros: Good!

Issue #8554, opened 2022-04-07
Impossible to replace "missing" media because DeleteFile throws a exception


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/ffb56e684ce39ef65387793730d3c6d5ecfcf3ed/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/ClientStorageController.cs

When trying to replace a "broken" media file (Action Replace in ClientStorageController), if the file has been deleted from filesystem, the process stops because a exception is thrown (by FileSystemStorageProvider.DeleteFile(string)), so the media can't be replaced.

Skrypt:
Could you share an error log example as a reference?

AndreaPiovanelli:
> Could you share an error log example as a reference?

The error status (the one displayed to the user when uploading the replacement file) is the following:

> File Foto\image.jpg does not exist

The error log is:

> 2022-04-08 09:04:31,656 [43] Orchard.MediaLibrary.Controllers.ClientStorageController - TenantName - Unexpected exception when uploading a media. [BaseHttpUrl/Orchard.MediaLibrary/ClientStorage/Replace]
> System.ArgumentException: File Foto\image.jpg does not exist
>    at Orchard.FileSystems.Media.FileSystemStorageProvider.DeleteFile(String path) in OrchardRoot\src\Orchard\FileSystems\Media\FileSystemStorageProvider.cs:line 254
>    at Orchard.MediaLibrary.Services.MediaLibraryService.DeleteFile(String folderPath, String fileName)
>    at Orchard.MediaLibrary.Controllers.ClientStorageController.Replace(Int32 replaceId, String type)

The file Foto\image.jpg is the original file, the one missing from filesystem. If you import a new media with the same file name, the "old" media (the broken one) works again properly (and you have two identical media files displayed, because both reference the same file in the filesystem). This is a sort of workaround to the issue, because afterwards you may delete the new media getting the old one to work again.

AndreaPiovanelli:
Closing the issue because PR #8555 has been successfully merged.

PR #8555, opened 2022-04-07
Fix to replace media when original file has been deleted from filesystem

AndreaPiovanelli:
Referencing issue #8554 
Added a try-catch statement around MediaLibraryService.DeleteFile call inside Replace action. ArgumentException thrown by FileSystemStorageProvider.DeleteFile in case the original file is missing interrupted the actual replace of the media, making how to solve this uncommon scenario unclear to the back office operator (he needed to delete the original media and import another media with the exact same name to make it work properly again). With this patch now the replace operation isn't stopped anymore, adding to the status message hints to the anomaly ("can't delete original file") but completing the replacement of the "broken" file.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/ClientStorageController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/ClientStorageController.cs
@@ -178,7 +178,14 @@ namespace Orchard.MediaLibrary.Controllers {
                                                                 .Where(x => x.FolderPath == replaceMedia.FolderPath && x.FileName == replaceMedia.FileName)
                                                                 .Count();
                     if (mediaItemsUsingTheFile == 1) { // if the file is referenced only by the deleted media content, the file too can be removed.
-                        _mediaLibraryService.DeleteFile(replaceMedia.FolderPath, replaceMedia.FileName);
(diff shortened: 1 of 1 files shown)


sebastienros (approved):

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/ClientStorageController.cs
```diff
@@ -178,7 +178,14 @@ public class ClientStorageController : Controller {
                                                                 .Where(x => x.FolderPath == replaceMedia.FolderPath && x.FileName == replaceMedia.FileName)
                                                                 .Count();
                     if (mediaItemsUsingTheFile == 1) { // if the file is referenced only by the deleted media content, the file too can be removed.
-                        _mediaLibraryService.DeleteFile(replaceMedia.FolderPath, replaceMedia.FileName);
+                        try {
+                            _mediaLibraryService.DeleteFile(replaceMedia.FolderPath, replaceMedia.FileName);
+                        } catch (ArgumentException) { // File not found by FileSystemStorageProvider is thrown as ArgumentException.
```
Skrypt, line 183: Code formatting. Use Visual Studio to reformat it properly.

Issue #8557, opened 2022-05-02
Taxonomy Id checking before executing queries in Orchard.Taxonomies (performance fix/upgrade).


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/aff9b2542cc8733c9fce1a1c6e6d172c0074cf7c/src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs#L67

Everytime a function in TaxonomyService has the taxonomy id parameter, performance would be improved checking id validity (id > 0) before executing queries. This mostly is a sanity check but, in some cases, "useless" queries are executed for id = 0.

PR #8558, opened 2022-05-02
8557 check valid taxonomy

AndreaPiovanelli:
#8557 
Added id checks on every function executing queries on taxonomy or term id to avoid executing the query when the parameter is not valid (id <= 0) and avoid some queries when they would clearly return a empty result.
This mostly is a sanity check but, in some scenario, id = 0 is passed as a parameter and a query on the database is executed, giving no results.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
@@ -223,6 +223,10 @@ namespace Orchard.Taxonomies.Services {
         }
 
         public TermPart GetTerm(int id) {
+            // If term id isn't valid, return null without executing the query.
+            if (id <= 0) {
+                return null;
+            }
             return GetTermsQuery()
                 .Where(x => x.Id == id).List().FirstOrDefault();
         }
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
```diff
@@ -169,6 +169,10 @@ public class TaxonomyService : ITaxonomyService {
         }
 
         public IEnumerable<TermPart> GetTerms(int taxonomyId) {
+            // If taxonomyId isn't valid, return a empty list without executing the query.
+            if (taxonomyId <= 0) {
+                return new List<TermPart>();
```
sebastienros: You could return `Array.Empty<TermPart>()`, for no allocations.

sebastienros (approved):

sebastienros (approved):

sebastienros:
Did you find out how these ids are < 0 ?

AndreaPiovanelli:
> Did you find out how these ids are < 0 ?

It is, in theory, possible to call those functions passing a negative int as a parameter, even if it should never happen and I have never actually seen it.
A zero value is passed, for instance, when building taxonomy related menus.

sebastienros:
Looks good

Issue #8569, opened 2022-06-16
Blog Search


AndreaPiovanelli:
Search feature in blogs does not exist, so we are going to implement it in a new feature inside Orchard.Search module (feature: Orchard.Search.Blogs).
![image](https://user-images.githubusercontent.com/83577153/174037267-c2c8e018-c083-46fa-9ea8-e3fd986607e0.png)

Inside the manage blog page in back office, there's going to be a "Search" tab:
![image](https://user-images.githubusercontent.com/83577153/174037069-ed6f6666-ddf1-4a5f-88f1-4d03e93b2eca.png)

New index used for the search implementation:
![image](https://user-images.githubusercontent.com/83577153/174036862-671e0b55-08cf-407e-9596-ae37524bbfea.png)

sebastienros:
It's a search feature just for admin?
Does O1 support multiple blogs? If so why a single index?
Aren't blog posts already indexes somewhere?

AndreaPiovanelli:
> It's a search feature just for admin?  

Yes, it is for back office only. Front end search already works with a standard index, with any content.

> Does O1 support multiple blogs? If so why a single index?

It supports multiple blogs. We use a single index because it's a single content type and, when searching, it's easier filtering results (for the needed blog) than having to create indexes for each blog. 

> Aren't blog posts already indexes somewhere?

Not yet. Admin search index at the moment can be used to index any content but, since blog search is on a separate controller, we chose to create a dedicated admin blog index.

PR #8570, opened 2022-06-16
Feature/blogposts search

AndreaPiovanelli:
As described in #8569


src/Orchard.Web/Modules/Orchard.Blogs/Views/BlobPosts_NavigationTabs.cshtml
```diff
@@ -0,0 +1 @@
+﻿aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
sebastienros: left-over

AndreaPiovanelli:
@sebastienros I removed the leftover you pointed out and answered your questions on issue #8569; should we merge this pr?

sebastienros:
Can't the feature be part of the Blogs module instead of the Search one? Just a matter of what depends on what.

AndreaPiovanelli:
> Can't the feature be part of the Blogs module instead of the Search one? Just a matter of what depends on what.

We chose to add the feature to Orchard.Search module because there already are features referencing third modules inside it (e.g. Orchard.Search.MediaLibrary, depending on Orchard.MediaLibrary).

sebastienros (approved):

PR #8574, opened 2022-06-21
Fix/cache by role exception

### LINKED ISSUE #8573: Cache By Role - Exception during CacheKeyGenerated
When generating a cache key based on a User belonging to a Role that has the SiteOwner permission we log a System.NotSupportedException: new UserPermission() {RoleName = [100001].Name} 
We are working on a PR: from what we see, probably it depends on how NHibernate prepare the query and when the query is concretly executed. An earlier .ToList() seems to solve the problem.

MatteoPiovanelli:
fixes #8573 
This also fixes a couple other issues we found while investigating that with "colliding" cache keys

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
@@ -24,6 +24,7 @@ namespace Orchard.OutputCache.Filters {
             IRepository<RoleRecord> roleRepo,
             IRepository<RolesPermissionsRecord> rolesPermissionsRepo,
             IRepository<PermissionRecord> permissionRepo) {
+
             _authenticationService = authenticationService;
             _authorizer = authorizer;
             _userRolesRepo = userRolesRepo;
@@ -33,6 +34,8 @@ namespace Orchard.OutputCache.Filters {
         }
 
         public void KeyGenerated(StringBuilder key) {
+            // Can the queries in this method be optimized away so that their results can be memorized
+            // at least within the scope of a request?
             List<UserPermission> userRolesPermissions = new List<UserPermission>();
             IQueryable<UserPermission> userRolesPermissionsQuery = Enumerable.Empty<UserPermission>().AsQueryable();
             IQueryable<UserPermission> permissionsQuery = Enumerable.Empty<UserPermission>().AsQueryable();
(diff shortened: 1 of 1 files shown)


AndreaPiovanelli:
@sebastienros any question on this pr?

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
```diff
@@ -85,25 +97,23 @@ public class CacheByRoleFilter : ICachingEventHandler {
...
+            }
+            else {
                 // the anonymous user has no roles, get its permissions
-                permissionsQuery = GetPermissioFromRole("Anonymous");
+                permissionsQuery = GetPermissionsFromRole("Anonymous");
             }
 
             if (userRolesPermissionsQuery.Any()) {
                 userRolesPermissions.AddRange(userRolesPermissionsQuery
-                    .OrderBy(urp => urp.RoleName)
```
sebastienros, line 95: I believe the order matter since it will change the key if the order changes.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
```diff
@@ -118,12 +128,37 @@ public class CacheByRoleFilter : ICachingEventHandler {
...
+                    .Distinct() // sanity check
+                    ;
+            }
+
+            var roleName = _siteOwnerRoleName;
+            if (_siteOwnerRoleNames.Any() && _siteOwnerRoleNames.Contains(roleName)) {
+                // compute unique and repeatable roleName
+                var i = 0;
+                do {
+                    roleName = $"{_siteOwnerRoleName}-{i}";
```
sebastienros, line 155: `i` never changes?

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
AndreaPiovanelli: Roles and permissions are ordered a few rows afterwards.

src/Orchard.Web/Modules/Orchard.OutputCache/Filters/CacheByRoleFilter.cs
AndreaPiovanelli: Fixed, thanks.

sebastienros (approved):

PR #8596, opened 2022-06-30
Fix/newtonsoft.json 13.0.1

HermesSbicego-Laser:
This PR Updates Newtonsoft.Json to 13.0.1: it updates web.config (binding redirects), packages.config (nuget) and *.csproj files (projects); It can be used instead of merging those 22 PR made by dependabot few weeks ago.

### DIFF
--- a/src/Orchard.Azure.Tests/App.config
+++ b/src/Orchard.Azure.Tests/App.config
@@ -29,7 +29,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-12.0.0.0" newVersion="12.0.0.0" />
+        <bindingRedirect oldVersion="0.0.0.0-13.0.0.0" newVersion="13.0.0.0" />
       </dependentAssembly>
       <dependentAssembly>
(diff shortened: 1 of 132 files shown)


src/Orchard.Azure.Tests/App.config
```diff
@@ -29,7 +29,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-12.0.0.0" newVersion="12.0.0.0" />
+        <bindingRedirect oldVersion="0.0.0.0-13.0.0.0" newVersion="13.0.0.0" />
```
sebastienros, line 32: should it be 13.0.1 ?

sebastienros (commented):
OK if the binding redirect is correct.

src/Orchard.Azure.Tests/App.config
HermesSbicego-Laser: 13.0.0.0 is the assembly version
13.0.1 is the nuget version.
So 13.0.0.0 is the good one for the binding redirect.

sebastienros: Good to go then

MatteoPiovanelli:
Just double-checking to confirm this is gtg to merge

Issue #8602, opened 2022-07-18
Path too long when processing image profiles


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/343ceb777b3aaa786ed6581e78ecd6a383c5fea1/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs#L236

When creating profile files for media, sometimes and exception is thrown because file path is longer then the maximum path length (260 characters, ref: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation).

The exception is the following:

> 2022-07-11 16:53:00,098 [5] Orchard.MediaProcessing.Services.ImageProfileManager - TenantName - A profile could not be processed: /Media/TenantName/Folder/VeryLongFileName.jpeg [https://thesite.com/Search/Media/MediaItems?folderPath=&skip=0&count=20&order=created&mediaType=&search=sacra&_=1657551171829]
System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\PathToOrchardWeb\Media\Tenant_Profiles\b0724b75\e03e630\VeryLongFileName.jpeg'.
in System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
in System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
in System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
in System.IO.File.InternalWriteAllBytes(String path, Byte[] bytes, Boolean checkHost)
in Orchard.FileSystems.Media.FileSystemStorageProvider.CreateFile(String path)
in Orchard.MediaProcessing.Services.ImageProfileManager.GetImageProfileUrl(String path, String profileName, ContentItem contentItem, FilterRecord[] customFilters)

PR #8603, opened 2022-07-18
Profile path too long when processing images

AndreaPiovanelli:
Reference issue #8602 

When generating a image profile, sometimes resulting path can be very long, especially when a user imports a media with a very long file name. As described into the referenced issue, this can throw exceptions and stop Orchard from processing, for instance, resized images displayed as thumbnails in the media gallery.
Since profile folders are already hashed, when file name is too long I hashed it the same way, eventually trimming it even more in extreme cases (this may happen when path is really long even before appending the file name).

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
@@ -171,6 +171,8 @@ namespace Orchard.MediaProcessing.Services {
                                             }
                                         }
                                     }
+                                    // the storage provider may have altered the filepath
+                                    filterContext.FilePath = newFile.GetPath();
                                 }
                             }
                             catch(Exception e) {
--- a/src/Orchard/FileSystems/Media/FileSystemStorageProvider.cs
+++ b/src/Orchard/FileSystems/Media/FileSystemStorageProvider.cs
@@ -34,9 +34,26 @@ namespace Orchard.FileSystems.Media {
             _publicPath = appPath + "Media/" + settings.Name + "/";
 
             T = NullLocalizer.Instance;
+            MaxPathLength = 260;
         }
 
         public Localizer T { get; set; }
+
+        public int MaxPathLength {
+            get; set;
+            // The public setter allows injecting this from Sites.MyTenant.Config or Sites.config, by using
+            // an AutoFac component:
+            /*
+             <component instance-scope="per-lifetime-scope"
+                type="Orchard.FileSystems.Media.FileSystemStorageProvider, Orchard.Framework"
+                service="Orchard.FileSystems.Media.IStorageProvider">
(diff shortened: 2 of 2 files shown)


src/Orchard.Web/Modules/Orchard.MediaProcessing/Services/ImageProfileManager.cs
```diff
@@ -238,6 +256,30 @@ public class ImageProfileManager : IImageProfileManager {
             var filenameWithExtension = Path.GetFileName(path) ?? "";
             var fileLocation = path.Substring(0, path.Length - filenameWithExtension.Length);
 
+            // If absolute path is longer than the maximum path length (260 characters), file cannot be saved.
+            // File name is hashed to avoid homonyms when trimming similar file names.
+            var filenameWithoutExtension = Path.GetFileNameWithoutExtension(path).GetHashCode().ToString("x");
```
sebastienros: `GetHashCode()` is not stable, it might return different values (even if it works locally most of the time).

You should use a stable hashing algorithm, like SHA256, MD5, ...

AndreaPiovanelli: I used GetHashCode() because it was already used to hash profile subfolders. The purpose is to avoid very large file and folder names; hashing with stable algorithms generates large strings (32 characters with MD5, 64 with SHA256). I may use them (and, for the same reason, I should use them to generate folder names too), but then I'd have to truncate those large strings to limit path length.

AndreaPiovanelli:
As we noticed issues when checking paths via out of context calls (e.g. calling the methods via rest apis), we realized ImageProfileManager was the wrong class to test for path format. For this reason, that logic has been moved inside the file storage provider. 

We used the same pr as it hasn't been merged yet. Can you check this, @sebastienros?

src/Orchard/FileSystems/Media/FileSystemStorageProvider.cs
```diff
@@ -335,6 +352,17 @@ public class FileSystemStorageProvider : IStorageProvider {
             if (!Directory.Exists(dirName)) {
                 Directory.CreateDirectory(dirName);
             }
+            //Path.GetFileNameWithoutExtension(fileInfo.Name)
+            // If absolute path is longer than the maximum path length (260 characters), file cannot be saved.
+            if (fileInfo.FullName.Length > MaxPathLength) {
+                var fileName = Path.GetFileNameWithoutExtension(fileInfo.Name);
+                var extension = fileInfo.Extension;
+                // try to generate a shorter path for the file
+                var nameHash = fileName.GetHashCode().ToString("x");
```
sebastienros, line 361: I think I already mentioned that but hashing is not stable. If the app restart the values might differ for the same input. There are stable hashes (SHA256, MD5) that can be used.

sebastienros (approved):

PR #8628, opened 2022-10-07
Prevent caching GetTaxonomy Action

HermesSbicego-Laser:
That action is called by LocalizedTaxonomyField editor view and need to not serve cached data


MatteoPiovanelli:
@sebastienros 
to clarify this:
If the results results from that action are cached, the backoffice UI for taxonomyfields in localizable contentItems may result inconsistent after an update where the selected terms have been changed.

sebastienros (approved):

PR #8632, opened 2022-10-11
Fix for issue 8631: field values are reset when imported

### LINKED ISSUE #8631: Field importing resets mlpf ContentItems if mlpf isn't included in imported xml
https://github.com/OrchardCMS/Orchard/blob/a402c82c23c726911dd0e704b3985e76de43c529/src/Orchard.Web/Modules/Orchard.MediaLibrary/Drivers/MediaLibraryPickerFieldDriver.cs#L75

When importing a content field (e.g. when updating the value of a field inside a recipe) and a MediaLibraryPickerField is present inside the content item, if mlpf is not included in the imported xml, its value gets reset.

Example xml snippet:
```
      [...]
      <BooleanField.SomeField Value="false" />
      <MediaLibraryPickerField.SecondImage ContentItems="/Identifier=something" />
      [...]
```
Note that inside the content item there is another MediaLibraryPickerField, called "FirstImage", which would get reset (field.Ids = new int[0]).

After analyzing and testing the code, we noticed that the same issue applies to ContentPickerField and NumericField import procedures. These field are going to be corrected with the same pr connected to the current issue.

AgostiniAlessandro:
This fixes #8631 , adding a check on the existence of the importing element inside the context. If the field isn't inside the context, the field is not imported (previously its value was being reset).

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.ContentPicker/Drivers/ContentPickerFieldDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.ContentPicker/Drivers/ContentPickerFieldDriver.cs
@@ -108,9 +108,7 @@ namespace Orchard.ContentPicker.Drivers {
                     .Select(context.GetItemFromSession)
                     .Select(contentItem => contentItem.Id).ToArray();
             }
-            else {
-                field.Ids = new int[0];
-            }
+            // If nothing about the field is inside the context, field is not modified.
         }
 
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/Orchard?pullRequest=8632) <br/>All CLA requirements met.

sebastienros (approved):

PR #8638, opened 2022-11-04
Added Username policies

AgostiniAlessandro:
Added custom policies for the username during registration (at frontend) and creation and edit (at backend).

The policies are:
- Minimum username length
- Maximum username length
- Username must not contain whitespaces
- Username must not contain special characters
- Emails are allowed as username (enabled only if the previous policy is selected)

Added specific ValidationAttributes to check the settings are configured correctly.


AndreaPiovanelli:
We are refining policy checks in different scenarios (backoffice user creation and edit, front-end registration, etc) and testing various situations. For this reason, even if pr may be approved, we are holding the actual merge until more corner cases are managed correctly.

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Users/Annotations/UsernameValidLengthAttribute.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Orchard.Localization;
+using Orchard.Users.Models;
+
+namespace Orchard.Users.Annotations {
```
sebastienros: You can delete it. It's ok to not validate this, and the way it's made as an attribute doesn't make sense since it's not reusable, it's checking explicitly which properties it's applied to. It's less ugly to just remove the check.

src/Orchard.Web/Modules/Orchard.Users/Services/UsernameValidationError.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web;
+using Orchard.Localization;
+
+namespace Orchard.Users.Services {
+
+    public enum Severity {
+        Warning,
```
sebastienros, line 10: There is no such concept in Orchard, please don't add this. A form is either valid or invalid.

src/Orchard.Web/Modules/Orchard.Users/Annotations/UsernameValidLengthAttribute.cs
AgostiniAlessandro: I removed the file and changed the service so if Maximum username length and Minimum username length values don't make sense those settings are ignored during the validation.

src/Orchard.Web/Modules/Orchard.Users/Services/UsernameValidationError.cs
AgostiniAlessandro: We made this because if we enable the custom policies some previously created usernames may be not valid anymore so they can't be edite. For example if you set the minimum username length to 10 you can't edit the admin user anymore so by doing this we still allow the backend user to edit and save users and show a warning while frontend users can't register with an invalid username since at frontend the warning are treated as errors.

We could have used two lists, one containing the warnings and one containing the errors but that would have required our username validation method signature to be completely different from the password validation method signature and the handling of the validation results to be different as well. By introducing the severity we can have consistency between the validation methods and the code that handles the validation results is simpler as well.

src/Orchard/Security/IMembershipSettings.cs
```diff
@@ -21,5 +21,13 @@ public interface IMembershipSettings {
         MembershipPasswordFormat PasswordFormat { get; set; }
         bool EnablePasswordHistoryPolicy { get; set; }
         int PasswordReuseLimit { get; set; }
+        bool EnableCustomUsernamePolicy { get; set; }
+        int MinimumUsernameLength { get; set; }
+        int MaximumUsernameLength { get; set; }
+        bool ForbidUsernameSpecialChars { get; set; }
+        bool AllowEmailAsUsername {get; set;}
```
sebastienros, line 28: This is introducing a security issue if the login allows for both, you could impersonate someone else (we changed it in OC).

src/Orchard.Web/Modules/Orchard.Users/Services/UsernameValidationError.cs
sebastienros: Understood, thanks

src/Orchard/Security/IMembershipSettings.cs
AgostiniAlessandro: To avoid that, if the username is an email we check it's the same email of the user.

For reference: 

In the UserService class

`if (usernameIsEmail && !username.Equals(email, StringComparison.OrdinalIgnoreCase)){
                validationErrors.Add(new UsernameValidationError(Severity.Fatal, UsernameValidationResults.UsernameAndEmailMustMatch,
                        T("If the username is an email it must match the specified email address.")));
                return false;
            }
`

sebastienros: Do you check the unicity of a new username to see if it's not an existing email address?

AgostiniAlessandro: Yes. I double checked and found a case where the backoffice user could change the username to an. already existing username and I fixed it. In every other case it was already checked

MatteoPiovanelli:
@sebastienros After the comments/reviews/changes, do you think this looks good to go?

sebastienros:
@MatteoPiovanelli-Laser yes, thanks

Issue #8648, opened 2023-02-20
Single upload batch of multiple files freezes the UI and takes a long time to process


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/7bf78ef852a80582495a5c2c7a2c67b3d358029d/src/Orchard.Web/Modules/Orchard.MediaLibrary/Views/ClientStorage/Index.cshtml#L81

When uploading multiple files, the UI sometimes freezes, failing to report the progress of single uploads and takes a long time to complete. For this reason, user has the perception no file has been uploaded until all uploads are complete.

PR #8649, opened 2023-02-20
Sequential uploads in media library import procedure

AndreaPiovanelli:
To solve issue #8648 , added the optional parameter "sequentialUploads: true" to fileupload options. This forces the sequential uploads of files, avoiding the "stuck progress bars" effect and speeding up the process when uploading multiple files.


AndreaPiovanelli:
Wrong branch

Issue #8661, opened 2023-03-22
SingleTermId is null when Term has no LocalizationPart


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/049b4a8f3582a7b9d8daf55d61a93dff275e6f48/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/LocalizedTaxonomyController.cs#L67

As the subject of the issue says, if TermPart has no LocalizationPart, no first term is found and the SingleTermId property of the TaxonomyFieldViewModel doesn't get to be valued. This causes a issue when a TaxonomyField is configured to accept the selection of on term only.

To clarify, a scenario where Taxonomy has the LocalizationPart and its relative Term hasn't can exist if terms aren't visible at front-end but need to be a group criteria for content (e.g. are used in queries for specific projection widgets).

PR #8662, opened 2023-03-22
Avoid SingleTermId being null if at least one term is checked

AndreaPiovanelli:
In reference to #8661 
If no term with the proper culture is found, SingleTermId is the id of the first valid checked term. This patch avoids TaxonomyViewModel.SingleTermId to be null when at least one term has been checked by the user. This was a bug happening when TaxonomyField was configured to allow the selection of a single term only.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/LocalizedTaxonomyController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/LocalizedTaxonomyController.cs
@@ -66,6 +66,13 @@ namespace Orchard.Taxonomies.Controllers {
                         var firstTermForCulture = appliedTerms.FirstOrDefault(x => x.As<LocalizationPart>() != null && x.As<LocalizationPart>().Culture != null && x.As<LocalizationPart>().Culture.Culture == culture);
                         if (firstTermForCulture != null) {
                             firstTermIdForCulture = firstTermForCulture.Id;
+                        } else {
+                            // If there is no valid localization, firstTermForCulture is null.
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Controllers/LocalizedTaxonomyController.cs
```diff
@@ -66,6 +66,13 @@ public class LocalizedTaxonomyController : Controller {
                         var firstTermForCulture = appliedTerms.FirstOrDefault(x => x.As<LocalizationPart>() != null && x.As<LocalizationPart>().Culture != null && x.As<LocalizationPart>().Culture.Culture == culture);
                         if (firstTermForCulture != null) {
                             firstTermIdForCulture = firstTermForCulture.Id;
+                        } else {
+                            // If there is no valid localization, firstTermForCulture is null.
+                            // To avoid that, use the first checked term (if any is checked).
+                            var firstTerm = terms.FirstOrDefault(t => t.IsChecked = appliedTerms.Select(x => x.Id).Contains(t.Id));
```
sebastienros:
```suggestion
                            var firstTerm = terms.FirstOrDefault(t => t.IsChecked = appliedTerms.Any(x => x.Id == t.Id));
```

this should be simpler.

Also this whole FirstOrDefault call is complex, it's actually doing an assignment and stops at the first one that is true. Was that intentional?

AndreaPiovanelli:
@sebastienros I forgot to apply the last update to Orchard's dev-derived branch, which now is pushed.
I simplified the subsequent query following your suggestions too.

sebastienros (approved):

Issue #8678, opened 2023-05-03
TinyMce new feature proposal: links to content


HermesSbicego-Laser:
When using TinyMce as HTML editor the only way to add a link (anchors) to an internal content is to insert the URL of the content. If the URL of the content changes, all links within the HTML results in a 404 error page.
The proposal is to add a new TinyMce plugin that gives the ability to create links to a content (of the site) instead of inserting a link to an URL. In this way, if the editor modifies the permalink of a content, all the links within an HTML input (TextFields, BodyParts, Html Elements, Html Widgets) result in a 200 status and no more in a 404 error status.
We have a PR ready.

PR #8679, opened 2023-05-03
Feature/tinymce contentlinks plugin

HermesSbicego-Laser:
Implements  what described into #8678


src/Orchard.Web/Modules/Orchard.Tokens/Providers/ContentTokens.cs
```diff
@@ -26,6 +26,9 @@ public class ContentTokens : ITokenProvider {
         public Localizer T { get; set; }
 
         public void Describe(DescribeContext context) {
+            context.For("ContentManager", T("Content Manager"), T("Content Manager"))
```
sebastienros: `ContentItem` -> `Query:`, `Id:`, `Route:`, `Identity:` ...

sebastienros (approved):

Issue #8681, opened 2023-05-09
Records in namespaces nested under Models are not included in NHibernate mappings


MpDzik:
When a record class is created in a namespace nested under Models, it is not included in the NHibernate mappings. When the record is used with the `IRepository` interface, NHibernate throws an exception that there is no persister for the record class.

PR #8682, opened 2023-05-09
Fix mapping records in nested namespaces (#8681)

MpDzik:
PR for #8681

### DIFF
--- a/src/Orchard/Environment/ShellBuilders/CompositionStrategy.cs
+++ b/src/Orchard/Environment/ShellBuilders/CompositionStrategy.cs
@@ -197,7 +197,7 @@ namespace Orchard.Environment.ShellBuilders {
         }
 
         private static bool IsRecord(Type type) {
-            return ((type.Namespace ?? "").EndsWith(".Models") || (type.Namespace ?? "").EndsWith(".Records")) &&
+            return ((type.Namespace ?? "").EndsWith(".Models") || (type.Namespace ?? "").EndsWith(".Records") || type.Name.EndsWith("Record")) &&
                    type.GetProperty("Id") != null &&
                    (type.GetProperty("Id").GetAccessors()).All(x => x.IsVirtual) &&
                    !type.IsSealed &&
--- a/src/Orchard/Environment/ShellBuilders/CompositionStrategy.cs
+++ b/src/Orchard/Environment/ShellBuilders/CompositionStrategy.cs
@@ -6,6 +6,7 @@ using System.Web.Mvc;
 using Autofac.Core;
 using Orchard.ContentManagement;
 using Orchard.ContentManagement.Records;
+using Orchard.Data;
 using Orchard.Environment.Configuration;
 using Orchard.Environment.Descriptor.Models;
 using Orchard.Environment.Extensions;
@@ -96,7 +97,7 @@ namespace Orchard.Environment.ShellBuilders {
                 }
 
                 var feature = availableFeatures[shellFeature];
-                
+
                 foreach (var childDependency in ExpandDependenciesInternal(availableFeatures, feature.Dependencies, dependentFeatureDescriptor: feature))
                     yield return childDependency;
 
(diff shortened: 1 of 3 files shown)


src/Orchard/Environment/ShellBuilders/CompositionStrategy.cs
```diff
@@ -197,7 +197,7 @@ public class CompositionStrategy : ICompositionStrategy {
         }
 
         private static bool IsRecord(Type type) {
-            return ((type.Namespace ?? "").EndsWith(".Models") || (type.Namespace ?? "").EndsWith(".Records")) &&
+            return ((type.Namespace ?? "").EndsWith(".Models") || (type.Namespace ?? "").EndsWith(".Records") || type.Name.EndsWith("Record")) &&
```
BenedekFarkas: Would it matter if we detect a class to be a record even when it isn't (i.e., false positive)?
In any case, we can just modify this to be more forgiving like this:
```suggestion
            var namespace = type.Namespace ?? string.Empty;
            return (namespace.Contains(".Models") || namespace.Contains(".Records")) &&
```

MatteoPiovanelli: The chance for false positive makes this PR potentially breaking. AFAIK it wouldn't break anything in the current Orchard codebase, nor ours, but the potential is there.

The approach using `Contains` may be more robust. The proposed snippet is missing a few corner cases (e.g. the "Example.ModelsX" namespace would pass the test now, while it never did and probably shouldn't). A possible middle ground could be to have both tests?
```
// Only for "Models" namespace and its sub-namespaces, but similarly for Records:
(namespace.Contains(".Models.") || namespace.EndsWith(".Models"))
```

MpDzik: It actually does cause issues if a non-record class is mapped. For example, if the class contains non-virtual methods or doesn't have a parameterless constructor and Orchard tries to map it with NHibernate, it will throw an exception. This already happens if you put a class which is not a valid record into the `Models` folder and add a virtual `Id` property.

I like @MatteoPiovanelli-Laser's approach with checking the exact name `Models` or `Records`. It is consistent with the current naming convention. If we want to by fully backward compatible, we could add an attribute to opt-in which classes should be mapped. The drawback of this approach is that it would be inconsistent with the records in non-nested namespaces, so we would need the attribute only for the nested namespaces. And frankly, why would anyone name a non-record class with a `Record` suffix?

Maybe it would make sense to use the `DoNotMap` attribute to exclude classes from mapping. This way, even if this change breaks any custom modules, it will be easy to resolve the issue by adding the attribute to the class that was mapped as a false positive.

MatteoPiovanelli: The opt-in attribute makes sense for classes outside the current naming convention, I think. 
Perhaps something like `[MapAsRecord]` for the classes that should be tentatively be considered records; when such class is in a Namespace that matches the conventions, such attribute should be redundant, for backwards compatibility.
At the same time `[MapAsRecord(false)]` could prevent classes that shouldn't be considered records to be considered as such even if they happened to satisfy all other constraints.

MpDzik: Yes, I was thinking about the same. It seems strange to me that it was not implemented this way from the beginning. Using an attribute or marker interface is a lot better than relying on the namespace layout, in my opinion.

sebastienros:
I'll let you all agree on the best solution (which doesn't break anything) since you are more knowledgeable.

sebastienros (approved):

MatteoPiovanelli:
@MpDzik could you update the PR with the things we discussed above?

MpDzik:
@MatteoPiovanelli-Laser How does this look to you?

src/Orchard/Environment/ShellBuilders/CompositionStrategy.cs
BenedekFarkas: Looks great, thanks!

BenedekFarkas (approved):

MatteoPiovanelli (approved):

MatteoPiovanelli:
sorry for the delay on my part, but I had missed the notification.

PR #8687, opened 2023-05-17
8686: Fixing build and tests on 1.10.x

### LINKED ISSUE #8686: Fixing build and tests on 1.10.x and dev
Steps:

- [x] **1. Fixing build and tests on `1.10.x`**:
PR #8687 (`issue/8686` branch) is to be merged first. When that happens, `1.10.x` is fully stabilized, but this issue shouldn't be resolved yet.
- [x] **2. Merge `1.10.x` into `dev`**:
Then merge `1.10.x` into `issue/8684` (issue #8684, PR #8685), validate that everything's working and then merge that PR (which shows the diff between the two branches). At this point `dev` is partially stabilized.
- [x] **3. Fixing build and tests on `dev`**:
Merge `dev` into `issue/8686-dev` (PR #8695), validate that everything's working and then merge that PR. At this point, `dev` is fully stabilized and this issue can be resolved.

BenedekFarkas:
Fixes #8686


.github/workflows/compile.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          name: Orchard-${{ github.sha }}-${{ github.run_number }}.binlog
+          path: Orchard.binlog
+          if-no-files-found: ignore
+
+      - name: Test
+        run: msbuild Orchard.proj /m /t:Test
+
+      - name: Spec
+        run: msbuild Orchard.proj /m /t:Spec
```
BenedekFarkas: This is a not-yet final version of the basic Compile workflow we should run for PRs (but from that we can exclude the Spec step).

Orchard.proj
```diff
@@ -181,7 +181,7 @@
     <NUnit Assemblies="@(TestAssemblies)" ToolPath="@(NUnitPackageToolsFolder)" WorkingDirectory="$(CompileFolder)" OutputXmlFile="$(BuildFolder)\Orchard.Tests.xml" ExcludeCategory="longrunning" />
   </Target>
 
-  <Target Name="Spec" DependsOnTargets="Package-Stage">
+  <Target Name="Spec" DependsOnTargets="Compile">
```
BenedekFarkas, line 184: Spec doesn't use anything from the output of Package-Stage, just Compile.

src/Orchard.Core.Tests/Common/Providers/CommonPartProviderTests.cs
```diff
@@ -215,11 +215,11 @@ class StubThemeService : IThemeManager {
 
             var user = contentManager.New<IUser>("User");
             _authn.Setup(x => x.GetAuthenticatedUser()).Returns(user);
-            _authz.Setup(x => x.TryCheckAccess(StandardPermissions.SiteOwner, user, item)).Returns(true);
+            _authz.Setup(x => x.TryCheckAccess(OwnerEditorPermissions.MayEditContentOwner, user, item)).Returns(true);
```
BenedekFarkas, line 218: Because the Owner editor was changed to check for this permission instead of SiteOwner.

src/Orchard.Specs/Bindings/Settings.cs
```diff
@@ -30,6 +22,11 @@ public class Settings : BindingBase {
                     }
 
                     orchardServices.WorkContext.CurrentSite.SiteCulture = cultureName;
+
+                    // Restarting the shell to reset the cache, because the cache entry storing the list of available
+                    // cultures isn't invalidated by the signal in DefaultCultureManager.ListCultures when running
+                    // inside the test webhost.
+                    MvcApplication.RestartTenant("Default");
```
BenedekFarkas, line 29: This is what I mentioned in my comment under the issue: https://github.com/OrchardCMS/Orchard/issues/8686#issuecomment-1562516639
IMO the workaround is harmless (although not ideal), but changing the culture settings works fine otherwise.

src/Orchard.Specs/Blogs.feature
```diff
@@ -171,6 +171,7 @@ Scenario: I can create browse blog posts on several pages
         And I go to "admin/blogs"
         And I follow "My Blog"
         And I follow "New Post" where class name has "primaryAction"
+        And I wait "1000"
```
BenedekFarkas, line 174: These waits are needed, because the tests are ran using SQL CE and that doesn't have millisecond-precision, so the blogposts might be out of order if they are created in rapid succession, causing false negative test result at the end.

src/Orchard.Specs/Blogs.feature
```diff
@@ -269,12 +281,12 @@ Scenario: I can create browse blog posts on several pages
     Then I should see "<h1[^>]*>.*?My Blog.*?</h1>"
         And I should see "<h1[^>]*>.*?My Post 12.*?</h1>"
         And I should see "<h1[^>]*>.*?My Post 11.*?</h1>"
-        And I should not see "<h1[^>]*>.*?My Post 10.*?</h1>"
+        And I should not see "My Post 2"
```
BenedekFarkas, line 284: This was a false positive, because "I should not see" uses a string contains check (vs. "I should see" is a regex match) and the value checked was also incorrect (since "My Post 10" should appear on the first page).

src/Orchard.Specs/DateTime.feature
```diff
@@ -184,15 +184,15 @@ Scenario: Creating and using date time fields in another culture
...
-        And I go to "Admin/ContentTypes/Edit/Event"
+    When I go to "Admin/ContentTypes/Edit/Event"
         And I fill in 
             | name                                     | value       |
             | Fields[EventDate].DateTimeFieldSettings.Display  | DateAndTime |
             | Fields[EventDate].DateTimeFieldSettings.Required | true        |
         And I hit "Save"
+
+    # Date & Time are validated based on current culture
+    When I have "fr-FR" as the default culture
```
BenedekFarkas, line 195: This is just some logical reordering.

src/Orchard.Specs/Hosting/Orchard.Web/Global.asax.cs
```diff
@@ -29,6 +31,7 @@ public class MvcApplication : HttpApplication {
 
         protected void Application_BeginRequest() {
             Context.Items["originalHttpContext"] = Context;
+            HttpContext.Current.Response.AddHeader("X-Frame-Options", "SAMEORIGIN");
```
BenedekFarkas, line 34: These two additions are a workaround for the MVC upgrade to 5.2.7 (the breaking release is actually 5.2.4) causing problem with the test web host: https://github.com/aspnet/AspNetWebStack/issues/162.
Even better would be to change the test setup to use integrated pipeline, but I haven't found yet how to do that.

src/Orchard.Tests/Storage/FileSystemStorageProviderTests.cs
```diff
@@ -335,8 +334,8 @@ public class FileSystemStorageProviderTests {
             _storageProvider.CreateFile(@"test.txt");
 
             using (Stream stream = GetFile("test.txt").OpenRead()) {
-                Assert.That(() => _storageProvider.SaveStream(@"../newTest.txt", stream), Throws.InstanceOf(typeof(ArgumentException)));
-                Assert.That(() => _storageProvider.SaveStream(@"../../newTest.txt", stream), Throws.InstanceOf(typeof(ArgumentException)));
+                Assert.That(() => _storageProvider.SaveStream(@"../newTest.txt", stream), Throws.InstanceOf(typeof(OrchardException)));
+                Assert.That(() => _storageProvider.SaveStream(@"../../newTest.txt", stream), Throws.InstanceOf(typeof(OrchardException)));
```
BenedekFarkas, line 338: I reviewed and merged the PR a few years ago that broke these tests. :(

src/Orchard.Specs/Media.feature
```diff
@@ -1,31 +1,13 @@
...
-    # Editing a media with limited rights
-    When I go to "admin/media/edit?name=..\..\bin&mediaPath=..\..\bin"
-        And I am redirected
-    Then I should see "Media"
-        And I should see "Editing failed: Invalid path"
-        And the status should be 200 "OK"
+    # Accessing the Media Library page
+    When I go to "Admin/Orchard.MediaLibrary"
+    Then I should see "Media Library"
+    And the status should be 200 "OK"
```
BenedekFarkas, line 13: The Media module is deprecated, but I didn't really find an easy way to adapt these tests to Media Library, so I just adapter the basics and removed the rest.

src/Orchard.Specs/Hosting/Orchard.Web/Web.config
```diff
@@ -141,4 +141,22 @@
             <mimeMap fileExtension=".woff2" mimeType="font/woff2" />
         </staticContent>
     </system.webServer>
+    <runtime>
+        <gcServer enabled="true" />
+        <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+            <probing privatePath="App_Data/Dependencies" />
```
BenedekFarkas, line 147: The removal of this line during the MVC upgrade was the root cause behind the error I mentioned under the issue:

> None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'Orchard.Environment.DefaultOrchardShell' can be invoked with the available services and parameters:
> Cannot resolve parameter 'Orchard.IWorkContextAccessor workContextAccessor' of constructor 'Void .ctor(Orchard.IWorkContextAccessor, System.Collections.Generic.IEnumerable1[Orchard.Mvc.Routes.IRouteProvider], System.Collections.Generic.IEnumerable1[Orchard.WebApi.Routes.IHttpRouteProvider], Orchard.Mvc.Routes.IRoutePublisher, System.Collections.Generic.IEnumerable1[Orchard.Mvc.ModelBinders.IModelBinderProvider], Orchard.Mvc.ModelBinders.IModelBinderPublisher, Orchard.Tasks.ISweepGenerator, System.Collections.Generic.IEnumerable1[Orchard.Owin.IOwinMiddlewareProvider], Orchard.Environment.Configuration.ShellSettings)'.

Some of the assembly binding redirects also had to be restored for the test app to be able to start.

src/Orchard.Web/Modules/Orchard.Workflows/Orchard.Workflows.csproj
```diff
@@ -70,7 +70,7 @@
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
     <Reference Include="System.Data.DataSetExtensions" />
-    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Net.Http, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
```
BenedekFarkas, line 73: This comment is relevant for all the changes across various files that have a change related to System.Net.Http:
It started with the static Razor compilation throwing a warning (which is now intentionally configured to be an error) that there are conflicting references with a different target framework. With some experimentation it turned out that pinning the 4.2.0.0 version fixes all the build warnings and the code that actually uses it (e.g., the WebRequestActivity in Workflows) still works. On the other hand, this and the other modules depending on Workflows also need this reference set in their web.config.

System.Net.Http is proven to be problematic in some cases - not sure if directly related, but still an interesting read: https://github.com/dotnet/runtime/issues/65005.

src/Orchard/Orchard.Framework.csproj
```diff
@@ -108,9 +108,7 @@
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
     <Reference Include="System.Data" />
-    <Reference Include="System.Net.Http">
-      <Private>True</Private>
-    </Reference>
```
BenedekFarkas, line 113: Not sure why, but setting this reference private (copy local) with the version number pinned to 4.2.0.0 causes the same error when running the "normal" application as the spec app had (not being able to resolve the WorkContext). 4.2.0.0 is a reference assembly in .NET 4.8 though., so we should be fine without copy local.

sebastienros (approved):
Nothing shocks me.

Issue #8696, opened 2023-06-12
Safe handle has been closed Exception


MatteoPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/9122abbdec97748fa8fb19ed5b66b3bde2c1495a/src/Orchard/Environment/Extensions/ExtensionManager.cs#LL110C19-L110C19

We found an issue with this, where it would sometimes return a "Safe handle has been closed".
This seems to be related to this answer from StackOverflow https://stackoverflow.com/a/21151576

We are working on a fix

PR #8697, opened 2023-06-12
Added lock on md5 variable to avoid concurrent hash generations for feature list

AndreaPiovanelli:
In refernce to issue #8696 , added a lock to _md5 variable to avoid concurrent has generations for features list, which brought to wrong feature lists being generated..

### DIFF
--- a/src/Orchard/Environment/Extensions/ExtensionManager.cs
+++ b/src/Orchard/Environment/Extensions/ExtensionManager.cs
@@ -25,7 +25,7 @@ namespace Orchard.Environment.Extensions {
         private readonly ICacheManager _cacheManager;
         private readonly IParallelCacheContext _parallelCacheContext;
         private readonly IEnumerable<IExtensionLoader> _loaders;
-
+       
         public Localizer T { get; set; }
         public ILogger Logger { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/Orchard/Environment/Extensions/ExtensionManager.cs
```diff
@@ -107,14 +107,16 @@ public class ExtensionManager : IExtensionManager {
...
-            var cacheKey = BitConverter.ToString(
-                _md5.ComputeHash(
-                    Encoding.UTF8.GetBytes(
-                        string.Join(";",
-                            featureDescriptors
-                                .Select(fd => fd.Id)
-                                .OrderBy(x => x)))));
-
+            string cacheKey;
+            lock (_md5) {
```
sebastienros, line 111: That could work but why not just create a custom instance every time and dispose it?

But if the issue happens it's because LoadFeatures is called concurrently, and that could limit the usage of the cache (cache stampede). So maybe create a new MD5 and also prevent multiple results for the same key from being created at the same time using this thing: https://github.com/dotnet/aspnetcore/blob/main/src/Middleware/OutputCaching/src/DispatcherExtensions.cs#L8

If that's too complex then feel free to merge what you just did.

PR #8700, opened 2023-06-12
Added flag to display a single model error for MediaLibraryPickerField

AndreaPiovanelli:
Added flag to display a single model error by LocalizationExtensionHandler if media have been removed from the field only. This ensures that no duplicate model error has displayed.

The flag is true when the field is set to remove items without localization and one or more media items cannot be localized.
Afterwards, to display a model error, this flag is checked in combination with the previously implemented checks (if field is required and any media item is selected).


AndreaPiovanelli:
Hi @sebastienros 
Any comment on this pr? Can we merge it?

src/Orchard.Web/Modules/Orchard.MediaLibrary/Handlers/MediaLibraryPickerFieldLocalizationExtensionHandler.cs
```diff
@@ -65,8 +70,7 @@ public class MediaLibraryPickerFieldLocalizationExtensionHandler : ContentHandle
                                             "{0}: the media item {1} was culture neutral and it has been localized",
                                             field.DisplayName,
                                             mediaItem.As<MediaPart>().FileName));
-                                    }
-                                    else {
+                                    } else {
```
sebastienros: nit: that's not how it should be formatted, it was fine on a separate line

sebastienros (approved):

Issue #8701, opened 2023-06-13
filterTaxonomyCulture function isn't unique if multiple TaxonomyFields are inside the content


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/2ef2ad8e99948625ebebb7e05ca609eaa85c2894/src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyFieldList.cshtml#L12

The javascript function referenced in the issue subject is called multiple times if there is more than one TaxonomyField inside the page / content. This has the effect of making the field losing user selection when the content isn't saved because of validation errors (e.g. another required field not compiled). On top of that, multiple GetTaxonomy calls are launched.

PR #8702, opened 2023-06-13
filterTaxonomyCulture function is now named with a proper unique name, based on part and field

AndreaPiovanelli:
After trying to force to check back selected terms, I noticed the problem was the one referenced in issue #8701 
Added part name and field name to the javascript function name, to avoid being called multiple times when culture is changed.
Added a parameter to the "change" event for "#Localization_SelectedCulture" control to ensure that, when triggered by TaxonomyFieldListShape, it triggers the filterTaxonomyCulture for the current TaxonomyField only.
Modified LocalizedTaxonomiesViewModel to add PartName property.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyFieldList.cshtml
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyFieldList.cshtml
@@ -5,7 +5,7 @@
     Script.Require("jQuery");
     var Taxonomyprefix = ((ViewData)).TemplateInfo.HtmlFieldPrefix.Replace('.', '_');
     using (Script.Foot()) {
-        <script type="text/javascript">
+    <script type="text/javascript">
             //<![CDATA[
             $(document).ready(function () {
                 var pageurl = '@Url.Action("GetTaxonomy", "LocalizedTaxonomy", new { area = "Orchard.Taxonomies" })';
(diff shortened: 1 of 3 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyFieldList.cshtml
```diff
@@ -5,16 +5,19 @@
     Script.Require("jQuery");
     var Taxonomyprefix = ((ViewData)).TemplateInfo.HtmlFieldPrefix.Replace('.', '_');
     using (Script.Foot()) {
-        <script type="text/javascript">
+    <script type="text/javascript">
             //<![CDATA[
             $(document).ready(function () {
                 var pageurl = '@Url.Action("GetTaxonomy", "LocalizedTaxonomy", new { area = "Orchard.Taxonomies" })';
-                function filterTaxonomyCulture(culture) {
+                function @("filterTaxonomyCulture_" + Model.PartName + "_" + Model.FieldName)(culture) {
```
sebastienros: this function name could be created once in a `var` in this view

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyFieldList.cshtml
AndreaPiovanelli: Suggestion applied, thanks.

PR #8705, opened 2023-07-03
Use ContentManagerSession in GetMany

MatteoPiovanelli:
This is the first hacky attempt. I'm not satisfied with the code: I look at it and it feels like it's doing some stuff too many times where it shouldn't.
Anyway, it does reduce the number of repeated calls to the database. In some setups, that may become significant.

I haven't had a chance to properly test this, or measure the timing performance, to make sure that we aren't losing time  in the work we do to avoid calls to the db.

To summarize, my concerns with my code in this PR right now:
 - it looks hacky to me.
 - I'm not 100% confident it will always 100% of the time in every condition return the exact same results as the live code, even though I can't think of a case where it wouldn't.
 - I have tested the timing of the methods I changed.


MatteoPiovanelli:
@BenedekFarkas forgot to tag you in the main post

BenedekFarkas:
I'll need a few days to get down to testing it properly, but it looks good on first glance!

MatteoPiovanelli:
@BenedekFarkas I just realized this is still open. My tests still look good, But I'm still not too fond of the code I wrote.

BenedekFarkas:
Well, DefaultContentManager is very old, so there's probably a lot of stuff we could refactor (which sort of affects what you can achieve there). I wouldn't merge it without unit/Spec testing completing on it first but being able to run tests is still blocked by me (https://github.com/OrchardCMS/Orchard/issues/8686). We could also add some specific tests to verify this change, if necessary.

MatteoPiovanelli:
this is targeted at 1.10.x, so I think specflows should be able to run... I'll get back to you on this.

MatteoPiovanelli:
I merged the latest 1.10.x into the branch for this PR and I was able to run all specflows.

MatteoPiovanelli:
I'm still not 100% on this solution.
Results look consistent, and I see fewer queries in the logs, but no great impact on final timings (in quick page load tests I did).
Perhaps doing a conditional and short circuiting the case when there is a single Id being queried would be useful, as that is an actual common case I see in our tenants, in ContentPickerFields and such

BenedekFarkas:
> I'm still not 100% on this solution. Results look consistent, and I see fewer queries in the logs, but no great impact on final timings (in quick page load tests I did).

I think if you see that it consistently reduces the number of queries, that's already a win. Did you test with SysCache enabled?

> Perhaps doing a conditional and short circuiting the case when there is a single Id being queried would be useful, as that is an actual common case I see in our tenants, in ContentPickerFields and such

Doesn't the current logic handle that situation just by nature of running as if there were more? BTW ContentManager.Get uses ContentManagerSession already, so it's possible to short-circuit to call that, if necessary.

AndreaPiovanelli:
I merged the latest 1.10.x inside the pr branch.
What I noticed is that DefaultContentManagerSessions dictionary doesn't consider QueryHints for its key. This raises an error because, in some cases (which I am describing later in this comment), stores inside the session dictionary an "incomplete" content item.

Example case: TaxonomyField.
At the line https://github.com/OrchardCMS/Orchard/blob/5c3d045aa1189e8760b5090ebb939ce571dccab0/src/Orchard.Web/Modules/Orchard.Taxonomies/Drivers/TaxonomyFieldDriver.cs#L89
following to https://github.com/OrchardCMS/Orchard/blob/5c3d045aa1189e8760b5090ebb939ce571dccab0/src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs#L206
Orchard reads the selected terms for current content item. Since there is no QueryHint, the data from TermPartRecord (i.e. term path) isn't loaded, so it stays empty and gets stored in session this way.
Since GetMany function doesn't read that term anymore in the future (as it is saved inside the session), path doesn't get to be corrected anymore (previously - and it is "luck" - after the GetAppliedTerms call, the GetTerms overwrote the terms inside the session with the proper QueryHints (https://github.com/OrchardCMS/Orchard/blob/5c3d045aa1189e8760b5090ebb939ce571dccab0/src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs#L167) but with current patch to GetMany, it's not the case anymore.
This issue raises and exception when sorting terms (here: https://github.com/OrchardCMS/Orchard/blob/5c3d045aa1189e8760b5090ebb939ce571dccab0/src/Orchard.Web/Modules/Orchard.Taxonomies/Models/TermPart.cs#L111) because Orchard is looking for a path that does not exist inside the dictionary (it searches key "/id1" and "/id2" but, since the path isn't correct for the selected terms, one of them isn't in the dictionary (generating a KeyNotFoundException).

It's not so easy to explain, but I hope I have been clear. Simple way to reproduce the error is to add a TaxonomyField to a content item, select a term, save and, when loading the editor, it crashes and no TaxonomyField is displayed.

BenedekFarkas:
1.10.4 will be released soon from 1.10.x, so if you're still working on this, please target the dev branch.

BenedekFarkas:
If you continue working on this, please open a new PR targeting `dev`.

PR #8707, opened 2023-07-06
Using MailKit instead of SmtpClient (ORCH-267)

### LINKED ISSUE #8691: Upgrade Orchard.Email from using SmtpClient
As stated on [`System.Net.Mail.SmtpClient`'s official documentation page](https://learn.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient?view=netframework-4.8#remarks):

> We don't recommend that you use the SmtpClient class for new development because SmtpClient doesn't support many modern protocols. Use [MailKit](https://github.com/jstedfast/MailKit) or other libraries instead. For more information, see [SmtpClient shouldn't be used](https://github.com/dotnet/platform-compat/blob/master/docs/DE0005.md) on GitHub.

For example, SendGrid deprecates support for TLS 1.0 and 1.1 on the 5th of June.

dministro:
fixes #8691

[Jira issue](https://lombiq.atlassian.net/browse/ORCH-267)

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
+++ b/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
@@ -1,14 +1,17 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Configuration;
-using System.Net;
+using System.Linq;
 using System.Net.Configuration;
 using System.Net.Mail;
 using System.Web.Mvc;
+using MailKit.Security;
+using MimeKit;
 using Orchard.ContentManagement;
 using Orchard.DisplayManagement;
-using Orchard.Logging;
 using Orchard.Email.Models;
+using Orchard.Logging;
+using SmtpClient = MailKit.Net.Smtp.SmtpClient;
 
 namespace Orchard.Email.Services {
     public class SmtpMessageChannel : Component, ISmtpChannel, IDisposable {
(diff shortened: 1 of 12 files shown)


sebastienros (approved):

BenedekFarkas:
@dministro I'll check the build error, I have an idea why is that.

dministro:
> @dministro I'll check the build error, I have an idea why is that.

Thank you

dministro:
@BenedekFarkas I can't repro this locally. Tried with both the command line and VS.

We could add `System.Net.Http` as a direct reference to `Orchard.Email` and make it private just for a test.

```diff
 src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj b/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj
index 975fba730..28ad95825 100644
--- a/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj
+++ b/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj
@@ -83,6 +83,9 @@
     </Reference>
     <Reference Include="System.Data" />
     <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="System.Net.Http">
+      <Private>True</Private>
+    </Reference>
     <Reference Include="System.Security" />
     <Reference Include="System.Web.ApplicationServices" />
     <Reference Include="System.Web.DynamicData" />
```

src/Orchard.Web/Modules/Orchard.Email/Views/EditorTemplates/Parts/SmtpSettings.cshtml
```diff
@@ -25,10 +26,29 @@
...
-        <span class="hint">@T("Check if the SMTP server requires SSL communications.")</span>
+        <label for="@Html.FieldIdFor(m => m.EncryptionMethod)">@T("Encryption method to use")</label>
+        <select for="@Html.FieldIdFor(m => m.EncryptionMethod)" name="@Html.NameOf(m => m.EncryptionMethod)" disabled="@Model.AutoSelectEncryption">
+            @Html.SelectOption(
+                Model.EncryptionMethod,
+                SmtpEncryptionMethod.None,
+                String.Join(String.Empty, T("None").ToString(), " - ", T("Connect to server using insecure connection.").ToString()))
+            @Html.SelectOption(
+                Model.EncryptionMethod,
+                SmtpEncryptionMethod.None,
```
AndreaPiovanelli: This may be a error, SmtpEncryptionMethod.None is the value of all three select options.

src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
```diff
@@ -133,35 +124,68 @@ public class SmtpMessageChannel : Component, ISmtpChannel, IDisposable {
...
+                    AutoSelectEncryption = !smtpSection.Network.EnableSsl,
+                    RequireCredentials = smtpSection.Network.DefaultCredentials || !String.IsNullOrWhiteSpace(smtpSection.Network.UserName),
+                    smtpSection.Network.UserName,
+                    smtpSection.Network.Password,
+                };
+            }
 
-            if (!smtpClient.UseDefaultCredentials && !String.IsNullOrWhiteSpace(_smtpSettings.UserName)) {
-                smtpClient.Credentials = new NetworkCredential(_smtpSettings.UserName, _smtpSettings.Password);
+            var secureSocketOptions = SecureSocketOptions.Auto;
```
AndreaPiovanelli, line 155: SecureSocketOptions defaulting to None may be a compatibility issue when SmtpSettingsPart.EnableSsl was previously set to true (the value of which is now based on SmtpSettingsPart.EncryptionMethod value only). In those cases, Ssl would be disabled. Can a migration properly set the EncryptionMethod property value to e.g. SmtpEncryptionMethod.SslOnConnect when "old" EnableSsl was true?

BenedekFarkas: @dministro how did we handle this for our sites? Anyway, a migration step for keeping the correct settings would be nice indeed.

BenedekFarkas:
> @BenedekFarkas I can't repro this locally. Tried with both the command line and VS.
> 
> We could add `System.Net.Http` as a direct reference to `Orchard.Email` and make it private just for a test.
> 
> ```diff
>  src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj | 3 +++
>  1 file changed, 3 insertions(+)
> 
> diff --git a/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj b/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj
> index 975fba730..28ad95825 100644
> --- a/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj
> +++ b/src/Orchard.Web/Modules/Orchard.Email/Orchard.Email.csproj
> @@ -83,6 +83,9 @@
>      </Reference>
>      <Reference Include="System.Data" />
>      <Reference Include="System.Data.DataSetExtensions" />
> +    <Reference Include="System.Net.Http">
> +      <Private>True</Private>
> +    </Reference>
>      <Reference Include="System.Security" />
>      <Reference Include="System.Web.ApplicationServices" />
>      <Reference Include="System.Web.DynamicData" />
> ```

Yeah, I don't have a repro either, which is really annoying. It probably comes down to some combination of SDKs installed. I'll play with the references a bit more - BTW this type of warning came up before: https://github.com/OrchardCMS/Orchard/pull/8687/commits/d266db61de22a30b440139e185b5b4884912e883#r1207070731

**Edit: This is fixed now.**

src/Orchard.Web/Modules/Orchard.Email/Views/EditorTemplates/Parts/SmtpSettings.cshtml
BenedekFarkas: @dministro and the correct value isn't persisted, even if I correct the technical values here, please check.

src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
dministro: This is handled a bit differently on our sites. I kept the `EnableSsl` and the [MailKit.MAilService.Connect(string host, int port, bool useSsl, CancellationToken cancellationToken)](https://github.com/jstedfast/MailKit/blob/7f3affdb9716bf84270f7d831e90a42a89c9d7f8/MailKit/MailService.cs#L975-L1035) is used to initiate the connection.

The solution implemented under this PR gives a bit more control over the connection settings.

dministro:
> Yeah, I don't have a repro either, which is really annoying. It probably comes down to some combination of SDKs installed. I'll play with the references a bit more - BTW this type of warning came up before: [d266db6#r1207070731](https://github.com/OrchardCMS/Orchard/commit/d266db61de22a30b440139e185b5b4884912e883#r1207070731)
> 
> **Edit: This is fixed now.**

Thank you for your support @BenedekFarkas .

src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
BenedekFarkas: Yeah, I was wondering, because the UI is different compared to our sites in Production now, but it's good to have better configurability. At the same time, what Andrea is saying is that the configuration that was set previously (including SSL and authentication) should carry over.

dministro: I added the migration step to migrate from `EnableSsl` to `EncryptionMethod`. See https://github.com/OrchardCMS/Orchard/pull/8707/commits/8a6958cec2edf28bfdabaadd19088c7a866be963

Regarding the `UseDefaultCredentials` there is no alternative in `MailKit` so removing this is a breaking change. Here is how this situation is handled in `OrchardCore.Email` module:

https://github.com/OrchardCMS/OrchardCore/blob/8cc817c13d12ec39f4dbdec6f9c71195c0c511c1/src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs#L295-L299

src/Orchard.Web/Modules/Orchard.Email/Views/EditorTemplates/Parts/SmtpSettings.cshtml
BenedekFarkas: @dministro just commenting a reminder here for your SendGrid testing.

BenedekFarkas: Sorry, I just noticed that it's fixed already!

src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
BenedekFarkas: I pushed some changes to simplify how the settings are migrated + the settings upgrade also includes setting AutoSelectEncryption to the opposite of what EnableSsl was (so it's more explicit).

BenedekFarkas (approved):

BenedekFarkas:
Here are the test cases (8 in total) I checked (including my latest changes):
1. Setup on issue branch, configure for localhost SMTP.
2. Setup on issue branch, configure for SendGrid account.
3. Setup on 1.10.x, configure for localhost SMTP, upgrade to issue branch.
4. Setup on 1.10.x, configure for SendGrid account, upgrade to issue branch.

Note that all the test cases work with either value of AutoSelectEncryption:
1. True: Port-based detection works fine.
2. False: EncryptionMethod has to be set to None for localhost and SSL for SendGrid.

I think it's ready to be merged! @AndreaPiovanelliLaser?

AndreaPiovanelli:
> Here are the test cases (8 in total) I checked (including my latest changes):
> 
> 1. Setup on issue branch, configure for localhost SMTP.
> 2. Setup on issue branch, configure for SendGrid account.
> 3. Setup on 1.10.x, configure for localhost SMTP, upgrade to issue branch.
> 4. Setup on 1.10.x, configure for SendGrid account, upgrade to issue branch.
> 
> Note that all the test cases work with either value of AutoSelectEncryption:
> 
> 1. True: Port-based detection works fine.
> 2. False: EncryptionMethod has to be set to None for localhost and SSL for SendGrid.
> 
> I think it's ready to be merged! @AndreaPiovanelliLaser?

Sorry for the late reply.
I think using infoset works without breaking compatibility with previously saved mail settings.

PR #8711, opened 2023-07-12
Consolidating Microsoft.Owin package versions on dev

### LINKED ISSUE #8710: Consolidate Owin package versions on dev
Enabling/disabling features or running a migration update fails with Orchard.OpenId (even if disabled) not being able to resolve one of the dependencies of `Microsoft.Owin.Security`, because it (and some other projects) use version 4.1.1 of `Microsoft.Owin.Host.SystemWeb` and `Microsoft.Owin.Security`, while the other Owin libraries are on version 4.2.2.

Fortunately, fixing it is as simple as a search and replace of the string "4.1.1" to "4.2.2".

BenedekFarkas:
Fixes #8710


BenedekFarkas:
The build fails due to a warning (error) in Razor compilation related to System.Net.Http references, but that's already fixed in #8685.

Issue #8712, opened 2023-07-20
Adding scale parameter to ResizeMediaUrl function


dbertolin:
https://github.com/OrchardCMS/Orchard/blob/274b4416ec2a9dd6794207ffd7868cf7f706f866/src/Orchard.Web/Modules/Orchard.MediaProcessing/Shapes/MediaShapes.cs#L31

In certain cases we got messed images from ResizeMediaUrl call: images were grainy when resized
We need to add scaling as an optional parameters to ResizeMediaUrl  function.

HermesSbicego-Laser:
fixed via #8713

PR #8713, opened 2023-07-20
Implementation of "scale" filter to avoid grained images

luciofassio:
In reference to #8712 
Added an optional paramater to MediaShapes.ResizeMediaUrl function to properly process image scaling, when original image size is less than display size.


luciofassio:
@dotnet-policy-service agree company="INVA Spa"

sebastienros (approved):

PR #8716, opened 2023-07-21
Updated XML helper to support storing and parsing of Guid

RichieOrtiz:
Updated XML helper to support storing and parsing of Guid.


RichieOrtiz:
@dotnet-policy-service agree

MpDzik:
Excellent addition, I would also include this in the `1.10.x` branch.

Issue #8718, opened 2023-08-08
Missing maximum length setting for TextField and TitlePart


AgostiniAlessandro:
As of now it's not possibile to specify a maximum length for TextFields and TitleParts, it would be useful to have a setting for the TextField and for the TitlePart to allow users to specify those values. 

Having these settings would make the data entry process more error-proof.

PR #8719, opened 2023-08-08
Feature/maxlengthsetting

AgostiniAlessandro:
Implemented what was described in the #8718 issue

### DIFF
--- a/src/Orchard.Web/Core/Common/Views/Body-Large.Editor.cshtml
+++ b/src/Orchard.Web/Core/Common/Views/Body-Large.Editor.cshtml
@@ -1,7 +1,14 @@
 ﻿@{
-    var htmlAttributes = new Dictionary<string, object> {
-        {"class", "text large"}
-    };
+    var htmlAttributes = (Dictionary<string, object>)Model.HtmlAttributes;
+
+    if (htmlAttributes == null) {
+        htmlAttributes = new Dictionary<string, object> {
+            {"class", "text large"}
+        };
+    }
+    else {
+        htmlAttributes["class"] = "text large";
+    }
 
     if (Model.Required == true) {
         htmlAttributes["required"] = "required";
--- a/src/Orchard.Web/Core/Common/Views/Body.Editor.cshtml
+++ b/src/Orchard.Web/Core/Common/Views/Body.Editor.cshtml
@@ -1,10 +1,16 @@
 ﻿@using Orchard.Utility.Extensions;
 @{
     string editorFlavor = Model.EditorFlavor;
+    var htmlAttributes = (Dictionary<string, object>)Model.HtmlAttributes;
 
-    var htmlAttributes = new Dictionary<string, object> {
-        {"class", editorFlavor.HtmlClassify()}
-    };
+    if (htmlAttributes == null) {
+        htmlAttributes = new Dictionary<string, object> {
+        {"class", editorFlavor.HtmlClassify()}};
+
+    }
+    else {
+        htmlAttributes["class"] = editorFlavor.HtmlClassify();
+    }
 
     if (Model.Required == true) {
         htmlAttributes["required"] = "required";
--- a/src/Orchard.Web/Core/Title/Settings/TitlePartSettings.cs
+++ b/src/Orchard.Web/Core/Title/Settings/TitlePartSettings.cs
@@ -9,6 +9,7 @@ namespace Orchard.Core.Title.Settings {
 
     
     public class TitlePartSettings {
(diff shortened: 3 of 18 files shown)


src/Orchard.Web/Core/Common/Views/Body.Editor.cshtml
```diff
@@ -1,10 +1,16 @@
 ﻿@using Orchard.Utility.Extensions;
 @{
     string editorFlavor = Model.EditorFlavor;
+    var htmlAttributes = (Dictionary<string, object>)Model.HtmlAttributes;
```
sebastienros, line 4: There should be a "behavior" attached to this HtmlAttributes property that will initialize the dictionary automatically to prevent such tests. Can you try?

src/Orchard.Web/Core/Title/Settings/TitlePartSettings.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel;
+using System.Linq;
+using System.Web;
+
+namespace Orchard.Core.Title.Settings {
+    public class TitlePartSettings {
+        [Range(0, 1024)]
```
sebastienros: why limit the limit?

sebastienros (approved):

src/Orchard.Web/Core/Title/Settings/TitlePartSettings.cs
MatteoPiovanelli: Maximum length for the Title property of the underlying record is set to 1024 characters both in the migration and the record itself already.

sebastienros: Whoever wrote this should have use a constant and shared it. Hint:

image

AgostiniAlessandro: I changed that value to a constant and modified the corresponding recipe too.

src/Orchard.Web/Core/Common/Views/Body-Large.Editor.cshtml
```diff
@@ -1,7 +1,14 @@
 ﻿@{
-    var htmlAttributes = new Dictionary<string, object> {
-        {"class", "text large"}
-    };
+    var htmlAttributes = (Dictionary<string, object>)Model.HtmlAttributes;
+
+    if (htmlAttributes == null) {
```
sebastienros, line 4: Can you try to set the attributes even if Model.HtmlAttributes is null, meaning don't check for it, just assign it and see if that works (some magic we made in Orchard1).

AgostiniAlessandro: I tried, it doesn't work. 
Anyway I found a bug introduced in the last commit and I fixed it and tested everything from scratch.

src/Orchard.Web/Core/Title/Settings/TitlePartSettings.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Web;
+
+namespace Orchard.Core.Title.Settings {
+
+    
+    public class TitlePartSettings {
+        // Whenever this constant is changed a new migration step must be created to update the length of the field on the DB
+        public const int MAX_TITLE_LENGTH = 1024;
```
sebastienros: Pascal case

```suggestion
        public const int MaxTitleLength = 1024;
```

src/Orchard.Web/Core/Common/Views/Body-Large.Editor.cshtml
sebastienros: I think it's because the existing property on the shapes is named `Attributes`. Can you try to use this one and even doing something like `Model.Attributes["required"] = "required"` for instance?
And it also already has a `IList<string> Classes` property which isn't null.

sebastienros:
@AndreaPiovanelliLaser let me know if you think we should review it again?

MatteoPiovanelli:
He only changed the the casing to PascalCase per your September recommendation. 
We've been testing this for a while on our test and staging environments, so if it's still fine for you, we are ok with it.

AndreaPiovanelli:
> @AndreaPiovanelliLaser let me know if you think we should review it again?

As Matteo said, I only changed the casing of the MaxTitleLength const, so I think this pr is safe to be merged.

PR #8723, opened 2023-09-01
Fixes null culture

### LINKED ISSUE #8722: Error during translation
Orchard.Localization.Services.GetCultureByName(string cultureName) throws an error when the cultureName is null
It happens during the translation phase

HermesSbicego-Laser:
fixes #8722

### DIFF
--- a/src/Orchard/Localization/Services/DefaultCultureManager.cs
+++ b/src/Orchard/Localization/Services/DefaultCultureManager.cs
@@ -89,7 +89,7 @@ namespace Orchard.Localization.Services {
         public CultureRecord GetCultureByName(string cultureName) {
             var cultures = GetAllCulturesByName();
             CultureRecord result;
-            cultures.TryGetValue(cultureName, out result);
+            cultures.TryGetValue(cultureName ?? "", out result);
 
             return result;
         }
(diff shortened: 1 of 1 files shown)


src/Orchard/Localization/Services/DefaultCultureManager.cs
```diff
@@ -89,7 +89,7 @@ public class DefaultCultureManager : ICultureManager {
         public CultureRecord GetCultureByName(string cultureName) {
             var cultures = GetAllCulturesByName();
             CultureRecord result;
-            cultures.TryGetValue(cultureName, out result);
+            cultures.TryGetValue(cultureName ?? "", out result);
```
sebastienros: If `""` doesn't have any meaning in terms of cultures and you just want a null record, then don't call `TryGetValue`, that will save two dictionary lookups.

sebastienros:
```c#
if (string.IsNullOrEmpty(cultureName)) {
  return null;
}
... same code as before
```

sebastienros: When will this cache be part of the dev branch?

HermesSbicego-Laser: We don't have plans to merge 1.10.x into dev but we could talk about it in the next weeks with @BenedekFarkas: if the merge results to be too complex we could try to cherry pick the commit?

sebastienros (approved):

Issue #8724, opened 2023-10-20
Exception for LazyFields: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/a4be7c68cb163fdc5655a3cd94d0e44341017835/src/Orchard.Web/Modules/Orchard.ContentPicker/Handlers/ContentPickerFieldHandler.cs#L33

Using the injected content manager inside the ContentPickerFieldHandler to get ContentItems property throws a System.ObjectDisposedException. This happens particularly often in pages where multiple ContentPickerFields are present (e.g. inside widgets).

AndreaPiovanelli:
Copying here the complete Exception for reference:

```
Orchard.Exceptions.DefaultExceptionPolicy - ComuneIvrea - An unexpected exception was caught [http://localhost/OrchardLocal/ComuneIvrea/vivere-il-comune]
System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it has already been disposed.
   in Autofac.Core.Lifetime.LifetimeScope.CheckNotDisposed()
   in Autofac.Core.Lifetime.LifetimeScope.ResolveComponent(IComponentRegistration registration, IEnumerable`1 parameters)
   in lambda_method(Closure )
   in Orchard.Data.TransactionManager.EnsureSession(IsolationLevel level) in D:\Laser.Orchard\Laser.Orchard\Orchard\src\Orchard\Data\SessionLocator.cs:riga 127
   in Orchard.Data.TransactionManager.Demand() in D:\Laser.Orchard\Laser.Orchard\Orchard\src\Orchard\Data\SessionLocator.cs:riga 60
   in Orchard.Data.TransactionManager.GetSession() in D:\Laser.Orchard\Laser.Orchard\Orchard\src\Orchard\Data\SessionLocator.cs:riga 56
   in Orchard.ContentManagement.DefaultContentManager.GetManyImplementation(QueryHints hints, Action`2 predicate) in D:\Laser.Orchard\Laser.Orchard\Orchard\src\Orchard\ContentManagement\DefaultContentManager.cs:riga 347
[...]

PR #8725, opened 2023-10-20
Lazy field loading using content item ContentManager instead of the injected one.

AndreaPiovanelli:
In refernce to issue #8724 , now lazy fields are loaded using the content item ContentManager (passed as a parameter to the Loading function inside the LoadContentContext. This avoids losing the database session when reading LazyFields.
The same patch has been applied to both ContentPickerFieldHandler and MediaLibraryPickerFieldHandler, because both were using the injected content manager when reading lazy fields.


src/Orchard.Web/Modules/Orchard.MediaLibrary/Handlers/MediaLibraryPickerFieldHandler.cs
```diff
@@ -37,7 +37,8 @@ public class MediaLibraryPickerFieldHandler : ContentHandler {
 
             foreach (var field in fields) {
                 var localField = field;
-                localField._contentItems = new Lazy<IEnumerable<MediaPart>>(() => _contentManager.GetMany<MediaPart>(localField.Ids, VersionOptions.Published, QueryHints.Empty).ToList());
```
sebastienros, line 40: Should the local _contentManager be removed then?

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.MediaLibrary/Handlers/MediaLibraryPickerFieldHandler.cs
AndreaPiovanelli: As you suggested, _contentManager variable has been removed from both ContentPickerFieldHandler and MediaLibraryPickerFieldHandler.

PR #8727, opened 2023-10-30
Prevent TermsFilter from having to deal with empty strings

MatteoPiovanelli:
We noticed there was a chance for the TermsFilter to throw an exception, locking backoffice users from being able to edit the filter and use it. This happened in some cases when the user would select the taxonomy rather than a term, because the value for the taxonomy is an empty string: that could result in the filter attempting to parse empty strings to integers.
So:

- filter them out before parsing them to int
- in the form, convert taxonomies to OptionGroups rather than options with no value, so they should not be selected so easily.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Projections/TermsFilter.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Projections/TermsFilter.cs
@@ -36,7 +36,10 @@ namespace Orchard.Taxonomies.Projections {
             var termIds = (string)context.State.TermIds;
 
             if (!String.IsNullOrEmpty(termIds)) {
-                var ids = termIds.Split(new[] { ',' }).Select(Int32.Parse).ToArray();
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Projections/TermsFilter.cs
```diff
@@ -36,7 +36,10 @@ public class TermsFilter : IFilterProvider {
             var termIds = (string)context.State.TermIds;
 
             if (!String.IsNullOrEmpty(termIds)) {
-                var ids = termIds.Split(new[] { ',' }).Select(Int32.Parse).ToArray();
+                var ids = termIds.Split(new[] { ',' })
```
sebastienros:
```suggestion
                var ids = termIds.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
```

sebastienros (approved):
Pending recommended changes

Issue #8742, opened 2023-12-18
Path generation is case sensitive and it potentially generates path that are actually the same path


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/9644ceda1f9b077d44aaa4ffaab103c19a59ddba/src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs#L243C25-L243C25

AutorouteService.GetSimilarPaths loads up all the DisplayAlias strings that start with provided path. After that, a unique slug is generated (https://github.com/OrchardCMS/Orchard/blob/9644ceda1f9b077d44aaa4ffaab103c19a59ddba/src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs#L188C25-L188C25) but the IEnumerable<string>.Contains(string) function used to check if the path doesn't exist yet is case sensistive. This causes the creation of conflicting aliases (i.e. a "/TEST-PAGE" alias can be created even if a "/test-page" alias already exists in the tenant).

PR #8743, opened 2023-12-19
Similar paths are now compared ignoring casing

AndreaPiovanelli:
Reference issue: #8742 
Added StringComparer and forced lowercase where needed to ensure matching paths to ignore casing.
This changes apply to new content or when autoroute is changed. This means that, for existing content, if user does not change the  current url textbox, the path generation is skipped, thus keeping ambiguous paths if present.
For this reason I opened a draft pull request for the moment, to find a way to avoid this ambiguity for existing content.  This can be solved mainly in two ways:
1) Forcing the slug / url generation even if the matching between url and content routevaluedictionary is passed (https://github.com/OrchardCMS/Orchard/blob/9644ceda1f9b077d44aaa4ffaab103c19a59ddba/src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs#L217)
2) Adding a migration step (in Orchard.Autoroute or Orchard.Alias) forcing the regeneration of all ambiguous urls.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Autoroute/Services/AutorouteService.cs
@@ -186,7 +186,7 @@ namespace Orchard.Autoroute.Services {
         }
 
         public string GenerateUniqueSlug(AutoroutePart part, IEnumerable<string> existingPaths) {
-            if (existingPaths == null || !existingPaths.Contains(part.Path))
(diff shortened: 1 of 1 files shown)


sebastienros (approved):

MatteoPiovanelli:
@sebastienros @AndreaPiovanelliLaser should we merge this in, and worry about possible pre-existing path conflicts some other time, since this should prevent any new ones from cropping up?

sebastienros:
I think you can merge, I approved the PR

PR #8748, opened 2024-01-23
#8747: Workaround for file extension with "." inside

### LINKED ISSUE #8747: Workaround for file extension out from filename.slice() with "." inside
https://github.com/OrchardCMS/Orchard/blob/97648ed5a2a6c43074216786126ff8a207e6ac2f/src/Orchard.Web/Modules/Orchard.Media/Views/Admin/Add.cshtml#L45

In some cases the file name out from filename.slice(filename.lastIndexOf(".")).toLowerCase() still contains the char "." and fails the next check, so replacing it with an empty string fix the problem

gdessimoneinvalle:
Fixes #8747 

replace "." char with "" in filename.lastIndexOf(".")).toLowerCase() to avoid failure in file type check (that happens sometimes)


gdessimoneinvalle:
@dotnet-policy-service agree company="INVA Spa"

BenedekFarkas:
See #8747.

Issue #8749, opened 2024-01-23
Performance issue when opening the widget admin page


gdessimoneinvalle:
https://github.com/OrchardCMS/Orchard/blob/97648ed5a2a6c43074216786126ff8a207e6ac2f/src/Orchard.Web/Modules/Orchard.Widgets/Controllers/AdminController.cs#L87

When many content pages contain widgets we experience perfomance issues as the widget admin page takes even some minutes to open

BenedekFarkas:
Continuing from #8750:

A generic-purpose solution to this could be to optionally move the layer filtering to the server-side. Right now, when you open the Widgets page and select any layer, you still see the widgets from each layer and you can hide/show the widgets of each layer on the client-side using the layer list on the right side.

This could be overridden for example by adding a site setting, which, when enabled would filter the displayed widgets down to the selected layer only by using an override of GetWidgets (with `currentLayer.Id` passed into it) at the place highlighted by @gdessimoneinvalle above.
Since this would make the client-side filtering useless, that should be hidden too. A new, "Show all" option (with value 0) needs to be added to the layer dropdown too.

It can be taken one step further by the above site setting just providing a default value and the Widgets screen having an additional checkbox for the same purpose, but its value stored in a cookie, so admins can switch between the two behaviors for themselves.

CC: @MatteoPiovanelli-Laser

MatteoPiovanelli:
Just for the sake of discussion, the things we could filter widgets on for that action are:

1. Layer: this is what we are discussing, and I think this is what makes the most sense to have in a site setting.
2. Zone: the usefulness of this in the context of this controller/action is debatable. Personally, I think the job of this controller is to display everything, but, since it's grouping things by Zone anyway, it may be worth it to set that up as a filter. Overall, I think this would be overengineering this feature.
3. ContentType: I think it makes little sense to have a setting to filter this page for this.

So I think the site setting should definitely let admins select default layers whose widgets are loaded/shown; probably nothing else.
I think the UI for client side filtering may still be there, but rather than operate entirely client-side, it may call a server-side api for the updated information.

I like the approach of using the cookie for admins so they may override their own setting, but I'm not sure I'd actually implement it, unless there is a trivial way to handle this:
[...]

PR #8750, opened 2024-01-24
Exclude ContentWidgets layer by default in order to increase performance

gdessimoneinvalle:
In reference to #8749

Added a filter to exlude widgets from list if the layer ContentWidgets is not explicity chosen by the user in order to open the widget page more quickly

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Widgets/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Modules/Orchard.Widgets/Controllers/AdminController.cs
@@ -63,6 +63,8 @@ namespace Orchard.Widgets.Controllers {
 
             IEnumerable<LayerPart> layers = _widgetsService.GetLayers().OrderBy(x => x.Name).ToList();
 
+
+
             if (!layers.Any()) {
                 Services.Notifier.Error(T("There are no widget layers defined. A layer will need to be added in order to add widgets to any part of the site."));
                 return RedirectToAction("AddLayer");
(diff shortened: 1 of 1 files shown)


sebastienros:
@MatteoPiovanelli-Laser

src/Orchard.Web/Modules/Orchard.Widgets/Controllers/AdminController.cs
```diff
@@ -84,14 +86,27 @@ public class AdminController : Controller, IUpdateModel {
             string zonePreviewImagePath = string.Format("{0}/{1}/ThemeZonePreview.png", currentTheme.Location, currentTheme.Id);
             string zonePreviewImage = _virtualPathProvider.FileExists(zonePreviewImagePath) ? zonePreviewImagePath : null;
```
BenedekFarkas, line 87:
```suggestion
            var zonePreviewImagePath = string.Format("{0}/{1}/ThemeZonePreview.png", currentTheme.Location, currentTheme.Id);
            var zonePreviewImage = _virtualPathProvider.FileExists(zonePreviewImagePath) ? zonePreviewImagePath : null;
```

src/Orchard.Web/Modules/Orchard.Widgets/Controllers/AdminController.cs
```diff
@@ -84,14 +86,27 @@ public class AdminController : Controller, IUpdateModel {
...
+                var IdsLayers = layers.Where(x => x.Name.ToLowerInvariant() != "contentwidgets").Select(x => x.Id).ToList().ToArray();
+                if (layerId != null && (int)layerId == IdLayerContentWidget) {
+                    //If ContentWidgets is chosen explicitly then IdsLayers is not filtered, note that the editors should be advised that this can take lot of time
+                    IdsLayers = layers.Select(x => x.Id).ToList().ToArray();
+                }
+                widgets = _widgetsService.GetWidgets(IdsLayers);
+            }
+            else {
+                widgets = _widgetsService.GetWidgets();
+            }
```
BenedekFarkas, line 103: Simplified logic with corrected variable naming.
```suggestion
            // Only listing the widgets of the "ContentWidgets" layer when it's explicitly chosen, not by default (for
            // performance reasons).
            IEnumerable<WidgetPart> widgets;
            var contentWidgetsLayer = layers.Where(x => x.Name.ToLowerInvariant() == "contentwidgets").FirstOrDefault();
            if (contentWidgetsLayer == null) {
                widgets = _widgetsService.GetWidgets();
            }
            else { // The "ContentWidgets" layer exists.
                var layerIds = layers.Select(x => x.Id).ToList();
                // There's no layer selected or it's not "ContentWidgets", so excluding it from the widget listing.
                if (!layerId.Equals(contentWidgetsLayer.Id)) {
                    layerIds.Remove(contentWidgetsLayer.Id);
                }

                widgets = _widgetsService.GetWidgets(layerIds.ToArray());
            }
```

src/Orchard.Web/Modules/Orchard.Widgets/Controllers/AdminController.cs
```diff
@@ -63,6 +63,8 @@ public class AdminController : Controller, IUpdateModel {
 
             IEnumerable<LayerPart> layers = _widgetsService.GetLayers().OrderBy(x => x.Name).ToList();
 
+
+
```
BenedekFarkas, line 67:
```suggestion
```

BenedekFarkas (changes requested):

BenedekFarkas:
When/what creates the ContentWidgets layer? On latest 1.10.x, I added a content widget to the home page, but there was no ContentWidgets layer created and I couldn't find any reference to other than this code.

MatteoPiovanelli:
> When/what creates the ContentWidgets layer? On latest 1.10.x, I added a content widget to the home page, but there was no ContentWidgets layer created and I couldn't find any reference to other than this code.

It's created by a feature. I think that's an issue with this PR. The optimization makes sense to me, because we also sometimes generate layers/widgets dynamically.
But the list of layers that are "filtered out" should be set in a Settings, rather than hardcoded, I think.

BenedekFarkas:
OK, but which feature and where? I can't find any clue to this layer existing in the code.

MatteoPiovanelli:
One of our own features (hence it wouldn't be in this repo) adds a layer with that same name, so it's likely to be that one. That feature can easily lead to having many widgets.

BenedekFarkas:
In that case this code cannot be accepted like this.

Out of curiosity, how many Content Widgets are we talking about? Isn't the performance issue a SELECT N+1 for each of them? Maybe it's an inherent issue with Content Widgets we can address.

MatteoPiovanelli:
Can be a few widgets, but it's easily hundreds. The feature lets you add widgets that are specific to specific ContentItems.
In the BO page that displays all widgets, I don't think we have an N+1 issue, because the queries are those from the Orchard.Widgets Controller/service, that simply gets all Widgets. 
It becomes a potential performance bottleneck when it fetches all those widgets (each one being a contentitem), fires off all their handlers, then renders them out (grouping them by zone and layer, if I remember correctly).

BenedekFarkas:
OK, thanks! I'll close this PR, but we can discuss under the issue if there's an Orchard-specific improvement that can help in this case too - I have an idea.

PR #8757, opened 2024-02-16
Added data annotation to have a proper validation on MenuText length 

### LINKED ISSUE #8756: MenuText throws an exception when user sets a very long text when creating menu items
https://github.com/OrchardCMS/Orchard/blob/90b104ed74234bdaaedf2b730166eddc66ad4d4c/src/Orchard.Web/Core/Navigation/ViewModels/MenuPartViewModel.cs#L11

When a user sets a MenuText longer than 255 characters, an exception is thrown because data would be truncated inside the database.
The MenuText column of Navigation_MenuPartRecord table has a datatype of nvarchar(255) but there is no validation.
I noticed that there is a proper data annotation for Orchard.Core.Navigation.Models.MenuPart but data annotation is missing for Orchard.Core.Navigation.ViewModels.MenuPartViewModel, which is the model used inside the EditorTemplates/Parts.Navigation.Menu.Edit.cshtml shape.

AndreaPiovanelli:
As the title says. I added a data annotation to validate MenuText length based on its maximum length on database (MenuPartRecord.DefaultMenuTextLength).

Reference issue: #8756

### DIFF
--- a/src/Orchard.Web/Core/Navigation/ViewModels/MenuPartViewModel.cs
+++ b/src/Orchard.Web/Core/Navigation/ViewModels/MenuPartViewModel.cs
@@ -1,5 +1,7 @@
 ﻿using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using Orchard.ContentManagement;
+using Orchard.Core.Navigation.Models;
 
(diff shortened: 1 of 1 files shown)


BenedekFarkas (approved):
Tested and can confirm it's working!

PR #8764, opened 2024-03-01
Updated ClickToBuild.cmd to support VS2019 (v16) and VS2022 (v17)

### LINKED ISSUE #8758: Building latest 1.10.x from command line
I'm just updating my orchard application. It compiles in Visual Studio 2022, but does not compiled using the command line build.

It give a syntax error message "; expected" for the following line

https://github.com/OrchardCMS/Orchard/blob/90b104ed74234bdaaedf2b730166eddc66ad4d4c/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs#L157

I believe that there is a version conflict between the compiler that is being called from the command line and the version of the compiler required to handle the c# language constructs used in SmtpMessageChannel.cs (version 8, I believe). I believe that ClickToBuild.cmd determines which version of msbuild to use, and in my case it is using a version that does not support the more advanced language feature. 

Any suggestions on whether this is due to version issues on my computer or if the ClickToBuild.cmd file can be updated to resolve the issue?

Thank you!

dmhiggins23:
Removed support for eariler MSBuild versions (they cannot compile SmtpMessageChannel.cs)

Updated lib/nuget/nuget.exe to latest version (6.9.1.3) (required for package restoration)

Modified Obsolete attribute in SmtpSettingsPart.cs so the solution builds with VS2019

### DIFF
--- a/ClickToBuild.cmd
+++ b/ClickToBuild.cmd
@@ -3,29 +3,16 @@
 REM Necessary for the InstallDir variable to work inside the MsBuild-finding loop below.
 SETLOCAL ENABLEDELAYEDEXPANSION
 
-for /f "usebackq tokens=1* delims=: " %%i in (`lib\vswhere\vswhere -latest -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild`) do (
+for /f "usebackq tokens=1* delims=: " %%i in (`lib\vswhere\vswhere -latest -version "[16.0,18.0)" -requires Microsoft.Component.MSBuild`) do (
   if /i "%%i"=="installationPath" (
 	set InstallDir=%%j
 	echo !InstallDir!
-	if exist "!InstallDir!\MSBuild\15.0\Bin\MSBuild.exe" (
-		echo "Using MSBuild from Visual Studio 2017"
-		set msbuild="!InstallDir!\MSBuild\15.0\Bin\MSBuild.exe"
+	if exist "!InstallDir!\MSBuild\Current\Bin\MSBuild.exe" (
+		echo "Using MSBuild from !InstallDir!"
+		set msbuild="!InstallDir!\MSBuild\Current\Bin\MSBuild.exe"
 		goto build
 	)
   )
-)
-
-FOR %%b in (
-       "%VS140COMNTOOLS%\vsvars32.bat"
-       "%VS120COMNTOOLS%\vsvars32.bat"
-       "%VS110COMNTOOLS%\vsvars32.bat"
-    ) do (
-    if exist %%b ( 
-		echo "Using MSBuild from %%b"
-		call %%b
-		set msbuild="msbuild"
-		goto build
-    )
 )
 
 echo "Unable to detect suitable environment. Build may not succeed."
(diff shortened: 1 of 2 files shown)


dmhiggins23:
@dotnet-policy-service agree

sebastienros:
This is removing support for VS 11,12,14,and 15? Is there a reason to not support them anymore?

dmhiggins23:
To my knowledge, the version of msbuild that comes with those earlier versions does not support the advanced language features that are currently in SmtpMessageChannel.cs 

https://github.com/OrchardCMS/Orchard/blob/50d416c9f2b1ebecd56bfbb0c8933768493fc89f/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs#L157

So, even if ClickToBuild.cmd found and used one of those earlier versions. The build would fail. As it was, I updated SmtpSettingsPart.cs so version 16 (vs 2019) would work

https://github.com/OrchardCMS/Orchard/blob/50d416c9f2b1ebecd56bfbb0c8933768493fc89f/src/Orchard.Web/Modules/Orchard.Email/Models/SmtpSettingsPart.cs#L34

I only tested vs 2017 and later. The current head of the dev branch compiles with Visual Studio 2017 (version 15). The current head of the 1.10.x branch does not. The problem appears to be restricted to the files I referenced above. 

If you want to keep the door open for the earlier versions of msbuild in ClickToBuild.cmd, I can redo it so that it just adds support for versions 16 and 17, and leave support for the other versions in there. Then you could defer the question of the use of the advanced c# features to a later time. It is probably the most conservative approach.

sebastienros:
Let's merge and see who would be broken. @MatteoPiovanelli-Laser are you using a recent version of VS ? > 15

sebastienros (approved):

sebastienros:
@MatteoPiovanelli-Laser please merge if this doesn't break you

BenedekFarkas:
Interpolated strings were introduced in C# 6 and we were supposed to support C# 7.3 - at least in dev. The above changes (= the fact that there was only one file that was incompatible with VS 2017) tell me that we intentionally did not use higher-version language features (but I don't remember :( ), but in that case the csprojs should reflect (by language version restriction) which VS version we want to remain compatible with.

Better yet, setting langversion to default will follow the .NET Framework version, which is already 4.8 on both 1.10.x and dev, where 7.3 is the default, which is compatible with VS 2019 (VS version 16).

We were actually restricting the language version specifically to 7.3 in each csproj, but it was overwritten to "latest" some time ago. That's a separate problem, because AFAIR it's still not recommended to go above 7.3 for .NET FW applications due to possible runtime issues, even if it compiles.

In any case, at least according to [this Wikipedia article](https://en.wikipedia.org/wiki/C_Sharp_(programming_language)#Versions), even VS 2015's MSBuild should be able to compile interpolated strings, so if that's true, then the changes in this PR are in conflict with each other.

dmhiggins23:
I believe the language features in question are:

- switch expressions (C# version 8?)
- Constant interpolated strings (required for attributes; C# version 10?)

BenedekFarkas:
Oh, I see, that's because the interpolated string is used in an attribute that requires a constant. I glossed over that, you're right! Thanks for the clarification.

Issue #8765, opened 2024-03-07
Orchard.Taxonomies.Handlers.TermsPartHandler: Lifetime scope already disposed when accessing LazyField


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/50d416c9f2b1ebecd56bfbb0c8933768493fc89f/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs#L67

When accessing TermsPart.Record (which is a LazyField - https://github.com/OrchardCMS/Orchard/blob/50d416c9f2b1ebecd56bfbb0c8933768493fc89f/src/Orchard/ContentManagement/ContentPart.cs#L127) inside a delegate, it sometimes throws a ObjectDisposedException because the lifetime scope has already been disposed.

PR #8766, opened 2024-03-07
TermsPart.Record usage made before the delegate to avoid accessing a lifetime scope which is already disposed.

AndreaPiovanelli:
In reference to issue #8765 , terms info is loaded outside of the delegate, to void accessing the LazyField Record later in the execution of the code, potentially throwing the ObjectDisposedException described in the issue.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs
@@ -60,13 +60,22 @@ namespace Orchard.Taxonomies.Handlers {
             var queryHint = new QueryHints()
                 .ExpandRecords("ContentTypeRecord", "CommonPartRecord", "TermsPartRecord");
 
+            // Get TermRecordIds for each field before the delegate inside the for iterator.
+            // This avoids the TermsPart.Record lifetime scope to be disposed when executed, since ContentPart.Record is a LazyField.
+            var groupedRecordIds = part.Record.Terms
+                .GroupBy(tci => tci.Field)
+                .ToDictionary(g => g.Key,
+                    g => g.Select(tci => tci.TermRecord.Id).ToArray());
 
             foreach (var field in part.ContentItem.Parts.SelectMany(p => p.Fields).OfType<TaxonomyField>()) {
                 var tempField = field.Name;
                 field.TermsField.Loader(() => {
-                    var fieldTermRecordIds = part.Record.Terms.Where(t => t.Field == tempField).Select(tci => tci.TermRecord.Id);
-                    // Using context content item's ContentManager instead of injected one to avoid lifetime scope exceptions in case of LazyFields.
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs
```diff
@@ -60,13 +60,22 @@ public class TermsPartHandler : ContentHandler {
             var queryHint = new QueryHints()
                 .ExpandRecords("ContentTypeRecord", "CommonPartRecord", "TermsPartRecord");
 
+            // Get TermRecordIds for each field before the delegate inside the for iterator.
+            // This avoids the TermsPart.Record lifetime scope to be disposed when executed, since ContentPart.Record is a LazyField.
+            var groupedRecordIds = part.Record.Terms
```
sebastienros, line 65: Could you check the SQL traces before and after the change? It is possible that now you are doing queries all the time instead of only when necessary, or even that you are loading records that would have never been before. Hence the reason of "lazy" loading.

AndreaPiovanelli: > Could you check the SQL traces before and after the change? It is possible that now you are doing queries all the time instead of only when necessary, or even that you are loading records that would have never been before. Hence the reason of "lazy" loading.

Here's the test I did on the matter, after enabling NHibernate logging:
I created a new content type, adding a TaxonomyField to it. After that, I displayed the content item at front end in the following four scenarios:
1) Old loader, TaxonomyField displayed
2) Old loader, TaxonomyField hidden (via placement.info)
3) New loader, TaxonomyField displayed
4) New loader, TaonomyField hidden (via placement.info)

In each case the logger logged the same following queries:
- One query on Orchard_Taxonomies_TermsPartRecord, filtered by column "id" = the id of the displayed content item.
- One query on Orchard_Taxonomies_TermContentItem, filtered by column "TermsPartRecord_id" = the id of the displayed content item.
- One query on Orchard_Framework_ContentItemVersionRecord joined with other tables, the notable of them are Orchard_Framework_ContentItemRecord and Orchard_Taxonomies_TermsPartRecord, where ContentItemRecord.Id is the id of the selected term for the displayed content item.

This means that the new loader doesn't add (or remove) any query, but caches some info (the TermsPartRecord) to avoid accessing a disposed session or transaction later in the call.

The query to get the record is executed inside the StorageFilter for the repository (of TermsPartRecord), which is added here:
https://github.com/OrchardCMS/Orchard/blob/50d416c9f2b1ebecd56bfbb0c8933768493fc89f/src/Orchard.Web/Modules/Orchard.Taxonomies/Handlers/TermsPartHandler.cs#L34
The Record is loaded in any case here:
https://github.com/OrchardCMS/Orchard/blob/50d416c9f2b1ebecd56bfbb0c8933768493fc89f/src/Orchard/ContentManagement/Handlers/StorageFilter.cs#L30

sebastienros (approved):

PR #8768, opened 2024-03-07
New pr to update MailKit, ideally replacing #8746

AndreaPiovanelli:
This should replace #8746 starting from the up-to-date 1.10.x branch.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
+++ b/src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
@@ -14,11 +14,10 @@ using Orchard.Logging;
 using SmtpClient = MailKit.Net.Smtp.SmtpClient;
 
 namespace Orchard.Email.Services {
-    public class SmtpMessageChannel : Component, ISmtpChannel, IDisposable {
+    public class SmtpMessageChannel : Component, ISmtpChannel {
         private readonly SmtpSettingsPart _smtpSettings;
         private readonly IShapeFactory _shapeFactory;
         private readonly IShapeDisplay _shapeDisplay;
-        private readonly Lazy<SmtpClient> _smtpClientField;
         public static readonly string MessageType = "Email";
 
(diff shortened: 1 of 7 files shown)


src/Orchard.Web/Modules/Orchard.Email/Services/SmtpMessageChannel.cs
```diff
@@ -116,7 +106,11 @@ public class SmtpMessageChannel : Component, ISmtpChannel, IDisposable {
                     mailMessage.ReplyTo.AddRange(ParseRecipients(emailMessage.ReplyTo));
                 }
 
-                _smtpClientField.Value.Send(mailMessage);
+                using (var smtpClient = CreateSmtpClient()) {
```
sebastienros, line 109: Might be worth having an overload which takes multiple emails to reuse the client, otherwise bulk-sending them will create as many connections.

sebastienros (approved):

BenedekFarkas:
Please merge the latest 1.10.x to fix the build error.

BenedekFarkas:
@AndreaPiovanelli @MatteoPiovanelli-Laser would you like to continue working on this or #8746?

MatteoPiovanelli:
I've rebased this one as well. At this point I'm just hoping either of the PRs passes github's checks

BenedekFarkas:
Is there any particular reason why you didn't update to the latest version at the time? It looks like that 4.4.0 or 4.5.0 was already out.
Anyway, can you please try upgrading 4.8.0 (latest)? The build should pass with that version (I just tested it on a runner VM).

EDIT: I see you wrote this in the other PR's description:

> We tried to upgrade the version of the NuGet package, but there were several issues with its dependencies (https://github.com/jstedfast/MailKit/issues/1462, https://github.com/jstedfast/MimeKit/issues/872, https://github.com/dotnet/sdk/issues/31147) so we were only able to advance it to 3.4.2.

BenedekFarkas:
Superseded by #8822.

PR #8775, opened 2024-03-12
Small optimization in TaxonomyService

MatteoPiovanelli:
Memorize results of queries to prevent repeated execution within a request.
We noticed a few instances where the same query would be invoked several times in the same request. For example, when several contentItems of a same contentType are in the same list.

This allows skipping a few of those queries.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
@@ -32,6 +32,9 @@ namespace Orchard.Taxonomies.Services {
 
         private readonly HashSet<int> _processedTermPartIds = new HashSet<int>();
 
+        private Dictionary<string, TaxonomyPart> _taxonomiesByName;
+        private Dictionary<string, TaxonomyPart> _taxonomiesBySlug;
+
         public TaxonomyService(
             IRepository<TermContentItem> termContentItemRepository,
             IContentManager contentManager,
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
```diff
@@ -54,13 +57,21 @@ public class TaxonomyService : ITaxonomyService {
...
+
+            // initialize memorization structures
+            _taxonomiesByName = new Dictionary<string, TaxonomyPart>();
+            _taxonomiesBySlug = new Dictionary<string, TaxonomyPart>();
         }
 
         public ILogger Logger { get; set; }
         public Localizer T { get; set; }
 
+        private IEnumerable<TaxonomyPart> _taxonomies;
```
sebastienros, line 69:
```suggestion
        private List<TaxonomyPart> _taxonomies;
```

Since the type is known and this is a private member

sebastienros: Unless it's not what the method returns

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Taxonomies/Services/TaxonomyService.cs
MatteoPiovanelli: `IContentQuery<T>.List()` (which is what's assigned and eventually returned here) returns an `IEnumerable<T>`. Definitely not a `List<T>`.

PR #8780, opened 2024-04-09
Minor syntax fix on TaxonomyFieldList.cshtml

AndreaPiovanelli:
Missing closing brackets after providing data for the ajax call


src/Orchard.Web/Modules/Orchard.Taxonomies/Views/EditorTemplates/Fields/TaxonomyFieldList.cshtml
BenedekFarkas, line 1: Nice catch! Please add an empty line at the end of the file.

BenedekFarkas (changes requested):

BenedekFarkas (approved):

PR #8781, opened 2024-04-10
8225: Adding a checkbox to StringFilterForm to control whether an empty value should cause the filter to be skipped

### LINKED ISSUE #8225: Return all items by projection in ContainsAll if provided value is empty

BenedekFarkas:
Fixes #8225

The changes from #8213 are also removed as that functionality is now covered, as well as that of #8226.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/StringFilterForm.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/StringFilterForm.cs
@@ -55,10 +55,6 @@ namespace Orchard.Projections.FilterEditors.Forms {
                     f._Operator.Add(new SelectListItem { Value = Convert.ToString(StringOperator.Ends), Text = T("Ends with").Text });
                     f._Operator.Add(new SelectListItem { Value = Convert.ToString(StringOperator.NotEnds), Text = T("Does not end with").Text });
                     f._Operator.Add(new SelectListItem { Value = Convert.ToString(StringOperator.NotContains), Text = T("Does not contain").Text });
-                    f._Operator.Add(new SelectListItem {
-                        Value = Convert.ToString(StringOperator.ContainsAnyIfProvided),
-                        Text = T("Contains any word (if any is provided)").Text
-                    });
 
                     return f;
                 };
@@ -103,14 +99,6 @@ namespace Orchard.Projections.FilterEditors.Forms {
                     return y => y.Not(x => x.Like(property, Convert.ToString(value), HqlMatchMode.End));
                 case StringOperator.NotContains:
                     return y => y.Not(x => x.Like(property, Convert.ToString(value), HqlMatchMode.Anywhere));
-                case StringOperator.ContainsAnyIfProvided:
-                    if (string.IsNullOrWhiteSpace((string)value))
-                        return x => x.IsNotEmpty("Id"); // basically, return every possible ContentItem
-                    var values3 = Convert.ToString(value)
-                        .Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
-                    var predicates3 = values3.Skip(1)
-                        .Select<string, Action<IHqlExpressionFactory>>(x => y => y.Like(property, x, HqlMatchMode.Anywhere)).ToArray();
-                    return x => x.Disjunction(y => y.Like(property, values3[0], HqlMatchMode.Anywhere), predicates3);
                 default:
                     throw new ArgumentOutOfRangeException();
             }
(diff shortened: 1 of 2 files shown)


sebastienros (approved):

BenedekFarkas:
@MatteoPiovanelli-Laser @LorenzoFrediani-Laser please see the changes here. If you're already running the code in your PRs linked above, then you can update your Queries with a recipe for example - adapting the filter configurations is a simple search and replace.
Basically, if you're using `ContainsAnyIfProvided` or `ContainsAllIfProvided`, then you should change it to use `ContainsAny` or `ContainsAll` respectively and have the checkbox enabled.

AndreaPiovanelli:
> @MatteoPiovanelli-Laser @LorenzoFrediani-Laser please see the changes here. If you're already running the code in your PRs linked above, then you can update your Queries with a recipe for example - adapting the filter configurations is a simple search and replace. Basically, if you're using `ContainsAnyIfProvided` or `ContainsAllIfProvided`, then you should change it to use `ContainsAny` or `ContainsAll` respectively and have the checkbox enabled.

I believe the safest way to ensure compatibility of the new StringFilterForm operators and to properly add the new boolean parameter where needed inside the FilterRecords is to add a migration step to Orchard.Projections, perhaps using Orchard.Forms.Services.FormParametersHelper to correct the form state.

BenedekFarkas:
> I believe the safest way to ensure compatibility of the new StringFilterForm operators and to properly add the new boolean parameter where needed inside the FilterRecords is to add a migration step to Orchard.Projections, perhaps using Orchard.Forms.Services.FormParametersHelper to correct the form state.

I didn't find a proper way to manipulate the form, but we can still adjust the filter in any way to cover the earlier cases.
BTW are you already using `ContainsAnyIfProvided` in production? Honestly I'm not really keen on adding a migration step just for this, especially on 1.10.x after the yak shaving I had to do with merging 1.10.x into dev the last time due to conflicting migrations.

Also, this is still unreleased code, so I'm not that worried about breaking changes. But if you're already running this in production and you can't update filters easily with recipes, then I'm happy to work towards a solution that works for you too.

src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/StringFilterForm.cs
```diff
@@ -36,8 +36,14 @@ public class StringFilterForm : IFormProvider {
                             Title: T("Value"),
                             Classes: new[] { "text medium", "tokenized" },
                             Description: T("Enter the value the string should be.")
-                            )
-                        );
+                        ),
+                        _IgnoreIfEmptyValue: Shape.Checkbox(
+                            Id: "IgnoreFilterIfValueIsEmpty",
+                            Name: "IgnoreFilterIfValueIsEmpty",
+                            Title: "Ignore filter if value is empty",
```
AndreaPiovanelli: Can we add T("Ignore filter if value is empty"), please?

AndreaPiovanelli:
> > I believe the safest way to ensure compatibility of the new StringFilterForm operators and to properly add the new boolean parameter where needed inside the FilterRecords is to add a migration step to Orchard.Projections, perhaps using Orchard.Forms.Services.FormParametersHelper to correct the form state.
> 
> I didn't find a proper way to manipulate the form, but we can still adjust the filter in any way to cover the earlier cases. BTW are you already using `ContainsAnyIfProvided` in production? Honestly I'm not really keen on adding a migration step just for this, especially on 1.10.x after the yak shaving I had to do with merging 1.10.x into dev the last time due to conflicting migrations.
> 
> Also, this is still unreleased code, so I'm not that worried about breaking changes. But if you're already running this in production and you can't update filters easily with recipes, then I'm happy to work towards a solution that works for you too.

I believe we may use a Repository<FilterRecord>, and we want to edit the column "State", which is similar to: 
`<Form><Description></Description><Operator>ContainsAllIfProvided</Operator><Value>{Request.QueryString:q}</Value></Form>`
These rows need to be updated, with the help of FormParametersHelper, adding the parameter "IgnoreIfEmptyValue" when needed.
The migration ensures every row is updated properly and in a multi-tenant, multi-instance with tens of tenants, using a recipe for each tenant is tricky.

If needed, I can implement the migration phase and give you a branch to cherry pick from or a gist.

src/Orchard.Web/Modules/Orchard.Projections/FilterEditors/Forms/StringFilterForm.cs
BenedekFarkas: Absolutely, good catch!

BenedekFarkas: Fixed in 9481a66fbf1ceef908e7161b4e2ced7a76800787.

BenedekFarkas:
That's all right and I knew how to migrate the data, I just wanted to avoid having it. :)
Please check the latest changes and let me know if it's working OK for you too!

AndreaPiovanelli:
> That's all right and I knew how to migrate the data, I just wanted to avoid having it. :) Please check the latest changes and let me know if it's working OK for you too!

I confirm migration works and query execution seems good.

BenedekFarkas:
> I confirm migration works and query execution seems good.

Awesome, thanks!

BenedekFarkas:
This change is now in `dev` too!

PR #8784, opened 2024-04-15
#6793: Adding a content-independent culture selector shape for the front-end

### LINKED ISSUE #6793: Localization Module don't provide a proper culture selector for front-end
Current Localization module provides a culture selector shape (FrontEndCultureSelector) for front-end that allows user to change to a different culture but it always redirects user to the home page instead of redirecting you to the culture version of the current page.
There is an alternative for pages with LocalizationPart that allows you to show links for the different cultures for which that page is available.
Problem is I have dynamic forms without LocalizationPart using a visibility rule based on the user language to decide which fields to show. 
Another problem of current FrontEndCultureSelector is unexpectedly it forces you to use LocalizationPart in your home page. Furthermore, any content you get through Ajax won't be localized because by default cookie by default it is not used, you need to customize it.
In a site with some pages with LocalizationPart and other without it but with different appearance depending on the language problems arise for usability. The point is FrontEndCultureSelector and LocalizationPart shape doesn't provide a consistent way for changing culture. One redirects the user to the home page, and the other is only available sometimes.

BenedekFarkas:
Fixes #6793

Originally developed by [Sergio](https://github.com/jersiovic) from [XK Project](https://github.com/xkproject).


src/Orchard/Utility/Extensions/NameValueCollectionExtensions.cs
BenedekFarkas, line 1: @sebastienros Re: https://github.com/OrchardCMS/Orchard/pull/6794#discussion_r1561371776

QueryStringBuilder is in ASP.NET Core (I didn't find a .NET FW-equivalent). :/
However, Orchard.DynamicForms already had this exact same extension method too, so I moved it to Orchard.Framework.

src/Orchard.Web/Modules/Orchard.Localization/Services/LocalizationService.cs
```diff
@@ -87,5 +91,51 @@ public class LocalizationService : ILocalizationService {
...
+            if (!routableContent.Parts.Any(p => p.Is<ILocalizableAspect>())) {
+                localizedRoute = null;
+                return false;
+            }
+
+            IEnumerable<LocalizationPart> localizations = ((ILocalizationService)this).GetLocalizations(routableContent, VersionOptions.Published);
+
+            ILocalizableAspect localizationPart = null, siteCultureLocalizationPart = null;
+            foreach (LocalizationPart l in localizations) {
+                if (l.Culture.Culture == cultureName) {
```
sebastienros: We are currently normalizing these in OC, should it compare with ignore case?

src/Orchard.Web/Modules/Orchard.Localization/Services/LocalizationService.cs
```diff
@@ -87,5 +91,51 @@ public class LocalizationService : ILocalizationService {
...
+                localizationPart = siteCultureLocalizationPart;
+            }
+
+            if (localizationPart == null) {
+                localizedRoute = null;
+                return false;
+            }
+
+            ContentItem localizedContentItem = localizationPart.ContentItem;
+            localizedRoute = localizedContentItem.Parts.Single(p => p is AutoroutePart).As<AutoroutePart>();
```
sebastienros: Single is evil, it will throw if it's not exactly one element. But you might also want that.

src/Orchard/Utility/Extensions/NameValueCollectionExtensions.cs
sebastienros: ok

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Localization/Services/LocalizationService.cs
BenedekFarkas: Thanks, I didn't notice it. As far as I can tell, it's not necessary to use it here. `.Single` is used in some other parts in Orchard (mostly in tests, but not only) to find a part by its name, but when the type is known, `.As<>` should be fine, since you can't have more than one of a (statically typed) part. Fixed in bf9b9b0c9eec78f27a82b633533e7e7f5d97aba2.

src/Orchard.Web/Modules/Orchard.Localization/Services/LocalizationService.cs
BenedekFarkas: Good idea! Updated in a5e4f1c147e732371a3966df8551e77c08260bbf.

Issue #8794, opened 2024-05-08
Form submit event triggering before content publish in Orchard.CustomForms causes issues when reading content items


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/86bcf413e703da88c21c5dbd329e76482fae7da4/src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs#L202

Since event triggering is made before the actual publishing of content items, the content passed in the event context (e.g. for workflows) isn't up to date. This causes, for example, some field values to be empty, making it impossible for the following workflow steps to actually use the content item.

AndreaPiovanelli:
Closed issue since #8795 has been merged into dev.

PR #8795, opened 2024-05-08
Content Item published before Custom Form submit event triggers

AndreaPiovanelli:
Moved content item publish before triggering form submit events.
Refreshed the content item passed in the context of rules manager and workflow manager.

This solves issue #8794

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
+++ b/src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
@@ -202,7 +202,7 @@ namespace Orchard.CustomForms.Controllers {
             if (customForm.SaveContentItem) {
                 conditionallyPublish(contentItem);
                 // Refresh content item
-                contentItem = _contentManager.Get(contentItem.Id);
+                contentItem = _contentManager.Get(contentItem.Id, VersionOptions.Latest);
             }
 
             // triggers any event
--- a/src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
+++ b/src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
@@ -198,6 +198,13 @@ namespace Orchard.CustomForms.Controllers {
 
             contentItem.As<ICommonPart>().Container = customForm.ContentItem;
 
+            // save the submitted form
+            if (customForm.SaveContentItem) {
+                conditionallyPublish(contentItem);
+                // Refresh content item
+                contentItem = _contentManager.Get(contentItem.Id);
+            }
+
             // triggers any event
             _rulesManager.TriggerEvent("CustomForm", "Submitted",
                     () => new Dictionary<string, object> { { "Content", contentItem } });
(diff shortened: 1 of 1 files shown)


src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
```diff
@@ -198,6 +198,13 @@ public class ItemController : Controller, IUpdateModel {
 
             contentItem.As<ICommonPart>().Container = customForm.ContentItem;
 
+            // save the submitted form
+            if (customForm.SaveContentItem) {
+                conditionallyPublish(contentItem);
+                // Refresh content item
+                contentItem = _contentManager.Get(contentItem.Id);
```
sebastienros: Does this have any effect? It should just return the same instance from the cache. Or is the publishing creating a new version.

src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
```diff
@@ -198,6 +198,13 @@ public class ItemController : Controller, IUpdateModel {
 
             contentItem.As<ICommonPart>().Container = customForm.ContentItem;
 
+            // save the submitted form
```
sebastienros, line 201: Is it worth creating a new "submitting" event instead? Can anything be done between them of value (The workflow is triggered so something could be done).

src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
AndreaPiovanelli: The publish procedure is triggering the drivers of all the fields and creating a new version so I noticed I get the "compiled" content only by reading it again from the content manager, because there there may be some LazyField inside the content item.

Before editing this controller, I experimented inside the razor executed in the workflow and the only way to make it work was to force the content to be published and reading it again from the content manager.

src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
AndreaPiovanelli: As explained here https://github.com/OrchardCMS/Orchard/pull/8795#discussion_r1596347011 , I tested how to make the workflow work and the only way to have the actual values of the fields was to publish the content item and reading it again.
For this reason, I moved the logic I tested from the RazorExecute activity to the controller to make it work as expected without publishing the content explicitly inside the external code.

For reference, this is how our worflow looks like:
image

The "SelezionaDestinatari.cshtml" razor reads a TaxonomyField inside the content item created by the form submit to make some operations. If content isn't published (and re-read), the TaxonomyField is empty.

src/Orchard.Web/Modules/Orchard.CustomForms/Controllers/ItemController.cs
BenedekFarkas: LGTM based on your answers! One question for line 205: If the item is conditionally published (line 203), then `_contentManager.Get(contentItem.Id)` will return null, if the item is not actually published, so VersionOptions needs to be taken into account.

AndreaPiovanelli: It makes sense, thanks; suggestion tested on my environment and applied on current pr branch.

sebastienros (approved):

BenedekFarkas (approved):

PR #8801, opened 2024-08-21
Upgrades to enumeration fields and queries related to them

### LINKED ISSUE #8800: Problems using EnumerationFields in queries
Since the format values are now saved, the Equals clause in queries doesn't work anymore.

This is a side effect of #8789 , which saves the EnumerationField value in a different way then before, as shown by the following screenshot from Orchard_Projections_StringFieldIndexRecord table:
image
As you can see, the new version of the Value column contains now the separator character ";" at the start and at the end of the actual value, even when only one value is selected. 
There are two topics to discuss about:
1. Values inside the table are now ambiguous and should be coherent, perhaps with a migration step replacing old values with new ones. As a side note: saving a content item is enough to replace the "old" value format with the new one and using the "Value" property of the field works as expected.
2. Queries filtering on the Value of an EnumerationField do not work anymore, because the value is now saved with ";" characters even if a single option is selected. For this reason, the Equals clause cannot be used and currently needs to be replaced by the Contains clause. Creating a migration step to change queries isn't a viable solution: any suggestions on how to make EnumerationFields compatible with the old query usage?

I add to the discussion @MatteoPiovanelli-Laser and @BenedekFarkas because they worked on the referenced pull request and @sebastienros who approved it.

AndreaPiovanelli:
In reference to issue #8800, which solves a compatibility issue added by #8789 , I added an Upgrade procedure (inside the Upgrade module) that saves / publishes again every content item which contains an EnumerationField in the definition of its parts. This is needed to ensure that both the content item's infoset and the string field index records used by the projections for the enumeration fields' values are in the new format (with ";" separator characters at the beginning and at the end of the value).
I didn't use a migration step because of the lenght of the procedure (potentially saving thousands of content items for each tenant) and to be able to schedule the long update script to be asynchronous.

As discussed and explained in the issue thread, I also implemented a new IFilterProvider, specific for EnumerationFields, that processes the values for the query in order to ensure that old queries still work. This solves an issue which is specific to single choice EnumerationFields, the value of which is now in the format of ";value;" instead of "value" (without the semicolons). Appropriate semicolons are added for Equals/NotEquals, StartsWith/NotStartsWith and EndsWith/NotEndsWith string operators, the ones who were affected by #8789 .
This new filter provider has the same category and name of the original ContentFieldsFilterProvider (modificed accordingly to exclude EnumerationFields from its execution) and, for this reason, old queries work properly without the need of a migration, using the new, specific provider.

The Upgrade step has to be executed manually for each tenant in order to make the new filter provider to work properly. If upgrade isn't executed, EnumerationFields with the old value format will not be "found" by the queries if they are filtered with Equals/NotEquals, StartsWith/NotStartsWith or EndsWith/NotEndsWith string operators.


src/Orchard.Web/Modules/Orchard.Projections/Providers/Filters/ContentFieldsFilter.cs
```diff
@@ -36,9 +36,12 @@ public void Describe(DescribeFilterContext describe) {
                     continue;
                 }
 
+                // Exclude EnumerationFields, that use a specific IFilterProvider
+                var fields = part.Fields.Where(fd => !fd.FieldDefinition.Name.Equals("EnumerationField"));
```
sebastienros, line 40: This can't be an acceptable solution. The upgrade should make things right such that not such hack is necessary.

BenedekFarkas (changes requested):
I commented on the issue, let's discuss there: https://github.com/OrchardCMS/Orchard/issues/8800#issuecomment-2486868481

BenedekFarkas:
Closing due to #8824.

Issue #8803, opened 2024-09-18
Downloading video meta from Vimeo fails


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/0d9fccb624e4c46ebe4e9f2d876b23896970118d/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs#L85

When importing a video inside the media library, if the video is from Vimeo, the download fails with a http 429 status (too many requests). This seems an issue related to the fact Vimeo forbids the scraping of its pages (which Youtube still consents), making it impossible to import its videos inside the media library.

BenedekFarkas:
#8827 moves the import logic's implementation to a service that can be overridden.

PR #8804, opened 2024-09-18
Fix to import Vimeo videos

AndreaPiovanelli:
In reference to #8803 , added a fix to avoid the scraping of the "vimeo.com/xxxx" page (which replies with a 429 http error) by forcing the direct api call if Vimeo domain is recognized.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
@@ -103,11 +103,12 @@ namespace Orchard.MediaLibrary.Controllers {
 
                 var source = "";
 
+                var uri = GetRedirectUrl(url);
+
                 // Vimeo doesn't consent anymore the scraping of web pages, so the direct api call has to be enforced.
                 // In this case, the downloaded string is already the expected xml, in the format that needs to be parsed.
                 // Legacy process is done for non-Vimeo content.
                 // First of all, url domain is checked.
-                var uri = GetRedirectUrl(url);
                 var vimeo = uri.Host.Equals("vimeo.com", StringComparison.OrdinalIgnoreCase);
 
                 // Youtube changed the markup of the page of its videos, so the direct api call has to be enforced
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
@@ -86,10 +86,8 @@ namespace Orchard.MediaLibrary.Controllers {
 
                 // Vimeo doesn't consent anymore the scraping of web pages, so the direct api call has to be enforced.
                 // In this case, the downloaded string is already the expected xml, in the format that needs to be parsed.
-                // Legacy process is done either way.
-                // At the end of it, oembedSignature value will still be -1.
-                // So the downloaded string is parsed.
-                // Check if url is inside Vimeo domain.
+                // Legacy process is done for non-Vimeo content.
+                // First of all, url domain is checked.
                 var uri = new Uri(url);
                 var vimeo = uri.Host.Equals("vimeo.com", StringComparison.OrdinalIgnoreCase) ||
                     uri.Host.Equals("www.vimeo.com", StringComparison.OrdinalIgnoreCase);
--- a/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
+++ b/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
@@ -82,26 +82,44 @@ namespace Orchard.MediaLibrary.Controllers {
             try {
                 // <link rel="alternate" href="http://vimeo.com/api/oembed.xml?url=http%3A%2F%2Fvimeo.com%2F23608259" type="text/xml+oembed">
 
-                var source = webClient.DownloadString(url);
+                var source = "";
 
-                // seek type="text/xml+oembed" or application/xml+oembed
-                var oembedSignature = source.IndexOf("type=\"text/xml+oembed\"", StringComparison.OrdinalIgnoreCase);
-                if (oembedSignature == -1) {
-                    oembedSignature = source.IndexOf("type=\"application/xml+oembed\"", StringComparison.OrdinalIgnoreCase);
-                }
-                if (oembedSignature != -1) {
-                    var tagStart = source.Substring(0, oembedSignature).LastIndexOf('<');
-                    var tagEnd = source.IndexOf('>', oembedSignature);
-                    var tag = source.Substring(tagStart, tagEnd - tagStart);
-                    var matches = new Regex("href=\"([^\"]+)\"").Matches(tag);
-                    if (matches.Count > 0) {
-                        var href = matches[0].Groups[1].Value;
-                        try {
-                            var content = webClient.DownloadString(Server.HtmlDecode(href));
-                            viewModel.Content = XDocument.Parse(content);
-                        }
-                        catch {
-                            // bubble exception
+                // Vimeo doesn't consent anymore the scraping of web pages, so the direct api call has to be enforced.
+                // In this case, the downloaded string is already the expected xml, in the format that needs to be parsed.
+                // Legacy process is done either way.
+                // At the end of it, oembedSignature value will still be -1.
+                // So the downloaded string is parsed.
+                // Check if url is inside Vimeo domain.
+                var uri = new Uri(url);
+                var vimeo = uri.Host.Equals("vimeo.com", StringComparison.OrdinalIgnoreCase) ||
+                    uri.Host.Equals("www.vimeo.com", StringComparison.OrdinalIgnoreCase);
+                if (vimeo) {
+                    // Add api url to original url provided as a parameter
+                    url = "http://vimeo.com/api/oembed.xml?url=" + url;
+                    source = webClient.DownloadString(url);
+
+                    viewModel.Content = XDocument.Parse(source);
+                } else {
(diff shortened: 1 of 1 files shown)


sebastienros:
I think this is really wrong. If Vimeo doesn't support OEmbed then it's better to have another feature dedicated to vimeo, or just drop vimeo support altogether.

AndreaPiovanelli:
> I think this is really wrong. If Vimeo doesn't support OEmbed then it's better to have another feature dedicated to vimeo, or just drop vimeo support altogether.

Vimeo still supports embed (and the xml format is still the same) but it doesn't include the link to the video metadata xml in its markup anymore, so it cannot be automatically imported by directly reading the video page content.

While investigating further, we also noticed Youtube stopped working too. This is caused by the fact it blocks video output when trying to watch a video from a server (a bot protection). This is what happened when I tried to debug the Youtube issue directly from the server: I was trying to find a way to get the meta link in the page, but that's what the page shows:
image

For this reason, we worked on a patch on the code, similar to what has already been done to get Vimeo's oembed xml data: to properly get the provider, we check for the url we're redirected to and, with that information, call the right api.

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -82,26 +82,54 @@ public ActionResult IndexPOST(string folderPath, string url, string type, string
             try {
                 // <link rel="alternate" href="http://vimeo.com/api/oembed.xml?url=http%3A%2F%2Fvimeo.com%2F23608259" type="text/xml+oembed">
```
BenedekFarkas, line 84:
```suggestion
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -82,26 +82,54 @@ public ActionResult IndexPOST(string folderPath, string url, string type, string
...
+                        var tagEnd = source.IndexOf('>', oembedSignature);
+                        var tag = source.Substring(tagStart, tagEnd - tagStart);
+                        var matches = new Regex("href=\"([^\"]+)\"").Matches(tag);
+                        if (matches.Count > 0) {
+                            var href = matches[0].Groups[1].Value;
+                            try {
+                                var content = webClient.DownloadString(Server.HtmlDecode(href));
+                                viewModel.Content = XDocument.Parse(content);
+                            } catch {
+                                // bubble exception
```
BenedekFarkas, line 131:
```suggestion
                                throw;
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -82,26 +82,54 @@ public ActionResult IndexPOST(string folderPath, string url, string type, string
...
+                // Api url is built based on the requested video
+                var youtube = uri.Host.Equals("www.youtube.com", StringComparison.OrdinalIgnoreCase);
+
+                if (vimeo) {
+                    // Add api url to original url provided as a parameter
+                    url = "https://" + uri.Host + "/api/oembed.xml?url=" + url;
+                    source = webClient.DownloadString(url);
+
+                    viewModel.Content = XDocument.Parse(source);
+                } else if (youtube) {
```
BenedekFarkas, line 106:
```suggestion
                }
                else if (youtube) {
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -82,26 +82,54 @@ public ActionResult IndexPOST(string folderPath, string url, string type, string
...
+                    // Add api url to original url provided as a parameter
+                    url = "https://" + uri.Host + "/oembed?format=xml&url=" + url;
+                    source = webClient.DownloadString(url);
+
+                    viewModel.Content = XDocument.Parse(source);
+                } else {
```
BenedekFarkas, line 112:
```suggestion
                }
                else {
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -82,26 +82,54 @@ public ActionResult IndexPOST(string folderPath, string url, string type, string
...
+                        var tagStart = source.Substring(0, oembedSignature).LastIndexOf('<');
+                        var tagEnd = source.IndexOf('>', oembedSignature);
+                        var tag = source.Substring(tagStart, tagEnd - tagStart);
+                        var matches = new Regex("href=\"([^\"]+)\"").Matches(tag);
+                        if (matches.Count > 0) {
+                            var href = matches[0].Groups[1].Value;
+                            try {
+                                var content = webClient.DownloadString(Server.HtmlDecode(href));
+                                viewModel.Content = XDocument.Parse(content);
+                            } catch {
```
BenedekFarkas, line 130:
```suggestion
                            }
                            catch {
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -137,8 +165,7 @@ public ActionResult IndexPOST(string folderPath, string url, string type, string
                     root.El("description", description);
                 }
                 Response.AddHeader("X-XSS-Protection", "0"); // Prevents Chrome from freaking out over embedded preview
-            }
-            catch {
+            } catch {
```
BenedekFarkas, line 168:
```suggestion
            }
            catch {
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -167,8 +194,7 @@ public ActionResult Import(string folderPath, string url, string document) {
 
             if (oembed.Element("title") != null) {
                 part.Title = oembed.Element("title").Value;
-            }
-            else {
+            } else {
```
BenedekFarkas, line 197:
```suggestion
            }
            else {
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -213,8 +239,7 @@ public ActionResult Replace(int replaceId, string url, string document) {
 
             if (oembed.Element("title") != null) {
                 replaceMedia.Title = oembed.Element("title").Value;
-            }
-            else {
+            } else {
```
BenedekFarkas, line 242:
```suggestion
            }
            else {
```

src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs
```diff
@@ -238,5 +263,22 @@ public ActionResult Replace(int replaceId, string url, string document) {
...
+            HttpWebRequest myHttpWebRequest = (HttpWebRequest)WebRequest.Create(myUri);
+            // Send the request and wait for response.
+            HttpWebResponse myHttpWebResponse = (HttpWebResponse)myHttpWebRequest.GetResponse();
+
+            if (!myUri.Equals(myHttpWebResponse.ResponseUri)) {
+                myUri = myHttpWebResponse.ResponseUri;
+            }
+
+            // Release resources of response object.
+            myHttpWebResponse.Close();
```
BenedekFarkas, line 279:
```suggestion
            var myHttpWebRequest = WebRequest.Create(myUri) as HttpWebRequest;
            // Send the request and wait for response.
            using (var myHttpWebResponse = myHttpWebRequest.GetResponse() as HttpWebResponse) {
                if (!myUri.Equals(myHttpWebResponse.ResponseUri)) {
                    myUri = myHttpWebResponse.ResponseUri;
                }
            }
```

BenedekFarkas (changes requested):

BenedekFarkas:
> > I think this is really wrong. If Vimeo doesn't support OEmbed then it's better to have another feature dedicated to vimeo, or just drop vimeo support altogether.
> 
> Vimeo still supports embed (and the xml format is still the same) but it doesn't include the link to the video metadata xml in its markup anymore, so it cannot be automatically imported by directly reading the video page content.
> 
> While investigating further, we also noticed Youtube stopped working too. This is caused by the fact it blocks video output when trying to watch a video from a server (a bot protection). This is what happened when I tried to debug the Youtube issue directly from the server: I was trying to find a way to get the meta link in the page, but that's what the page shows: image
> 
> For this reason, we worked on a patch on the code, similar to what has already been done to get Vimeo's oembed xml data: to properly get the provider, we check for the url we're redirected to and, with that information, call the right api.

@sebastienros what do you think? BTW I tested the YouTube import (but couldn't test Vimeo), works fine.

sebastienros:
Let's keep oembed as is, it's oembed. Add some other option called youtube, another one for vimeo and they are dedicated. Or a single one called "video" which would support whatever we want. Copy as much as you want from the current oembed.

AndreaPiovanelli:
@sebastienros what if we start from here: https://github.com/OrchardCMS/Orchard/blob/d26e8e47f132b1fb210e244735ec755d9d948df7/src/Orchard.Web/Modules/Orchard.MediaLibrary/Controllers/OEmbedController.cs#L85
And we implement a new service that does the job of getting the href variable (used at line 98 of the same class).
This enable us to implement dedicated services for Vimeo and Youtube cases while keeping the Media Library as a reference (and possibly the service we implement inside Media Library as a base class to fallback to).

sebastienros:
Sounds good

AndreaPiovanelli:
> Sounds good

To have a cleaner situation, I created a new pr: #8827

PR #8805, opened 2024-10-14
Internal page links validation for LinkField

AndreaPiovanelli:
We need to validate internal page references to a tag id in LinkFields, so I patched the LinkFieldDriver to validate urls like "mysite/mypage#mydiv" or just "#mydiv".
I also changed the hint in the editor shape to add "#someId" to the examples for valid field values.

The field value is divided in prefix and anchor to be sure to properly validate the format of the actual page the field is linking to:
A value like "http://mysite.com/page#div" is divided in two separate validations:
- "http://mysite.com/page" is validated with the original Uri.IsWellFormedUriString(string) function.
- "div" is validated by checking for the presence of white spaces inside the string (white space is the only forbidden character for html 5 tag ids).

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Fields/Drivers/LinkFieldDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Fields/Drivers/LinkFieldDriver.cs
@@ -57,11 +57,30 @@ namespace Orchard.Fields.Drivers {
                 if (settings.Required && String.IsNullOrWhiteSpace(field.Value)) {
                     updater.AddModelError(GetPrefix(field, part), T("Url is required for {0}", T(field.DisplayName)));
                 }
-                else if (!String.IsNullOrWhiteSpace(field.Value) && !Uri.IsWellFormedUriString(field.Value, UriKind.RelativeOrAbsolute)) {
-                    updater.AddModelError(GetPrefix(field, part), T("{0} is an invalid url.", field.Value));
-                }
                 else if (settings.LinkTextMode == LinkTextMode.Required && String.IsNullOrWhiteSpace(field.Text)) {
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Fields/Drivers/LinkFieldDriver.cs
```diff
@@ -57,11 +57,30 @@ protected override DriverResult Editor(ContentPart part, LinkField field, IUpdat
...
-                    updater.AddModelError(GetPrefix(field, part), T("{0} is an invalid url.", field.Value));
-                }
                 else if (settings.LinkTextMode == LinkTextMode.Required && String.IsNullOrWhiteSpace(field.Text)) {
                     updater.AddModelError(GetPrefix(field, part), T("Text is required for {0}.", T(field.DisplayName)));
+                } else if (!String.IsNullOrWhiteSpace(field.Value)) {
+                    // Check if it's a valid uri, considering that there may be the link to an anchor only
+                    // e.g.: field.Value = "#divId"
+                    // Take everything before the first "#" character and check if it's a valid uri.
+                    // If there is no character before the first "#", consider the value as a valid one (because it is a reference to a div inside the same page)
+                    var index = field.Value.IndexOf('#');
```
sebastienros: A technique I like to use for these cases is `field.Value.Split('#', 2)`. It will only give you an array or 1 or 2 values even if there are more `#`. This way you have access to both parts directly.

AndreaPiovanelli: As suggested, refactored the code to use string split function.

sebastienros (approved):

PR #8813, opened 2024-11-27
#8232: Adding warnings when a Projection Query has an empty Filter Group

### LINKED ISSUE #8232: Projections: Empty Query returns every content item
Which can pretty much kill a site or even a whole server, even if you don't have many content items, just some Layout (as in Orchard.Layouts) that renders the Query.

BenedekFarkas:
Fixes #8232

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
+++ b/src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
@@ -140,11 +140,6 @@ namespace Orchard.Projections.Services {
 
             var contentItems = new List<ContentItem>();
 
-            // If there are no filters, return the empty list of content items instead of returning all content items.
-            if (queryRecord.FilterGroups.SelectMany(group => group.Filters).Count() == 0) {
-                return contentItems;
-            }
-
             // prepares tokens 
             Dictionary<string, object> tokens = new Dictionary<string, object>();
             if (part != null) {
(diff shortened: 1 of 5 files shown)


src/Orchard.Web/Modules/Orchard.Projections/Services/ProjectionManager.cs
```diff
@@ -204,11 +209,11 @@ public IEnumerable<IHqlQuery> GetContentQueries(QueryPartRecord queryRecord, IEn
 
             var version = queryRecord.VersionScope.ToVersionOptions();
 
-            // pre-executing all groups
-            foreach (var group in queryRecord.FilterGroups) {
+            // Iterate over each filter group, but ignore empty ones, because they'd cause all content items to be returned.
+            foreach (var group in queryRecord.FilterGroups.Where(group => group.Filters.Count > 0)) {
```
BenedekFarkas: This is the only functional change, the rest is just code styling and comment rephrasing.

sebastienros: I am suggesting adding a limit filter instead (100, 500). And add a notification.

BenedekFarkas: I didn't find a suitable way to apply such a limit due to how the query is constructed (there is no explicit limit filter) without hurting the abstractions. It is possible to intervene towards the end of the query evaluation when Slice is called (and apply a limit there), but that would mess up paging.
I added a notification instead, when a Query or a ProjectionPart is saved, IMO that's good enough.

PR #8824, opened 2024-12-09
#8800: Reverting breaking changes to Enumeration Fields caused by #8789

BenedekFarkas:
This is an alternative to #8801. It reverts the changes merged in #8789, except those that fix the editor template of the EnumerationField. See Option 1 in https://github.com/OrchardCMS/Orchard/issues/8800#issuecomment-2486868481 for an explanation.


sebastienros (approved):

MatteoPiovanelli:
lgtm

Issue #8056, opened 2018-05-15
767 length limitation in mysql


HengzheLi:
Create a new tenant using mysql as database. Occur below error while setting it up
```
2018-05-15 23:47:49,304 [42] Orchard.Data.Migration.DataMigrationManager - fanyong - Error while running migration version 0 for Orchard.Autoroute. 
System.Reflection.TargetInvocationException:  ---> 
MySql.Data.MySqlClient.MySqlException: Specified key was too long; max key length is 767 bytes
   在 MySql.Data.MySqlClient.MySqlStream.ReadPacket()
   在 MySql.Data.MySqlClient.NativeDriver.GetResult(Int32& affectedRow, Int64& insertedId)
   在 MySql.Data.MySqlClient.Driver.NextResult(Int32 statementId, Boolean force)
   在 MySql.Data.MySqlClient.MySqlDataReader.NextResult()
   在 MySql.Data.MySqlClient.MySqlCommand.ExecuteReader(CommandBehavior behavior)
   在 MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQuery()
   在 Orchard.Data.Migration.Interpreters.DefaultDataMigrationInterpreter.RunPendingStatements() 
```

Got the sql from vs debugger, it is: 
`create index `fy_IDX_AutoroutePartRecord_DisplayAlias` on `fy_Orchard_Autoroute_AutoroutePartRecord` (DisplayAlias(767))`

From SO answer https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes/31474509#31474509. I guess may be
`create index `fy_IDX_AutoroutePartRecord_DisplayAlias` on `fy_Orchard_Autoroute_AutoroutePartRecord` (DisplayAlias(255))`
can work

sebastienros:
Dev branch?

HengzheLi:
Yes

BenedekFarkas:
Does this mean that the length of the DisplayAlias column should be restricted to 255 characters?

sebastienros:
Or not index this field on mysql

BenedekFarkas:
OK, so a generic solution might be that SchemeBuilder.CreateIndex ignores columns longer than 255 characters for MySQL databases.

BenedekFarkas:
@HengzheLi which MySQL version are you using?

BenedekFarkas:
[...]

PR #8826, opened 2025-01-21
#8056: Adapting `MySqlCommandInterpreter` to 767 character length limitation

mazuryv:
https://github.com/OrchardCMS/Orchard/issues/8056
767 length limitation in mysql


sebastienros:
LGTM. But needs @BenedekFarkas and @MatteoPiovanelli

src/Orchard/Data/Migration/Interpreters/MySqlCommandInterpreter.cs
```diff
@@ -91,20 +91,47 @@ public string[] CreateStatements(AddIndexCommand command) {
             using (var sqlCommand = session.Connection.CreateCommand()) {
                 var columnNames = String.Join(", ", command.ColumnNames.Select(c => string.Format("'{0}'", c)));
                 var tableName = PrefixTableName(command.TableName);
+                var columnList = command.ColumnNames.ToList();
+                var indexMaximumLenght = 767;
```
agriffard: indexMaximumLenght => indexMaximumLength

mazuryv:
Without these changes recipe was fallen for me in Orchard.Recipes.Migrations
  table.CreateIndex("IDX_RecStResRec_ExecId_StepN", "ExecutionId", "StepName");
MySqlServer 8.0.40/8.0.41

sebastienros:
LGTM but someone should validate it too. Or add a functional test.

BenedekFarkas:
> LGTM but someone should validate it too. Or add a functional test.

I'll test it locally.

BenedekFarkas:
I'm trying to test this, but having problems completing the setup with a MySQL version (5.7.44) that [is supposed to be compatible](https://dev.mysql.com/blog-archive/mysql-connectornet-6-7-9-6-8-7-and-6-9-8-have-been-released/) with the MySQL connector we're using (6.7.9).

**Docker:**

Run: `docker run --name mysql -p 3307:3306 -e MYSQL_ROOT_PASSWORD=password -e MYSQL_DATABASE=Orchard -d mysql:5.7.44`
Connection string: `Server=localhost;Port=3307;Database=Orchard;Uid=root;Pwd=password;SslMode=None`

**Local server:**

Connection string: `Server=localhost;Port=3306;Database=orchard;Uid=root;Pwd=password;SslMode=None`

**Error (same for both):**

> **MySql.Data.MySqlClient.MySqlException**: 'You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(NOLOCK) inner join Title_TitlePartRecord titlepartr2_ WITH(NOLOCK) on this_.Id=' at line 1'

Docker images with more recent versions have different errors.

@mazuryv what version of MySQL are you using? Did you experience any similar errors?

mazuryv:
Hello.
I have updated MySQL connector MySql.Data nuget package up to 9.2.0 to be compatible with current MySqlServer 8.0.40/8.0.41+
Thank you.

BenedekFarkas:
OK, that's a pretty important detail. :)

Are there any other changes to your code/application that isn't in the PR branch? I can't get the setup to complete unless `NoLockInterceptor` is "neutralized" by not modifying the SQL statement.

BenedekFarkas:
We need additional verification from someone who already uses MySQL to make sure we don't break existing applications + I still can't get the setup to work in this state, so any additional testing and extra set of eyes are welcome.

BenedekFarkas:
I managed to test this and verify that the changes work, however, only if:
- `NoLockInterceptor` is changed not to modify commands if the database provider is not SQL Server or SQL CE. This is already included in #8861.
- `DistributedLockSchemaBuilder` is modified so that the `Name` column has only a length of 255. Anything above that throws an error, see #8859.

I'll look into upgrading `MySql.Data` in #8859.

Issue #8828, opened 2025-02-03
Orchard.Redis does not provide an API to clear keys by wildcard


MpDzik:
`ICacheStorageProvider.Remove` deletes a single key and `ICacheStorageProvider.Clear` deletes all keys. There is no way to remove multiple keys using the Redis wildcard syntax.

BenedekFarkas:
Fixed in #8829.

PR #8829, opened 2025-02-03
#8828: Expose Redis API to clear multiple keys using wildcard syntax

MpDzik:
This change adds a new `IRedisCacheStorageProvider.Clear` API that can be used to remove multiple keys from the cache using wildcard syntax.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Redis/Caching/RedisCacheStorageProvider.cs
+++ b/src/Orchard.Web/Modules/Orchard.Redis/Caching/RedisCacheStorageProvider.cs
@@ -10,9 +10,13 @@ using System;
 
 namespace Orchard.Redis.Caching {
 
+    public interface IRedisCacheStorageProvider : ICacheStorageProvider {
+        void Clear(string key);
+    }
+
     [OrchardFeature("Orchard.Redis.Caching")]
     [OrchardSuppressDependency("Orchard.Caching.Services.DefaultCacheStorageProvider")]
-    public class RedisCacheStorageProvider : Component, ICacheStorageProvider {
+    public class RedisCacheStorageProvider : Component, IRedisCacheStorageProvider {
         public const string ConnectionStringKey = "Orchard.Redis.Cache";
 
         private readonly ShellSettings _shellSettings;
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Redis/Caching/RedisCacheStorageProvider.cs
```diff
@@ -10,9 +10,13 @@
...
 namespace Orchard.Redis.Caching {
 
+    public interface IRedisCacheStorageProvider : ICacheStorageProvider {
+        void Clear(string key);
+    }
+
     [OrchardFeature("Orchard.Redis.Caching")]
     [OrchardSuppressDependency("Orchard.Caching.Services.DefaultCacheStorageProvider")]
-    public class RedisCacheStorageProvider : Component, ICacheStorageProvider {
+    public class RedisCacheStorageProvider : Component, IRedisCacheStorageProvider {
```
sebastienros: Will this resolve the service on both interfaces?

src/Orchard.Web/Modules/Orchard.Redis/Caching/RedisCacheStorageProvider.cs
```diff
@@ -10,9 +10,13 @@
 
 namespace Orchard.Redis.Caching {
 
+    public interface IRedisCacheStorageProvider : ICacheStorageProvider {
```
sebastienros: Could be a more general interface instead, like `ICacheStorageProviderWithKeyPrefix`

sebastienros (approved):

src/Orchard.Web/Modules/Orchard.Redis/Caching/RedisCacheStorageProvider.cs
MpDzik: Yes, it does. I use this pattern very often in Orchard and never had any issues with it.

src/Orchard.Web/Modules/Orchard.Redis/Caching/RedisCacheStorageProvider.cs
BenedekFarkas: What do you think @MpDzik?

MpDzik: Fixed in 138b068e24a5980e740e078e823a6ce9cd1e7f9f.

BenedekFarkas (approved):

Issue #8830, opened 2025-04-02
Saving a menu item doesn't update the fields in the parts attached to it


AndreaPiovanelli:
https://github.com/OrchardCMS/Orchard/blob/3561be6e37b7ec1e610b375de9775965a929d047/src/Orchard.Web/Core/Navigation/Controllers/AdminController.cs#L246-L248

When saving an existing menu item, the content doesn't get published (the handlers aren't called for the "published" event, even if menu item is not draftable), leaving the string field index record table with the LatestValue column updated only (and not the Value column, which doesn't get changed).
This happens because the menu item doesn't get to be "Published" when checking if it is (`contentItem.IsPublished()`); basically, content item version record "Published" property is false in that moment, because it's just been "saved" (so a record with Latest = true and Published = false is the current version record).
A couple screenshots to explain the effect:
1) Editing a Menu Item with some fields added to the MenuPart (e.g. the ElementAttribute TextField):

![Image](https://github.com/user-attachments/assets/3ab04481-24f0-4a9b-94fd-6cb2f4f648bc)
2) Row from Orchard_Projections_StringFieldIndexRecord involved:

![Image](https://github.com/user-attachments/assets/b24f4b0f-8fa7-4b96-9cc4-ddbd762fd42b)

Since I noticed this behaviour in Menu Items only, I checked how the AdminController of Core/Contents works and, in fact, when saving it doesn't explicitly check if the content is published:
https://github.com/OrchardCMS/Orchard/blob/3561be6e37b7ec1e610b375de9775965a929d047/src/Orchard.Web/Core/Contents/Controllers/AdminController.cs#L336

Removing the `contentItem.IsPublished()` check from Navigation/AdminController seems to solve the issue: does anybody see any shortcomings or unexpected effects to this solution?

BenedekFarkas:
Changing the Contents AdminController that way would probably have significant unwanted side-effects, but I see that in the PR you've changed the Navigation AdminController instead. I wouldn't do that either, because if a menu item is unpublished, it should stay that way after editing.

The problem here is two-fold:
1. Saving an item creates a draft version with `ContentManager.GetDraftRequired` in the private `EditPOST` function. I'm not sure why `GetDraftRequired` is used there, but it looks like that it's used in some of the built-in modules that have feature-specific content parts/types. I haven't really used it in custom code before and I'm not sure what the benefit is - it basically forces versioning on a type that isn't draftable.
[...]

PR #8831, opened 2025-04-11
Forced the handler to "publish" the fields of a menu to ensure "Value" is updated in Orchard Projections FieldIndexRecords instead of "LatestValue" only.

AndreaPiovanelli:
Following issue #8830 , this is the patch to avoid checking for the actual menu item publish state and save "Value" column inside StringFieldIndexRecords tables.


AndreaPiovanelli:
I marked the pull request as draft because I am working on a proper solution to save the menu item fields on menu item creation.

BenedekFarkas:
1.10.4 will be released soon from 1.10.x, so if you're still working on this, please target the dev branch.

BenedekFarkas:
If you continue working on this, please open a new PR targeting `dev`.

PR #8837, opened 2025-06-20
#8836: Sanitize scripts in TinyMCE

### LINKED ISSUE #8836: Set sanitization of script tags in TinyMCE
Content creators should not have the ability to add script tags to pages. TinyMCE should be set to sanitize script tags.

mwentz:
Fixes #8836

This change to the TinyMCE module will set TinyMCE to remove scripts that are added in any TinyMCE blocks.

This is a breaking change since users were previously able to add <script> tags to content using TinyMCE.

This fixes https://nvd.nist.gov/vuln/detail/CVE-2022-37720


mwentz:
@dotnet-policy-service agree

sebastienros:
Sorry, I commented on the issue but this should be here:

> Do you think you could provide a PR that would make the setting configurable from the admin?

mwentz:
Great idea. Might be a while until I can carve out time but I'll give it a shot when I can.

I3undy:
Here is a PR where the setting added as a configurable site setting:
https://github.com/OrchardCMS/Orchard/pull/8872

PR #8840, opened 2025-09-04
#8839 Orchard Layouts Bootstrap 4 columns fix

### LINKED ISSUE #8839: Column widths can't be adjusted in Layouts interface
Using the latest Dev branch there is an issue with display of columns in the admin related to Orchard Layouts. The columns all show as full-width in the interface instead of being displayed in their columns with the correct width on a 12-column grid. I'm seeing two issues with this:

1) It is confusing for editors. There is no indication of column widths to the editors unless they hover or click on the right place to see the "Column (X)" associated with each column. If this was a design decision, there should be always-visible indicators indicating columns and their widths.

2) I can't find a way to adjust the width of a column. If I have three 4-column items (4 - 4 - 4) and I want to adjust them to have small outer columns with a large middle column (2 - 8 - 2) there doesn't seem to be a way to adjust widths.

Reproduction steps:

- pull down latest dev branch
- run and create a site (default recipe)
- in the admin, open the "Welcome to Orchard" content item
- within layouts, you'll see the second row that includes columns are all shown full width. You can find ways to do things such as get a column down to 1 width, but the size adjustment generally does not work.

mwentz:
#8839 update all col-xs-* in Orchard.Layouts module to just col-* to fix columns. col-xs-* was removed in Bootstrap 4. display:flex was also added to Row.less to counteract the display:block on all the divs between .row and the .col-*

Gulp is broken per #8522 so all compiled files were manually updated. These files are included in the source, so this should work just fine.


mwentz:
The display:flex rule is causing issues and will need more specificity. Update coming.

sebastienros (approved):

BenedekFarkas:
The client-side asset build pipeline (with Gulp) is supposed to be working, but you need to use Node v7 (I know, ancient) for it. Unfortunately, I just noticed that the workflow job that checks the integrity of the client-side assets is broken, so I'll fix that first and then review this PR.

mwentz:
Per my previous comment, there were some issues in the editor. I added a new commit that fixes the display of rows / columns and child elements.

BenedekFarkas (approved):

PR #8844, opened 2025-09-24
#8830: Fixing that fields attached to a Menu Item are not updated

### LINKED ISSUE #8830: Saving a menu item doesn't update the fields in the parts attached to it

BenedekFarkas:
Fixes #8830

### DIFF
--- a/src/Orchard.Web/Core/Navigation/Controllers/AdminController.cs
+++ b/src/Orchard.Web/Core/Navigation/Controllers/AdminController.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Web.Mvc;
-using System.Web.Routing;
 using Orchard.ContentManagement;
 using Orchard.ContentManagement.Aspects;
 using Orchard.ContentManagement.Handlers;
@@ -172,7 +171,8 @@ namespace Orchard.Core.Navigation.Controllers {
                 return new HttpUnauthorizedResult();
 
             // create a new temporary menu item
-            var menuPart = _contentManager.New<MenuPart>(id);
+            var contentItem = _contentManager.New(id);
+            var menuPart = contentItem.As<MenuPart>();
 
             if (menuPart == null)
                 return HttpNotFound();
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Core/Navigation/Controllers/AdminController.cs
```diff
@@ -308,16 +322,15 @@ private MenuItemEntry CreateMenuItemEntries(MenuPart menuPart) {
         }
 
         private ActionResult EditPOST(int id, string returnUrl, Action<ContentItem> conditionallyPublish) {
-            var menuPart = _contentManager.GetDraftRequired<MenuPart>(id);
+            var contentItem = _contentManager.GetLatest(id);
```
sebastienros, line 330: Test: could you before and now after edit a menu item but it's not published yet.

BenedekFarkas: From this perspective, it works the same way as before:
1. Create a menu item -> it starts out as published.
2. Unpublish it.
3. Edit it and see the changes persisted on the admin, but the item remains unpublished.
4. Publish it -> see the changes made in step 3 on the frontend too.

PR #8856, opened 2025-10-12
Updating NuGet packages to latest minor/patch versions

BenedekFarkas:
## Updated NuGet packages

- `Castle.Core`: Upgraded from 3.3.1 to 3.3.3. (0d3001a5a1d22fb5c0a164f08a7f73d54f78b566)
- `DocumentFormat.OpenXml` and `DocumentFormat.OpenXml.Framework`: From 3.0.2 to 3.3.0 (latest). (1f20a4e8b200116bf3668b9b8058a590da31a5d1)
  - Only used by Dynamic Forms for the "Export Submissions" functionality.
- `FluentNHibernate`: From 3.1.0 to 3.4.1. (b469231a79747c684ab783066ff24e6f4730108d)
- `log4net`: From 2.0.12 to 2.0.17 (latest 2.x). (e170d31160690aeec150fae4d4c28ed354cfec9b)
  - Subsequently upgraded to 3.2.0 (latest). The major version bump is just due to dropping old build targets. (6e33894df7dc5e544e221678ba4ba1ea011c43ec)
- `MailKit`: From 4.9.0 and 4.14.0 (latest). (84dc53ace5061dbf022e15c39b8e97bda89c5c94)
  - Dependencies: `MimeKit` from/to the same versions and `BouncyCastle.Cryptography` from 2.5.0 to 2.6.1 (almost latest)
  - Only used by Orchard.Email.
  - `MailKit` subsequently updated to 4.14.1. (cdf58c682503b3efaf47e58f9d70db8b212bdcac)
- `Markdig.Signed`: From 0.22.1 to 0.42.0 (latest). (d46bf8fc0dcbc1431267c431e8706b5da25a3bef)
  - Only used by Markdown.
- `Microsoft.AspNet.*` packages from 5.2.7/3.2.7 to 5.3.0/3.3.0, except `Microsoft.AspNet.WebApi.Client` to 6.0.0. (3ca5dd381fc0b310b2d96207c24d9848301913db)
  - Also updated their dependencies to latest.
- `Moq`: From 4.2.1510.2205 to 4.5.30. Higher versions would require Castle.Core 4.0.0, but we use 3.3.3. (d020137dd45ad5c775e5365b82edae928a94d948)
- `NHibernate`: Updated from 5.5.2 to 5.6.0. (b469231a79747c684ab783066ff24e6f4730108d)
  - `Iesi.Collections` updated as a dependency from 4.0.4 to 4.1.1 (latest).
- `NHibernate.Caches.SysCache2`: Updated from 5.7.0 to 5.9.0. (b469231a79747c684ab783066ff24e6f4730108d)
- `RestEase`: From 1.4.9 to 1.6.4. (da9c4eec2c2e3c32d6b179166d2e90dc044c649c)
- `Redis`: From 1.0.481 to 1.2.6 (latest 1.x). (3291199606f88c87aa28cf81f60c9840e59a0abe)
  - It's compatible with the latest Docker image. 2.x would bring in a lot of new dependencies.
- `YamlDotNet`: From 11.1.1 to 16.3.0 (latest). Only used by Orchard.Layouts. (9ee8217c55a7b9a4fe97f3b10629f634338efa54)
- Orchard.Azure packages: (83de6289b81883ad14728f5e1af5d2e5ff699445)
These packages are all deprecated (but at least not vulnerable), a proper migration to the new Azure.* packages would be nice to have, but not a must.
  - `Microsoft.Azure.KeyVault.Core`: From 1.0.0 to 3.0.5.
  - `Microsoft.WindowsAzure.ConfigurationManager`: From 3.1.0 to 3.2.3.
  - `WindowsAzure.Storage`: From 5.0.2 to 9.3.3.
  - New dependencies:
    - `Microsoft.Rest.ClientRuntime`, also updated to 2.3.24 (latest).
    - `Microsoft.Rest.ClientRuntime.Azure`, also updated to 3.3.19 (latest).
- Orchard.Azure.Tests added to the solution and cleaned up. Since the old Azure SDK and the Azure Storage Emulator can't be installed anymore, Azurite should be running when running these tests. (11bf8f4a0632edd84e0f69fa7c983546c028d61c)
  - `npm install azurite -g; azurite --blobHost 127.0.0.1` or
  - `docker run -d --name azurite -p 10000:10000 mcr.microsoft.com/azure-storage/azurite `

## Assembly binding redirects
- Removed every assembly binding redirect and re-added them based on what VS detects in build warnings + a few more based on runtime errors + more for the test projects based on the errors seen there.

## Removed NuGet packages
- `Microsoft.TypeScript.MSBuild` was only added to Orchard.DynamicForms, but not actually used. (eff88ce6f8cebfc1cae5698d8e01bc7c24745e88)
- `Microsoft.WindowsAzure.Caching` was removed on dev in c6c667d180d741ff54aca804ae7842194d4e05f0 and then added back by mistake in ca86ec98316f6fa3a3b561cf83eae98b14988dae. (de006d30150b13b306f622fed5cf07fc40ba43f7)
- `Orchard.WindowsAzure.ServiceRuntime` (e3ea926ba1b093d862c7c8a5df43799b7dd87eb7)

## Already up-to-date

- `NUnit` and `NUnit.Runners`: 2.7.1 is latest 2.x version, recently updated. (0e71cf3c9defc637a3d3e9a347e314545ee0437f and d66c6b8919d2ecdf20cd5a3bc7fe0ae6d371cceb)
- `SpecFlow` and `SpecFlow.Tools.MsBuild.Generation`: 2.4.1 is latest 2.x version, recently updated (but the whole library is deprecated). (e42519c98807c3a5d273e9f7fcab3c24e2890e78 and 99a500376452610c8e36e8584d9637e0b4657543)

## Not updated intentionally
- `Autofac.*`: Only major update is available (4.0.0) with breaking changes.
- `BouncyCastle.Cryptography`: See `MailKit` above.
- `HtmlAgilityPack`: Can be upgraded from 1.4.9.5 to 1.12.4 (latest), but even some (not all!) of the close versions break tests in various ways that don't make sense. Maybe try [AngleSharp](https://github.com/AngleSharp/AngleSharp) instead?
  - Only used by Orchard.Specs. If it ain't broken, don't fix it.
- `Microsoft.Data.Edm`, `Microsoft.Data.OData`, `Microsoft.Data.Services.Client` and `System.Spatial` are intertwined and cannot be upgraded from 5.8.4 to 5.8.5, because `Microsoft.Data.OData` being as marked as [deprecated due to broken thread safety](https://github.com/OData/odata.net/issues/2452).
- `Microsoft.IdentityModel.*`: Upgrading to the nearest non-deprecated version would bring in a ton of other dependencies and they are only used by Orchard.OpenId.
- `Microsoft.Web.Infrastructure`: Upgrading from 1.0.0.0 to 2.0.0.0 (11 years between them!) breaks the Spec app, maybe worth investigating later.
- `MySql.Data`: DB server support is a can of worms I won't open now, maybe later.
- `Newtonsoft.Json`: Existing packages (e.g. MVC) depend on 13.0.3, no need to manually upgrade to 13.0.4.
- `Newtonsoft.Json.Bson`: Existing packages (MVC and friends) depend on 1.0.2, no need to manually upgrade to 1.0.3.
- `Npgsql`: See `MySql.Data`. 
- `System.Formats.Asn1`: Dependency of `MailKit` and nothing else uses it.
- `System.IdentityModel.Tokens.Jwt`: Nearest non-vulnerable/non-deprecated version is 7.7.1, but other `Microsoft.IdentityModel.*` packages are currently on the same version (5.7.0).
- `System.ValueTuple`: Dependency of multiple packages, no need to update directly.


.github/workflows/specflow.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches:
       - 1.10.x
       - dev
+      - task/nuget-updates
```
BenedekFarkas, line 10: Change to be reverted before merge.
```suggestion
```

BenedekFarkas:
@sebastienros regarding this change (see description):

> Removed every assembly binding redirect and re-added them based on what VS detects in build warnings + a few more based on runtime errors + more for the test projects based on the errors seen there.

Do I assume correctly that if I enable every feature and there are no assembly binding errors, then that completely verifies that all the assembly bindings are OK?

domonkosgabor (approved):

PR #8860, opened 2025-10-16
#8859: `NoLockInterceptor` should modify the SQL statement only if the database provider supports NOLOCK

### LINKED ISSUE #8859: Fixing setup errors with PostgreSQL and MySQL
- [x] Fixed that `NoLockInterceptor` breaks database providers that don't support it (PostgreSQL and MySQL), because it was always active.
- [x] Fixed that an error was logged when running the setup with PostgreSQL. The error itself was necessary and expected to happen, but actually surfaced in the logs, which is a false positive.
- [x] MySQL setup breaks on creating the `Name` column in `DistributedLockSchemaBuilder`. Marking a column `Unique()` automatically indexes it in MySQL (but we do create one manually later on). However, indexed columns can't be larger than 255 characters and currently that column's length is 512 characters. Unfortunately, upgrading `MySql.Data` to the latest 8.x or 9.x version doesn't help.

Also updated `MySql.Data` to the nearest minor version (from 6.7.9 to 6.10.9).

```
MySql.Data.MySqlClient.MySqlException (0x80004005): BLOB/TEXT column 'Name' used in key specification without a key length
```

Ideally, we should intercept the `Unique()` calls in `MySqlCommandInterpreter` or the CreateColumn/AlterColumn commands to modify the column and create the index as above instead of database-specific code in migrations, but here's an example of what works:
```
public void CreateSchema() {
    _schemaBuilder.CreateTable(TableName, table => table
        .Column<int>("Id", column => column.PrimaryKey().Identity())
        // Keep 512 chars but avoid inline .Unique() to prevent MySQL TEXT/BLOB key-length errors.
        .Column<string>("Name", column => column.NotNull().WithLength(512))
        .Column<string>("MachineName", column => column.WithLength(256))
        .Column<DateTime>("CreatedUtc")
        .Column<DateTime>("ValidUntilUtc", column => column.Nullable()));

    var prefixedTableName = _schemaBuilder.TableDbName(TableName);
    var isMySql = string.Equals(_shellSettings.DataProvider, "MySql", StringComparison.OrdinalIgnoreCase);

    if (isMySql) {
        // Force VARCHAR(512) on MySQL (some interpreters map >255 to TEXT).
        _schemaBuilder.ExecuteSql(
            $"ALTER TABLE {prefixedTableName} MODIFY COLUMN Name VARCHAR(512) NOT NULL;",
            cmd => cmd.ForProvider("MySql"));

        // Create a provider-specific unique index on Name.
        _schemaBuilder.ExecuteSql(
            $"CREATE UNIQUE INDEX UQ_DistributedLockRecord_Name ON {prefixedTableName} (Name);",
            cmd => cmd.ForProvider("MySql"));
    }
    else {
        // Provider-agnostic unique constraint for other database providers.
        _schemaBuilder.AlterTable(TableName, table => {
            table.AddUniqueConstraint("UQ_DistributedLockRecord_Name", "Name");
        });

        // Keep the secondary (non-unique) index for non-MySQL only to avoid duplicate indexes.
        _schemaBuilder.AlterTable(TableName, table => {
            table.CreateIndex("IDX_DistributedLockRecord_Name", "Name");
        });
    }
}
```

BenedekFarkas:
Fixes #8859.

### DIFF
--- a/.github/workflows/specflow.yml
+++ b/.github/workflows/specflow.yml
@@ -7,6 +7,7 @@ on:
     branches:
       - 1.10.x
       - dev
+      - task/nuget-updates
   schedule:
     - cron: '0 0 * * 1' # Every Monday midnight.
 
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -9,3 +9,6 @@ trim_trailing_whitespace = true
 [*.cs]
 csharp_new_line_before_open_brace = none
 dotnet_sort_system_directives_first = true
+
+[*.config]
+indent_size = 2
--- a/src/Libraries/NHibernate/NHibernate.Linq/NHibernate.Linq.csproj
+++ b/src/Libraries/NHibernate/NHibernate.Linq/NHibernate.Linq.csproj
@@ -45,12 +45,12 @@
     <Reference Include="Antlr3.Runtime, Version=3.5.0.2, Culture=neutral, PublicKeyToken=eb42632606e9261f, processorArchitecture=MSIL">
       <HintPath>..\..\..\packages\Antlr3.Runtime.3.5.1\lib\net40-client\Antlr3.Runtime.dll</HintPath>
     </Reference>
-    <Reference Include="Iesi.Collections, Version=4.0.0.4000, Culture=neutral, PublicKeyToken=aa95f207798dfdb4, processorArchitecture=MSIL">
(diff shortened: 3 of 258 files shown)


BenedekFarkas:
Accidentally opened the issue branch from another issue branch.

PR #8870, opened 2025-11-05
#8835: Merging imported SearchFields from recipes instead of overwriting them (Lombiq Technologies: ORCH-310)

### LINKED ISSUE #8835: SearchSettingsPart SearchFields property is overwritten when multiple recipes add search fields
When defining SearchSettingsPart settings through recipes, the SearchFields property is overwritten instead of merged when multiple recipes attempt to add search fields for different indexes.
This results in losing previously defined search fields if a later recipe sets its own.

### Steps to reproduce the behavior

1. Create a recipe that adds the following SearchFields:
```
<SearchSettingsPart SearchFields="
    Admin:
    UserPart.Email,
    UserPart.FirstName,
    UserPart.LastName" 
    DisplayType="Summary" FilterCulture="false" SearchIndex="" />
```
2. Create another recipe that adds:

```
<SearchSettingsPart SearchFields="
    UserSearch:
    UserPart.UserName,
    UserPart.Email,
    UserPart.RegistrationStatus,
    UserPart.CreatedUtc,
    UserPart.LastLoginUtc,
    UserRolesPart.Roles" 
    DisplayType="Summary" FilterCulture="false" SearchIndex="" />
```
3. Execute both recipes.

Expected behavior
Both Admin and UsersReport search fields should coexist after applying both recipes.

Actual behavior
The last applied recipe overwrites the previous SearchFields settings. Only the fields from the latest recipe are present.

The SearchFields is treated as a single text field and fully replaced rather than merged by the index name. Ideally, the system should merge entries or provide a way to append settings without overwriting previous ones.

domonkosgabor:
Fixes: https://github.com/OrchardCMS/Orchard/issues/8835

You can test the changes by executing the following recipes with some test content:
[Search2.recipe.xml](https://github.com/user-attachments/files/23580538/Search2.recipe.xml)
[Search3.recipe.xml](https://github.com/user-attachments/files/23580539/Search3.recipe.xml)
[Search4.recipe.xml](https://github.com/user-attachments/files/23580540/Search4.recipe.xml)

This PR also fixes if you enable the **Search** module without an index implementation, navigating to **/Admin/Settings/Search** will throw an exception.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Search/Drivers/SearchSettingsPartDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Search/Drivers/SearchSettingsPartDriver.cs
@@ -36,7 +36,9 @@ namespace Orchard.Search.Drivers
                     var model = new SearchSettingsIndexViewModel
                     {
                         SelectedIndex = part.SearchIndex,
-                        AvailableIndexes = _indexManager.HasIndexProvider() ? _indexManager.GetSearchIndexProvider().List().ToList() : new List<string>()
+                        AvailableIndexes = _indexManager.HasIndexProvider()
+                            ? _indexManager.GetSearchIndexProvider().List().ToList()
+                            : new List<string>()
                     };
 
                     if (updater != null)
--- a/src/Orchard.Web/Modules/Orchard.Search/Drivers/SearchSettingsPartDriver.cs
+++ b/src/Orchard.Web/Modules/Orchard.Search/Drivers/SearchSettingsPartDriver.cs
@@ -5,6 +5,7 @@ using Orchard.ContentManagement.Drivers;
 using Orchard.ContentManagement.Handlers;
 using Orchard.Indexing;
 using Orchard.Localization;
+using Orchard.Search.Helpers;
 using Orchard.Search.Models;
 using Orchard.Search.ViewModels;
 
@@ -35,7 +36,7 @@ namespace Orchard.Search.Drivers
                     var model = new SearchSettingsIndexViewModel
                     {
                         SelectedIndex = part.SearchIndex,
-                        AvailableIndexes = _indexManager.GetSearchIndexProvider().List().ToList()
+                        AvailableIndexes = _indexManager.HasIndexProvider() ? _indexManager.GetSearchIndexProvider().List().ToList() : new List<string>()
                     };
 
                     if (updater != null)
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.Search/Drivers/SearchSettingsPartDriver.cs
```diff
@@ -35,7 +36,7 @@ protected override DriverResult Editor(SearchSettingsPart part, IUpdateModel upd
                     var model = new SearchSettingsIndexViewModel
                     {
                         SelectedIndex = part.SearchIndex,
-                        AvailableIndexes = _indexManager.GetSearchIndexProvider().List().ToList()
+                        AvailableIndexes = _indexManager.HasIndexProvider() ? _indexManager.GetSearchIndexProvider().List().ToList() : new List<string>()
```
BenedekFarkas:
```suggestion
                        AvailableIndexes = _indexManager.HasIndexProvider()
                            ? _indexManager.GetSearchIndexProvider().List().ToList()
                            : new List<string>()
```

src/Orchard.Web/Modules/Orchard.Search/Helpers/SearchSettingsHelper.cs
```diff
@@ -37,5 +37,53 @@ public static string[] GetSearchFields(this SearchSettingsPart part, string inde
         {
             return part.SearchFields.ContainsKey(index) ? part.SearchFields[index] : new string[0];
         }
+
+        public static string MergeSearchFields(IDictionary<string, string[]> existingSearchFields, IDictionary<string, string[]> searchFieldsToAdd)
```
BenedekFarkas, line 44: Please add a few comments on the most important parts of the logic to annotate what's happening.

BenedekFarkas, line 44: If I import search fields for an index that doesn't include a field that is currently included, then that field should be removed. Simply put, the import process should overwrite the given index's configuration, just not drop the others (if they aren't included). Otherwise there will be no way to remove a field from indexing via a recipe.

BenedekFarkas (changes requested):

BenedekFarkas:
See the review suggestions. BTW thanks for the test recipes - please also include the content definitions in the first one, because the search field configuration references fields that don't exist in a fresh setup.

src/Orchard.Web/Modules/Orchard.Search/Drivers/SearchSettingsPartDriver.cs
BenedekFarkas: Please use the commit suggestion feature next time.

BenedekFarkas (approved):

PR #8873, opened 2025-11-24
#8268: Getting rid of [Obsolete] stuff (Lombiq Technologies: ORCH-308)

### LINKED ISSUE #8268: Getting rid of [Obsolete] stuff
We have a few dozen `[Obsolete]` components. This issue is about removing as many of them as possible, the deciding condition being that the component was marked obsolete more than a release ago.

domonkosgabor:
Fixes: [#8268](https://github.com/OrchardCMS/Orchard/issues/8268)

I compared the current dev branch with the Orchard 1.10.3 release. I got rid of the code that was obsoleted in that release as well. This involves the following changes:

- Using Orchard.Mvc.FormValueRequiredAttribute instead of the FormValueRequiredAttribute class of the AdminController in the Orchard.Core.Contents namespace.
- Removing obsoleted Spam from the Orchard.Comments.Models.CommentStatus enum.
- Removing obsoleted LazyField<T> class from the Orchard.Core.Common.Utilities namespace.
- Removing obsoleted methods from OrchardLog4netLogger.
- Removing obsoleted Loader(Func<T, T> loader) method from LazyField.cs in the Orchard.ContentManagement.Utilities namespace.
- Removing obsoleted ContentShapeResult ContentShape(IShape shape) method from ContentPartDriver.
- Removing obsoleted ISessionLocator interface.
- Removing obsoleted ContentFieldTemplate methods from ContentFieldDriver<TField>.
- Removing obsoleted IMessagingChannel and IMessageManager interfaces, along with the migration step that updates the existing Send Email activities to use the new Message Queueing functionality.
- Removing obsoleted CustomStepsStep. Don't need the CustomStepsStep RecipeBuilderStep because you can implement IRecipeBuilderStep and IRecipeExecutionStep instead of implementing custom export steps.
- Removing obsoleted CustomPropertiesPart. It used to be necessary for the List feature (also obsolete) to be able to sort on custom data.
- Removing the obsoleted ITaskLeaseService with its default implementation. Removing deprecated Orchard.TaskLease module because it only contains this interface with its default implementation. Use Orchard.Tasks.Locking.IDistributedLockService instead.
- Removing LegacyRulesEvaluator. This condition provider was there for backwards compatibility during the deprecation period.
- Removing IRuleManager with its default implementation. Use Orchard.Conditions.Services.ConditionManager instead.
- Removing unused RequiresOwnership permission.


BenedekFarkas:
> I compared the current dev branch with the Orchard 1.10.3 release.

Would it change anything if we compare to 1.10 instead?

domonkosgabor:
> > I compared the current dev branch with the Orchard 1.10.3 release.
> 
> Would it change anything if we compare to 1.10 instead?

I assume it means less obsolete code that we need to take care of. Should I compare the current dev branch with the Orchard 1.10 release to see the changes?

BenedekFarkas:
> I assume it means less obsolete code that we need to take care of. Should I compare the current dev branch with the Orchard 1.10 release to see the changes?

If it's simple to compare, then yes please.

domonkosgabor:
> > I assume it means less obsolete code that we need to take care of. Should I compare the current dev branch with the Orchard 1.10 release to see the changes?
> 
> If it's simple to compare, then yes please.

It seems that Orchard 1.10 and Orchard 1.10.3 have the exact same code marked as obsolete.

BenedekFarkas:
Perfect, thanks!

src/Orchard.Web/Modules/Upgrade/Controllers/MessagingController.cs
BenedekFarkas, line 1: This shouldn't be removed as it could leave users stuck with having to upgrade manually. This feature can depend on other features (it already does) and you can move the minimum required code from the other module to this one for the upgrade process to keep working.

BenedekFarkas (changes requested):

BenedekFarkas (approved):

PR #8874, opened 2025-11-26
#8266: Security sensitive permissions (Lombiq Technologies: ORCH-214)

### LINKED ISSUE #8266: Security sensitive permissions
Counterpart of https://github.com/OrchardCMS/OrchardCore/issues/2177.

Suggested changes:

1. Add `Hint` property to the `Permissions` class that should be used to explain in detail how that permission affects the Users' access.
2. Add `IsSecurityCritical` property to the `Permissions` class that should be set to true for Permissions that can be used by Users to elevate their permissions (either explicitly by changing Settings/Roles/Permissions or implicitly by changing the site's behavior).
I'd consider the following permissions to be marked: `Site Owners Permission`, `Manage Settings`, `Managing Users`, `Managing Roles`, `Assign Roles`, `Edit content types`, `Apply a Theme`, `Manage packages`, `Manage Features`, `Import Data`, `Export Data`, `Managing Templates` (Razor) and everything in Audit Trail.
3. Security critical permissions should be indicated in the following places:
    1. Role editor: A not too distracting highlight and an icon with a hint or a new column with disabled checkboxes next to "Effective". On a related note, the text "Effective" should have a hint to explain that Permissions are affected by their `ImpliedBy` property.
    2. Role Admin list: Roles that have at least one security sensitive permission should be marked with an icon (with a hint).
    3. UserRolesPart editor: Same as Role Admin list.

domonkosgabor:
Fixes: https://github.com/OrchardCMS/Orchard/issues/8266

Here, I tried to mimic what we currently have in Orchard Core. It means we have a badge with the "_Security Critical_" text and a tooltip that reads "_This permission could allow a user to elevate their other permissions. Grant it with extreme consideration._". It appears before each _Allow_ checkbox when setting up a role's permissions. We also have a hint on that page describing the differences between the _Allow_ and _Effective_ settings.

In this PR, I have added the same hint to describe the _Allow_ and _Effective_ settings. We have an icon with an exclamation mark instead of a badge, and I used the exact text for the tooltip, as you can see in Orchard Core.

The tooltip text is the same as in Orchard Core.

I didn't want to include too much styling, so I used an icon from Font Awesome with the title HTML attribute.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.Roles/Services/RoleService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Roles/Services/RoleService.cs
@@ -203,6 +203,22 @@ namespace Orchard.Roles.Services
             return installedPermissions;
         }
 
+        public IEnumerable<string> GetSecurityCriticalPermissions()
+        {
+            var securityCriticalPermissions = new List<string>();
+            foreach (var permissionProvider in _permissionProviders)
+            {
+                var permissions = permissionProvider.GetPermissions();
+                foreach (var permission in permissions)
+                {
+                    if (permission.IsSecurityCritical)
+                        securityCriticalPermissions.Add(permission.Name);
+                }
+            }
+
+            return securityCriticalPermissions;
+        }
+
         public IEnumerable<string> GetPermissionsForRole(int id)
         {
             var permissions = new List<string>();
--- a/src/Orchard.Web/Modules/Orchard.Roles/Services/RoleService.cs
+++ b/src/Orchard.Web/Modules/Orchard.Roles/Services/RoleService.cs
@@ -203,21 +203,8 @@ namespace Orchard.Roles.Services
             return installedPermissions;
         }
 
-        public IEnumerable<string> GetSecurityCriticalPermissions()
-        {
-            var securityCriticalPermissions = new List<string>();
-            foreach (var permissionProvider in _permissionProviders)
-            {
-                var permissions = permissionProvider.GetPermissions();
-                foreach (var permission in permissions)
-                {
-                    if (permission.IsSecurityCritical)
-                        securityCriticalPermissions.Add(permission.Name);
-                }
-            }
-
-            return securityCriticalPermissions;
-        }
+        public IEnumerable<string> GetSecurityCriticalPermissions() =>
+            _permissionProviders.SelectMany(pp => pp.GetPermissions().Where(p => p.IsSecurityCritical)).Select(p => p.Name);
 
         public IEnumerable<string> GetPermissionsForRole(int id)
         {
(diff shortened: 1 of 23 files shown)


src/Orchard.Web/Modules/Orchard.Roles/Services/RoleService.cs
```diff
@@ -203,6 +203,22 @@ public IDictionary<string, IEnumerable<Permission>> GetInstalledPermissions()
...
+                var permissions = permissionProvider.GetPermissions();
+                foreach (var permission in permissions)
+                {
+                    if (permission.IsSecurityCritical)
+                        securityCriticalPermissions.Add(permission.Name);
+                }
+            }
+
+            return securityCriticalPermissions;
+        }
```
BenedekFarkas:
```suggestion
        public IEnumerable<string> GetSecurityCriticalPermissions() =>
            _permissionProviders.SelectMany(pp => pp.GetPermissions().Where(p => p.IsSecurityCritical)).Select(p => p.Name);
```

src/Orchard.Web/Modules/Orchard.Roles/Views/Admin/Edit.cshtml
```diff
@@ -1,67 +1,92 @@
...
+                            <td style="width:60px;">
+                                @if (Model.CurrentPermissions.Contains(permission.Name))
+                                {
+                                    <input type="checkbox" value="true" name="Checkbox.@permission.Name" checked="checked" />
+                                }
+                                else
+                                {
+                                    <input type="checkbox" value="true" name="Checkbox.@permission.Name" />
+                                }
+                                @if (permission.IsSecurityCritical && !string.IsNullOrEmpty(permission.Hint))
```
BenedekFarkas: Modify the display logic to put the default hint in the title if the hint is empty, then revert adding the default hint in the permission providers.

src/Orchard/Security/Permissions/Permission.cs
```diff
@@ -11,6 +11,13 @@ public class Permission
 
         public IEnumerable<Permission> ImpliedBy { get; set; }
 
+        /// <summary>
+        /// Gets a value indicating whether the permission is security critical.
```
BenedekFarkas: Please write a meaningful summary instead of self-referencing.

BenedekFarkas (changes requested):

src/Orchard/Security/Permissions/Permission.cs
```diff
@@ -11,6 +11,13 @@ public class Permission
 
         public IEnumerable<Permission> ImpliedBy { get; set; }
 
+        /// <summary>
+        /// Indicates whether this permission could allow a user to elevate their other permissions.
+        /// </summary>
+        public bool IsSecurityCritical { get; set; }
+
+        public string Hint { get; set; }
```
BenedekFarkas: It occurred to me that if I look at the Permission class, I would assume that if I provide a hint, then it will be displayed. That's not the case, because it's currently tied to the IsSecurityCritical being set to true.
Having a generic hint functionality here would be nice to have, but there was no request for it and I'd rather keep it simple, so we can just display the default hint if the permission is security critical (and then it will become a T-string too) and we're done.

BenedekFarkas (changes requested):

BenedekFarkas (approved):

PR #8875, opened 2025-12-02
#8862: Preventing the SQL Server Service Broker feature from being enabled for tenants that are not using Microsoft SQL Server as their database provider (Lombiq Technologies: ORCH-309) 

### LINKED ISSUE #8862: SQL Server Service Broker breaks the app when enabled with port parameter in the connection string
For example features can't be enabled/disabled properly until the application is restarted, because the tenant fails to start. Tested with MySQL.

```
2025-10-17 20:50:28,467 [7] Orchard.Environment.DefaultOrchardHost - (null) - A tenant could not be started: Default Attempt number: 0 [(null)]
System.ArgumentException: Keyword not supported: 'port'.
   at System.Data.Common.DbConnectionOptions.ParseInternal(Hashtable parsetable, String connectionString, Boolean buildChain, Hashtable synonyms, Boolean firstKey)
   at System.Data.Common.DbConnectionOptions..ctor(String connectionString, Hashtable synonyms, Boolean useOdbcRules)
   at System.Data.SqlClient.SqlConnectionString..ctor(String connectionString)
   at System.Data.SqlClient.SqlDependency.Stop(String connectionString, String queue, Boolean useDefaults, Boolean startFailed)
   at Orchard.MessageBus.Brokers.SqlServer.SqlServerBroker.Dispose()
   at Autofac.Core.Disposer.Dispose(Boolean disposing)
   at Autofac.Util.Disposable.Dispose()
   at Autofac.Core.Lifetime.LifetimeScope.Dispose(Boolean disposing)
   at Autofac.Util.Disposable.Dispose()
   at Orchard.Environment.ShellBuilders.ShellContextFactory.CreateShellContext(ShellSettings settings) in E:\DEV\Orchard-1\Orchard\src\Orchard\Environment\ShellBuilders\ShellContextFactory.cs:line 74
   at Orchard.Environment.DefaultOrchardHost.CreateShellContext(ShellSettings settings) in E:\DEV\Orchard-1\Orchard\src\Orchard\Environment\DefaultOrchardHost.cs:line 240
   at Orchard.Environment.DefaultOrchardHost.<CreateAndActivateShells>b__41_1(ShellSettings settings) in E:\DEV\Orchard-1\Orchard\src\Orchard\Environment\DefaultOrchardHost.cs:line 166
```

I3undy:
Fixes #8862.

### DIFF
--- a/src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
+++ b/src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
@@ -12,14 +12,13 @@ using Orchard.UI.Notify;
 namespace Orchard.MessageBus.Services
 {
     /// <summary>
-    /// Prevents the SQL Server Service Broker feature from being enabled for tenants that are not using
-    /// Microsoft SQL Server as their database provider.
+    /// Prevents the SQL Server Service Broker feature from being enabled for tenants that are not using Microsoft SQL
+    /// Server as their database provider.
     /// </summary>
     /// <remarks>
-    /// The implementation is hackish but there seems to be no other way: if it would use <see cref="Orchard.UI.Notify.INotifier"/>
-    /// it wouldn't work since <see cref="Orchard.UI.Notify.NotifyFilter"/> that writes out notifications to TempData 
-    /// runs before feature events are raised. Thus we have to manually add the messages to TempData, just as the filter 
-    /// would do.
+    /// The implementation is hackish but there seems to be no other way: if it would use <see cref="INotifier"/> it
+    /// wouldn't work since <see cref="NotifyFilter"/> that writes out notifications to TempData runs before feature
+    /// events are raised. Thus we have to manually add the messages to TempData, just as the filter would do.
     /// </remarks>
     public class SqlServerServiceBrokerFeatureGuard : FilterProvider, IFeatureEventHandler, IActionFilter
     {
@@ -82,7 +81,6 @@ namespace Orchard.MessageBus.Services
 
         public void Uninstalling(Feature feature) { }
 
-
         public void Uninstalled(Feature feature) { }
 
         private void AddNotification(LocalizedString message, NotifyType notifyType = NotifyType.Warning)
@@ -92,12 +90,7 @@ namespace Orchard.MessageBus.Services
             if (tempDataDictionary == null) return;
 
             ((TempDataDictionary)tempDataDictionary)["messages"] +=
-                notifyType.ToString() +
-                ":" +
-                message.Text +
-                System.Environment.NewLine +
-                "-" +
-                System.Environment.NewLine;
+                $"{notifyType}:{message.Text}{System.Environment.NewLine}-{System.Environment.NewLine}";
         }
     }
 }
--- a/src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
+++ b/src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
@@ -1,0 +1,103 @@
+using System;
+using System.Web.Mvc;
+using Orchard.Environment;
+using Orchard.Environment.Configuration;
+using Orchard.Environment.Extensions.Models;
+using Orchard.Environment.Features;
+using Orchard.Localization;
+using Orchard.Mvc;
+using Orchard.Mvc.Filters;
+using Orchard.UI.Notify;
+
+namespace Orchard.MessageBus.Services
+{
+    /// <summary>
+    /// Prevents the SQL Server Service Broker feature from being enabled for tenants that are not using
+    /// Microsoft SQL Server as their database provider.
+    /// </summary>
+    /// <remarks>
+    /// The implementation is hackish but there seems to be no other way: if it would use <see cref="Orchard.UI.Notify.INotifier"/>
+    /// it wouldn't work since <see cref="Orchard.UI.Notify.NotifyFilter"/> that writes out notifications to TempData 
+    /// runs before feature events are raised. Thus we have to manually add the messages to TempData, just as the filter 
+    /// would do.
+    /// </remarks>
+    public class SqlServerServiceBrokerFeatureGuard : FilterProvider, IFeatureEventHandler, IActionFilter
+    {
+        private const string FeatureId = "Orchard.MessageBus.SqlServerServiceBroker";
+        private const string TempDataKey = "Orchard.MessageBus.SqlServerServiceBroker.TempData";
+
+        private readonly ShellSettings _shellSettings;
+        private readonly IFeatureManager _featureManager;
+        private readonly IHttpContextAccessor _hca;
+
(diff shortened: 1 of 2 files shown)


src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// <summary>
+    /// Prevents the SQL Server Service Broker feature from being enabled for tenants that are not using
+    /// Microsoft SQL Server as their database provider.
+    /// </summary>
+    /// <remarks>
+    /// The implementation is hackish but there seems to be no other way: if it would use <see cref="Orchard.UI.Notify.INotifier"/>
+    /// it wouldn't work since <see cref="Orchard.UI.Notify.NotifyFilter"/> that writes out notifications to TempData 
+    /// runs before feature events are raised. Thus we have to manually add the messages to TempData, just as the filter 
+    /// would do.
+    /// </remarks>
```
BenedekFarkas, line 22: Namespace and rewrap:
```suggestion
    /// <summary>
    /// Prevents the SQL Server Service Broker feature from being enabled for tenants that are not using Microsoft SQL
    /// Server as their database provider.
    /// </summary>
    /// <remarks>
    /// The implementation is hackish but there seems to be no other way: if it would use <see cref="INotifier"/> it
    /// wouldn't work since <see cref="NotifyFilter"/> that writes out notifications to TempData runs before feature
    /// events are raised. Thus we have to manually add the messages to TempData, just as the filter would do.
    /// </remarks>
```

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        public void Disabling(Feature feature) { }
+
+        public void Disabled(Feature feature) { }
+
+        public void Uninstalling(Feature feature) { }
+
+
```
BenedekFarkas:
```suggestion
```

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (feature.Descriptor.Id != FeatureId)
+                return;
+
+            if (!_shellSettings.DataProvider.Equals("SqlServer", StringComparison.OrdinalIgnoreCase))
+            {
+                _featureManager.DisableFeatures(new[] { FeatureId }, force: true);
+
+                AddNotification(
+                    T("The SQL Server Service Broker cannot be enabled because it requires Microsoft SQL Server."),
+                    NotifyType.Error);
```
BenedekFarkas, line 74: Can you please check if you see the same behavior: When enabling `SqlServerServiceBroker` while `Orchard.MessageBus` is also disabled, then the notification will not be displayed. If I disable `SqlServerServiceBroker` and then enable it again (and thus `Orchard.MessageBus` is already enabled) then the notification is displayed.

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (httpContext == null) return;
+
+            httpContext.Items[TempDataKey] = filterContext.Controller.TempData;
+        }
+
+        public void Installing(Feature feature) { }
+
+        public void Installed(Feature feature) { }
+
+        public void Enabling(Feature feature) { }
```
BenedekFarkas, line 61: Please check if we can store here whether `SqlServerServiceBroker`'s dependencies (there's only `Orchard.MessageBus`, but don't wire that in) are already enabled or not. Those that aren't should also be disabled along with `SqlServerServiceBroker` in the `Enabled` handler.

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            if (tempDataDictionary == null) return;
+
+            ((TempDataDictionary)tempDataDictionary)["messages"] +=
+                notifyType.ToString() +
+                ":" +
+                message.Text +
+                System.Environment.NewLine +
+                "-" +
+                System.Environment.NewLine;
```
BenedekFarkas:
```suggestion
                $"{notifyType}:{message.Text}{System.Environment.NewLine}-{System.Environment.NewLine}";
```

BenedekFarkas (changes requested):

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
I3undy: There is no built-in feature in O1 for this. We only have `GetDependentFeatures`.
We can add `GetDependencies` to the `IFeatureManager`, but that is a complicated topic. 
E.g., How deep and wide should the graph traversal be?

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
I3undy: This happens because the message is added too late.
When the feature is enabled while `Orchard.MessageBus` is still off, Orchard has already finished the part where it shows notifications.

If `Orchard.MessageBus` is already on, the message gets added earlier, so it shows up right away.

Putting it in Session/TempData doesn’t fix it, because the feature check doesn’t always run during a normal page request.

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
BenedekFarkas: For suggested changes, use the commit suggestion feature next time, so the conversation is also resolved.

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
BenedekFarkas: OK, let's leave it at that. Not enabling `SqlServerServiceBroker` is already an improvement.

src/Orchard.Web/Modules/Orchard.MessageBus/Services/SqlServerServiceBrokerFeatureGuard.cs
BenedekFarkas: OK, let's skip this too, not worth the effort.

BenedekFarkas (approved):

PR #8885, opened 2026-01-07
#8884: Adding content type filter to the Recycle Bin

### LINKED ISSUE #8884: Add content type filter to the Recycle Bin

BenedekFarkas:
Fixes #8884


BenedekFarkas:
The changes at the current state have a flaw in the query, which I managed to fix, but only in a way that is incompatible with NoLockInterceptor, so tabling this for now.

